PR #8, opened 2019-02-28
Order datasets selectors

jameshadfield:
Previously the dataset selectors were not in the desired order due to the default value appearing first. This commit fixes this for the nextstrain.org server. Note that this functionality will be moved to the client when the new server API is implemented (see https://github.com/nextstrain/auspice/issues/687).

This commit closes Auspice issue https://github.com/nextstrain/auspice/issues/696.

### DIFF
--- a/auspice/server/getDatasetHelpers.js
+++ b/auspice/server/getDatasetHelpers.js
@@ -31,32 +31,47 @@ const decideSourceFromPrefix = (prefix) => {
  * "default" one.
  */
 const correctPrefixFromAvailable = (source, prefixParts) => {
+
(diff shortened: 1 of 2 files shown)


jameshadfield:
[nextstrain.org PR: order dataset selectors](https://trello.com/c/Lls48rio/11-nextstrainorg-pr-order-dataset-selectors)

trvrb:
This looks good to me. Code looks look and behavior is as expected. It's nice having this ordering fixed for flu 6m, 2y, etc... This is good to merge from my perspective.

PR #25, opened 2019-08-27
Implement server redirects in nextstrain.org to handle incomplete request urls

kairstenfay:
This is the nextstrain.org side of nextstrain/auspice#778

See also auspice issue [753](https://github.com/nextstrain/auspice/issues/753)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,16 +6,11 @@ Nextstrain is an open-source project to harness the scientific and public health
 
 
 This repo builds [nextstrain.org](https://nextstrain.org) and comprises:
-* The splash page & documentation, which is built using Gatsby
-* Customisations for [auspice](github.com/nextstrain/auspice)
-* Server which, amongst other things, handles auspice data JSONs
+  1. The [splash & documentation pages](#Splash--documentation-pages), which are built using Gatsby
+  2. Customisations for [Auspice](#Auspice)
+  3. [Server](#Server) which, amongst other things, handles auspice data JSONs
 
-
-## Building nextstrain.org locally
-
-1. `npm install` (from the "nextstrain.org" directory)
-2. `npm run set-up`, which runs `./set-up.sh` to build both the static site & auspice.
-3. `npm run server` will then start a local instance, by default available at [localhost:5000](http://localhost:5000).
+This repository provides the tools you need to [build nextstrain.org locally](#build-nextstrainorg-locally) and [deploy nextstrain.org](#deploy-nextstrainorg).
 
 ---
 ## Splash & documentation pages
@@ -24,13 +19,15 @@ This is found at `./static-site/`.
 See [static-site/README.md](./static-site/README.md) for instructions on how to add documentation and develop.
 This is built using Gatsby during deployment and served via `./server.js`.
 
-
+---
 ## Auspice
-Auspice is the software used to visualise & interact with phylogenomic data.
-The auspice customisations are found in `./auspice/client/`, with the server code for handling auspice-derived requests for datasets etcetera in `./auspice/server`.
+[Auspice](https://github.com/nextstrain/auspice) is the web-facing (e.g. "client") software used to visualise & interact with phylogenomic data.
+
+### Auspice client customisations
+The auspice customisations specific to nextstrain.org are found in `./auspice/client/`.
 Please see [the auspice documentation](https://nextstrain.github.io/auspice/customisations/introduction) for more information.
 
-#### Testing locally:
+### Testing locally:
 > Auspice must be installed globally -- check that `auspice -h` works.
 
 Production mode:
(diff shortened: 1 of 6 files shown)


jameshadfield:
Thanks @kairstenfay -- i'll look at this today 👍 

Before I do so:
* The changes to auspice (https://github.com/nextstrain/auspice/pull/778) will get merged into the v2 branch (somehow), so the code changes which change behavior here are best to go into a `auspice-v2` branch or similar.
* Given that, would you like to separate out the docs changes into a separate PR which can go into master straight away (I don't mind)
* Is commit 1a10803, whose message starts with "DROP" intended to remain?

jameshadfield:
> Do you mean this for nextstrain.org or auspice?

I meant that for nextstrain.org, but you could also do for auspice if you'd like 😄

README.md
```diff
@@ -52,12 +53,34 @@ Note that the auspice development mode uses the auspice splash page which won't
...
+Therefore, auspice installations do not come with any pre-packaged data.
+Auspice requires interaction with a server to send and receive data.
+The [auspice repo on GitHub](http://github.com/nextstrain/auspice) contains a NodeJS backend that serves the bundled auspice JavaScript file to the server and also provides web API endpoints (prefixed with '`/charon`).
+The auspice front-end (e.g. "client") is server-agnostic. Currently, however, the auspice backend (e.g. "server") is not implemented in an importable or flexible way that allows usage across servers.
+This means that nextstrain.org, auspice.us, and your local auspice installation all use different servers.
 
+[./server.js](server.js) decides, based on the path, whether to serve the bundled auspice JavaScript file (built from running `auspice build`), the (pre-built) static [splash & documentation pages](#Splash--documentation-pages), or the auspice-specific code from `./auspice/server` to process auspice requests for data from the `/charon` web API endpoints (including data transforms).
 
+---
+## Build nextstrain.org locally
```
jameshadfield: These are great! Can we separate out "Build nextstrain.org locally mirroring the deployed (live) website", which _literally_ runs these 3 steps, and "Build nextstrain.org locally using a custom version of auspice" which is only really useful for development purposes.

README.md
```diff
@@ -52,12 +53,34 @@ Note that the auspice development mode uses the auspice splash page which won't
...
+
+1. `npm install` (from the "nextstrain.org" directory)
+2. `npm run set-up`, which runs `./set-up.sh` to build both the static site & auspice.
+3. `npm run server` will then start a local instance, by default available at [localhost:5000](http://localhost:5000).
+
+Because the nextstrain.org server and the auspice backend are completely independent, you cannot currently run `auspice view` (with hot reloading) to serve the nextstrain.org server.
```
jameshadfield: some clarifications:

*  They're independent in the sense that they're both "servers" for an auspice client -- i.e. they both serve the auspice client JS code, and listen for GET requests (`/charon/...`) -- so it doesn't make sense to try to combine them. 

* `auspice view` doesn't have hot reloading - that's `auspice develop`.

* The actual code that handles `/charon/...` requests in nextstrain.org (see https://github.com/nextstrain/nextstrain.org/blob/master/auspice/server/index.js) has been written in such a way that that it can be imported by 
    (i) the nextstrain.org server
    (ii) `auspice view --handlers ./server/index.js`
    (iii) `auspice develop --handlers ./server/index.js`

README.md
```diff
@@ -24,13 +19,19 @@ This is found at `./static-site/`.
 See [static-site/README.md](./static-site/README.md) for instructions on how to add documentation and develop.
 This is built using Gatsby during deployment and served via `./server.js`.
 
-
+---
 ## Auspice
-Auspice is the software used to visualise & interact with phylogenomic data.
-The auspice customisations are found in `./auspice/client/`, with the server code for handling auspice-derived requests for datasets etcetera in `./auspice/server`.
+[Auspice](https://github.com/nextstrain/auspice) is the web-facing (e.g. "client") software used to visualise & interact with phylogenomic data.
+When nextstrain.org is built, it uses a JavaScript bundle from a previously installed version of auspice.
```
jameshadfield: Could you clarify this a little bit? 

do you mean while running `./set-up.sh`, or while running `node server.js`? The former will install a fresh copy auspice and build the client (i.e. the bundle) using nextstrain.org provided customisations. The latter isn't building anything, and expects a built client to exist (which it'll serve).

README.md
```diff
@@ -24,13 +19,19 @@ This is found at `./static-site/`.
...
+By default, `nextstrain.org` is built using a globally installed version of auspice specified in [set-up.sh](/set-up.sh).
+To develop `nextstrain.org` using a local development version of auspice, do not run `set-up.sh`.
+Instead, follow [these auspice installation instructions](https://github.com/nextstrain/auspice/blob/master/docs-src/docs/installation.md#installing-from-source) to set up your desired auspice environment.
```
jameshadfield: Can you change the link to https://nextstrain.org/docs/getting-started/local-installation/#install-auspice-from-source -- thanks 🙏

README.md
```diff
@@ -24,13 +19,19 @@ This is found at `./static-site/`.
...
+
+### Auspice customisations
```
jameshadfield: Due to the folder structure placing both auspice client customisation code & the code related to handling auspice-derived GET requests, can we rename this "Auspice client customisations"

static-site/README.md
```diff
@@ -4,10 +4,11 @@
 
 <br/>
 
-
+# Nextstrain static site
 This directory contains the documentation files and static site generator for nextstrain.
+For general `nextstrain.org` documentation including the nextstrain server (charon) and auspice customizations, see [nextstrain.org/README.md]((./static-site/README.md)).
```
jameshadfield: Given the recent confusion over what "charon" is, can we reword this? (Charon's not the "nextstrain server" -- it's just the word which refers to the auspice-server API)

jameshadfield:
Thanks @kairstenfay -- i've tested this with https://github.com/nextstrain/auspice/pull/778 and it works 👍 💯 

All of the above comments are about the provided documentation, which I think should be merged into master ASAP as they really help explain things.

The refactoring does make things easier to read, thanks for that 👍 

As for the redirect-related code, could you explain a bit what it's doing? Here's what I think's happening, but am not entirely sure:
1. When a `getDataset` GET request is picked up with a (query defined) `prefix` which doesn't match a dataset, but we know which one it should, so the server sends `res.redirect` containing a new GET request with the correct `prefix`
2. This is never actually seen by the client
3. The server picks up this redirect as a "new" GET request, processes things as normal, and responds with the dataset JSON
4. The client sees this JSON and is happy. This is the only reponse the client sees for its initial request. (It uses the `prefix` of the GET URL to modify its own URL.)

kairstenfay:
> All of the above comments are about the provided documentation, which I think should be merged into master ASAP as they really help explain things.
README updates separated in PR #26 

> The refactoring does make things easier to read, thanks for that +1
Anytime :smile: 

> As for the redirect-related code, could you explain a bit what it's doing? Here's what I think's happening, but am not entirely sure:
Yep, that's my understanding as well. I added a bit of documentation in the docstring of `getDataset`

jameshadfield:
Thanks! Just a bit of follow-up on the redirect & the client:

The added docstring states: "then [the server] sends the client a redirect to the new URL." but my (brief) testing indicates that "[the redirect] is never actually seen by the client"...

kairstenfay:
Ah, right. I think I'm confusing what "client" means here.

The final result of the `fetch` call in auspice (loadData.js) is a 200 response.
`fetch` quietly handles the redirect and by default does not inform the client of anything odd going on.
However, you can catch the redirect in two ways:
1. In your browser dev tools, go to the Network tab (I'm on Firefox, it might be something different in Chrome). If you filter by XHR requests, you can see that there is an initial `/getDataset` request that results in a 302 (not 301, I have to update auspice code). Then there is a second request immediately following that with an updated URL which returns a status code of 200. 
2. If you go into the `fetch` call in `loadDataset.js` in auspice and `console.log()` the response, you should see it in the browser first as a redirect and then as a final response. 
^ I believe this second bit to be true but I seem to have just broken my local auspice environment. I'll edit this comment if I'm wrong.

jameshadfield:
> fetch quietly handles the redirect and by default does not inform the client of anything odd going on. However, you can catch the redirect in your browser dev tools.
Go to the Network tab (I'm on Firefox, it might be something different in Chrome). If you filter by XHR requests, you can see that there is an initial /getDataset request that results in a 302. Then there is a second request immediately following that with an updated URL which returns a status code of 200.

Brilliant, that explains why my testing on the client never saw the redirect! 

> not 301, I have to update auspice code

If you're referring to https://github.com/nextstrain/auspice/pull/778/files#r318390786, then I don't think you can include 301 or 302 here, as "fetch quietly handles the redirect and by default does not inform the client of anything odd going on.".

jameshadfield:
Thanks @kairstenfay -- i've put these changes onto the auspice-v2 branch, and you can see the result by going to nextstrain-dev.herokuapp.com/flu/ and seeing it redirect  👍

PR #27, opened 2019-08-31
Private datasets for INRB

tsibley:
Details in the commit messages and our ongoing Slack convos.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ auspice/dist/
 node_modules/
 npm-debug.log
 scratch
+
+# local sessions
+/sessions/
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ python:
 install:
   - npm install
 script:
-  - npm run set-up
+  - npm run build
 after_success:
   - if [[ "$TRAVIS_PULL_REQUEST" == "false" && "$TRAVIS_BRANCH" == "master" ]]; then
       npm run redeploy-site;
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ This means that nextstrain.org, auspice.us, and your local auspice installation 
 ## Build nextstrain.org locally
 ### Build locally mirroring the deployed (live) website
 1. `npm install` (from the "nextstrain.org" directory)
-2. `npm run set-up`, which runs `./set-up.sh` to build both the static site & auspice with nextstrain.org customisations.
+2. `npm run build`, which runs `./build.sh` to build both the static site & auspice with nextstrain.org customisations.
 3. `npm run server` will then start a local instance, by default available at [localhost:5000](http://localhost:5000).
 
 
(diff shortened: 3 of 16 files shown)


tsibley:
This is now deployed to nextstrain.org and works in my testing. On the Heroku side, I added a few config vars (visible in the web console and with `heroku config`) and enabled the session affinity feature (`heroku feature:enable http-session-affinity`).

@trvrb:

> I imagine adding a source for "inrb-drc", "gisrs", "wa-doh" is sustainable, but it won't work to have a separate source per user. But I see how this is a clear first step.

I've been thinking of sources as per team or per organization. Dataset-level permissions can be layered on top of this.

It would also be possible to define a source as a prefix in a single bucket (so multiple sources share a bucket) instead of as an entire bucket. The point of the source abstraction is that this can all be done at the source level, and transparent to the rest of Auspice.

Also keep in mind that while we'll maintain a static list of sources in the short term, the source abstraction lets us very easily make it a dynamic, generated list of sources.

@jameshadfield:

> This PR implements a number of changes both related to -- and tangential to -- INRB authentication. @tsibley please make a PR (or multiple) into branch `auspice-v2` (which will become master when we release auspice v2) so we can discuss them pray

I'll make PR which merges `master` into `auspice-v2`. This will make the ultimate merge of `auspice-v2` back into `master` clean.

> 1. The auspice import mechanism has changed from a global install to a local npm install. This gave plenty of subtle bugs during my testing (albeit a while ago) -- see here [nextstrain/auspice#689](https://github.com/nextstrain/auspice/issues/689).

We can ultimately revert this if we want, but I think it is safe enough and comes along with benefits as I've described in that issue.

> 2. A lot of the (auspice-related) server code has been refactored, including code which @kairstenfay refactored last week. A PR into auspice-v2 will allow us to not diverge here.

Yep, I'm aware of her work there (we've chatted a bit) but haven't had time to review it myself. In the coming week or two, I plan to dig into it in detail and address any conflicts that now exist.

> 3. I don't have time now to fully comprehend all the AWS pieces involved and what this means for the future. A later PR into auspice-v2 will allow @tsibley to expand on his "future visions in mind with this work." and describe all the moving parts.

:+1: This definitely deserves more explication, which I'm happy to provide.

> 4. this PR makes heavy use of the "sources" concept in nextstrain.org / auspice client. Auspice v2 will do away with this -- see [nextstrain/auspice#783](https://github.com/nextstrain/auspice/issues/783) -- so a PR into auspice-v2 should accommodate that.

The two uses of "source" are very much unrelated conceptually. What the Auspice client means is basically completely independent of what the nextstrain.org server means. I am aware of the `_source` key's pending removal, and I don't see any conflicts here.

> 5. The design decisions related to users, the makeup of URLs etc can be discussed in a separate PR. It's not a problem to change it and maintain a redirect from the URL structure introduced in this PR.

:+1: I agree.

PR #35, opened 2019-10-03
new docs page describing how to clade label

alliblk:
This is a docs page that I created to explain how to add clade labeling to a nextstrain workflow. I made it because when I wanted to do this for ebola-drc, I had to scavenge a bit for examples and work out what the input files and rules should be like. I'm assuming that this is a capability that a decent chunk of users might want to have, and so I think it would be useful to have this tutorial in the docs.


jameshadfield:
Thanks @alliblk -- this content is a great addition to the docs 👍 

There's a couple of things that we'll want to change when auspice v2 is released (e.g. no need to toggle entropy panel to `nt` anymore), but that's separate to this PR.

On a side note, what steps did you take to successfully get this repo running on your laptop? I remember there were some issues in the DRC.

@milcisin can you please copy edit these since you're currently working on docs. Thanks!

alliblk:
@jameshadfield I didn't ever get the repo fully running to the point where I could see how the markdown would be compiled for the docs site. I just looked at how the markdown was getting shown in Atom's compiler, and worked with it until I thought it would look good.

milcisin:
@jameshadfield @alliblk gave it a copy edit to match the other Nextstrain docs. Looks good! I only made minor changes (like switching "I" to "we" and capitalizing Nextstrain). Happy to make/un-make changes.

jameshadfield:
Thanks so much @alliblk (and @misjailcisin for the copy editing!)

PR #36, opened 2019-10-04
Integrate changes on master into Auspice v2 branch

tsibley:
@jameshadfield I integrated the changes on `master` into `auspice-v2` by rebasing the latter onto the former. (Rebasing turned out to be easier and cleaner than merging `master` into `auspice-v2`.) The result is now pushed to `auspice-v2-rebased`, and everything seems to work in my testing. It would be really good to get your eyes on both the code and the functionality too.

Using [git range-diff](https://git-scm.com/docs/git-range-diff) is really useful for seeing the relative changes between the original version of the branch and the rebased version. (A range diff is essentially a diff of a diffs.) Unfortunately GitHub doesn't have a range-diff view, but you can produce one locally (after fetching) by running:

     git range-diff --color origin/{master,auspice-v2,auspice-v2-rebased}

I've also generated one and [nopasted it here](https://tsibley.net/paste/2019-10-04Ic_mLddM), which might be more convenient.


tsibley:
@kairstenfay and @joverlee521, you might also be interested in looking over this as some of your work was involved as well.

kairstenfay:
> @kairstenfay and @joverlee521, you might also be interested in looking over this as some of your work was involved as well.

It's been a long time since I looked at this code, and this PR contains many changes. I'm OK with it but didn't look in depth. Thanks for tagging me, Thomas.

jameshadfield:
Thanks @tsibley -- looking forward to diving into this. A few requests before I do:

1. Could you document how we should now run using a from-source version of auspice? The instructions [here](https://github.com/nextstrain/nextstrain.org/tree/auspice-v2-rebased#build-locally-using-a-custom-version-of-auspice) are out-of-date as the npm-dependency version of auspice will be imported by the server code  (e.g. [here](https://github.com/nextstrain/nextstrain.org/blob/auspice-v2-rebased/auspice/server/getDataset.js#L5)) rather than the desired global version of auspice. 

2. Please fix the linting errors (or at least most of them). Happy to have rule changes proposed if you believe the rules are the problem. 

3. Could you point me to the documentation / make issues to create the required documentation around (a) the AWS pieces we're using for private datasets & (b) the heroku settings necessary to make this work (see https://github.com/nextstrain/nextstrain.org/pull/27#pullrequestreview-282288888 & https://github.com/nextstrain/nextstrain.org/pull/27#issuecomment-526794827)

tsibley:
@jameshadfield I can do all three of those properly, but not immediately. The short answers:

1. `npm install <path to local auspice repo>`

2. Taking a quick look, it'll be some combination of fixing linting issues and adjusting the rules. Once this repo is lint-free, I would like to add an eslint test to Travis so branches with errors are flagged.

3. AWS setup or new Heroku settings are not currently documented except in meeting notes and whiteboard sketches. Some of the Heroku settings will be rendered unnecessary with session lifetime fixes.

jameshadfield:
Great, thanks @tsibley, I know seattle flu is very busy at the moment.

Here are my comments after I was able to test the branch today:

1. Is it preferable for `build.sh` to use `../node_modules/.bin/auspice` instead of `npx auspice`?

2. Unfortunately I can't build auspice as it currently stands. This is related to how auspice client customisations import libraries, which I remember was a source of quite a few bugs when I build that functionality. I'm sure there's better ways for auspice to handle this than the way I did it using [webpack alias'](https://github.com/nextstrain/auspice/blob/master/webpack.config.js#L17). These bugs led me to write documentation recommending auspice be installed globally. See also https://github.com/nextstrain/auspice/issues/689.

When building nextstrain.org after installing auspice using `npm install <path to local auspice repo>` I get the following errors:
```
npm run build auspice
...
ERROR in ../nextstrain.org/auspice/client/navbar.js
Module not found: Error: Can't resolve 'react' in '/Users/naboo/github/nextstrain/nextstrain.org/auspice/client'
 @ ../nextstrain.org/auspice/client/navbar.js 30:0-26 97:9-14 105:9-14 110:11-16 120:9-14 157:13-18 231:2-7 244:9-14 246:5-10 250:5-10 254:61-66 258:6-11 260:5-10 264:5-10 268:90-95
 @ ../nextstrain.org/auspice/client sync ^\.\/.*$
 @ ./src/util/extensions.js
 @ ./src/components/splash/index.js
 @ ./src/root.js
 @ ./src/index.js
 @ multi babel-polyfill ./src/index
[error] Webpack built with errors. Exiting.
```
These occur with both auspice 1.39.0 and 2.0.0-beta.2 (the latter is triggered from a different js file but it's the same error). Since it works on heroku and on your setup, it may be a MacOS thing.

Changing the import statement in `navbar.js` to 
```js
import React from "@libraries/react";
```
which uses auspice webpack alias' allows auspice to build for me, but it may cause errors on other setups (it has in similar situations in the past).

3. Running locally (after building auspice as above) gives a stream of identical errors. Have I missed something related to the authentication introduced in #27?
```
[session-file-store] will retry, error on last attempt: Error: ENOENT: no such file or directory, open 'sessions/yKvYHgTNzV4pyk8i4NkztGmwP2vD-cHO.json'
```

4. After auspice v2 is released (freeing up the nextstrain-dev server) and the heroku settings documented then I'll test this PR on that heroku server.

5. "Once this repo is lint-free, I would like to add an eslint test to Travis so branches with errors are flagged." Sounds great, thanks!

6. If you don't disagree strongly, then I'd much prefer multi-line ternaries to be written as 
```js
a = condition ?
    b :
    c;
```
as opposed to 
```js
a = condition
    ? b
    : c;
```

7. `git range-diff` is a great way to see how the commits on `auspice-v2` have changed when they were rebased onto master (i.e. this PR), but doesn't (unless i'm mistaken) address the commits of #27 which were going to be discussed in the merge of `master` into `auspice-v2`. That's no problem -- the current solution is working & being used. When the AWS documentation is complete i'll be able to give more thorough feedback, but it looks good for now.

tsibley:
@jameshadfield 

> 1. Is it preferable for build.sh to use ../node_modules/.bin/auspice instead of npx auspice?

Yes, so npx doesn't have to be installed.

> 2. Unfortunately I can't build this using a local-source version of auspice. … I'm not sure if you tried using `npm install <local path>`?

I didn't try `npm install <local path>` recently. I suspect it can be made to work, but without digging into it, don't know the problem.

> 3. Running locally (after building auspice as above) gives a stream of identical errors. Have I missed something related to the authentication introduced in #27?

No, those log messages are harmless. They arise because your browser remembers a session id that is not on your filesystem. I suspect there are configuration knobs to reduce their chattiness.

> 6. If you don't disagree strongly, then I'd much prefer multi-line ternaries to be written as …

I disagree—the `?` and `:` mark the two branches when they're leading—but not strongly. I'd be interested to learn why you prefer them that way.

jameshadfield:
Thanks for all the work here @tsibley (and @misjailcisin for the copy editing). I'll get auspice v2 merged in via the "upgrade-to-auspice-v2" branch (up soon) and make issues arising from the discussion in this PR.

PR #43, opened 2019-11-11
Track the Auspice dependency in package.json like a normal dep

tsibley:
Normal deps are preferrable because they:

- use the standard dependency tooling everything else does
- don't require additional manual resolution or installation
- don't require a build process which performs installation steps
- don't overwrite the global Auspice install on build, which is
  surprising at best and rude at worst

See the changes to the README for information on using local copies of
Auspice with `npm link`.

This commit reverts the commits "install auspice globally" (4e52c21) and
"allow auspice to be imported from a global npm installation" (f702df7).

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,8 @@ npm link <path to auspice repo>
 
 This uses symlinks both globally and within the local `node_modules/` directory to point the local Auspice dependency to your local Auspice source.
 
+> Using `npm install <path to auspice repo>` will _not_ work to use a local development copy, as dependencies are handled differently with `npm install` vs. `npm link` in this case.
+
 ---
 ## Nextstrain.org server
 `npm run server` runs `./server.js` which serves all the content on [nextstrain.org](https://nextstrain.org) & handles authentication.
--- a/README.md
+++ b/README.md
@@ -82,9 +82,11 @@ npx auspice develop --verbose --extend ./client/config.json --handlers ./server/
 
 Sometimes it is useful to change the version of Auspice that is used.
 Perhaps you're upgrading the version nextstrain.org uses or you want to test changes in a local copy of Auspice against nextstrain.org.
-First change the installed Auspice version to either a local development version with `npm install <path to auspice repo>` or a different released version with `npm install auspice@...` (replace `...` with the [semantic version specifier](https://docs.npmjs.com/about-semantic-versioning) you want).  
+
+If you're upgrading the version of Auspice used, install the new version with `npm install auspice@...`.
+Replace `...` with the [semantic version specifier](https://docs.npmjs.com/about-semantic-versioning) you want.
 This will change `package.json` and `package-lock.json` to reflect the new version you installed.
-If you are preparing to commit your Auspice version change, you'll want to include those changed files too.
+When preparing to commit your Auspice version change, you'll want to include those changed files too.
 Run `npx auspice --version` to check that (a) it's installed correctly and (b) which version you have.
 
 Then build and run the server as above with either:
--- a/README.md
+++ b/README.md
@@ -74,7 +74,8 @@ cd auspice
 npx auspice develop --verbose --extend ./client/config.json --handlers ./server/index.js
 ```
 
-> If you don't have [`npx`](https://www.npmjs.com/package/npx) installed, you can either install it first with `npm install --global npx` or just use `../node_modules/.bin/auspice` in its place.
+> If you're not familiar with [`npx`](https://www.npmjs.com/package/npx), it's a command to easily run programs installed by `npm`.
+> Running `npx auspice` above is equivalent to `../node_modules/.bin/auspice`.
 
 
 ### Using a different version of Auspice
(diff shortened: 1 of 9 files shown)


jameshadfield:
Thanks @tsibley for you work here. My thoughts are best summed up in the commit message of 4e52c210d60c8855e3153035c95cb71581065909

> Eventually it would be preferable to have auspice as a (normal) dependency, but this is currently problematic. While it _mostly_ worked, it wasn't possible to install auspice from source using `npm install <local path>` and have the client-customisations work. This is important for testing purposes. 

I'll test this PR out and make sure that everything works locally and get back to you if there are any shortcomings.

jameshadfield:
The following steps are problematic for me -- does it work on your system @tsibley? Note that the corresponding steps worked for a global auspice install (indeed they were part of the reason I had to switch back to a global install)

```
npm install <path to auspice src>
cd auspice
npx auspice develop --verbose --extend ./client/config.json --handlers ./server/index.js
```
Running the client results in:
```
ERROR in ./client/navbar.js
Module not found: Error: Can't resolve 'react' in '/Users/naboo/github/nextstrain/nextstrain.org/auspice/client'
 @ ./client/navbar.js 30:0-26 97:9-14 105:9-14 110:11-16 120:9-14 157:13-18 231:2-7 242:9-14 244:5-10 248:5-10 252:61-66 256:6-11 258:5-10 262:5-10 266:90-95
 @ ./client sync ^\.\/.*$
 @ /Users/naboo/github/nextstrain/auspice/src/util/extensions.js
 @ /Users/naboo/github/nextstrain/auspice/src/util/googleAnalytics.js
 @ /Users/naboo/github/nextstrain/auspice/src/index.js
 @ multi babel-polyfill webpack-hot-middleware/client ./src/index
```

The following also gives the same error:
```bash
npm install <path to auspice src>
./build.sh auspice
npm run server
```

README.md
```diff
@@ -65,34 +64,49 @@ npm run server
...
+This means that you will see a different splash page & the documentation will not work.
+It is not currently possible to run auspice in development mode & see the nextstrain.org splash page & docs.
+Read the section below on the nextstrain.org server for more context.
+
+```bash
+cd auspice
+npx auspice develop --verbose --extend ./client/config.json --handlers ./server/index.js
+```
+
+> If you don't have [`npx`](https://www.npmjs.com/package/npx) installed, you can either install it first with `npm install --global npx` or just use `../node_modules/.bin/auspice` in its place.
```
jameshadfield: I don't think this is necessary. `npx` comes with `npm` starting at version 5.2 (~2 years ago). Our install instructions are for `nodejs=10` which will bring in (at least) npm 5.8. If a user doesn't have `npx` then I think they should be advised to update node & npm.

tsibley: Oh, that's great!

tsibley:
> The following steps are problematic for me -- does it work on your system @tsibley? Note that the corresponding steps worked for a global auspice install (indeed they were part of the reason I had to switch back to a global install)

My understanding of `npm install <folder>` is that it isn't quite intended for this use case. Specifically, the dependency trees of the installed `<folder>` and the current project (where you're executing `npm install`) aren't merged as with a normal dep.

However, there is an npm feature for the use case we have here: npm link. It worked well in my testing. To use it, run `npm link <path to auspice>` from within the nextstrain.org repo. The nextstrain.org code will now use your local copy of Auspice. The `npx auspice develop` commands and normal build process will both work. You can revert the local link using `npm unlink <path to auspice>`. I documented `npm link` in the readme, but the readme is still a bit messy so you may have easily missed it.

Let me know how it works?

README.md
tsibley: Pushed an adjustment to this as a50fdf0.

jameshadfield:
> My understanding of `npm install <folder>` is that it isn't quite intended for this use case. 

Ahh ok, interesting. I was following your instructions added to the README in this PR where you specifically said to use `npm install <path>`. Please amend the README documentation to replace `npm install <path>` with the appropriate `npm link` commands, and specifically say that users _should not_ use `npm install <path>` as it does not work for auspice with customisations. I'll try this out and report back.

tsibley:
> I was following your instructions added to the README in this PR where you specifically said to use `npm install <path>`.

Whoops. That was held over from the previous README content and not removed after testing and adding the `npm link` section. I've removed it now.

package.json
```diff
@@ -21,6 +21,7 @@
   },
   "dependencies": {
     "argparse": "^1.0.10",
+    "auspice": "~2.0.3",
```
jameshadfield: @tsibley this is a change from previous behavior which specified the exact auspice version. Could you explain your reasoning here?

tsibley: Ah, yeah, I should have called this change out separately. Since we can be fairly confident in Auspice's adherence to semantic versioning—being it's controlled by us—it seemed useful to allow bug fix releases to happen ~automatically if nextstrain.org is being redeployed anyway.

That said, I'm also completely ok with using a pinned exact version and making any upgrade of Auspice, even a bug fix one, always be an explicit process.

jameshadfield: Sounds fine to me. Can easily pin if the need arises.

trvrb: This change makes sense to me.

trvrb:
Awesome! I think this is a nice improvement.

PR #45, opened 2019-11-19
authn: Switch session storage to Redis from (temporarily disabled) Memcached

tsibley:
Switching to Redis means we can avoid stunnel buildpacks and use a Redis
client which natively supports TLS connections.  There don't appear to
be any Memcached clients which support TLS connections.

One big caveat is that the Heroku Redis cert is self-signed, so we have
to disable peer verification.  This is documented by Heroku and a known
issue that doesn't seem to have a fix on the horizon.  I'm less than
impressed.  The risk is mitigated somewhat by both our dyno and Redis
host being within the broader AWS network.

In production, REDIS_URL is defined by the official Heroku Redis add-on.
In development, a file-based session store is still used for persistence
across local server restarts.  Redis support can be tested locally by
grabbing the Redis URL from the Heroku config.

This is deployed on https://nextstrain-dev.herokuapp.com/.


trvrb:
Thank you Tom! This works well for me.

jameshadfield:
Note: https://nextstrain-dev.herokuapp.com/ is now being deployed from another branch. Hopefully not a problem, as the implementation of this PR has been tested and works well.

tsibley:
> I can understand the desire to keep code comments brief, but we should ensure things such as the RedisStore's prefix are documented internally.

Hmm, documented for what purpose? Nothing else I can think of needs to care about the prefix other than that line of code.

> Only change to implementation is to remove the now unused `connect-memjs` dep

Done.

jameshadfield:
> Nothing else I can think of needs to care about the prefix other than that line of code.

Current and future developers. Hence the request to have these things documented internally.

tsibley:
> I can understand the desire to keep code comments brief

FWIW, I don't actually think code comments should be brief. :-)

> I'm presuming the heroku add-on sets the url as `redis:` which we can change to `rediss:` (and bump the port) as we want to use TLS?

Yep, [as I commented](https://github.com/nextstrain/nextstrain.org/pull/45/commits/f90231f6c3398a17771a8f0feed7fd7ca85bfc16#diff-85d3a43a3025d8c8b7741677a95ef441R100-R101).

tsibley:
> Current and future developers. Hence the request to have these things documented internally.

I'm not sure I understand what you're asking to be documented, but I pushed a commit adding a comment about why a prefix is used. Is that what you're looking for?

jameshadfield:
Absolutely. Comments such as these are incredibly helpful to others.

tsibley:
Great! For the future, a review note like the following:

> Please add a comment on why `RedisStore`'s `prefix` is used.

would be much more clear and helpful to me than

> [W]e should ensure things such as the `RedisStore`'s `prefix` are documented internally.

especially given the recent creation of https://github.com/nextstrain/documentation for internal documentation.

PR #46, opened 2019-11-21
Customisable auspice splash pages

jameshadfield:
This commit sets up a prototype implementation of a custom auspice splash
page, by adding an auspice built-time extension for the splash page.
No changes to the auspice source code were needed, this is all done as
a build-time extension, which I think is really cool.

The auspice splash page is accessed when the nextstrain.org router decides
a path should be handled by auspice, but auspice's request for the dataset
returns 500 or 204 (i.e. it doesn't exist). This commonly occurs in the
following cases (with this response code):
* nextstrain.org/doesnotexist (500)
* nextstrain.org/staging (204)
* nextstrain.org/<group_name> (204)
* nextstrain.org/community/<org_name> (500)
* nextstrain.org/community/<org_name>/<repo_name> (behavior has changed in this PR, it is now 204 if there is not an appropriate dataset in the repo)

Here I implement a custom nextstrain.org auspice splash page which makes
a API call to `/charon/getSourceInfo` (and implement the relevant handlers).
The response of this API call defines fields to be rendered, such as title,
byline etc (the exact structure of this is to be finalised, but note that
we can iterate on this after it is live but before publicising).

The intent of this (not yet implemented) is to allow different sources (e.g.
community GitHub repo, "nextstrain group") to define a JSON specifying this
which will be displayed via this splash page.

Screenshots:
image
image
image
image

(and there is a default implementation for groups too, no screenshot included here)

This PR isn't intended for merge currently. There are still some parts of the code to improve, and I'm hoping to get feedback about this general direction. But I think it's a step in the right direction, and the fact that this requires zero modifications to auspice means we deploy this to nextstrain.org and implement additional functionality in future commits.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -51,6 +51,7 @@ rules:
   react-hooks/rules-of-hooks: "error"
   react-hooks/exhaustive-deps: "warn"
   react/no-danger: off # gatsby uses this a lot
+  no-use-before-define: ["error", { "functions": false }]
 parserOptions:
   ecmaVersion: 6
   sourceType: module
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@ If you're installing from a local development copy of Auspice's source, you can 
 
 ```bash
 npm link <path to auspice repo>
+npm link <path to auspice repo>/node_modules/react
 ```
 
 This uses symlinks both globally and within the local `node_modules/` directory to point the local Auspice dependency to your local Auspice source.
--- a/auspice/client/config.json
+++ b/auspice/client/config.json
@@ -9,6 +9,7 @@
     "unselectedBackground": "#888"
   },
   "navbarComponent": "navbar.js",
+  "splashComponent": "splash.js",
   "browserTitle": "nextstrain",
   "googleAnalyticsKey": "UA-92687617-1"
 }
--- a/auspice/client/splash.js
+++ b/auspice/client/splash.js
@@ -1,0 +1,187 @@
+import React, {useState, useEffect} from "react"; // eslint-disable-line
+import styled from 'styled-components'; // eslint-disable-line
+import NavBar from "./navbar";
+
+/* WHy is this here? See https://reactjs.org/warnings/invalid-hook-call-warning.html
+Running locally, I had multiple versions of react running.
+This could be due to `npm link`.
+See https://github.com/nextstrain/auspice/issues/689
+Fixed by running `npm link ../auspice/node_modules/react`
+Need to test via a npm install from the registry.
+THIS SHOULD BE REMOVED BEFORE MERGING INTO MASTER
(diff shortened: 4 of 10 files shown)


trvrb:
Awesome work James. Of course, I'm a fan of the general direction of enabling group level splash pages. I do wonder about the logic behind having this injected into Auspice as a build time extension, vs having nextstrain.org server catch these paths and field its own page. The latter seems cleaner to me, but I don't fully understand how all this is put together. I would wonder about @tsibley's thoughts here as well.

jameshadfield:
I agree. This is exactly why it's so good to implement this as an auspice build-time customisation. As we add more capability to nextstrain.org, it's trivial to shift this splash page to a different part of nextstrain.org as appropriate and implement a simple auspice splash page which redirects to that. This is an implementation detail within nextstrain.org and involves no modifications of the auspice source code.

trvrb:
Got it. Thanks James. Then I fully approve of direction.

tsibley:
> This is exactly why it's so good to implement this as an auspice build-time customisation. As we add more capability to nextstrain.org, it's trivial to shift this splash page to a different part of nextstrain.org as appropriate and implement a simple auspice splash page which redirects to that. This becomes an implementation detail contained within nextstrain.org and involves no modifications of the auspice source code.

It'd be much simpler to have nextstrain.org field the pages directly rather than round-trip through Auspice. The described setup above looks like:

1. Request hits nextstrain.org server
2. nextstrain.org server routes it to Auspice client
3. Auspice client returns custom splash page
4. Custom splash page renders components provided by nextstrain.org repo

whereas having nextstrain.org field such pages directly looks like:

1. Request hits nextstrain.org
2. nextstrain.org responds with the appropriate page

This also aligns well with nextstrain.org's server eventually not falling back to Auspice for all requests, which will be important once we have more Nextstrain UI. I realize that fallback exists because of the history of Auspice vs. nextstrain.org, but it no longer makes as much sense.

jameshadfield:
@tsibley glad that you agree with me. The nextstrain.org server routing capabilities you speak of don't exist, but once they do (perhaps this week) then the plan is indeed for the the UI implemented here to indeed move out of auspice.

tsibley:
@jameshadfield Great. I think I misunderstood what you meant by "implement a simple auspice splash page which redirects to that", but glad we're on the same page after all.

jameshadfield:
I'm going to leave this PR open for another week to see if we can get the nextstrain.org website to a state where this design is implemented there rather than as an auspice customisation, as that would render this PR unnecessary.

jameshadfield:
Merged in 3b9f688a4a28e227315dd673ee802011cc51b644. In the future this capability should be refactored to be part of the gatsby section, and we can expand the functionality there.

PR #47, opened 2019-11-25
Stream v2 JSONs

jameshadfield:
By streaming the v2 JSONs to the client we reduce the memory needed by the server. This is made possible by the design of the v2 JSONs and auspice v2 which doesn't require any server-side modifications to the JSON data. 

See https://github.com/nextstrain/auspice/pull/825 for the same concept implemented in auspice.

Testing shows the following memory improvements (averages of a few runs & rounded to nearest 5Mb):
Baseline server memory usage (no requests): ~60Mb memory (unchanged). 
H3N2 12y dataset, minified (2.6Mb JSON): 75Mb -> 60Mb memory.
H3N2 12y dataset, indented (75Mb JSON): 290Mb -> 65Mb memory.

### DIFF
--- a/auspice/server/getDataset.js
+++ b/auspice/server/getDataset.js
@@ -3,6 +3,7 @@ const utils = require("./utils");
 const helpers = require("./getDatasetHelpers");
 const {NoDatasetPathError} = require("./exceptions");
 const auspice = require("auspice");
+const request = require('request');
 
 /**
  *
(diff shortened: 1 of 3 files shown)


auspice/server/getDataset.js
```diff
@@ -18,26 +19,45 @@ const requestCertainFileType = async (res, req, additional, query) => {
...
+ * If this request 404s, then we attempt to fetch a v1-version
+ * of the dataset (i.e. meta + tree JSONs) and convert it
+ * to v2 format for the response.
+ */
+const requestMainDataset = async (res, fetchUrls) => {
+  /* try to stream the (v2+) dataset JSON as the response */
+  const req = request
+    .get(fetchUrls.main)
+    .on("response", (response) => {
+      if (response.statusCode === 200) {
```
tsibley: I think it'd be more robust to check `if (response.ok)` instead since 200 is just one possible success code.

auspice/server/getDataset.js
```diff
@@ -18,26 +19,45 @@ const requestCertainFileType = async (res, req, additional, query) => {
...
+        utils.verbose(`Successfully streaming ${fetchUrls.main}.`);
+        req.pipe(res);
```
tsibley: The way `req`'s initialization closes over `req` itself is a little weird to me.

I think the `getDataset`, `requestMainDataset`, and `requestV1Dataset`  functions might read a bit cleaner if the `request*` functions returned values instead of being passed the `res`ponse object from `getDataset`. This way the response object stays within the `getDataset` function and all possible responses are readable there.

tsibley:
@jameshadfield I think this change would be great to get into production. Do you want to follow up this week on the changes requested above?

auspice/server/getDataset.js
jameshadfield: The `response` object here has no `ok` property, so I'm going to leave as is. Please feel free to improve this if you have time.

auspice/server/getDataset.js
jameshadfield: Agreed. I've done this for `requestV1Dataset` but the others are a bit trickier, as I couldn't work out how to implement streaming here without closing over the `request` object. Very happy for you to create an issue detailing where we can improve here.

Issue #50, opened 2019-12-09
Add groups url prefix


jameshadfield:
A group name `abc` should be accessible via the URL nextstrain.org/groups/abc. 

We currently have one active group. Unsure whether to redirect this to the "new" URL or not, but this can be handled separately.

tsibley:
Redirecting should be possible without breaking the existing URLs at all, and I think will be most consistent and leave less historical baggage for the future.

jameshadfield:
Closed by #59

PR #59, opened 2019-12-11
Add "groups" prefix to Nextstrain groups URLs

kairstenfay:
Solves https://github.com/nextstrain/nextstrain.org/issues/50

Add a "/groups" prefix to Nextstrain groups URLs, like for `inrb-drc`,
to distinguish from and avoid namespace collisions with core Nextstrain
data sources.
Redirect old "inrb-drc" URLs to the new `/groups` endpoints.

### DIFF
--- a/auspice/server/getDataset.js
+++ b/auspice/server/getDataset.js
@@ -86,6 +86,16 @@ const getDataset = async (req, res) => {
   if (!query.prefix) {
     return helpers.handleError(res, `getDataset request must define a prefix`);
   }
+
+  /*
+   * "inrb-drc" was the first of the Nextstrain groups. Groups now live at
+   * `nextstrain.org/groups`, but we want to support old URLs for INRB DRC by
+   * redirecting requests from "/inrb-drc" to "/groups/inrb-drc".
+   */
+  if (query.prefix.startsWith('/inrb-drc')) {
+    return res.redirect("getDataset?prefix=/groups" + query.prefix);
+  }
+
   utils.log(`Getting (nextstrain) datasets for: ${req.url.split('?')[1]}`);
 
   // construct fetch URL
--- a/auspice/server/getDatasetHelpers.js
+++ b/auspice/server/getDatasetHelpers.js
@@ -36,9 +36,21 @@ const splitPrefixIntoParts = (prefix) => {
   /* The first part of the prefix is an optional source name.  The rest is the
    * source path parts.
    */
-  const sourceName = sources.has(prefixParts[0])
-    ? prefixParts.shift()
-    : "core";
+  let sourceName;
+
+  switch (prefixParts[0]) {
+    case "community":
+    case "staging":
+      sourceName = prefixParts.shift();
+      break;
+    case "groups":
+      prefixParts.shift();
+      sourceName = prefixParts.shift();
+      break;
+    default:
+      sourceName = "core";
+      break;
+  }
 
   const isNarrative = prefixParts[0] === "narratives";
 
@@ -72,8 +84,15 @@ const splitPrefixIntoParts = (prefix) => {
 const joinPartsIntoPrefix = ({source, prefixParts, isNarrative = false}) => {
   const leadingParts = [];
 
-  if (source.name !== "core") {
-    leadingParts.push(source.name);
+  switch (source.name) {
+    case "core":
+      break;
+    case "community":
+    case "staging":
+      leadingParts.push(source.name);
+      break;
+    default:
+      leadingParts.push("groups", source.name);
   }
 
   if (isNarrative) {
--- a/auspice/server/getDatasetHelpers.js
+++ b/auspice/server/getDatasetHelpers.js
@@ -210,4 +210,3 @@ module.exports = {
   unauthorized,
   parsePrefix
 };
-
--- a/server.js
+++ b/server.js
@@ -125,18 +125,14 @@ const auspicePaths = [
   "/narratives/*",
   "/staging",
   "/staging/*",
+  "/groups",
+  "/groups/*",
 
   /* Auspice gets specific /community paths so it can show an index of datasets
    * and narratives, but Gatsby gets top-level /community.
    */
   "/community/:user/:repo",
   "/community/:user/:repo/*",
-
-  /* XXX TODO: Remove the following two lines once PR #59 is merged and replace
-   * with an appropriate /groups/… path pattern.
-   */
-  "/inrb-drc",
-  "/inrb-drc/*",
 ];
 
 app.route(auspicePaths).get((req, res) => {
@@ -144,6 +140,10 @@ app.route(auspicePaths).get((req, res) => {
   res.sendFile(auspiceAssetPath("index.html"));
 });
 
+/* handle redirects for inrb-drc (first of the Nextstrain groups) */
+app.get("/inrb-drc*", (req, res) => {
+  res.redirect(`/groups${req.originalUrl}`);
+});
 
 /* Everything else hits our Gatsby app's entrypoint.
  */
--- a/server.js
+++ b/server.js
@@ -15,7 +15,7 @@ const version = utils.getGitHash();
 const nextstrainAbout = `
   Nextstrain is an open-source project to harness the scientific and public health potential
   of pathogen genome data. This is the server behind nextstrain.org.
-  It delivers the static content (https://github.com/nextstrain/static) as well as the interactive 
+  It delivers the static content (https://github.com/nextstrain/static) as well as the interactive
   visualisation app auspice (https://github.com/nextstrain/auspice), with customisations.
 `;
 const parser = new argparse.ArgumentParser({
--- a/auspice/server/getNarrative.js
+++ b/auspice/server/getNarrative.js
@@ -12,6 +12,15 @@ const getNarrative = async (req, res) => {
     return helpers.handleError(res, "No prefix in narrative URL query");
   }
 
+  /*
+   * "inrb-drc" was the first of the Nextstrain groups. Groups now live at
+   * `nextstrain.org/groups`, but we want to support old URLs for INRB DRC by
+   * redirecting requests from "/inrb-drc" to "/groups/inrb-drc".
+   */
+  if (prefix.startsWith('/inrb-drc')) {
+    return res.redirect("getNarrative?prefix=/groups" + prefix);
+  }
+
   const {source, prefixParts} = helpers.splitPrefixIntoParts(prefix);
 
   // Authorization
--- a/package-lock.json
+++ b/package-lock.json
@@ -5658,7 +5658,8 @@
         },
         "ansi-regex": {
           "version": "2.1.1",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "aproba": {
           "version": "1.2.0",
@@ -5676,11 +5677,13 @@
         },
         "balanced-match": {
           "version": "1.0.0",
-          "bundled": true
+          "bundled": true,
+          "optional": true
         },
         "brace-expansion": {
           "version": "1.1.11",
           "bundled": true,
+          "optional": true,
           "requires": {
             "balanced-match": "^1.0.0",
             "concat-map": "0.0.1"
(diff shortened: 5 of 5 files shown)


tsibley:
@kairstenfay 

> The only thing I'm unsure about is how to break Nextstrain group URLs that don't specify "groups".

Is this still true after we talked at the end of day yesterday?

auspice/server/getAvailable.js
```diff
@@ -7,7 +7,7 @@ const getAvailable = async (req, res) => {
   const prefix = queryString.parse(req.url.split('?')[1]).prefix || "";
   utils.verbose(`getAvailable prefix: "${prefix}"`);
 
-  const {source} = splitPrefixIntoParts(prefix);
+  const {source, isGroup} = splitPrefixIntoParts(prefix);
```
tsibley: Instead of threading through a boolean (as another variable that needs preservation and passing around), I think `splitPrefixIntoParts` and `joinPartsIntoPrefix` can do what they need via introspection of the `source` instance.

auspice/server/getDatasetHelpers.js
```diff
@@ -36,6 +36,12 @@ const splitPrefixIntoParts = (prefix) => {
...
    */
+
+  const isGroup = prefixParts[0] === "groups";
+  if (isGroup) {
+    prefixParts.shift();
+  }
+
   const sourceName = sources.has(prefixParts[0])
     ? prefixParts.shift()
     : "core";
```
tsibley: As a reminder for what we talked about yesterday, the `isGroup` conditional and `sources.has` conditional need to be combined in order to "break" the old paths in the way you want. Happy to chat more today.

kairstenfay:
I implemented the changes you requested, @tsibley. 
I'm still encountering one bug that I'm not sure how to fix: I get a 404 on narratives that are part of a group. I'm unsure what's causing that problem.

Also, is there somewhere I can find "staging" URLs to test?

jameshadfield:
> Also, is there somewhere I can find "staging" URLs to test?

https://nextstrain.org/staging/

kairstenfay:
> > Also, is there somewhere I can find "staging" URLs to test?
> 
> https://nextstrain.org/staging/

Thanks. Staging URLs (including narratives) work.

jameshadfield:
> Groups Narratives are still suffering from the aforementioned bug.

This is probably to do with how the datasets are defined within the narrative markdown files within that group. The markdown (YAML) frontmatter defines a dataset URL, which results in a `charon/getDataset` call from auspice. You'll probably have to allow the server to react appropriately to these.

auspice/server/getDatasetHelpers.js
```diff
@@ -36,9 +36,26 @@ const splitPrefixIntoParts = (prefix) => {
   /* The first part of the prefix is an optional source name.  The rest is the
    * source path parts.
    */
-  const sourceName = sources.has(prefixParts[0])
-    ? prefixParts.shift()
-    : "core";
+  let sourceName;
+
+  switch (prefixParts[0]) {
+    case "core":
```
tsibley: This makes URLs like `/core/zika` valid, which I don't think we actually want to do.

auspice/server/getDatasetHelpers.js
```diff
@@ -71,9 +88,22 @@ const splitPrefixIntoParts = (prefix) => {
  */
 const joinPartsIntoPrefix = ({source, prefixParts, isNarrative = false}) => {
   const leadingParts = [];
+  console.log("source", source);
+  console.log("prefixParts", prefixParts);
```
tsibley: These look like debugging leftovers that should be removed.

auspice/server/getDatasetHelpers.js
```diff
@@ -143,7 +173,7 @@ const guessTreeName = (prefixParts) => {
  */
 const parsePrefix = (prefix, otherQueries) => {
   const fetchUrls = {};
-  let {source, prefixParts} = splitPrefixIntoParts(prefix);
+  let {source, prefixParts, isGroup} = splitPrefixIntoParts(prefix);
```
tsibley: This looks like a leftover from a previous version of the PR.

auspice/server/getDatasetHelpers.js
```diff
@@ -172,7 +202,7 @@ const parsePrefix = (prefix, otherQueries) => {
   }
 
   // The URL to be displayed in Auspice, tweaked below if necessary
-  let auspiceDisplayUrl = joinPartsIntoPrefix({source, prefixParts});
+  let auspiceDisplayUrl = joinPartsIntoPrefix({source, prefixParts, isGroup});
```
tsibley: Ditto.

auspice/server/getDatasetHelpers.js
```diff
@@ -36,9 +36,26 @@ const splitPrefixIntoParts = (prefix) => {
...
+  switch (prefixParts[0]) {
+    case "core":
+      sourceName = prefixParts.shift();
+      break;
+    case "community":
+      sourceName = prefixParts.shift();
+      break;
+    case "staging":
+      sourceName = prefixParts.shift();
+      break;
```
tsibley: In case you don't know, when two `case` statements in a `switch` should do the same thing, you can combine them. For example:

```js
case "community":
case "staging":
  sourceName = prefixParts.shift();
  break;
```

This is because `case` statements that don't include a `break` statement will "fall through" to the next `case`.

auspice/server/getDatasetHelpers.js
```diff
@@ -71,9 +88,22 @@ const splitPrefixIntoParts = (prefix) => {
...
+    case "community":
+      leadingParts.push(source.name);
+      break;
+    case "staging":
+      leadingParts.push(source.name);
+      break;
+    default:
+      leadingParts.push("groups");
+      leadingParts.push(source.name);
```
tsibley: FYI, `Array.push` accepts multiple arguments, e.g. `leadingParts.push("groups", source.name)`.

auspice/server/getDatasetHelpers.js
kairstenfay: I didn't know that. Thanks!

kairstenfay:
> > Groups Narratives are still suffering from the aforementioned bug.
> 
> This is probably to do with how the datasets are defined within the narrative markdown files within that group. The markdown (YAML) frontmatter defines a dataset URL, which results in a `charon/getDataset` call from auspice. 

@jameshadfield Where is this YAML file you're referring to? 

> You'll probably have to allow the server to react appropriately to these.

Do you mean I shouldn't put groups narratives under /groups?

tsibley:
At the top of a narrative Markdown file, there is a block of embedded YAML. This YAML provides some structured data for the narrative and is stripped out before the Markdown parser sees the rest of the file. (Some Markdown parsers also support reading YAML front matter blocks.)

For an example, look at the top of https://raw.githubusercontent.com/nextstrain/narratives/master/intro-to-narratives.md. For groups, those narrative files are in the S3 bucket.

The term "front matter" comes from the publishing industry, where it's used to describe the informational pages in books before page 1.

Narratives should definitely stay under `/groups/`.

jameshadfield:
@kairstenfay always happy to have a video call to help you with any of these complexities

kairstenfay:
The Nextstrain groups narratives bug is fixed thanks to a change in the nextstrain.org auspice server's `getDatset` functionality.

auspice/server/getDataset.js
```diff
@@ -40,6 +40,20 @@ const requestMainDataset = async (res, req, fetchUrls) => {
...
+ * "inrb-drc" was the first of the Nextstrain groups. Groups now live at
+ * `nextstrain.org/groups`, but we want to support old URLs for INRB DRC by
+ * redirecting requests from "/inrb-drc" to "/groups/inrb-drc".
+ *
+ * @param {Response} res
+ * @param {Object} query
+ */
+const redirectInrbDrc = (res, query) => {
+  if (query.prefix.startsWith('/inrb-drc')) {
+    res.redirect("getDataset?prefix=/groups" + query.prefix);
```
jameshadfield: Nice work fixing this bug @kairstenfay. As a heads up for future nextstrain work, a solution such as this would break the sourcing of any sidecar files, which are defined by additional parts of the query (see https://nextstrain.github.io/auspice/server/api#charon-getdataset). As this group doesn't use any sidecar files (e.g. tip-frequencies) I would expect this solution to be perfectly acceptable.

tsibley: The `query.prefix` value needs to be escaped before being added to the URL. While you can do this with `encodeURIComponent()`, using `"getDataset?" + queryString.stringify(query)` will both do the escaping and address the sidecar issue James mentioned (since it would include `query.type` too). Note that `query` originally came from `queryString.parse(…)`; `stringify` is the inverse operation.

auspice/server/getDataset.js
```diff
@@ -58,6 +72,9 @@ const getDataset = async (req, res) => {
   if (!query.prefix) {
     return helpers.handleError(res, `getDataset request must define a prefix`);
   }
+
+  redirectInrbDrc(res, query);
```
tsibley: If a redirect is sent in this function call, then `getDataset` should return instead of continuing to process the request. The simplest way to do that in this case is going to be to inline the function into the callsite and add a `return` statement if the redirect needs to happen.

You can see the evidence of the continued request processing in the server logs. For example, requesting `/getDataset?prefix=/inrb-drc/ebola/2019-12-09` results in logs for both paths:

```
Getting (nextstrain) datasets for: prefix=/inrb-drc/ebola/2019-12-09
[verbose]       Guessing tree name -- this should be improved
[verbose]       Dataset:  { fetchUrls:
   { main: 'http://data.nextstrain.org/inrb-drc_ebola_2019-12-09.json',
     tree:
      'http://data.nextstrain.org/inrb-drc_ebola_2019-12-09_tree.json',
     meta:
      'http://data.nextstrain.org/inrb-drc_ebola_2019-12-09_meta.json' },
  auspiceDisplayUrl: 'inrb-drc/ebola/2019-12-09',
  treeName: undefined,
  secondTreeName: undefined,
  source: CoreSource {} }
[verbose]       Fetching http://data.nextstrain.org/inrb-drc_ebola_2019-12-09.json

Getting (nextstrain) datasets for: prefix=/groups/inrb-drc/ebola/2019-12-09
[verbose]       Can't compare against available datasets as there are none!
[verbose]       Guessing tree name -- this should be improved
[verbose]       Dataset:  { fetchUrls:
   { main:
      'https://nextstrain-inrb.s3.amazonaws.com/ebola_2019-12-09.json?…',
     tree:
      'https://nextstrain-inrb.s3.amazonaws.com/ebola_2019-12-09_tree.json?…',
     meta:
      'https://nextstrain-inrb.s3.amazonaws.com/ebola_2019-12-09_meta.json?…' },
  auspiceDisplayUrl: 'groups/inrb-drc/ebola/2019-12-09',
  treeName: undefined,
  secondTreeName: undefined,
  source: InrbDrcSource {} }
[verbose]       Fetching https://nextstrain-inrb.s3.amazonaws.com/ebola_2019-12-09.json?…
```

server.js
```diff
@@ -58,6 +58,11 @@ app.get("/auspice*", (req, res) => {
   res.redirect('https://nextstrain.github.io/auspice/');
 });
 
+/* handle redirects for inrb-drc (first of the Nextstrain groups) */
+app.get("/inrb-drc*", (req, res) => {
+  res.redirect(`/groups${req.path}`);
```
tsibley: This redirect doesn't preserve query parameters, which are used to define parts of the Auspice visual state. For example, going to `/inrb-drc/ebola/2019-12-09?c=num_date` redirects to `/groups/inrb-drc/ebola/2019-12-09`.

kairstenfay: So _that's_ the difference between `req.path` and `req.originalUrl`. Got it. Thanks.

auspice/server/getDataset.js
kairstenfay: How are you seeing verbose logging output? I've tried setting a few `DEBUG` type variables to verbose as well as using a `--verbose` flag on `npm run start` (similar to what we use for auspice), but to no avail.

kairstenfay:
@tsibley I have implemented your requested changes. 

> You can see the evidence of the continued request processing in the server logs. For example, requesting /getDataset?prefix=/inrb-drc/ebola/2019-12-09 results in logs for both paths:

As I mentioned above, I am unsure how to see the verbose logging output you are seeing. However, I don't see the lookup for both paths (initial and rediretced).

auspice/server/getDataset.js
jameshadfield: @kairstenfay `node server.js --verbose` or `npm run start -- --verbose`

joverlee521:
@jameshadfield I believe the problems with the charon API redirects are now fixed. 

I've also rebased this PR onto master, it is now ready for re-review.

Issue #52, opened 2019-12-09
User page to replace stub


tsibley:
When logged in, clicking on your username in the top right nav leads to a very bare bones page that includes the list of groups you're in and a logout link. This was due to time constraints during initial implementation. We'd like to replace this with a user page that fits in with the rest of the site. To start with, it doesn't have to have any more content then it currently does. I think the URL should change from `/whoami` to `/users/<name>`. (The `/whoami` API endpoint still needs to stay for the client.)

jameshadfield:
Closed by #71

PR #63, opened 2019-12-14
WIP: User page to replace stub

kairstenfay:
I have started a new "Users" page for logged in users of Nexstrain.org.

This PR will eventually solve https://github.com/nextstrain/nextstrain.org/issues/52. 
It does not yet redirect from `/whoami` to `/users/<name>`, as Tom suggested. Instead, it redirects to `/users/user?=<name>&groups=<groups>` as an intermediary solution (e.g. the one I could get working with limited knowledge of Gatsby).

My guess is that an endpoint like `/users/<name>` will require the use of [Reach router](https://www.gatsbyjs.org/docs/reach-router-and-gatsby/). I want to adopt this solution, as the current query-string parameterized solution is very buggy.


jameshadfield:
> My guess is that an endpoint like /users/<name> will require the use of Reach router. I want to adopt this solution, as the current query-string parameterized solution is very buggy.

That's my understanding too, based on https://www.gatsbyjs.org/docs/client-only-routes-and-user-authentication/#handling-client-only-routes-with-gatsby. I think this will be a lot easier were gatsby to have been updated to the latest version before attempting this work. Note that `@reach/router` was called "react router" back when this version of gatsby was released.

However I don't see why this will automatically get rid of bugs -- it's possible to implement the current approach without bugs. 

@kairstenfay would you like a review on anything specifically here?

joverlee521:
I've updated the Users page to fetch user data from the `/whoami` API endpoint so we don't need the user & group query parameters in the redirect URL. 

It now redirects from `/whoami` to `/users`. When I try to redirect to `/users/<username>`, it goes to the auspice 404 page. This might be because this PR doesn't include https://github.com/nextstrain/nextstrain.org/pull/64, or due to my limited knowledge of Gatsby.

jameshadfield:
@joverlee521 can you rebase onto master before review so that we can test the behavior with #64 please. Happy to review at any stage.

authn.js
```diff
@@ -212,10 +212,7 @@ function setup(app) {
   // Provide the client-side app with info about the current user
   app.route("/whoami").get((req, res) => {
```
jameshadfield: Can this be done simply via `.get((req, res) => res.redirect('/users'))`? (I haven't tested, but think this should work)

joverlee521:
Rebased on to master. 

Now redirects from `/whoami` to `/users/<username>` if a user is signed in. (turns out I had to update `gatsby-node.js` to create the users page during build time so it doesn't go to 404 page). 

The only thing is the splash pages loads for a split second before the user page loads. I'm assuming this is another gastby issue, haven't really figured it out yet...

jameshadfield:
Sorry @joverlee521, this shouldn't have been closed and I can't reopen for some reason!

image

PR #64, opened 2019-12-14
server: Invert routing order so Gatsby is the default app, not Auspice

### LINKED ISSUE #49: Add server routing capabilities.
### Current situation

The nextstrain.org website consists of two "apps" or sections:
(i) The splash page & all the documentation. Located in https://github.com/nextstrain/nextstrain.org/tree/master/static-site and built using gatsby via `npm run build -- static` or similar. 
(ii) auspice. Located in https://github.com/nextstrain/nextstrain.org/tree/master/auspice & built using `npm run build -- auspice` or similar.

The server contains some extremely simple logic for deciding which "app" to serve to the client here: https://github.com/nextstrain/nextstrain.org/blob/master/server.js#L66.

### Where are we going?

We want a router which can respond to requests and deliver the appropriate "page" or "app". I'm not sure what forms each of these "pages" or "apps" should take, and likely they'll all be different, but a general guide for how interactions should be handled between the server & each app would be beneficial at this stage. 

The capabilities that this new router should allow us to implement are:
(a) Decoupling the documentation from the splash page
(b) requests for user information e.g. `/whoami` 
(c) Requests to the "main group page" (e.g. nextstrain.org/groups/group_name) can be interpreted as such by the server and a custom page/app sent, rather than making the client load auspice.
(e) A new "app" can be implemented allowing manipulation (upload/download) of UI can be 
(d) future capabilities, such as cloud augur etc
(e) Dataset redirection (e.g. /flu -> /flu/seasonal/h3n2...) could be done before auspice is ever sent to the client. 

_(Note that this issue isn't about the implementation of each of these, but making it possible to implement them)_

tsibley:
This lets us route dataset and narrative paths to Auspice (for core,
staging, community, and groups), but handle cases like group index pages
either in Gatsby or with a separate React app.

Most of this change is laying a foundation to make future changes
easier.  The only intended user-visible change is that Gatsby's 404 page
is now shown instead of Auspice's.  I think this is probably a slight
improvement—the splash page is a nicer error page than a dataset
index—but there's definitely still room to improve the 404 UX in the
future.

Resolves #49.


server.js
```diff
@@ -42,69 +54,96 @@ app.set('port', process.env.PORT || 5000);
...
+  "/flu",
+  "/lassa",
+  "/measles",
+  "/mers",
+  "/mumps",
+  "/seattleflu", // XXX TODO: Remove once PR #60 is merged.
+  "/tb",
+  "/WNV",
+  "/yellow-fever",
+  "/ZEBOV",
```
jameshadfield: ~As far as i'm aware this isn't a nextstrain.org dataset. https://nextstrain.org/ZEBOV doesn't work (although why the client crashes rather than displaying a 404 is a different matter).~ This dataset shouldn't be on nextstrain-data & can be removed from this list.

server.js
```diff
@@ -42,69 +54,96 @@ app.set('port', process.env.PORT || 5000);
...
+  .all((req, res) => res.status(404).end());
+
+
+/* Dataset and narrative paths hit Auspice's entrypoint.
+ */
+const coreBuilds = [
+  "/dengue",
+  "/ebola",
+  "/enterovirus",
+  "/FFV",
```
jameshadfield: This dataset has now been removed from the "core" bucket & should be removed from here. (It may come back as a community / group dataset.)

tsibley: Ok, done.

server.js
tsibley: Done.

PR #65, opened 2019-12-16
Return v2 JSONs for private S3 sources

jameshadfield:
Previously `availableDatasets` for (private) S3 sources were not returning v2 JSONs. This commit creates a helper function to perform this task as we use it in multiple places.

### DIFF
--- a/auspice/server/utils.js
+++ b/auspice/server/utils.js
@@ -59,6 +59,39 @@ const printStackTrace = (err) => {
   }
 };
 
+/**
+ * Given a list of files, return a list of URL pathnames of
+ * datasets which can be fetched
+ * @param {Array} files. Array of strings.
+ * @returns {Array}
+ */
+const getDatasetsFromListOfFilenames = (filenames) => {
+  const jsonFiles = filenames
+    .filter((file) => file.endsWith(".json"));
+
+  // All JSON files which aren't a sidecar file with a known suffix are assumed to
+  // be v2+ JSONs (aka "unified" JSONs)
+  const sidecarSuffixes = ["meta", "tree", "root-sequence", "seq", "tip-frequencies"];
+  const datasets = jsonFiles
+    .filter((filename) => !sidecarSuffixes.some((suffix) => filename.endsWith(`_${suffix}.json`)))
+    .map((filename) => filename.replace(/[.]json$/, ""));
+
+  // All *_meta.json files which have a corresponding *_tree.json are assumed to
+  // be v1 JSONs.
+  jsonFiles
+    .filter((filename) => filename.endsWith("_meta.json"))
+    .filter((filename) => jsonFiles.includes(filename.replace(/_meta[.]json$/, "_tree.json")))
+    .map((filename) => filename.replace(/_meta[.]json$/, ""))
+    .filter((filename) => !datasets.includes(filename))
+    .forEach((filename) => datasets.push(filename));
+
+  // modify the filenames to represent URL pathnames not filenames
+  return datasets.map((filename) => filename
+    .split("_")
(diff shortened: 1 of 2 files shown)


auspice/server/utils.js
```diff
@@ -59,12 +59,46 @@ const printStackTrace = (err) => {
...
+    .map((filename) => filename.replace(/[.]json$/, ""));
+
+  // All *_meta.json files which have a corresponding *_tree.json are assumed to
+  // be v1 JSONs.
+  jsonFiles
+    .filter((filename) => filename.endsWith("_meta.json"))
+    .filter((filename) => jsonFiles.includes(filename.replace(/_meta[.]json$/, "_tree.json")))
+    .map((filename) => filename.replace(/_meta[.]json$/, ""))
+    .filter((filename) => !datasets.includes(filename))
+    .forEach((filename) => datasets.push(filename));
```
tsibley: I think that doing `datasets.push` within an `Array.forEach` like this weirdly mixes a functional, stream-based approach with a procedural approach. It hides the mutation that the entire statement ends up making. This is the reason I built the two arrays and then combined them with `Set([...v2, ...v1])` to process them together. That's not the only approach which I think would be better at preserving the importance of the mutation though; for example, making the `datasets.push` call the top-level statement and feeding it the return value of the `jsonFiles.filter.filter.map.filter` chain would convey similar importance. I typically try not to write void statements because it means something is having a side effect somewhere.

In general, I'm not sure sure why this bit changed during the refactor (similar with the removal of the `notSidecar` filter helper above). I think it's best to refactor things as-is as much as possible, and then make any other changes to them in subsequent commits. Separating out the refactor from other changes makes it easier for other people reading and understanding how and why something changed.

jameshadfield: Sounds like you have some great ideas here. Could you give me a concrete list of things you'd like changed?

tsibley: I'm asking why the implementation of this routine changed during the refactor. I was surprised to see that it was, because my expectation was that it wouldn't. (This is my expectation for a refactor in general, and so I tried to explain why in my comment above.)

jameshadfield: We [nextstrain] haven't been opinionated about not changing any details of the implementation in work such as this. If you'd like to write down your expectations then that may help.

If you have concrete points (e.g. keep the named function `notSidecar` rather than inlining it, or not using `forEach` to modify the `datasets` list) then I'd be happy to change the code, and appreciate your explanations as we all are trying to improve here.

tsibley: James, I'm not attacking you or your work. I'm not saying your changes are wrong. I haven't asked you change them back and did not plan to do so. I did express my surprise that they changed, and why I think the previous versions are more clear and reflective of standard practices. But people have different points of view, so I just wanted to know why you made the changes so I could understand the difference. It seems clear to me that there was thought behind it because the default, easiest action would be to just copy the code as-is. That's not what happened, so I was genuinely interested in understanding why. I apologize if I've inadvertently pissed you off in the process. :-/

tsibley: (and I guess to be as clear as possible, I would still appreciate knowing why, but I don't consider this blocking merge.)

trvrb:
V2 JSONs for public groups haven't been showing up on the live site: https://nextstrain.org/groups/blab. However, this PR fixes this. See: https://nextstrain-s-get-v2-dat-pr98r0.herokuapp.com/groups/blab. 🎉🎉🎉

Can this be merged? It would be nice to have these datasets showing.

tsibley:
As I noted in my last comment, I don't think there are any merge blockers. Will be nice to have this fixed.

PR #71, opened 2020-01-03
Dynamic user splash page

joverlee521:
This PR builds on top of #69 

If a user is signed in, the splash page has a groups section that dynamically creates tiles for the user's groups. These tiles are empty images with a background color from `theme.titleColors`.

In order to pass the user data to the index page, I created a `UserDataWrapper` that fetches user data from `/whoami` and passes the user state to each child element through the `user` prop. I'm not entirely sure how to manage a "global" user data store within gatsby when the authentication is handled on the server side.


jameshadfield:
👍 I'm going to close #69 and review this PR instead, which is now up on https://nextstrain-dev.herokuapp.com

static-site/src/pages/404.jsx
```diff
@@ -5,24 +5,33 @@ import styled from "styled-components";
 import config from "../../data/SiteConfig";
 import NavBar from '../components/nav-bar';
 import Splash from "../components/splash";
+import UserDataWrapper from "../layouts/userDataWrapper";
 
 
 const FourOhFour = () => {
+  // Workaround so 404 page doesn't flash when pages are redirecting
```
jameshadfield: This is nice! ~It removes the flash for me when going straight to docs pages (with or without redirects) but the~ (this was a flash of the splash page, not the 404 page)

The 404 page still flashes when I click "login" or "logout". In particular, clicking logout causes 2-3 page flashes, including at least one 404. Can you replicate?

static-site/package.json
```diff
@@ -31,6 +31,7 @@
     "heroku-postbuild": "gatsby build"
   },
   "dependencies": {
+    "@reach/router": "^1.2.1",
```
jameshadfield: This dep should be dropped -- it was only used in `withLocation` which was removed in 26cbaa2f132cd164ea0e3635f12683726139819f.

Note also that gatsby v1 uses `react-router` and so adding this dep introduces the possibility of unnecessary library duplication. Moving to gatsby v2+ will give access to @reach/router out of the box.

static-site/src/components/splash/userGroups.jsx
```diff
@@ -0,0 +1,45 @@
...
+        title: group,
+        color: groupColor
+      }
+    );
+  });
+
+  return (
+    <Fragment>
+      <ScrollableAnchor id={'groups'}>
+        <Styles.H1> {props.user.username}&apos;s Groups</Styles.H1>
```
jameshadfield: Please change this heading to be "Private nextstrain groups" and the subtext to be "Nextstrain groups are collections of datasets with controlled access. You (userName) have access to the following groups:". 

In the future we need to add documentation around groups and link to that here, but that's not part of this PR. I think it'd be great to add a lock icon to further indicate that they are not public, but that can be a separate PR.

static-site/src/pages/users.jsx
```diff
@@ -0,0 +1,74 @@
...
+  display: table;
+  margin: 0 auto !important;
+`;
+
+const Users = () => {
+  return (
+    <div className="index-container">
+      <main>
+        <UserDataWrapper>
+          <NavBar minified/>
```
jameshadfield: Please remove the `minified` prop here as this will add in the nextstrain logo which matches the splash page

static-site/src/pages/users.jsx
```diff
@@ -0,0 +1,74 @@
...
+import NavBar from '../components/nav-bar';
+import { Logos } from "../components/logos";
+
+const UserPage = (props) => {
+  const { user } = props;
+
+  const LoggedIn = () => (
+    <p>
+      You&apos;re logged in as <strong>{user.username}</strong>.<br />
+      You have access to the following groups:
```
jameshadfield: Add some text here explaining that the following are only private groups (e.g. `seatteflu` will be a public group and won't show up here as it won't have a corresponding cognito group attached to the user). E.g. "You have access to the following private nextstrain groups, which each contain a collection of datasets and/or narratives:"

static-site/src/pages/404.jsx
kairstenfay: After logging in and logging out, I no longer see 404 flashes.

kairstenfay:
@jameshadfield I've incorporated your requested changes except for:
* adding a padlock icon to Nextstrain groups header on the splash page (to further encode the concept of "private" datasets)
* using Context over Jover's dataWrapper component to pass state to the new user components

static-site/src/pages/404.jsx
jameshadfield: They're still there for me. This issue isn't blocking for this PR (the behavior's not introduced here) so i'm going to make an issue with GIFs so that you can see what i'm on about.

jameshadfield:
Thanks @kairstenfay i've retested and it's working well. This is a nice step forward for the website!

Note that because cognito groups are now filtered against those specified in `sources.js` at authentication time, if you're already logged in from a previous deployment of nextstrain.org then you may get tiles when you shouldn't (see screenshot below, `blab` isn't a valid nextstrain group (yet)). Looking through the cognito groups we have this should only impact users with `blab` or `seattleflu` cognito groups. It's resolved by logging-out & logging in.

image

> I've incorporated your requested changes except for ... [padlock icon & using `Context`]

I never requested those two changes! I only mentioned `Context` as an alternate solution (and specifically said that I was happy with the current solution) in response to Jover [saying she wasn't sure if this was the way to do it](https://github.com/nextstrain/nextstrain.org/pull/71#issue-359090182).

authn.js
```diff
@@ -211,18 +211,16 @@ function setup(app) {
 
   // Provide the client-side app with info about the current user
   app.route("/whoami").get((req, res) => {
+    if (req.user && req.user.groups) {
+      // Filter user groups to match available Nextstrain sources
+      req.user.groups = req.user.groups.filter((group) => sources.has(group));
```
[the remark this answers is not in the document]
tsibley: [Just made some comments](https://github.com/nextstrain/nextstrain.org/commit/1a65e10b0094e74642668cea2abd58ae91420c19#commitcomment-36703817) on the commit addressing this review note. Cross-linking now for completeness. I couldn't initially find this review note because resolved notes are collapsed by default, and I missed it.

kairstenfay:
> I never requested those two changes! I only mentioned `Context` as an alternate solution (and specifically said that I was happy with the current solution) in response to Jover [saying she wasn't sure if this was the way to do it](https://github.com/nextstrain/nextstrain.org/pull/71#issue-359090182).

Understood. I simply wanted to raise the two remaining _potential_ changes in a single comment. 

Additionally, it just occurred to me that we never implemented a redirect from `/users` to `/users/{username}`.

PR #74, opened 2020-01-09
Detangle Cognito, sources, & group logic on authn

kairstenfay:
Centralize the computation of the list of sources visible to the current
user, but using a different property that doesn't conflate Cognito
groups with our internal sources.

Additionally, centralize the authz decisions to stay within the Source
object.

This change was originally [requested by Thomas on a recently closed
PR](https://github.com/nextstrain/nextstrain.org/commit/1a65e10b0094e74642668cea2abd58ae91420c19#commitcomment-36703817).

### DIFF
--- a/authn.js
+++ b/authn.js
@@ -73,12 +73,17 @@ function setup(app) {
           Username: profile.username
         }).promise();
 
-        const groups = response.Groups
-          .map((g) => g.GroupName)
-          .filter((g) => sources.has(g));
+        const groups = response.Groups.map((g) => g.GroupName);
 
         // All users are ok, as we control the entire user pool.
-        return done(null, {...profile, groups});
+        const user = {...profile, groups};
+
+        user.visibleGroups = Array.from(sources.values())
+          .filter(source => source.visibleToUser(user))
+          .map(source => source._name)
+          .filter(source => groups.includes(source));
+
+        return done(null, user);
       }
     )
   );
--- a/static-site/src/components/splash/userGroups.jsx
+++ b/static-site/src/components/splash/userGroups.jsx
@@ -9,7 +9,7 @@ const UserGroups = (props) => {
 
   const colors = [...theme.titleColors];
 
-  const groupCards = props.user.groups.map((group) => {
+  const groupCards = props.user.visibleGroups.map((group) => {
     const groupColor = colors[0];
     colors.push(colors.shift());
 
--- a/static-site/src/pages/users.jsx
+++ b/static-site/src/pages/users.jsx
@@ -16,7 +16,7 @@ const UserPage = (props) => {
       </SubText>
 
       <UserGroupsList>
-        {user.groups.map((group) => (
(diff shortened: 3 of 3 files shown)


static-site/src/components/splash/userGroups.jsx
```diff
@@ -9,7 +9,7 @@ const UserGroups = (props) => {
 
   const colors = [...theme.titleColors];
 
-  const groupCards = props.user.groups.map((group) => {
+  const groupCards = props.user.visibleSources.map((group) => {
```
kairstenfay: how do we feel about calling the mapping variable `group` still?

tsibley: There's a conflict of terms here: the array is one of sources, but the map function calls them groups.

It may be that my off-the-cuff suggestion of `visibleSources` was a poor one and `visibleGroups` would be better.

authn.js
```diff
@@ -73,12 +73,17 @@ function setup(app) {
...
+        const groups = response.Groups.map((g) => g.GroupName);
 
         // All users are ok, as we control the entire user pool.
-        return done(null, {...profile, groups});
+        const user = {...profile, groups};
+
+        user.visibleSources = Array.from(sources.values())
+          .filter(source => source.visibleToUser(user))
+          .map(source => source._name)
+          .filter(source => groups.includes(source));
```
tsibley: Is the purpose of this last line to filter out the non-group sources (e.g. `CoreSource`, `CoreStagingSource`, and `CommunitySource`)?

If so, then I think there are more future-proof ways of doing this which delegates that decision to the source itself and avoids deepening the existing hardcoded assumption/conflation that the group name matches the source name exactly.

I don't think :point_up: needs to block merge, but I'll think about it some more and maybe push a commit up.

static-site/src/components/splash/userGroups.jsx
tsibley: Just commented on that separately before I saw your note here. :-)

authn.js
kairstenfay: Yes, it's to remove the groups/tiles that appear as `core`, `staging`, and `community` on nextstrain.org.

kairstenfay: I'm curious to know what a better solution would be.

kairstenfay:
@tsibley I renamed `visibleSources` to `visibleGroups`.

authn.js
tsibley: > I'm curious to know what a better solution would be.

Not entirely sure! :-) The equating of group names with source names in this part of the code raises red flags for me, but I think avoiding that here would require broader changes than I think are worth implementing right now. So I would merge this as-is and we can revisit later during other work on Groups if it arises again.

tsibley:
Ah, sorry, I was doing one last test and ran into an issue:

```
TypeError: "e.user.visibleGroups is undefined"
```

PR #79, opened 2020-01-11
Upgrade auspice to v2.2.4

trvrb:


trvrb:
Tested via Heroku review app.

PR #82, opened 2020-01-14
Identify public & private groups for logged in users

### LINKED ISSUE #80: Re-title "Private Nextstrain Groups" section header on splash page or restrict tiles to private groups
@jameshadfield [brought this up in our #nextstrain-general Slack channel](https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1578945475043700) when he realized he was seeing `blab` under Private Nextstrain Groups. 

In that Slack thread, @trvrb describes how he wants `blab` to exist as a public Nextstrain group. 

Ideally, we would show users their "Nextstrain Group Membership" including public & private with something to make it clear that links to private groups are only showing up because a user (with access) is logged in.

jameshadfield:
See commit messages for more details.

image

@tsibley can you please review b70246f as it adds some static methods to the `Source` class(es) and modifies the `visibleGroups` function.

@trvrb are you happy with the UI here? It would be simple to show this `Nextstrain Groups` section on the splash page for everyone (i.e. non logged in users), and this wouldn't be too bad with the small number of groups we currently have, but I don't think it should be above the core tiles. Let me know if you want me to do this.

Available for testing at https://nextstrain-dev.herokuapp.com/

Closes #80

### DIFF
--- a/auspice/server/sources.js
+++ b/auspice/server/sources.js
@@ -26,6 +26,9 @@ class Source {
   get baseUrl() {
     throw InvalidSourceImplementation("baseUrl() must be implemented by subclasses");
   }
+  static isGroup() { /* is the source a "nextstrain group"? */
+    return false;
+  }
   dataset(pathParts) {
     return new Dataset(this, pathParts);
   }
@@ -371,6 +374,9 @@ class PrivateS3Narrative extends Narrative {
 
 class PublicGroupSource extends S3Source {
   get bucket() { return `nextstrain-${this.name}`; }
+  static isGroup() {
+    return true;
+  }
 }
 
 class PrivateGroupSource extends PrivateS3Source {
@@ -378,6 +384,9 @@ class PrivateGroupSource extends PrivateS3Source {
 
   static visibleToUser(user) {
     return !!user && !!user.groups && user.groups.includes(this._name);
+  }
+  static isGroup() {
+    return true;
   }
 }
 
--- a/authn.js
+++ b/authn.js
@@ -210,25 +210,29 @@ function setup(app) {
   });
 
   /**
-   * Returns an array of names of Nextstrain groups that are visible to a
-   * given *user*.
+   * Returns an array of Nextstrain groups that are visible to a
+   * given *user* (or a non-logged in user). The order of groups returned
+   * matches the `sources` array in `sources.js`.
    *
    * FIX: Contains a hard-coded assumption that all Nextstrain groups match
    * their corresponding group name exactly.
    * See <https://github.com/nextstrain/nextstrain.org/issues/76> for more
    * context and to track this issue.
    *
-   * @param {Object} user
-   * @returns {Array}
+   * @param {Object | undefined} user. `undefined` represents a non-logged-in user
+   * @returns {Array} Each element is an object with keys `name` -> {str} (group name) and
+   *                  `private` -> {bool} (group is private)
    */
-  const visibleGroups = (user) => Array.from(sources.values())
-    .filter(source => source.visibleToUser(user))
-    .map(source => source._name)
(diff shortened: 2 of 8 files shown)


auspice/server/sources.js
```diff
@@ -26,6 +26,12 @@ class Source {
   get baseUrl() {
     throw InvalidSourceImplementation("baseUrl() must be implemented by subclasses");
   }
+  static isGroup() { /* is the source a "nextstrain group"? */
+    return false;
+  }
+  static isPrivate() { /* does the source require authentication? */
+    return false;
+  }
```
tsibley: The `isPrivate()` methods are duplicative. Calling `visibleToUser()` or `visibleToUser(null)` is valid and used elsewhere in the codebase to see if a Source is visible to an anonymous user.

authn.js
```diff
@@ -210,25 +210,33 @@ function setup(app) {
...
-  const visibleGroups = (user) => Array.from(sources.values())
-    .filter(source => source.visibleToUser(user))
-    .map(source => source._name)
-    .filter(source => user.groups.includes(source));
+  const visibleGroups = (user) => Array.from(sources)
+      .filter(([, source]) => source.isGroup())
+      .filter(([, source]) =>
+        source.isPrivate() ?
+          source.visibleToUser(user) :
+          true /* allow all public sources regardless of user data */
```
tsibley: The use of `isPrivate()` here subverts the intent that `visibleToUser(…)` is the sole authorizer of a (anonymous or logged in) user's ability to see a Source. As noted above, the `isPrivate()` method is unnecessary, and I think it should be removed.

static-site/src/layouts/userDataWrapper.jsx
```diff
@@ -11,7 +11,7 @@ export default class UserDataWrapper extends React.Component {
   }
 
   loadUser() {
-    console.log("Fetching user data");
+    // console.log("Fetching user data");
```
tsibley: Looks like an unrelated change made during dev slipped into this commit.

auspice/server/sources.js
jameshadfield: How can I use `visibleToUser` (or another method) to know whether a group (source) is visible to a user because it is public or because it's private and the user is authenticated? I could check for it being an instance of `PrivateGroupSource` but this seems more complex than the solution I implemented.

jameshadfield: 👍 resolved

authn.js
jameshadfield: 👍 resolved

static-site/src/layouts/userDataWrapper.jsx
jameshadfield: resolved

jameshadfield:
👍 thanks @tsibley -- have resolved all requested changes.

---

> I think the addition of isGroup() conflates the concept of a Nextstrain Group (for which we don't have first class concepts/objects/records for in the codebase) with the Source concept. 

Our current implementation is that nextstrain groups _are_ sources (although not all sources are groups) and I think it's necessary for a source to be able to interrogate itself to know if it's a "nextstrain group" or not. From your answer it's clear to me that you've thought about this and have a clear plan for how we these concepts are going to be implemented in the future. Looking forward to discussing this plan in the future (if there are docs I must have missed them and would appreciate you pointing me to them).

tsibley:
> From your answer it's clear to me that you've thought about this and have a clear plan for how we these concepts are going to be implemented in the future. Looking forward to discussing this plan in the future (if there are docs I must have missed them and would appreciate you pointing me to them).

Your intuition is correct, I do have a plan in my head. I look forward to (hopefully sooner than later) having time to put it down in writing!

PR #85, opened 2020-01-19
Include nCoV as core build and include tile

trvrb:


trvrb:
Reviewed via Heroku review app. Everything looks good.

PR #88, opened 2020-01-25
Upgrade auspice to 2.3.2

trvrb:


trvrb:
Tested via Heroku review app.

PR #89, opened 2020-01-25
Update splash page with nCoV heading

trvrb:


trvrb:
Reviewed via Heroku review app.

PR #90, opened 2020-01-26
add card for spanish ncov narrative

rneher:


jameshadfield:
👍 Added this card to the main nCoV set using the language of this PR

PR #96, opened 2020-01-27
Don't show nav-bar buttons when viewing narratives

jameshadfield:
Before:
image

After:
image

_P.S. See https://github.com/nextstrain/auspice/pull/853 for an improvement in the title display_


trvrb:
Looks great! Thanks James.

PR #101, opened 2020-01-28
upgrade auspice to 2.3.5

jameshadfield:


tsibley:
@jameshadfield Would you amend the commit to update `package-lock.json` to match? Otherwise it is out of sync with `package.json` and will be modified by the next `npm install`, leading to a dirty git checkout. This situation also breaks tools like `npm ci` which require the lock file is in sync.

PR #102, opened 2020-01-28
Support alternate Git branches for community sources

tsibley:
Community URLs may now include an alternate branch name with the syntax
`/community/<user>/<repo>@<branch>/…` to view datasets and narratives on a
branch other than "master".  If no branch is specified, "master" will be
used.

This change was made to allow translators of the 2019-nCoV narratives to
preview their work on their topic branches.


tsibley:
> The `getInfo` method of `CommunitySource` should be modified so that the byline indicates which branch is being accessed.

Done, thanks for the catch. While I was looking at `getInfo`, I fixed up a couple other things in separate commits.

PR #103, opened 2020-01-28
add card for german sit-rep

rneher:


rneher:
looks good. merging now.

PR #107, opened 2020-01-29
charon: Cache fetches to the GitHub API

tsibley:
This avoids repeatedly getting the same data from GitHub when nothing
has changed in the repo and thus avoids their unauthenticated API
rate-limit of 1 request per minute.  It makes loading of available
datasets and narrative lists much quicker too after the first time.

The make-fetch-happen package implements transparent caching just like
fetch() in the browser does.  The provided disk-based cache manager
should be sufficient for our needs.  It will be wiped every 24h by the
regular Heroku dyno restarts and by any deploy, but requests will just
be recached.  If it starts to have limitations, we can switch to caching
in Redis.

Additional verbose logging messages make it easier to troubleshoot
request or caching issues if they arise.

### DIFF
--- a/auspice/server/fetch.js
+++ b/auspice/server/fetch.js
@@ -1,0 +1,22 @@
+const __fetch = require("make-fetch-happen");
+const utils = require("./utils");
+
+const _fetch = __fetch.defaults({
+  cache: "default",
+  cacheManager: process.env.FETCH_CACHE || "/tmp/fetch-cache",
+});
+
+const fetch = async (url, options) => {
+  utils.verbose(`[fetch] ${options && options.method || "GET"} ${url}`);
+
+  const response = await _fetch(url, options);
+  const cachedAt = response.headers.get("X-Local-Cache-Time"); // documented by make-fetch-happen
+
+  utils.verbose(`[fetch] ${response.status} ${response.statusText} ${response.url} ${cachedAt ? `(from cache, timestamp ${cachedAt})` : ''}`);
+
+  return response;
+};
+
+module.exports = {
+  fetch,
+};
--- a/auspice/server/sources.js
+++ b/auspice/server/sources.js
@@ -1,6 +1,6 @@
 /* eslint-disable no-use-before-define */
 const AWS = require("aws-sdk");
-const fetch = require("node-fetch");
+const {fetch} = require("./fetch");
 const queryString = require("query-string");
 const {NoDatasetPathError, InvalidSourceImplementation} = require("./exceptions");
 const utils = require("./utils");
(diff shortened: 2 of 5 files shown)


trvrb:
This is a great idea Tom!

auspice/server/fetch.js
```diff
@@ -0,0 +1,22 @@
+const __fetch = require("make-fetch-happen");
+const utils = require("./utils");
+
+const _fetch = __fetch.defaults({
+  cache: "default",
+  cacheManager: process.env.FETCH_CACHE || "/tmp/fetch-cache",
```
jameshadfield: Is the plan to configure this variable for the heroku servers?

tsibley: No, the default should be fine. I did intend to document it but forgot, so I just pushed a commit noting it in the infrastructure docs.

tsibley:
> How long are requests cached? Fetching https://api.github.com/repos/nextstrain/ncov/contents/auspice (via browser) indicates a `Cache-Control` with `max-age=60` (i.e. 1min) -- is this what's used by `make-fetch-happen`?

HTTP caching is a pretty complex interaction; that `Cache-Control` header is used, but it is not the only thing effecting cache behaviour. I'll try to summarize.

The `Cache-Control` header sets how long compliant caches (of which `make-fetch-happen` is documented to be) are allowed to consider a resource _fresh_. Fresh resources can be returned from a cache without first checking the origin/upstream server for an updated version. After that period, a cached resource becomes _stale_, and the cache must check the origin for an updated resource the next time it is requested. It will do so using a conditional request by including the request headers `If-Modified-Since` and/or `If-None-Match`. Those headers are set to the values extracted from the `Last-Modified` and `ETag` headers of the cached resource. If the cached resource is still the most recent version, the origin will respond with 304 Not Modified and not send the resource content. Importantly for us, the GitHub API doesn't count 304 responses against rate limits. If the cached resource _is_ out of date, however, the origin will respond with a normal 200 Ok and the updated resource content. In either case, the case has validated (and potentially updated) its stale copy of the resource as fresh again and can return it directly for the next `Cache-Control: maxage` period without re-validating.

So as for how long requests are cached in this case: they're cached for up to the point a resource changes + 60 seconds (the `Cache-Control: maxage`). They may be cached for a shorter time if the cache is cleared by a dyno restart or other reasons (e.g. evicting entries to reduce cache size).

tsibley:
I should also mention that I've been saying the unauthenticated rate limit is 1 request per minute, but it's really 60 per hour. This makes a difference! You are allowed to make all 60 requests in, say, the first minute of the hour, you'll just have none left for the rest of the hour.

jameshadfield:
Thanks for the explanation Tom -- tested via review app & found no bugs. 👍 

> It makes loading of available datasets and narrative lists much quicker too after the first time.

Presumably this improvement is negligible after 1min, as I imagine a request + 304 response takes about the same time as a 200 response. If so, it may be an idea to modify the `max-age` of the cache (assuming one can do so) to ~5min, but I'll leave this up to you if you think it's a good idea.

tsibley:
> Presumably this improvement is negligible after 1min, as I imagine a request + 304 response takes about the same time as a 200 response. If so, it may be an idea to modify the `max-age` of the cache (assuming one can do so) to ~5min, but I'll leave this up to you if you think it's a good idea.

The request + response is very quick anyway, yeah.

`max-age` is set by the upstream server (i.e. GitHub in this case), not the client (i.e. our code). I don't think it would be a good idea to make the caching layer ignore it.

PR #108, opened 2020-01-30
fix(static-site): remove gatsby-inlined-css

ivan-aksamentov:
Since Gatsby v2  CSS is inlined automagically and the explicit setup relying on existence of `public/styles.css` (previously containing a portion of styles to be inlined) is no longer necessary, and, even worse, fails the production build with an error: 

   > Cannot find module '!raw-loader!../public/styles.css'.

because `public/styles.css` is no longer provided by Gatsby v2.

This PR removes the explicit setup and resolves the error. Complements #75. Resolves #104.

See discussion in:
 - https://github.com/nextstrain/nextstrain.org/issues/104
 - https://github.com/nextstrain/nextstrain.org/pull/75#issuecomment-574019845

Relevant gatsby docs:
 - https://www.gatsbyjs.org/docs/migrating-from-v1-to-v2/#remove-inlined-css-in-htmljs

Issue #104, opened 2020-01-28
Build error: Cannot find module '!raw-loader!../public/styles.css'


tsibley:
I get this error during the `static-site/` build when running `npm run build`. The build does seem to finish successfully despite it, but I'm not sure if `public/styles.css` is missing or not because of the error.

```
success Building production JavaScript and CSS bundles - 10.811s
success Rewriting compilation hashes - 0.004s
warn code block or inline code language not specified in markdown. applying generic code block
success run queries - 14.970s - 32/32 2.14/s
success Generating image thumbnails - 50.678s - 379/379 7.48/s
[                            ]   0.001 s 0/32 0% Building static HTML for pages
{ Error: Cannot find module '!raw-loader!../public/styles.css'
    at webpackMissingModule (/home/tom/nextstrain/nextstrain.org/static-site/public/render-page.js:6760:79)
    at Module../src/html.jsx (/home/tom/nextstrain/nextstrain.org/static-site/public/render-page.js:6760:186)
    at __webpack_require__ (/home/tom/nextstrain/nextstrain.org/static-site/public/render-page.js:30:30)
    at Module../.cache/static-entry.js (/home/tom/nextstrain/nextstrain.org/static-site/public/render-page.js:646:222)
    at __webpack_require__ (/home/tom/nextstrain/nextstrain.org/static-site/public/render-page.js:30:30)
    at ../favicon.png.module.exports (/home/tom/nextstrain/nextstrain.org/static-site/public/render-page.js:94:18)
    at /home/tom/nextstrain/nextstrain.org/static-site/public/render-page.js:97:10
    at webpackUniversalModuleDefinition (/home/tom/nextstrain/nextstrain.org/static-site/public/render-page.js:3:20)
    at Object.<anonymous> (/home/tom/nextstrain/nextstrain.org/static-site/public/render-page.js:10:3)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18) code: 'MODULE_NOT_FOUND' }
[...]

PR #109, opened 2020-01-30
fix(static-site): add polyfills and target config for Internet Explorer

ivan-aksamentov:
This resolves nexstrain.org main page crash with console errors when using IE browser.

Changes:
 - added polyfill for `fetch`
 - changed babel's transpilation target to cover 99% of active browsers, including IE versions above 9 (inclusive).

Note: On this particular commit build is failing, until https://github.com/nextstrain/nextstrain.org/issues/104 is resolved. However I made it independent on the fix https://github.com/nextstrain/nextstrain.org/pull/108 to avoid collisions in case someone will have the build issues resolved earlier.

Verified with
 - Internet Explorer 10.0.9200.17457
 - Internet Explorer 11.0.9600.18920

### DIFF
--- a/static-site/.browserslistrc
+++ b/static-site/.browserslistrc
@@ -1,1 +1,6 @@
+[development]
 Last 2 versions
+
+[production]
+cover 99%
+ie >= 9
--- a/static-site/gatsby-browser.js
+++ b/static-site/gatsby-browser.js
@@ -1,0 +1,5 @@
+export function onClientEntry() {
+  if (typeof fetch === 'undefined') {
+    require('whatwg-fetch'); // eslint-disable-line global-require
+  }
+}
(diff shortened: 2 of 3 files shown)


ivan-aksamentov:
There are still styling issues on internet explorer, such as 

 - pagination dots are misplaced: https://github.com/nextstrain/auspice/issues/655
 - also the heading text in the right panel of the narratives is partially obscured.

but at least it does not crash on the main page.

jameshadfield:
Tested with no regressions on MacOS / Firefox 72. I don't have IE / Edge to test on, but it's important for Nextstrain to work here as they are used in various public health agencies, so merging. Thanks @ivan-aksamentov

ivan-aksamentov:
@jameshadfield it is also worth checking how Safari is managing.

It is not uncommon for researchers, medical professionals and even some engineers to use outdated versions of MacOS and Safari, because they disable updates right after purchase.

For example, Safari only shipped `fetch()` in March 2017.
https://caniuse.com/#feat=fetch
So the Safari folks on pre-2017 Macs with updates disabled were unable to use nextstrain.org until this PR. There are many more missing features and bugs in Safari.

> "Safari is the new IE" (C) everybody

PR #110, opened 2020-01-30
Prototype auto language redirect & language selector

joverlee521:
This is specifically for nCoV narratives.

1. Will try to redirect to a translated narrative if the user has a language preference set. 
2. Displays a dropdown that shows all available languages for a dated narrative. 

No styling has been done for the dropdown yet.

### DIFF
--- a/auspice/client/languageSelector.js
+++ b/auspice/client/languageSelector.js
@@ -1,5 +1,6 @@
 import React from "react"; // eslint-disable-line
 import ISO6391 from "iso-639-1";
+import {parseNarrativeLanguage} from "../server/utils";
 
 class LanguageSelector extends React.Component {
     state = {
@@ -14,11 +15,7 @@ class LanguageSelector extends React.Component {
     }
 
     getNarrativeLanguage(narrative) {
-      const narrativeParts = narrative.split("/");
-      let narrativeLanguage = narrativeParts[narrativeParts.length - 2];
-      if (narrativeLanguage === 'sit-rep') {
-        narrativeLanguage = 'en';
-      }
+      const narrativeLanguage = parseNarrativeLanguage(narrative);
       return ISO6391.getNativeName(narrativeLanguage);
     }
 
--- a/auspice/client/languageSelector.js
+++ b/auspice/client/languageSelector.js
@@ -1,0 +1,78 @@
+import React from "react"; // eslint-disable-line
+import ISO6391 from "iso-639-1";
+
+class LanguageSelector extends React.Component {
+    state = {
+      currentNarrative: undefined,
+      availableNarratives: undefined
+    }
+
+    async getAvailableNarratives() {
+      const response = await fetch("/charon/getAvailable");
+      const available = await response.json();
+      return available.narratives.map((n) => n.request);
+    }
+
+    getNarrativeLanguage(narrative) {
+      const narrativeParts = narrative.split("/");
+      let narrativeLanguage = narrativeParts[narrativeParts.length - 2];
+      if (narrativeLanguage === 'sit-rep') {
+        narrativeLanguage = 'en';
+      }
+      return ISO6391.getNativeName(narrativeLanguage);
+    }
+
+    async componentDidMount() {
+      const currentNarrative = window.location.pathname.replace(/^\//, '');
+      const currentNarrativeDate = currentNarrative.split('/').pop();
+      this.getAvailableNarratives().then((availableNarratives) => {
+        const translatedNarratives = [];
+        availableNarratives.forEach(narrative => {
+          if (narrative.startsWith('narratives/ncov/sit-rep/')) {
+            const date = narrative.split('/').pop();
+            // Insert 'en' into narrative url to force English version of narrative
+            // if user has a separate language preference
+            if (narrative === 'narratives/ncov/sit-rep/' + date) {
+              const narrativeParts = narrative.split('/');
+              narrativeParts.splice(-1, 0, 'en');
+              narrative = narrativeParts.join('/');
+            }
+            if (date === currentNarrativeDate) {
+              translatedNarratives.push(narrative);
+            }
+          }
+        });
+        this.setState({
+          currentNarrative: currentNarrative,
+          availableNarratives: translatedNarratives
+        });
+      }).catch((error) => {
+        console.log(error);
+      });
+    }
+
+    handleLanguageChange(event) {
+      const narrativeUrl = event.target.value;
+      window.location.assign(window.location.origin + '/' + narrativeUrl + window.location.search);
+    }
+
+    render() {
+      if (this.state.availableNarratives && this.state.currentNarrative) {
+        return (
+          <select value={this.state.currentNarrative} onChange={this.handleLanguageChange}>
+            {this.state.availableNarratives.map((narrative) => {
+              const narrativeLanguage = this.getNarrativeLanguage(narrative);
+              return (
+                <option value={narrative}>
+                  {narrativeLanguage}
+                </option>
+              );
+            })}
+          </select>
+        );
+      }
+      return null;
+    }
+}
+
+export default LanguageSelector;
--- a/redirects.js
+++ b/redirects.js
@@ -1,3 +1,4 @@
+const acceptedLanguage = require("accept-language");
 const helpers = require("./auspice/server/getDatasetHelpers");
 
 const setup = (app) => {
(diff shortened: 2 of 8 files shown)


jameshadfield:
Thanks @joverlee521 -- could you provide some advice for how to test this?

joverlee521:
> Thanks @joverlee521 -- could you provide some advice for how to test this?

I believe all browsers have the ability to set a preferred language. I've set Chinese as my first language preference in Chrome settings to see the auto-redirect

jameshadfield:
https://nextstrain-s-auto-langu-y9rldh.herokuapp.com/

joverlee521:
> * Language doesn't automatically switch to for me on Firefox 72 (spanish is listed in the dropdown for the narrative I'm trying to load)
>   image

Ah, this is because the language code has to match exactly what we have in the url (so just `es` for Spanish). I can work on improving this.

server.js
```diff
@@ -139,8 +140,31 @@ const auspicePaths = [
...
+/*
+ * Redirect to translations of narratives if the client has
+ * set language preference and the translation is available
+ *
+ * Currently only used for nCoV narratives and requires the
+ * language code to be the second to last part of the url.
+ */
+const redirectToTranslatedNarratives = async (req, res) => {
+  const acceptedLanguages = req.acceptsLanguages();
+  if (acceptedLanguages && !acceptedLanguages[0].startsWith("en")) {
```
tsibley: We will have a much better chance of successfully identifying a matching language if we support the full `Accept-Language` negotiation standard. This includes matching on preferred languages other than the first as well as matching our language-only codes (`de`, `zh`) to language-region codes (`de-CH`, `de-DE`, `zh-CN`), which are commonly found in `Accept-Language` headers sent by browsers.

Taking a look at NPM, I expect the [accept-language](https://www.npmjs.com/package/accept-language) package will make this match easy to make compliant.

server.js
```diff
@@ -139,8 +140,31 @@ const auspicePaths = [
...
+      res.redirect(301, "/narratives/" + potentialNarrative);
+    }
+  }
+};
+
 app.route(auspicePaths).get((req, res) => {
   utils.verbose(`Sending Auspice entrypoint for ${req.originalUrl}`);
+  if (req.url.startsWith("/narratives/ncov")) {
+    redirectToTranslatedNarratives(req, res);
+  }
```
tsibley: It would be nice to move this alongside the other redirects in (the recently added) `redirects.js` file.

server.js
```diff
@@ -139,8 +140,31 @@ const auspicePaths = [
...
+ */
+const redirectToTranslatedNarratives = async (req, res) => {
+  const acceptedLanguages = req.acceptsLanguages();
+  if (acceptedLanguages && !acceptedLanguages[0].startsWith("en")) {
+    const {source, prefixParts} = helpers.splitPrefixIntoParts(req.url);
+    prefixParts.splice(-1, 0, acceptedLanguages[0]);
+    const potentialNarrative = prefixParts.join("/");
+    const availableNarratives = await source.availableNarratives();
+    if (availableNarratives.includes(potentialNarrative)) {
+      res.redirect(301, "/narratives/" + potentialNarrative);
```
tsibley: 301 is Moved Permanently, which will cause most browsers to remember the redirection and never request the first URL again. It is appropriate for when resources move, but not for something dynamic like language detection. 302 Found is a more appropriate redirection status that won't lead to weird bugs.

auspice/client/languageSelector.js
```diff
@@ -0,0 +1,78 @@
...
+
+    getNarrativeLanguage(narrative) {
+      const narrativeParts = narrative.split("/");
+      let narrativeLanguage = narrativeParts[narrativeParts.length - 2];
+      const languageMap = {
+        en: "English",
+        zh: "Chinese",
+        es: "Spanish",
+        pt: "Portuguese",
+        de: "German"
```
tsibley: Best practice is to present language names in the language itself, and this is what the existing translated narrative links do too.

tsibley:
@joverlee521 

> Depending on how long it takes to return the availableNarratives, may sendFile run before the redirect happens, leading to an Unhandled Promise error.

This is because `redirectToTranslatedNarratives()` is an async function but it is never awaited. If it is awaited, then the `sendFile` call won't run before it returns.

server.js
joverlee521: Looking at the [express-request-language](https://www.npmjs.com/package/express-request-language) which is the express middleware of accept-language, it only returns a single language instead of a list of _all_ accepted languages. So, using this we would match language-region codes, but only take into account the users first language preference.

tsibley: 1. I don't think the express-request-language middleware is appropriate here since the list of supported languages isn't static.

2. A single language match is what we want, yes? Given a list of all a user's preferred languages, in order of preference, and a list of the languages we support (e.g. have narratives written in), we want to find the supported language which is most preferred by the user.

joverlee521: I thought we wanted to serve the user's second choice in language if the first language translation was not available?

tsibley: (@joverlee521 and I just talked about this in person, but for posterity and anyone reading, I'll finish the reply I was in the middle of writing.)

Yes, exactly. :-)

When I say "we want to find the supported language which is most preferred by the user", I mean that if we support (in no particular order) A, B, C and the user prefers (in order) D, E, C, A then the ranked intersection of those is C, A and we want to use C. Using accept-language to do this would look like:

```js
acceptLanguages.languages(['A', 'B', 'C']);
acceptLanguages.get('D, E, C, A') → C
```

If the intersection is the empty set, then `acceptLanguages.get` returns nothing and we pick a default (probably English).

joverlee521:
@jameshadfield @tsibley Does this look good to merge?

redirects.js
```diff
@@ -19,6 +20,37 @@ const setup = (app) => {
...
+      const availableLanguages = new Set(availableNarratives.map((narrative) => {
+        if (narrative.startsWith('ncov/sit-rep/')) {
+          const narrativeParts = narrative.split("/");
+          const language = narrativeParts[narrativeParts.length - 2];
+          if (language !== 'sit-rep') return language;
+        }
+        return null;
+      }));
+      acceptedLanguage.languages(['en'].concat([...availableLanguages]));
+      const languageChoice = acceptedLanguage.get(req.headers['accept-language']);
```
tsibley: I realize only now, looking at [the docs for `req.acceptsLanguage()`](https://expressjs.com/en/api.html#req.acceptsLanguages), that it's not just a parsed header accessor but can be used to actually match accepted by supported languages by providing arguments:

```js
req.acceptsLanguages("en", "zh-cn", "de-ch", "es", "pt", "ru", …)
```

Seeing the switch this PR made from `req.acceptsLanguage()` to `req.headers['accept-language']` made me realize I might be missing something. @joverlee521 Were you aware of that? :-)

auspice/client/languageSelector.js
```diff
@@ -0,0 +1,78 @@
...
+      return available.narratives.map((n) => n.request);
+    }
+
+    getNarrativeLanguage(narrative) {
+      const narrativeParts = narrative.split("/");
+      let narrativeLanguage = narrativeParts[narrativeParts.length - 2];
+      if (narrativeLanguage === 'sit-rep') {
+        narrativeLanguage = 'en';
+      }
+      return ISO6391.getNativeName(narrativeLanguage);
```
tsibley: :sparkle: Awesome!

auspice/client/languageSelector.js
```diff
@@ -0,0 +1,78 @@
...
+      const available = await response.json();
+      return available.narratives.map((n) => n.request);
+    }
+
+    getNarrativeLanguage(narrative) {
+      const narrativeParts = narrative.split("/");
+      let narrativeLanguage = narrativeParts[narrativeParts.length - 2];
+      if (narrativeLanguage === 'sit-rep') {
+        narrativeLanguage = 'en';
+      }
```
tsibley: There's at least 3 or 4 places where this sort of path processing is done. For maintainability, it'd be really nice to centralize it.

redirects.js
```diff
@@ -19,6 +20,37 @@ const setup = (app) => {
...
+    .get(async (req, res, next) => {
+      const {source, prefixParts} = helpers.splitPrefixIntoParts(req.url);
+      const availableNarratives = await source.availableNarratives();
+      const availableLanguages = new Set(availableNarratives.map((narrative) => {
+        if (narrative.startsWith('ncov/sit-rep/')) {
+          const narrativeParts = narrative.split("/");
+          const language = narrativeParts[narrativeParts.length - 2];
+          if (language !== 'sit-rep') return language;
+        }
+        return null;
```
tsibley: This is trying to use a map as both a map and a filter, but will include `null` in the final `Set`.

redirects.js
joverlee521: Oh wow. completely missed that! 🤦‍♀

Issue #113, opened 2020-01-31
IE: crash on "let" keyword (related to gatsby-remark-autolink-headers)


ivan-aksamentov:
Internet Explorer prints console error:

 > `SCRIPT1004: Expected ';` 

I consider this a bug in Gatsby and opened an issue upstream: 
https://github.com/gatsbyjs/gatsby/issues/21058

And here is a fix:
https://github.com/gatsbyjs/gatsby/pull/21083

See there for full description.
 
Actions advised: monitor the issue and, if fix is merged, update `gatsby-remark-autolink-headers` package.

PR #115, opened 2020-02-02
fix: crash on "let" keyword in IE

ivan-aksamentov:
Resolves https://github.com/nextstrain/nextstrain.org/issues/113

Upgrades `gatsby-remark-autolink-headers` to version 2.1.24, which contains the necessary fix.


jameshadfield:
👍 tested via heroku review app on non-IE browsers with no regressions

PR #116, opened 2020-02-03
static pages for nCov

rneher:
exploring static pages -- not meant for merging -- made PR to trigger heroku review app.


jameshadfield:
Content moved over to #118

PR #119, opened 2020-02-22
Workaround to ensure that ES6 classes don't leak into the bundle from iso-639-1 package

ivan-aksamentov:
This does not solve the problem, but temporarily patches  https://github.com/nextstrain/auspice/issues/902, unlocking nextstrain.org users with some of the older browsers.

This works by forcing the import of ES5 version of `iso-639-1`

### DIFF
--- a/auspice/client/languageSelector.js
+++ b/auspice/client/languageSelector.js
@@ -1,5 +1,5 @@
 import React from "react"; // eslint-disable-line
-import ISO6391 from "iso-639-1";
+import ISO6391 from "iso-639-1/build/index";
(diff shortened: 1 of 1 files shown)


jameshadfield:
👍 thanks @ivan-aksamentov 

The `const` statements are coming from JS code inside `/nextstrain.org/auspice/server`. These aren't getting transpiled (which is somewhat intentional) as they shouldn't be included in the auspice bundle at all! I remember similar situations from developing auspice extensions, so i'll see if there's a quick (temporary) fix.

PR #120, opened 2020-02-26
change auspice folder structure to prevent auspice build bugs

### LINKED ISSUE #111: nextstrain splash page doesn't work with auspice develop
Being able to run `auspice develop --extend <path to config.json>` is useful for development. Unfortunately the nextstrain.org custom (auspice) splash page has errors in this environment (perhaps because of a failed API call?) and therefore doesn't list the datasets / narratives. Datasets / narratives load fine if you know the URLs.

Not a high priority.

jameshadfield:
This moves the "auspice" server code to `./src` so that all the auspice client code is in it's own directory (`./auspice-client`). This helps prevent `auspice build` pulling in code it shouldn't and which it doesn't correctly transpile. Note that this is a work-around for the deeper problem explained at https://github.com/nextstrain/auspice/issues/902. Without this PR `const` statements, and code from `./auspice/server` was ending up in the transpiled auspice client bundles. 

Moving server code to `./src` makes sense now that the nextstrain.org capabilities have advanced. Some file renaming may be in order here for clarity, and we should consider moving `authn.js` and `redirects.js` into `./src`.

@ivan-aksamentov would you mind taking a look at this please?

Closes #111


ivan-aksamentov:
Yep, looks like that fixed it.

Here is IE 10, IE 10 again, Chrome 54 and Firefox 45:
<a href="https://user-images.githubusercontent.com/9403403/75568929-7848a180-5a54-11ea-91a5-15a192d22d30.png">

</a>

LGTM

ivan-aksamentov:
@tsibley Thomas, there is some sort of magic going on with this issue, not sure if I or James can give you any more understanding :)

Anyway,  take a look if you have a moment.

Eliminating the import from server might have helped, indeed. `const` keywords disappeared, even though client code has them. This means that at least this part is transpiled correctly.

Additionally, I have just tested that reverting https://github.com/nextstrain/nextstrain.org/pull/119 breaks everything again (`class`es reaper). This means client dependencies are not picked up and/or transpiled correctly still.

Does it mean imports outside of client directory are not processed correctly?

jameshadfield:
>  But the change that actually fixes the issue is the copying of `parseNarrativeLanguage` instead of importing it from server → client code, right? 

Perhaps. Definitely more server code than just `parseNarrativeLanguage` was getting into the client bundle, but it's very plausible that this function import was the root cause. I remember from development that having sister directories like we had (server + client code) was problematic, and I'm sure I wrote something about it in the docs but can't seem to find it now. 

> Does it mean imports outside of client directory are not processed correctly?

Seems likely. I spent a long time on webpack imports during implementation of auspice's extension machinery and it was far from straightforward. There's plenty of room for improvement here, but we don't have the time currently. See also problems such as https://github.com/nextstrain/auspice/issues/689 which is still a concern to me.

tsibley:
@jameshadfield @ivan-aksamentov Thanks! It was useful to hear a little more context.

PR #125, opened 2020-03-05
fix: Display 404 page when needed

### LINKED ISSUE #123: Nextstrain gatsby section is missing a 404 page
When we upgraded to gatsby v2 recently, we lost our 404 page!

To test -- https://nextstrain.org/doesNotExist goes straight to https://nextstrain.org/ and there is no 404 message shown. The original code 404 page [is here](https://github.com/nextstrain/nextstrain.org/blob/master/static-site/src/pages/404.jsx).

ssong:
Not too sure how to feel about this particular fix but from my local testing it seems that any pages that aren't one of the matching routes won't have the `/page-data/` prefix.

fixes #123

### DIFF
--- a/server.js
+++ b/server.js
@@ -153,7 +153,11 @@ app.get("/inrb-drc*", (req, res) => {
  */
 app.get("*", (req, res) => {
   utils.verbose(`Sending Gatsby entrypoint for ${req.originalUrl}`);
-  res.sendFile(gatsbyAssetPath("index.html"));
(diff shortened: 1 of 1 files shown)


jameshadfield:
Thanks @ssong -- we've merged a very similar fix which closes this issue. I'm going to make an issue about how the server deals with the gatsby vs auspice routing decisions as there are improvements we could make here I think -- please comment there if you can help 😄

PR #129, opened 2020-03-11
fix 404 issue

### LINKED ISSUE #123: Nextstrain gatsby section is missing a 404 page

technicalciso:
bit of a hackup, but should fix the issue

fixes #123

### DIFF
--- a/server.js
+++ b/server.js
@@ -153,7 +153,9 @@ app.get("/inrb-drc*", (req, res) => {
  */
 app.get("*", (req, res) => {
   utils.verbose(`Sending Gatsby entrypoint for ${req.originalUrl}`);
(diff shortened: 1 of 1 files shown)


jameshadfield:
Thanks @technicalciso -- merged & released after fixing the eslint errors which prevented the build succeeding. I'm going to make an issue about how the server deals with the gatsby vs auspice routing decisions as there are improvements we could make here I think.

Issue #121, opened 2020-02-26
getAvailable API request bug for community datasets


jameshadfield:
The `request` field is not correct for `getAvailable` API requests with community datasets. This manifests in auspice not correctly displaying the "built with" text.

E.g. https://nextstrain.org/charon/getAvailable?prefix=/community/blab/zika-colombia returns:
```json
"datasets": [
    {"request": "community/blab/zika-colombia/zika-colombia"
```
But the request should be `community/blab/zika-colombia`.

Bisecting indicates this bug arrived in 94973fc2017071dabb01d3e3cc6c65bbfaafee66

victorlin:
Just made a PR with a potential fix for this. A better example might be one with a non-master branch (e.g. https://nextstrain.org/charon/getAvailable?prefix=/community/blab/ebola-narrative-ms@new-narrative):

    "datasets": [
        {"request": "community/blab/ebola-narrative-ms@new-narrative/ebola-narrative-ms",
         "buildUrl": "https://github.com/blab/ebola-narrative-ms"

The PR fixes it to return:

    "datasets": [
        {"request": "community/blab/ebola-narrative-ms@new-narrative",
         "buildUrl": "https://github.com/blab/ebola-narrative-ms"

This is probably a separate issue, but might as well mention it here: should `buildUrl` reflect the branch as well? For example:

    "datasets": [
        {"request": "community/blab/ebola-narrative-ms@new-narrative",
         "buildUrl": "https://github.com/blab/ebola-narrative-ms/tree/new-narrative"

PR #135, opened 2020-03-17
Remove duplication of repo name in getAvailable request field

victorlin:
#121 seems to be caused by having the repo name in both `source.repoNameWithBranch` and passed into `joinPartsIntoPrefix()` as an item in `prefixParts`. By treating this as a special case affecting community sources only, we can ignore `prefixParts` entirely for these sources.

### DIFF
--- a/src/getDatasetHelpers.js
+++ b/src/getDatasetHelpers.js
@@ -103,11 +103,10 @@ const joinPartsIntoPrefix = ({source, prefixParts, isNarrative = false}) => {
     // Community source requires an owner and repo name
     case "community":
       leadingParts.push(source.owner, source.repoNameWithBranch);
-      break;
-    // no default
+      return leadingParts.join("/");
+    default:
+      return [...leadingParts, ...prefixParts.filter((x) => x.length)].join("/");
   }
-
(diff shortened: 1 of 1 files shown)


jameshadfield:
@tsibley / @joverlee521 / @kairstenfay could you review this PR if you have time? Thanks in advance

tsibley:
This floated onto my radar again. A throwback to before you were on the team, @victorlin. :-) Sorry we never reviewed this! :disappointed: 

In any case, I took a look today, and while discarding `prefixParts` for community sources solves the particular example in #121, it breaks using nested datasets in community repo, e.g.:

    https://nextstrain.org/community/nextstrain/community-test/zika-tutorial

For that dataset, `prefixParts` contains `zika-tutorial` too.

However, the observation that the `repoName` is included in `prefixParts` is correct and that's a bug in `CommunitySource.availableDatasets()`. I'll submit another PR fixing it.

victorlin:
@tsibley haha, I forgot about this one. No worries, I can only imagine how crazy those times were for you all.

Thanks for picking this back up in another PR!

PR #139, opened 2020-03-29
Travis CI updates

peterbecich:
A couple of updates to the Travis CI configuration.  I noticed Node 9 is no longer supported.

### Testing

Builds succeed for Node 10 and 12: https://travis-ci.com/github/peterbecich/nextstrain.org/builds/156292305

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,6 @@ dist: bionic
 node_js:
   - "10"
   - "12"
-python:
-  - "3.8"
 install:
   - npm install
 script:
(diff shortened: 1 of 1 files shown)


jameshadfield:
Thanks @peterbecich -- I believe we don't need Python at all anymore (see https://github.com/nextstrain/nextstrain.org/pull/140). We used to need py27 with an earlier version of Gatsby, but no longer. Could you amend this commit to drop python altogether (unless you think it's needed).

PR #141, opened 2020-03-31
charon/getDataset: Use (cached) fetch() instead of "request"

tsibley:

Switching to fetch makes it easier to handle low-level network errors
(i.e. not HTTP errors) fetching from upstream, like connect(2) timeouts,
rather than crashing the app.  This was an issue when CloudFront (for
data.nextstrain.org) was having latency issues earlier today.  Thanks to
async/await, such errors will be properly caught by the try/catch
surrounding the requestMainDataset() call.  The code is also less
tangled by leveraging async/await and removing callbacks that close over
the objects they're attached to.

In addition, I used our cached version of fetch() to avoid unnecessary
network transfers with upstream, which seems like an obvious win.  I'm
presuming it's faster to send data directly from our disk cache rather
than stream it off an upstream socket.  This should be true, but I guess
it's possible the link to CloudFront is really blazing fast and Heroku's
using really slow disks.  Profiling would confirm if we have doubts.

### Testing
I tested locally to confirm that streaming still appears to be working, the the newly added caching works, that fall backs to v1 datasets work, and that low-level errors are actually caught.

### DIFF
--- a/src/getDataset.js
+++ b/src/getDataset.js
@@ -3,7 +3,7 @@ const utils = require("./utils");
 const helpers = require("./getDatasetHelpers");
 const {NoDatasetPathError} = require("./exceptions");
 const auspice = require("auspice");
-const request = require('request');
+const {fetch} = require("./fetch");
 
 /**
  *
@@ -54,25 +54,24 @@ const requestV1Dataset = async (metaJsonUrl, treeJsonUrl) => {
  * then the promise will reject.
  */
 const requestMainDataset = async (res, fetchUrls) => {
-  return new Promise((resolve, reject) => {
-    /* try to stream the (v2+) dataset JSON as the response */
-    const req = request
-      .get(fetchUrls.main)
-      .on("response", async (response) => { // eslint-disable-line consistent-return
-        if (response.statusCode === 200) {
-          utils.verbose(`Successfully streaming ${fetchUrls.main}.`);
-          req.pipe(res);
-          return resolve();
-        }
-        utils.verbose(`The request for ${fetchUrls.main} returned ${response.statusCode}.`);
-        const [success, dataToReturn] = await requestV1Dataset(fetchUrls.meta, fetchUrls.tree);
-        if (success) {
-          res.send(dataToReturn);
-          return resolve();
-        }
-        return reject(dataToReturn);
-      });
-  });
+  /* try to stream the (v2+) dataset JSON as the response */
+  const response = await fetch(fetchUrls.main);
+
+  if (response.status === 200 || response.status === 304) {
+    utils.verbose(`Successfully streaming ${fetchUrls.main}`);
+    res.set("Content-Type", "application/json");
+    response.body.pipe(res);
+    return;
+  }
+
+  utils.verbose(`The request for ${fetchUrls.main} returned ${response.status}.`);
+  const [success, dataToReturn] = await requestV1Dataset(fetchUrls.meta, fetchUrls.tree);
+
+  if (!success) {
+    throw new Error("Failed to fetch either a v2 or v1 dataset");
+  }
+
+  res.send(dataToReturn);
 };
 
 /**
--- a/src/getDataset.js
+++ b/src/getDataset.js
@@ -3,7 +3,7 @@ const utils = require("./utils");
 const helpers = require("./getDatasetHelpers");
 const {NoDatasetPathError} = require("./exceptions");
 const auspice = require("auspice");
-const {fetch} = require("./fetch");
+const fetch = require('node-fetch');
 
 /**
  *
--- a/package.json
+++ b/package.json
@@ -38,7 +38,6 @@
     "passport": "^0.4.0",
     "passport-oauth2": "^1.5.0",
     "query-string": "^4.2.3",
-    "request": "^2.88.0",
     "serve-favicon": "^2.5.0",
     "session-file-store": "^1.3.1",
     "yaml-front-matter": "^4.0.0"


jameshadfield:
Thanks @tsibley -- code is much cleaner 👍 

Testing via the review app, I get stochastic failures resulting in auspice falling back to its 404 page.

Examples:

Loading https://nextstrain-s-upstream-e-7jqxvm.herokuapp.com/narratives/ncov/sit-rep/2020-03-27.

Server logs:
```
[verbose]	[fetch] GET http://data.nextstrain.org/ncov_2020-03-27.json
[verbose]	[fetch] 200 OK http://data.nextstrain.org/ncov_2020-03-27.json 
[verbose]	Successfully streaming http://data.nextstrain.org/ncov_2020-03-27.json
```

Error in browser console:
```
SyntaxError: "JSON.parse: bad control character in string literal at line 903 column 20 of the JSON data" JSON.parse: bad control character in string literal at line 903 column 20 of the JSON data
```

Refreshing the page works, with one salient change in the log:
```
[verbose]	[fetch] 200 OK http://data.nextstrain.org/ncov_2020-03-27.json (from cache, timestamp Tue, 31 Mar 2020 02:48:43 GMT)
```

tsibley:
@jameshadfield Thanks for testing this! Strangely, I can't replicate any of the errors you're seeing. All links to the review app you sent consistently work fine for me with repeated refreshing, both with my browser cache enabled and disabled.

Can you catch more details on the failed requests/responses from your browser's network log? If you can export and share a HAR of the requests from a failed run, that might help with reproducing or debugging.

jameshadfield:
It's the server cache, not the browser cache. I believe it only happens the first time the dataset is fetched (from cloudfront) by the server (or whenever the cache is invalidated, but it's not clear to me when that is).

How about trying another of the (big) avian / seasonal flu datasets? I imagine if you tried the same datasets as me then the server has already cached them so it works...

tsibley:
Yep, I know we're talking about the server cache. :+1:  (I mentioned the browser cache because it affects client requests to the server and I wanted to eliminate it from the middle. ETOOMANYCACHES.)

I tried several other seasonal flu and avian datasets and could not replicate on the review app.

I switched to trying to reproduce locally, and after a bunch of clicking around was finally able to get a similar SyntaxError from JSON.parse.

This sounds absurd, but I can reliably reproduce locally by:

1. Making sure the server's cache is clear (`rm -rf /tmp/fetch-cache`)
2. Opening the (Firefox) dev tools
3. Requesting http://localhost:5000/flu/avian/h5n1/ha

If I don't open the dev tools, I can't reliably reproduce. It does happen though, just very rarely for me (1 in 20 at best, maybe). I don't know if the dev tools' observation of the requests affects them, or what the heck is going on there.

If I make the /charon/getDataset request with curl, I can't reproduce. I always seem to get a perfectly fine JSON, even with clearing the server-cache before every request.

I'll try to investigate more tomorrow but would appreciate if you wanted to spend some time debugging too. My best guess right now is mumblemumble something about `response.body.pipe(res)` not getting to finish because some object gets destroyed mumblemumble, but it's a pretty tenuous guess.

Looking at the stack trace of the SyntaxError, I did immediately want to get a debugging build of Auspice so that the symbols weren't mangled, but wasn't sure how to do that off-hand.

jameshadfield:
Will test more. Given that cloudfront seems to have sorted it out there's no immediate hurry, but this PR represents definite improvements. 

> I did immediately want to get a debugging build of Auspice

Can't do this without modifying the webpack config I believe. I imagine this could be a nice flag for `auspice build`. If an issue were made I think this could be added pretty quickly.

I also want to test this in order to understand when server caches are invalidated so as to not serve datasets which are out-of-date.

tsibley:
It might be worth testing this out without the cached `fetch()`, just normal `fetch()`. Switching is a one-line change and if that resolves the truncation issue, then we'd still get the proper error handling benefits. Caching could be re-added later after further investigation.

jameshadfield:
Notes on retesting this using a new review app. Browser: Firefox with cache disabled & dev tools open.

**Failed example number 1** https://nextstrain-s-upstream-e-qx0bjj.herokuapp.com/flu/avian.  The get request for /charon/getDataset?prefix=/flu/avian/h5n1/ha is a 302 as expected and the following get request for charon/getDataset?prefix=/flu/avian/h5n1/ha is 200 and just seems to end part way through the response -- here's the tail of the JSON from firefox dev tools

```
                                                  },
                                                  "region": {
                                                    "confidence": {
                                                      "China": 1.0
                                                    },
                                                    "entropy": -1.000088900581841e-12,
                                                "value": "Southeast Asia"
                                                                                                              },
                                                       
```
Here are the server logs

```
[verbose]	[fetch] GET http://data.nextstrain.org/flu_avian_h5n1_ha.json
[verbose]	[fetch] 200 OK http://data.nextstrain.org/flu_avian_h5n1_ha.json (from cache, timestamp Thu, 07 May 2020 02:04:16 GMT)
[verbose]	Successfully streaming http://data.nextstrain.org/flu_avian_h5n1_ha.json

heroku[router]: at=info method=GET path="/charon/getDataset?prefix=/flu/avian/h5n1/ha" host=nextstrain-s-upstream-e-qx0bjj.herokuapp.com request_id=b7322170-2823-4970-bb0f-05ae6a2c3160 fwd="203.173.142.4" dyno=web.1 connect=1ms service=576ms status=200 bytes=673318 protocol=https
```

As before, refreshing the page works (implicating something related to the first time the server fetches a file, or its cache. I don't know). Working server logs:

```
[verbose]	[fetch] GET http://data.nextstrain.org/flu_avian_h5n1_ha.json
[verbose]	[fetch] 200 OK http://data.nextstrain.org/flu_avian_h5n1_ha.json (from cache, timestamp Thu, 07 May 2020 02:04:16 GMT)
[verbose]	Successfully streaming http://data.nextstrain.org/flu_avian_h5n1_ha.json
```
---

**Failure example no. 2** Similar to above (but different dataset, so there's no server cache yet). Firefox without dev tools open. Attempt to load https://nextstrain-s-upstream-e-qx0bjj.herokuapp.com/charon/getDataset?prefix=/flu/avian/h5n1/na directly from firefox. Invalid JSON (incomplete). Server logs:
```
2020-05-07T02:20:31.333511+00:00 app[web.1]: [verbose]	[fetch] GET http://data.nextstrain.org/flu_avian_h5n1_na.json

2020-05-07T02:20:31.525386+00:00 app[web.1]: [verbose]	[fetch] 200 OK http://data.nextstrain.org/flu_avian_h5n1_na.json 

2020-05-07T02:20:31.525585+00:00 app[web.1]: [verbose]	Successfully streaming http://data.nextstrain.org/flu_avian_h5n1_na.json

2020-05-07T02:20:32.525282+00:00 heroku[router]: at=info method=GET path="/charon/getDataset?prefix=/flu/avian/h5n1/na" host=nextstrain-s-upstream-e-qx0bjj.herokuapp.com request_id=50be80e1-e540-4395-b84f-4a25e91d53bd fwd="203.173.142.4" dyno=web.1 connect=1ms service=1194ms status=200 bytes=546951 protocol=https
```

Refreshing the page (i.e. making another request to the same endpoint) works. Server logs:

```
2020-05-07T02:22:38.735220+00:00 app[web.1]: [verbose]	[fetch] GET http://data.nextstrain.org/flu_avian_h5n1_na.json

2020-05-07T02:22:38.737938+00:00 app[web.1]: [verbose]	[fetch] 200 OK http://data.nextstrain.org/flu_avian_h5n1_na.json (from cache, timestamp Thu, 07 May 2020 02:20:32 GMT)

2020-05-07T02:22:38.738016+00:00 app[web.1]: [verbose]	Successfully streaming http://data.nextstrain.org/flu_avian_h5n1_na.json

2020-05-07T02:22:39.103426+00:00 heroku[router]: at=info method=GET path="/charon/getDataset?prefix=/flu/avian/h5n1/na" host=nextstrain-s-upstream-e-qx0bjj.herokuapp.com request_id=b3ce1303-0a15-4aa1-a85d-c393e0b07e0c fwd="203.173.142.4" dyno=web.1 connect=1ms service=372ms status=200 bytes=545531 protocol=https
```

(Note that the number of bytes are differ, but not in the way I expected)

---

I'll see if I can disable the cache, relaunch the review app, and see if these bugs persist.

src/getDataset.js
```diff
@@ -54,25 +54,24 @@ const requestV1Dataset = async (metaJsonUrl, treeJsonUrl) => {
  * then the promise will reject.
```
jameshadfield, line 54: P.S. the comment immediately preceding the function is now out-of-date with this PR

jameshadfield:
Switching from `make-fetch-happen` to `node-fetch` (41320f2) and running similar test examples as above (but with different datasets, in case the cache persists on the heroku box between redeploys) works first time. Tested URLs:

* https://nextstrain-s-upstream-e-qx0bjj.herokuapp.com/flu/avian/h5n1/pb1 with devtools open & browser cache disabled
* https://nextstrain-s-upstream-e-qx0bjj.herokuapp.com/charon/getDataset?prefix=/flu/avian/h5n1/pb2 with devtools open & cache disabled

@tsibley I'll let you choose the implementation details (feel free to chuck 
41320f2) , but this PR seems good to go from my end _if_ the cache behavior is disabled (or fixed if you want to invest time there). Thanks!

tsibley:
Never got back to this, and its now obsoleted by https://github.com/nextstrain/nextstrain.org/pull/557.

PR #148, opened 2020-04-10
Add regional builds to splash page & squash tiles.

jameshadfield:
Here we add the regional builds as tiles to the splash page. 

As we now have so many nCoV cards (due to all the amazing translations) we set these cards to be "squashed" so that they take up less vertical space. 

We also make minor changes to the title: 
```diff
- Novel coronavirus (2019-nCoV)
+ Novel coronavirus

- We are incorporating nCoV genomes as soon as they are shared and providing analyses and situation reports.
- Please see below for the latest updates.
+ We are incorporating SARS-CoV-2 genomes as soon as they are shared and providing analyses and situation reports.
+ Please see below for the latest analyses and situation reports.
```

### DIFF
--- a/static-site/src/components/Cards/coreCards.js
+++ b/static-site/src/components/Cards/coreCards.js
@@ -2,7 +2,7 @@ const coreCards = [
   {
     img: "ncov.png",
     url: "/ncov",
-    title: "nCoV 2019-20"
+    title: "SARS-CoV-2"
   },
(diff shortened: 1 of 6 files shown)


trvrb:
Looks great! Would be nice to also update the tile below from "nCoV 2019-20" to "SARS-CoV-2".

PR #156, opened 2020-04-29
Show previous COVID-19 situation reports

### LINKED ISSUE #153: Show previous COVID-19 situation reports on Nextstrain website
**Context**  
Now that we're publishing situation reports focused on specific regions, accessing previous reports would be really useful. For example, the North American reports are not updated every week. But if I want to read the most recent analysis on North America, I need to be able to navigate to a previous week's report. From the splash page, readers cannot do that. Instead, they have to know the URL of the report they want to read.  

**Description**   
From the splash page, I'd like to add a link to another web page showing previous situation reports with their translations. Because of the sheer number of these reports, this might be best done with a click and scroll box for date and a click and scroll box for language. Choosing a report by topic rather than date would also be a useful feature.

jameshadfield:
This PR introduces the machinery for how to show all (previous) hCoV situation reports. It implements it via a (gatsby) page available at `/ncov-sit-reps`. 

This PR is meant as a starting point for @eharkins (and/or others), and is not ready to be merged. Big unimplemented features:
* Presentation (styling). It would be cool to use elements from the wonderful mocks that Lia has made for this. See [these](https://bedfordlab.slack.com/archives/C116R9PR7/p1586372885144200) [two](https://bedfordlab.slack.com/archives/C116R9PR7/p1586381547148200) slack messages.
* Error handling
* Better Integration / linking with the overall splash page

Current (very basic) state:
image

Will close #153. 
cc @cassiawag


eharkins:
@cassiawag redeployed this and it should be a little more fleshed out! Take a look when you get a chance and let me know if you have any feedback on the page layout or its linking with other pages in the static site (even though we have yet to implement any graceful error handling).

emmahodcroft:
This is looking really good @eharkins !

Just a couple of very minor comments - I think English might should go to the top of the languages, just due to traffic (I imagine - but if I'm wrong here, please ignore!). 

Also, we seem to maybe have a type in our situation report files in the directory (as I think these are being pulled in automatically) where the 1 May SitRep is in there twice, but once without a year? This isn't a problem with this code, I don't think - but the code is making it visible. @cassiawag maybe you can check this? We probably want to delete the file without a year!

image

eharkins:
Thanks @emmahodcroft! I'll defer to you and others more in tune with the sit reps traffic but can easily move English to the top if everyone agrees it should go there. 

If anyone is getting lots of duplicate English narratives like this please let me know:

Screen Shot 2020-05-07 at 10 29 37 AM

I can't reproduce it but when it happened I clicked on some duplicates and they did seem to render all as the same English narrative (as opposed to some other language that had defaulted to the english category or a genuinely different version of the same narrative with either the same name or a name not different enough to be caught by the parsing code).

This leads me to believe it requested duplicate narratives in error somehow (since there don't seem to be more than just the dozen or so English narratives in nextstrain/narratives).

If anyone has other theories I'm all ears!

emmahodcroft:
Hmm.. for me seems all good now - certainly not as extreme as you saw - incredible! 😆 
image

cassiawag:
I think this looks great @eharkins! I'm not seeing the bug that you describe.

I would change the text on the splash page to: "See previous situation reports" as it's unclear if "See all situation reports" refers to translations or previous week's reports. I'd also considering changing the text "translated into a number of different languages by volunteers and Google" to something like "...and Google-provided translators." I want it to be clear that we didn't put the sit-reps in Google translate, lol.

I also agree with Emma that English should be at the top. But for the rest, alphabetically is great. I really need to reorganize the splash page to be alphabetic as well! It's currently in order of translations received, which nobody actually knows.

We're adding on 6 new languages this week. Will this automatically update with those new languages?

cassiawag:
@emmahodcroft I deleted the file without a year. No idea how it got there, probably some inadvertent copying and pasting tom-foolery on my part.

eharkins:
@cassiawag thanks for looking through! I will make those changes to the text and move English to the top. New languages should require no changes to the code so long as they are in the [dictionary of languages I added](https://github.com/nextstrain/nextstrain.org/blob/show-all-sit-reps/static-site/src/components/Cards/languages.js) and their corresponding narratives follow the same pattern of:
`narratives/ncov/sit-rep/<ISO 2 Letter Language Code>/<narrative version/date name>`
e.g.
`narratives/ncov/sit-rep/zh/2020-04-10`

jameshadfield:
Thanks @eharkins -- this is looking really good. 

**Ordering of languages**

I agree English should be on top. I'm fine with then using alphabetical ordering, but @cassiawag may want to use a different ordering? 

I'd make this PR start with the English section "open". A future PR (or this one if it's easy) could use Jover's code which guesses the browser language and opens the corresponding section as applicable. 

**Stylistic changes**

I played around for 5 minutes to make the titles-lines look a bit more like Lia's mocks and came up with
```js
// componants/Misc/collapse-title.jsx
const TitleContainer = styled.div`
  display: flex;
  justify-content: space-between;
  border-top: 1px solid #CCC;
  margin: 0px;
  padding: 15px 10px 10px 10px;
`;
/* The icon container should use an actual icon, not text as it currently does */
const IconContainer = styled.span`
  font-size: 24px;
  font-weight: 500;
`;
const Name = styled.span`
  font-size: 20px;
  font-weight: 500;
`;
const CollapseTitle = ({name, isExpanded=false}) => (
  <TitleContainer>
    <Name>{name}</Name>
    <IconContainer>{isExpanded ? "x" : ">"}</IconContainer>
  </TitleContainer>
);

// ncov-sit-reps.jsx
<Collapsible
    triggerWhenOpen={<CollapseTitle name={language.nativeName} isExpanded />}
    trigger={<CollapseTitle name={language.nativeName} />}
    triggerStyle={{cursor: "pointer", textDecoration: "none"}}
  >
```
image

I think this is a lot cleaner. I'd also be in favor of the following changes:
* Adding an icon to each of the sit-rep links (see Lia's mocks)
* Making slightly smaller text for those links
* Removing the "(LANGUAGE)" from those links, as we've now grouped by language
* Intro paragraph text made smaller
* Actual icons used for drawer open/close indicators (but this isn't blocking, can do later)

**Main Nextstrain Splash Page**

@cassiawag @emmahodcroft what are your thoughts on removing the language cards on the main splash and having a link along the lines of "See all translated situation reports, including historical reports". That would mean a single card on the splash page with the latest sit-rep in English. The link could be more prominent than it currently is.

**Error Handling**

Reminder for this. A simple rendering of `<CenteredFocusParagraph>` with something along the lines of "There appears to have been an issue accessing the list of available situation reports! Please [get in touch)(mailto) or [make an issue on GitHub]() if this error persists!" would be ok for now. I don't think it's worth implementing a loading spinner or message as the API call will return almost instantaneously in most cases.

cassiawag:
Awesome! A couple of languages are not precisely listed in the 2-letter code. For example, we're translating into Filipino but that's a dialect of Tagalog, so it's 2-letter code is tl. For Portuguese, we're translating into PT Brazil, but the 2-letter code does not make that distinction. 

I'm trying to think how to best handle these exceptions; should we customize the dictionary to refer to the appropriate dialects that we're using? I'm still thinking on it.

jameshadfield:
> I'm trying to think how to best handle these exceptions; should we customize the dictionary to refer to the appropriate dialects that we're using? I'm still thinking on it.

I'd recommend us not using a "catch all" dictionary approach, rather maintain a small dictionary specific to this page mapping 2-letter-keys to display-title. Can fallback to using the 2-letter-key as title if we haven't updated the dictionary. We don't have enough language support in nextstrain.org to invest in a universal approach yet, in my opinion.

eharkins:
Sorry, didn't see the latest comments from @jameshadfield and @cassiawag after my last comment until pushing these commits ^! My latest changes should:
-  fix the duplication which was caused by not starting with a fresh copy of the language dictionary each time the page was rendered, allowing the same narratives to keep getting added to the same language object.
- put English first and the rest of the languages alphabetically afterward
- make some of Cassia's recommended wording changes (I'll stay tuned for any more updates to this wording here)
- add some basic error handling as James suggested.

Thanks for the continued feedback, it all makes sense and I'll get started on making it all happen first thing tomorrow.

emmahodcroft:
> I'd make this PR start with the English section "open". A future PR (or this one if it's easy) could use Jover's code which guesses the browser language and opens the corresponding section as applicable.

This would be awesome!

>Adding an icon to each of the sit-rep links (see Lia's mocks)
Making slightly smaller text for those links
Removing the "(LANGUAGE)" from those links, as we've now grouped by language
Intro paragraph text made smaller
Actual icons used for drawer open/close indicators (but this isn't blocking, can do later)

I'm in favour of all of these. By icons, we mean the little page symbols, right? 

>what are your thoughts on removing the language cards on the main splash and having a link along the lines of "See all translated situation reports, including historical reports". That would mean a single card on the splash page with the latest sit-rep in English. The link could be more prominent than it currently is.

Hopefully picturing the same thing are you, but what you mean is to have one 'card' which is to the English Sit Rep, and then a link to all the others? I think that would be fine. The sit rep translations are taking up quite a lot of the main page at this point and I think it would be good to free up some of the space. 


Finally, couple of last suggestions:

Can we reverse-order the past translation links? So that the first one is the most recent (2020-05-01) and then they go backwards to the oldest one. This way it's always easy to find the most recent - always in the same place (first).

I also wonder - this may be too hard, but could we bold or otherwise highlight the most recent one? I just think this is useful to call attention to it, since I imagine in most cases that's what people want to reach. But maybe that's difficult! Not a big deal.

cassiawag:
Sorry @jameshadfield, just now seeing your comment:
- Alphabetical is great! I've been meaning to reorder the splash page to this for a while.
- Really like all of your design suggestions.
- I'm okay with moving translations to this page as long as the link is quite prominent. They currently take up an absurd amount of space on the main page. It will be more difficult to find translated reports if you don't speak English because the text to navigate to those reports will be English. I don't have any good solutions to this though -- hope that browser's have a decent auto translate?
- For a custom dictionary, the languages we're adding on this week are Bengali (bn), Filipino (tl), Urdu (ur), Swahili (sw), Hindi (hi), Vietnamese (vi).

emmahodcroft:
We are incredibly timely! 😆 
https://twitter.com/RELenski/status/1258773895773069313

cassiawag:
Just spent some time with Google Analytics, as best as I can tell, people are navigating to sit-reps through direct links rather than through the splash page. So it should be okay if'ts more difficult to navigate to translations from the splash page.

eharkins:
^ should add all the recommended style and ordering changes from @jameshadfield and @emmahodcroft.

If that looks good, I'll move on to:
- switch from universal language dictionary to custom one
- a first version of:
>  removing the language cards on the main splash and having a link along the lines of "See all translated situation reports, including historical reports". That would mean a single card on the splash page with the latest sit-rep in English. The link could be more prominent than it currently is.
- using Jover's code to set the open language according to browser settings. @jameshadfield would you mind pointing me to this code when you get a chance?

eharkins:
Also I think languages are being sorted alphabetically by their English name rather than by country code. I'm assuming the later is more standard?

eharkins:
> 1. switch from universal language dictionary to custom one
> 2. a first version of:
removing the language cards on the main splash and having a link along the lines of "See all translated situation reports, including historical reports". That would mean a single card on the splash page with the latest sit-rep in English. The link could be more prominent than it currently is.
> 3. using Jover's code to set the open language according to browser settings. @ jameshadfield would you mind pointing me to this code when you get a chance?

Above changes should do all but 3 (for which it looks like the relevant code is [here](https://github.com/nextstrain/nextstrain.org/blob/master/redirects.js#L37) @jameshadfield?).

cassiawag:
Looks so great @eharkins! 

One final change from me if others are okay with it, change text on splash page to say:
"See all translated situation reports and all previous reports"

Otherwise, I'm not sure it's clear that historical reports includes English reports as well.

emmahodcroft:
This is looking great to me @eharkins ! Thank you for all the hard work on this!

eharkins:
> For example, we're translating into Filipino but that's a dialect of Tagalog, so it's 2-letter code is tl. For Portuguese, we're translating into PT Brazil, but the 2-letter code does not make that distinction.

@cassiawag here are the entries for:
- [Tagalog](https://github.com/nextstrain/nextstrain.org/pull/156/files#diff-8e56725b3447679dcf1c6b67981162afR96-R99)
- [Brazilian Portuguese](https://github.com/nextstrain/nextstrain.org/pull/156/files#diff-8e56725b3447679dcf1c6b67981162afR80-R83) - to show up separately from other `pt` narratives, they should use `pt-br`. I think this is ok with narrative parsing code to use the hyphen, but if not we can just go with `ptb`?

cassiawag:
@eharkins, I'm not concerned about separating `pt` from 'pt-br` as both languages are mutually intelligible, and there was one or two weeks that were Portuguese. 

I've been thinking about dictionary; can you just drop whatever is after the `,` in nativeName? This would be similar to what Jover did with the [language drop down](https://nextstrain.org/narratives/ncov/sit-rep/2020-05-08?f_region=Asia) present in sit-reps. I don't think that level of distinction is necessary for most translations. The only other change is to update "Tagalog" to "Wikang Filipino" as that's the most accurate way of representing our translation.

I _think_ that's all my comments on the language dictionary. Sorry they've come in waves and not all at once.

eharkins:
>  I'm not concerned about separating pt from 'pt-br` as both languages are mutually intelligible, and there was one or two weeks that were Portuguese.

Ah I see, I thought we were just now adding in a separate Brazil translation and hoping to have it show up separately. I'll remove the pt-br entry.

> I've been thinking about dictionary; can you just drop whatever is after the , in nativeName? This would be similar to what Jover did with the language drop down present in sit-reps. I don't think that level of distinction is necessary for most translations. The only other change is to update "Tagalog" to "Wikang Filipino" as that's the most accurate way of representing our translation.

Thanks for pointing me to the language drop down! Some of the code I wrote (including the dictionary) seems redundant now that I've taken a look at that [drop down code](https://github.com/nextstrain/nextstrain.org/blob/master/auspice-client/customisations/languageSelector.js). 

I think it makes sense to reuse some of those functions which get the native name etc which should make them consistent between the drop down and the all-sit-reps page.

eharkins:
Ok I made it so this page uses the same language parsing code as the language dropdown from the narratives navbar. @cassiawag this should solve the issues with language dialects after commas / mismatches in general with the dropdown.

static-site/src/components/splash/index.jsx
```diff
@@ -65,6 +66,15 @@ class Splash extends React.Component {
           </Styles.CenteredFocusParagraph>
         </FlexCenter>
 
+        <FlexCenter>
+          <Styles.CenteredFocusParagraph theme={{niceFontSize: "20px"}}>
+            <Link to="/ncov-sit-reps">
+              <MdTranslate style={{fontSize: "28px"}}/>
```
jameshadfield, line 72: I'm not familiar with this icon, but happy to have it if it helps things.

static-site/src/pages/ncov-sit-reps.jsx
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  render() {
+    return (
+      <MainLayout>
+        <div className="index-container">
+          <Helmet title={config.siteTitle} />
+          <main>
+
+            <NavBar location={this.props.location} />
```
jameshadfield, line 79: The nav-bar component needs to be wrapped as follows such that it renders appropriatly for logged-in users

```js
import UserDataWrapper from "../layouts/userDataWrapper";
...
<UserDataWrapper>
      <NavBar location={this.props.location} />
</UserDataWrapper>
```

P.S. @eharkins ping me on slack & we can make you a username for nextstrain.org

auspice-client/customisations/languageSelector.js
```diff
@@ -1,16 +1,50 @@
 import React from "react"; // eslint-disable-line
 import ISO6391 from "iso-639-1/build/index";
 
-/* This function is duplicated from ../../src/utils but included here to
-(a) keep server & client code seperate and (b) avoid any transpiling errors
-during client building */
-const parseNarrativeLanguage = (narrative) => {
-  const urlParts = narrative.split("/");
+/* This and some of the following functions are reused
```
jameshadfield, line 4: (Notes from a conversation Eli & I had today) -- This solution is fine for now. As auspice's extension machinery improves (there are some promising PRs at the moment) then these functions can be lifted to a more central place within the repo.

static-site/src/components/splash/index.jsx
eharkins: Sorry, I meant to explicitly mention this!

Screen Shot 2020-05-13 at 9 17 01 AM


Adding this icon ^ was my attempt at a compromise solution to this issue brought up by @cassiawag about removing translated sitreps from the splash page:
> Though it will be more difficult to find translated reports if you don't speak English. Because the text to navigate to those reports will be English

I was hoping the icon might encourage non-english speakers to click the link despite it being in english. 

1. I don't know if that will really work 
2. It could be that this is misleading if that icon has traditionally meant "translate this page into current browser settings language" (although if that's the case, linking to the translated sitreps page seems like the next best thing to translating the splash page).
3. Maybe the issue itself isn't make or break as Cassia suggested

So I'm happy to remove it if we don't think it serves as a good solution.

emmahodcroft: It looks like even if not this exact icon, variations of this are commonly used to mean translations. There are some nice other options on Google (With arrows going between the two letters, for example) that might help convey the idea, if we worry about it!

jameshadfield: Great -- I think it's should stay as is :)

jameshadfield:
Thanks @eharkins, @cassiawag & @emmahodcroft!

PR #168, opened 2020-05-15
link out data formats to augur and auspice docs

eharkins:
Link out any details on augur output and auspice input to their respective docs so they don't get out of sync with those docs, while providing the necessary high-level info about how the two relate when it comes to data formats. The ideal goal would be to inject that content into nextstrain.org docs to avoid users having to follow too many links.

Also adds a link to [the new page describing ways to share results](https://nextstrain.org/docs/contributing/sharing-data)

### Related issue(s) 
#164
https://github.com/nextstrain/auspice/pull/1127

### Testing
Tested links worked locally with Gatsby.


jameshadfield:
Brief message to say that Eli & I chatted about this PR and are going to re-frame this PR so that v2 datasets are front-and-center, with v1 datasets getting a small mention to indicate they are still supported but you're encouraged to update.

eharkins:
Should be clear now that Auspice uses v2 and v1 should only be used for existing builds requiring back compatibility.

PR #170, opened 2020-05-22
link out instructions from defining clades doc to augur docs

eharkins:

1. Replace the defining clades tutorial with links to [defining clades](https://nextstrain-augur.readthedocs.io/en/stable/faq/clades.html) faq page in augur and to [the augur docs tb tutorial](https://nextstrain-augur.readthedocs.io/en/stable/tutorials/tb_tutorial.html#export-the-results) as an example of how to export clade information.
2. Move the defining clades doc to the bioinformatics section from the tutorials section
3. Remove some unused docs images

I removed a solid chunk of explanation about defining clades from the nextstrain.org defining clades page, not all of which is made up for in a one-to-one way in the linked pages that replace 
it. We should make sure no valuable information was lost and if it was to include it in the augur pages.

### Related issue(s)  
#164 

### Testing
Locally with gatsby


eharkins:
@jameshadfield sorry for the delay!
> After the changes of #195 do you think the best path forward would be to add any necessary information to augur's "defining clades" page, and then remove the tutorial currently found on nextstrain.org? We can add a redirect and an external link in the bioinformatics section of the nextstrain.org/docs sidebar.

That makes sense to me for now, although it seems there is some information re: interacting with clades in the UI that feels like it belongs more in nextstrain.org (perhaps in /help) or auspice docs than in augur:
> Then, head back to the tree. When you hover over a branch in the tree, a dark grey window pops up. Inside this window, you’ll see that there’s a header that says Nucleotide Mutations:. In the image below, the clade descended from this branch is defined by a mutation from G to A at site 9417 in the genome. This is the information that you’ll need for defining your clades.

This is a good example of a concept that touches multiple parts of the nextstrain stack - so long term it will probably fit best located at the top level of our docs, adjacent to, instead of within docs headings about augur or auspice.

I'm happy to submit a PR merging them into augur docs for now.

eharkins:
Moved to https://github.com/nextstrain/docs.nextstrain.org/issues/35

Issue #158, opened 2020-04-30
Allow gatsby dev server to use auspice server 


eharkins:
**Context + Description**  
It's not currently possible to run a gatsby dev server (to develop things like the splash page) and also have auspice serve and visualize datasets. That is, live development of the static site excludes any use of the rest of the site.

This means options for developing the static site are:
- run `cd static-site && gatsby develop` but not have the datasets, narratives be served by auspice
- run `npm run server` and rebuild gatsby each time you want to see changes to stuff in `static-site`

**Examples**  
This would make developing pages like https://github.com/nextstrain/nextstrain.org/pull/156 much easier where we are developing the static site's presentation of available sit rep narratives and so we need to be able to call `/charon/getAvailable`(which only comes from the nextstrain.org server, not gatsby) to get those narratives but also want to not have to rebuild the site to see our cosmetic changes.

**Possible solution**   
@jameshadfield said
> There are a couple of options worth exploring:
(1) Run both — gatsby dev server on port 8000 and npm run server on 5000 and try to get the API request from gatsby to go to port 5000. Is this possible? Not sure
(2) Mock the api call by replacing the fetch call in the React component with a hardcoded array which is similar to what the API would have returned. Should be pretty easy for a one-off.
(3) See if it’s possible to run the gatsby dev server from within the nextstrain.org server. I imagine this is hard.

Personally I'm using 2. at the moment but James notes that long term, implementing 1. would be 
> pretty cool as we do more and more development, and shouldn’t be too hard with a nginx like reverse proxy. But not really worth investing in time at the moment.

However, if there are gatsby pros out there reading this, have at it 🏄

jameshadfield:
This is becoming a higher priority. 

A bit of history: There used to be a clear-cut separation between the "static-site" (built using gatsby) and the customised auspice client. The former could be developed using the gatsby dev server, while the latter could be developed [using the auspice dev server](https://github.com/nextstrain/nextstrain.org#testing-locally). 

As we've added functionality (e.g. authentication, a page which lists all the nCoV sit-reps), the "static-site" relies on API endpoints implemented in the nextstrain.org server, thus making the gatsby dev server inadequate. 

There's no hard necessity to have a hot-reloading dev-server, but it's quite a roadblock to future development.

didithilmy:
Hello, I'm new here, but I think I might have a solution.

[...]

PR #172, opened 2020-05-26
Add blurb clarifying various project docs

eharkins:
1. Add a blurb making it clear what information should go in the nextstrain.org documentation vs augur, auspice.
2. fix local static site build instructions. Despite my changes to the static site build instructions, I think this [importing code from auspice-client/customisations into the static-site](https://github.com/nextstrain/nextstrain.org/pull/156/files?file-filters%5B%5D=.json&file-filters%5B%5D=.jsx#diff-8e56725b3447679dcf1c6b67981162afR14 ) means we have to install the root repo node dependencies in addition to the static-site ones even if we just want to run the static site (i.e. if someone is following the instructions in the [document in question](https://nextstrain.org/docs/contributing/documentation#building-the-documentation-locally) here with a fresh clone). Although perhaps there is a way to set up node dependencies so this isn't necessary? If not we should change the instructions to something like `npm ci && cd static-site && npm ci`

### Related issue(s)  
#164 

### Testing
Locally with gatsby + testing links work

### DIFF
--- a/static-site/content/docs/05-contributing/03-documentation.md
+++ b/static-site/content/docs/05-contributing/03-documentation.md
@@ -5,9 +5,21 @@ title: "Contributing to Documentation"
 Documentation is crucial to a project's success, but often painful to write!
 We're always trying to improve our documentation here and would appreciate any and all contributions, whether you are a collaborator or simply a user.
 
+### Where documentation lives
+
+The docs you are reading here on nextstrain.org provide information on how the nextstrain.org website works.
+
+Augur and Auspice are tools we use to power nextstrain.org, and so they are introduced here in these docs. However, they are their own projects, with their own documentations and we make an effort to keep detailed usage documentation for each of those tools in their respective documentations:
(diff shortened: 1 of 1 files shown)


eharkins:
> Despite my changes to the static site build instructions, I think this importing code from auspice-client/customisations into the static-site means we have to install the root repo node dependencies in addition to the static-site ones even if we just want to run the static site (i.e. if someone is following the instructions in the document in question here with a fresh clone).

Given this, @jameshadfield and I have been discussing a good way to have the instructions until we have a better solution (see #158).

James suggests something along the lines of:
```
cd nextstrain.org
npm run install
npm run build # builds both auspice _and_ the static site
npm run server # see essentially a nextstrain.org clone
```

Along with a separate section for the dev server that says something like: “there’s a development server version as well, but it’s not perfect… here’s how you can use this”

tsibley:
:+1: for the install + build + server approach. That's what I've always done locally, as running the two "apps" separately doesn't realistically test changes anyway. (It's also possible to setup automatic re-runs of the build + server steps when files change, which I've used from time to time instead of hot reloading.)

PR #174, opened 2020-05-26
Link out duplicate narratives doc to auspice docs

eharkins:
Replace most of the content on https://nextstrain.org/docs/narratives/introduction with a link to https://nextstrain.github.io/auspice/narratives/introduction so they don't get out of sync.

I added a picture of an example COVID-19 narrative, but I'm open to removing it or replacing it with a link to that narrative (or another one) since there are a couple of potential issues with just using a screenshot:

* It doesn't show more than just the one page of the narrative and so users might miss the whole point of narratives being that you can have many pages to tell a story.
* It might be assumed that it is a live narrative embedded into the page and folks will be sad when they try to click on it.

### Related issue(s)  
#164

### Testing
Locally with gatsby + testing that links work


eharkins:
A gif might solve the issues listed above with the screenshot if those are the only concerns.

emmahodcroft:
Gif seems like a good compromise to me!

jameshadfield:
Thanks @eharkins -- I agree with this direction, apologies it's taken me so long to review this.

PR #195 has resulted in https://nextstrain.org/docs/narratives/introduction redirecting to https://nextstrain.github.io/auspice/narratives/introduction, as part of efforts to remove duplicated content. I see your point that it would be a good idea to add some SARS-CoV-2 specific text to that page and have created https://github.com/nextstrain/auspice/issues/1184 as a reminder

PR #178, opened 2020-05-29
add download data documentation

jameshadfield:
Copy edits etc welcome!

This PR must be merged after #176 as the downloading of diversity (entropy) data is only available with auspice 2.16.


emmahodcroft:
I'm not the best detailed copy-editor but I read through this and it looks good to me, certainly from a technical standpoint! Thanks for doing this James!

Issue #180, opened 2020-06-11
All ncov builds page


eharkins:
**Context**  
As discussed on a call today, we would like to have a page like https://nextstrain.org/ncov-sit-reps but that shows all the builds for ncov (global and regional).

**Description**   
UI details are an ongoing conversation and this issue is place to discuss them. Some suggestions included:
* a more or less identical UI to https://nextstrain.org/ncov-sit-reps
* a table showing various meta info about each build
* a searchable list
* an interactive map with links to builds

Perhaps we can also solicit more designs/advice from the UI designer who helped with designs for the splash page and https://nextstrain.org/ncov-sit-reps.
**Examples**  
https://nextstrain.org/ncov-sit-reps
https://bedfordlab.slack.com/archives/C116R9PR7/p1586372885144200

**Possible solution**   
Hopefully reuse lots of code from https://nextstrain.org/ncov-sit-reps

trvrb:
Here's my suggested UI:

![sars-cov-2-select](https://user-images.githubusercontent.com/1176109/84455869-53e39000-ac13-11ea-9828-4c71996e6f39.png)

Borrowing from my thoughts on generic dataset select from before https://github.com/nextstrain/nextstrain.org/issues/62.

emmahodcroft:
The downside of this is that I feel it's really not useful to figure out where I need to go if I want to see a dataset of Spain, Switzerland, or Connecticut. I think we need to give some more geographical guidance here. I have to say, this is a pretty big turn-off for me. I think unless you know these groups to start with (and what region they're likely to work with), this isn't very usable to find a dataset.

I recognise this means there'll possibly have to be come kind of more manual mapping between the groups/communities & what countries/states offer, but I think we kind of need to do this somehow - otherwise you have to expand every single one of these to figure out what locations are available.

I also think it's kind of weird that the main Nextstrain builds are listed at the same level as community & other builds. I think it would be ok to highlight these a little more - these are probably still what a large number of people want - maybe through the half-size tiles that James made? Or even quarter-size. 

Though this definitely won't be sustainable long-term, I think I'd rather have:
Nextstrain builds at the top, as some kind of quarter-size tiles or similar.
Grouping by region, and within each of those, the builds available, saying what country/state they are for. Possibly groups by country/state in another expandable, if needed.
The group who is maintaining it should be listed, but the focus should be on the region covered, IMO.

tsibley:
[...]

PR #185, opened 2020-06-19
first draft of sars-cov-2 builds page;

eharkins:
see https://github.com/nextstrain/nextstrain.org/issues/180

I will fill this description out with more TODOs etc but just creating the draft for now so that we have a place to iterate on the design, etc. 

### Related issue(s)  

Fixes #  
Related to #  

### Testing

### DIFF
--- a/static-site/src/pages/sars-cov-2.jsx
+++ b/static-site/src/pages/sars-cov-2.jsx
@@ -18,8 +18,8 @@ import allSARSCoV2Builds from "../../content/allSARS-CoV-2Builds.yaml";
 */
 
 const buildComponent = (build) => (
-  <div key={build.url}>
-    <splashStyles.SitRepTitle >
+  <splashStyles.SitRepTitle >
+    {build.url === null ? build.name : <div>
       <a href={build.url}>
         <FaFile />
         {` ${build.name} `}
@@ -28,8 +28,9 @@ const buildComponent = (build) => (
       {build.org.url === null ? build.org.name : <a href={build.org.url}>{build.org.name}</a>
       }
       )
-    </splashStyles.SitRepTitle>
-  </div>);
+    </div>}
+  </splashStyles.SitRepTitle>
+);
 
 // eslint-disable-next-line react/prefer-stateless-function
 class Index extends React.Component {
@@ -38,6 +39,27 @@ class Index extends React.Component {
     this.state = {
       hasError: false};
     this.buildsForGeo = this.buildsForGeo.bind(this);
+    this.subBuilds = this.subBuilds.bind(this);
+    this.buildTree = this.buildTree.bind(this);
+  }
+
+  subBuilds(parentBuild) {
+    const children = allSARSCoV2Builds.builds
+      .filter((b) => b.parentGeo === parentBuild.geo);
+    return (
+      <div key={parentBuild.url+parentBuild.name}>
+        {buildComponent(parentBuild)}
+        {children.length > 0 && children.map((child) => (
+          <div key={`${child.url+child.name}-children`} style={{marginLeft: "20px"}}>
+            {this.subBuilds(child)}
+          </div>
+        ))}
+      </div>);
+  }
+
+  buildTree() {
+    const roots = allSARSCoV2Builds.builds.filter((b) => b.parentGeo === null);
+    return roots.map((parentBuild) => this.subBuilds(parentBuild));
   }
 
   buildsForGeo(geo) {
@@ -76,7 +98,7 @@ class Index extends React.Component {
                                   Something went wrong getting data.
                                   Please <a href="mailto:hello@nextstrain.org">contact us at hello@nextstrain.org </a>
                                   if this continues to happen.</splashStyles.CenteredFocusParagraph>}
-                  {allSARSCoV2Builds.builds.map((build) => buildComponent(build))}
+                  {this.buildTree()}
                 </div>
               </div>
 
--- a/static-site/src/pages/sars-cov-2.jsx
+++ b/static-site/src/pages/sars-cov-2.jsx
@@ -43,23 +43,25 @@ class Index extends React.Component {
     this.buildTree = this.buildTree.bind(this);
   }
 
-  subBuilds(parentBuild) {
+  subBuilds(header, fontSize=20) {
     const children = allSARSCoV2Builds.builds
-      .filter((b) => b.parentGeo === parentBuild.geo);
+      .filter((b) => b.geo === header.geo && b.url !== null);
+    const subHeaders = allSARSCoV2Builds.builds
+      .filter((b) => b.parentGeo === header.geo && b.url === null);
     return (
-      <div key={parentBuild.url+parentBuild.name}>
-        {buildComponent(parentBuild)}
-        {children.length > 0 && children.map((child) => (
-          <div key={`${child.url+child.name}-children`} style={{marginLeft: "20px"}}>
-            {this.subBuilds(child)}
-          </div>
-        ))}
+      <div key={header.name}>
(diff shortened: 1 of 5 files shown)


static-site/src/pages/sars-cov-2.jsx
```diff
@@ -0,0 +1,130 @@
...
+              <div className="row">
+                <MediumSpacer />
+                <div className="col-md-1"/>
+                <div className="col-md-10">
+                  { this.state.hasError && <splashStyles.CenteredFocusParagraph>
+                                  Something went wrong getting data.
+                                  Please <a href="mailto:hello@nextstrain.org">contact us at hello@nextstrain.org </a>
+                                  if this continues to happen.</splashStyles.CenteredFocusParagraph>}
+                  {/* hardcoded structure for now */}
+                  <splashStyles.H3 left>Global</splashStyles.H3>
```
jameshadfield, line 81: I think we'll want to represent a lot of this structure in the YAML itself, rather than as JSX, as then data updates are as simple as updating the YAML and require no changes to the page itself. We could start with a solution as simple as maintaining the order of entries in the YAML, and implement more complicated display as needed.

static-site/src/pages/sars-cov-2.jsx
```diff
@@ -0,0 +1,130 @@
...
+      )
+    </splashStyles.SitRepTitle>
+  </div>);
+
+// eslint-disable-next-line react/prefer-stateless-function
+class Index extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      builds: allSARSCoV2Builds.builds,
```
jameshadfield, line 39: Is this data stored as `state` in preparation for a future where we may fetch the YAML rather than import it?

jameshadfield:
This is a great start -- thanks @eharkins 👍

static-site/src/pages/sars-cov-2.jsx
eharkins: Yes, but maybe we can just cross that bridge when we get there instead and leave it out of the component state.

static-site/src/pages/sars-cov-2.jsx
eharkins: Makes sense. I was trying to adhere to the hierarchy proposed by @tsibley [here](https://github.com/nextstrain/nextstrain.org/issues/180#issuecomment-645657684) without over developing the YAML design or the parsing code just yet. But I agree that this 
> represent a lot of this structure in the YAML itself, rather than as JSX, as then data updates are as simple as updating the YAML and require no changes to the page itself

seems like a priority over having a complex structure / hierarchy displayed

tsibley:
Agreed, great start on this!

static-site/src/pages/sars-cov-2.jsx
tsibley: :+1:

emmahodcroft:
So I'll admit, I didn't really understand the discussion above until I tried to modify the page today 😆 I got the hang of it (I think) but it might be more intuitive to have this all in the YAML somehow. However, what we've got is also fine, especially since we think this is not intended to be long-term :)

emmahodcroft:
I wasn't sure how to handle having 2 California builds & 1 San Diego build, either. I suppose it would also be able to be good to have a clickable header (example - if Massachusetts got a build) rather than having to list states under states - but when we have two... unsure!

static-site/src/pages/sars-cov-2.jsx
eharkins: b7c79d6

eharkins:
>  I suppose it would also be able to be good to have a clickable header (example - if Massachusetts got a build) rather than having to list states under states - but when we have two... unsure!

@emmahodcroft I removed the structure built into the component (d0424bc) so they just show up in the order they are listed without any structure for now. I agree with @jameshadfield saying that we can build structure into the YAML so that we don't have to edit the code to update the hierarchy. Want to allow for discussion of the YAML design here by everyone and not just assert something myself.

I agree about clickable headers which we can totally do!

emmahodcroft:
> @emmahodcroft I removed the structure built into the component (d0424bc) so they just show up in the order they are listed without any structure for now.

While I think it would be great to iterate on other designs/ideas, I'd be in favour of keeping this in place until we have another idea! Particularly if we want to make this public before the SPHERES call on Wednesday :) I think it does make it easy to interpret, and we can still play around with ideas as they come!

I do think we should maybe remove the redirect from this PR, we can make this publicly-accessible sooner (while we continue to iterate) but not redirect from `/ncov` until we've got a 'final product' :D

eharkins:
Two commits above ^:
- took out the redirect from /ncov -> /sars-cov-2 for now
- added in a `parentGeo` field to each build in the YAML, so that we can specify the structure this way. If there are any issues with this and we need to just go back to the initial commit's implementation, I'll gladly revert my changes! This allows for 
> clickable headers

as we have been calling them, aka nesting builds directly under other builds. This is actually the default unless you specify an empty/header build [like this](https://github.com/nextstrain/nextstrain.org/pull/185/files#diff-249fe4d939f224507f671ed1c289b115R58-R62) to be used as a header _without_ a build link.

emmahodcroft:
Thanks @eharkins !! This is looking fantastic - and nice idea with the `parentGeo` - I think this'll make the nesting very intuitive!

trvrb:
Nice work here! My only real comments are aesthetic:

1. I'd better organize headings like what @tsibley has here: https://github.com/nextstrain/nextstrain.org/issues/180#issuecomment-643433439, where there is an obvious hierarchy given by font size.

2. I also like @tsibley's example where the geo descriptor (say "North America") is separate from the build label itself. Otherwise, I find it strange that "California" gets non-linked text, while "Wisconsin" is a direct link.

3. The folded document icon works for a "situation report". I don't think it makes sense for an interactive visualization. This is semantically more like "view" in my mind. Something like: https://fontawesome.com/icons/project-diagram?style=solid or https://fontawesome.com/icons/chart-area?style=solid.

eharkins:
That all makes sense to me @trvrb! With respect to 
> I also like @tsibley's example where the geo descriptor (say "North America") is separate from the build label itself. Otherwise, I find it strange that "California" gets non-linked text, while "Wisconsin" is a direct link.

I think this allows for more consistency across regions with multiple builds (e.g. California) and regions with just one top level regional build (e.g. Wisconsin). California gets "non-linked text" in the current implementation because as @emmahodcroft [alluded to](https://github.com/nextstrain/nextstrain.org/pull/185#issuecomment-647019424), I couldn't think of a good way to resolve how to nest sub-California builds (which build do they go under?) when there are multiple California top-level builds. This results in the California level builds being defined in the YAML inconsistent with Wisconsin. 

Here is what it looks like (along with your other suggested changes) when we define/parse the YAML such that each geo region gets its own non-linked header:

Screen Shot 2020-06-22 at 6 12 31 PM

 The way I did this means that all headers have to be defined in the YAML [similar to how non-linked headers must be defined in the current implementation](https://github.com/nextstrain/nextstrain.org/pull/185/files#diff-249fe4d939f224507f671ed1c289b115R58-R62). If this makes sense and is what we want I can push these changes up.

jameshadfield:
This looks great @eharkins

emmahodcroft:
I like the look of this very much @eharkins ! I think we're just about ready to merge. You want to push up when you're ready and we can peek at the deploy? But I think apart from a check this is looking great!

eharkins:
Do we want to add a link somewhere on the splash page or will we just share the /sars-cov-2 path with a limited audience at first?

emmahodcroft:
Thank you so much @eharkins ! Looking fantastic!

PR #187, opened 2020-06-19
Add page where you can search samples across datasets

jameshadfield:
This introduces functionality which we've talked about previously but was sorely needed by me today! It works, but there are plenty of optimisations and improvements we could make. Consider this a Friday afternoon project (which is what it was). 

During `npm run build` we scan all the nCoV datasets (currently looking at the staging bucket) and generate a git-ignored JSON file mapping the sample names to which dataset they appear in.

A new page /ncov-sequence-to-dataset provides a search box For a single sample name, you'll get a chronological list of datasets which it appears in. For multiple samples, you'll get a list of datasets where they all appear together.

Limitations:
* React Select is slow with this many items
* The JSON file is only regenerated when the server restarts (when `npm run build` is run)
* Currently looking at the staging bucket only
* Styling needs improvement

Needs some more testing but works for the ones I was looking at so far. 

image
image
image

### DIFF
--- a/scripts/collect-strains-sars-cov-2.js
+++ b/scripts/collect-strains-sars-cov-2.js
@@ -1,0 +1,75 @@
+#!/usr/bin/env node
+
+const AWS = require("aws-sdk");
+const fetch = require("node-fetch");
+const fs = require('fs');
+
+const collectStrainNames = (name, json) => {
+  const strains = [];
+  const recurse = (node) => {
+    if (node.children) {
+      for (let i=0; i<node.children.length; i++) recurse(node.children[i]);
+    } else {
+      strains.push(node.name);
+    }
+  };
+  try {
+    recurse(json.tree);
+  } catch (err) {
+    console.log(`Error collecting strain names for ${name}`);
+    return [];
+  }
+  return strains;
+};
+
+(async () => {
+  const S3 = new AWS.S3();
+  const BUCKET = `nextstrain-staging`;
+  const NEXTSTRAIN_URL_PREFIX = `https://nextstrain.org/staging/`;
+  let s3Objects;
+  try {
+    s3Objects = await S3.listObjectsV2({Bucket: BUCKET}).promise();
+  } catch (err) {
+    console.log("Error listing objects via the S3 API -- were credentials correctly set?");
+    console.log(err.message);
+    process.exit(2);
+  }
+  if (s3Objects.isTruncated) console.log("WARNING: S3 listing is truncated. Results will be incomplete.");
+  s3Objects = s3Objects.Contents
+    .filter((s3obj) => {
+      if (!s3obj.Key.endsWith(".json")) return false;
+      if (s3obj.Key.endsWith("_meta.json") || s3obj.Key.endsWith("_tree.json")) return false;
+      if (s3obj.Key.endsWith("_frequencies.json")) return false;
+      if (s3obj.Key.endsWith("_tip-frequencies.json")) return false;
+      if (s3obj.Key.endsWith("_sequences.json")) return false;
+      if (s3obj.Key.endsWith("_entropy.json")) return false;
+      if (s3obj.Key.endsWith("_root-sequence.json")) return false;
+      return true;
+    })
+    .filter((s3obj) => {
+      return s3obj.Key.startsWith("ncov_");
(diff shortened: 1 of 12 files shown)


emmahodcroft:
You are a mind-reader James! I was just thinking about this in the last few weeks. I think this idea is great and will be really useful for people! If we expanded what we sent to S3 (or git), to include `ncov/results/flagged-sequences.tsv` could we also search this and let people know if the sequence they're looking for got dropped due to diverge issues? That would also be useful to know!

emmahodcroft:
This would also be super useful for 'build shepherds', as we are often looking to see which seqs from a run 'made it in' and where (to get photos).

jameshadfield:
Great suggestion @emmahodcroft -- I've modified things so that we can now indicate if a sample is excluded because it is present in [exclude.txt](), and the reason why:

image

It's not clear to me the best way to add data such as `ncov/results/flagged-sequences.tsv` which are dynamically generated per-run and not really stored anywhere that's accessible. I think it'd be a good future inclusion useful for data producers as well as build shepherds.

jameshadfield:
While the travis-CI is failing, this is available to interact with via https://nextstrain-s-strain-sea-bi7owm.herokuapp.com/ncov-sequence-to-dataset/

emmahodcroft:
Awesome!! Thanks James! This is sooo useful.

Yes, Unsure how best to tackle the `flagged_sequences.tsv` file. I'd be happy to add this to things that are 'pushed' to github, though perhaps that's weird since it's in `results`, which, in general, we like to ignore. Another option would be to push this up to S3 with our 'official' runs (since this is what determines if it's in our datasets, after all). However, for this to be most likely to be remembered, we'd probably have to copy it over to `auspice` (so it goes up with the usual `deploy` command, rather than needing two)... then this would happen for everyone else's runs too (harmless, but kinda weird...). 

We could also make `flagged_sequences.tsv` go to some other folder instead, and then push to git. Not really any easy ideal solution here, but I think this would be great to include if we could!

scripts/collect-strains-sars-cov-2.js
```diff
@@ -0,0 +1,75 @@
...
+
+(async () => {
+  const S3 = new AWS.S3();
+  const BUCKET = `nextstrain-staging`;
+  const NEXTSTRAIN_URL_PREFIX = `https://nextstrain.org/staging/`;
+  let s3Objects;
+  try {
+    s3Objects = await S3.listObjectsV2({Bucket: BUCKET}).promise();
+  } catch (err) {
+    console.log("Error listing objects via the S3 API -- were credentials correctly set?");
```
eharkins, line 33: Should we add something to docs for open source contributors about how unless you have nextstrain bucket AWS creds you should run `npm run build static && npm run build auspice` since `npm run build` will here crash otherwise (as it did for me since I have none set up for nextstrain right now)?

jameshadfield: Good idea -- documentation added and exit codes modified (which were one reason TravisCI was failing)

jameshadfield:
Thanks @eharkins & @emmahodcroft -- I've updated it to look at the `nextstrain-data` bucket, renamed the page contents & URL (now /sars-cov-2-sequence-search), and improved docs / error messages. I also added a fix for the heroku build failures when sourcing datasets from nextstrain-data, which were due to memory concerns.

With regards to the per-run exclusions, I see two paths going forward:
1. Maintain some database (or file on S3) that we update every time we release a build (i.e. run `nextstrain deploy` or similar) and maps sample name to a list of datasets where it was excluded. This could be updated via a AWS Lambda web-API or similar. This requires developer work to set up. 
2. Shift the dynamic exclusions into the static `exclude.txt` periodically. This requires manual work, and doesn't allow for the case where the data updates some time in the future such that the samples pass the dynamic QC steps.

emmahodcroft:
Thanks James, this is good to know. Sounds like having it live-update might be a bit more work, but there's no rush to this. I think the plan has always been to update `flagged-sequences` to `exclude` periodically, but of course this is one of those things that's hard to actually remember to do regularly. In fact, we really should do a run where we take a lot of things out of `exclude` (the ones with reasons like 'maybe too diverged?') and check them with the diagnostics...

Issue #154, opened 2020-04-24
API to access data on private buckets


jameshadfield:
**Context**  
Currently narratives (and the footer of datasets) can define images in the markdown for view, and often these are stored on the same S3 bucket as the dataset (for convenience, not necessity). For instance the 2020-01-23 narrative defines an image via
```html
<div>
  <img
        alt="cartoon of phylogenetic tree and corresponding alignment, with samples labelled A-E" 
        width="500"
        src="http://data.nextstrain.org/toy_alignment_tree.png"
   />
</div>
```
which can be seen rendered here: https://nextstrain.org/narratives/ncov/sit-rep/2020-01-23?n=4

If we are serving a narrative from a private group, then these URLs don't work (because the bucket requires authorization to access). 

**Description**   
A nextstrain.org API to access images (or other data files) from the bucket, iff the logged in user's authentication allows it.

tsibley:
Ideally the "API" here would just be relative URLs inside the Markdown. This is slightly hindered by the `foo_bar_baz.md` → `foo/bar/baz` rewriting, so images would have to share the same prefix as the Markdown file, but that's teneable.

PR #188, opened 2020-06-22
S3Source: return `group-logo.png` as avatar if found in S3 bucket

joverlee521:
This will allow Nextstrain groups to customize their logo/avatar on
the groups splash page by uploading `logo.png` to their S3 bucket.

### Testing
I've added `logo.png` for SeattleFlu, which can be seen at `localhost:5000/groups/seattleflu`

### DIFF
--- a/src/sources.js
+++ b/src/sources.js
@@ -1,5 +1,6 @@
 /* eslint-disable no-use-before-define */
 const AWS = require("aws-sdk");
+const zlib = require("zlib");
 const {fetch} = require("./fetch");
 const queryString = require("query-string");
 const {NoDatasetPathError, InvalidSourceImplementation} = require("./exceptions");
@@ -315,6 +316,13 @@ class S3Source extends Source {
         .split("_")
         .join("/"));
   }
+  async getAndDecompressObject(key) {
+    const object = await S3.getObject({ Bucket: this.bucket, Key: key}).promise();
+    if (object.ContentEncoding === 'gzip') {
+      object.Body = zlib.gunzipSync(object.Body);
+    }
+    return object.Body;
+  }
   /**
    * Get information about a (particular) source.
    * The data could be a JSON, or a markdown with YAML frontmatter. Or something else.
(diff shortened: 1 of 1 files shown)


src/sources.js
```diff
@@ -333,7 +338,8 @@ class S3Source extends Source {
         title: `"${this.name}" Nextstrain group`,
         byline: `The available datasets and narratives in this group are listed below.`,
         showDatasets: true,
-        showNarratives: true
+        showNarratives: true,
+        avatar: await this.availableLogo()
```
jameshadfield, line 342: It feels strange for this to be in the "Appropriate fallback if no customised data is available" catch-block rather than the "attempt to fetch customisable information from S3 bucket" block.

src/sources.js
```diff
@@ -315,6 +315,11 @@ class S3Source extends Source {
         .split("_")
         .join("/"));
   }
+  async availableLogo() {
+    const objects = await this._listObjects();
```
jameshadfield, line 319: In future PRs we'll want to (try to) get a customised title, byline etc which will also be stored in the bucket and therefore will also use the output of `this._listObjects()`. Very happy to defer that to future PRs if you'd prefer?

src/sources.js
joverlee521: Yup, updated in the new commit

src/sources.js
joverlee521: I moved it to `getInfo()` in the new commit

joverlee521:
> Actually, I've just realized that this approach (having the client get the PNG via an additional fetch straight to the bucket) won't work for private groups -- see #154. A simple solution would be to fetch it server-side and return the image within the `getSourceInfo` JSON response.

Thanks for pointing this out! I didn't realize that would be a problem. I've made the suggested solution part of the new commit.

tsibley:
@joverlee521 @jameshadfield Rather than fetch, decompress, and encode as a data: URI, I think it'd be much simpler to send the client an https: URI which points directly to the S3 bucket. For private groups, the URL would be pre-signed to allow the browser to fetch it (as [src/sources.js](https://github.com/nextstrain/nextstrain.org/blob/c71de94030fb9e1ba1e19136bb0df95b7ecbfaaf/src/sources.js#L378-L394) does for dataset and narrative URLs is passes to Auspice).

joverlee521:
> @joverlee521 @jameshadfield Rather than fetch, decompress, and encode as a data: URI, I think it'd be much simpler to send the client an https: URI which points directly to the S3 bucket. For private groups, the URL would be pre-signed to allow the browser to fetch it (as [src/sources.js](https://github.com/nextstrain/nextstrain.org/blob/c71de94030fb9e1ba1e19136bb0df95b7ecbfaaf/src/sources.js#L378-L394) does for dataset and narrative URLs is passes to Auspice).

Thank you @tsibley! That is much simpler, I'll update this in the new PR #193

PR #189, opened 2020-06-23
Add links in documentation pointing to discussion.nextstrain.org

jameshadfield:
Here we add links to the (new) discussion site in the majority of the places where we provide a contact email address. There are exceptions to this, such as the nextstrain.org splash page where as we wanted to direct more casual users to use the email for the time being.


trvrb:
This is perfect. Thanks so much James!

PR #192, opened 2020-06-24
Generalise strain search functionality

jameshadfield:
This generalises previous work (#187) to create a better framework for searching datasets for a particular strain (sample) name. 

The "databases", which are a single JSON file per pathogen, are created by the `./scripts/collect-search-results.js` and uploaded to a S3 bucket rather than stored in this repo. This has a number of advantages: it separates out the logic for creating the databases from building this site which both reduces the build time and allows for future work to automatically update these databases (JSON files) when the datasets update. Currently these are hosted on the nextstrain-scratch bucket, and updates must be done manually. Please note that currently this script only searches across the nextstrain-data bucket; future improvement would be to add more pathogen-specific logic - for example searching the different nCoV datasets detailed in #185. 

There is a simple nextstrain.org/search page which lists a number of per-pathogen search pages, which are dynamically defined by the data in `static-site/search_pages.json`. Each search page fetches the necessary data from a url defined in that JSON. Currently seasonal-flu and nCoV are included, but future additions will be simple to add.


jameshadfield:
@trvrb @emmahodcroft this is now working as alluded to on slack 🎉  https://nextstrain-s-search-ywpzukbtni.herokuapp.com/search provides a landing page which links out to a list of pathogens currently supported (seasonal flu + nCoV).

In terms of keeping the database up to date, we need to run a small script each time any dataset changes. This is less problematic for nCoV, where most datasets have time-stamped filenames and don't change, but is very problematic for flu. Long term, we could explore things such as [Lambda functions](https://docs.aws.amazon.com/lambda/latest/dg/with-s3.html) which automatically run whenever a database is uploaded to the bucket, or a cron job or similar. If the flu builds would be useful in the short term, then @joverlee521 would you be able to run this script when you update the influenza builds each week? What's required is to run `./scripts/collect-search-results.js --pathogen seasonal-flu` and then upload the resulting file to a bucket via the instructions that script will print out.

If we wish to expand the reach beyond datasets on nextstrain-data (e.g. to use the nCoV datasets we started to track in #185) then we'll need to have thought through an approach to keep things up to date.

joverlee521:
> If the flu builds would be useful in the short term, then @joverlee521 would you be able to run this script when you update the influenza builds each week? 

I can definitely do this as part of the weekly flu build update.

Side note: Would it help to take advantage of the versioning enabled in the `nextstrain-data` bucket? I believe you can [retrieve specific versions of an object from S3](https://docs.aws.amazon.com/AmazonS3/latest/dev/list-obj-version-enabled-bucket.html)

jameshadfield:
> Side note: Would it help to take advantage of the versioning enabled in the nextstrain-data bucket?

This would be interesting for auspice & nextstrain to explore -- essentially a way to access old datasets without having made a separate, timestamped file. If we go this way (it's larger and separate to this PR) then we'll want the search functionality to index them also :)

emmahodcroft:
This is so cool James!! Suuuuper cool. 

What is the script that we should be running when we deploy? We could maybe make this a 'nextstrain' snakemake rule so that it's easy for build shepards to run it after deploying?

Just one comment which is small and not very important - I notice results are sorted by date, which is fantastic, but was wondering if within that they could be sorted by the build name - so that for this:

image

They came out in order:
- ncov/asia (no trailing name)
- ncov/asia/gisaid (alphabetical, trailing letters)
- ncov/asia/zh (alphabetical, trailing letters)
- ncov/asia/2020-06-23 (trailing numbers)

Just because this would then put any results in the 'main' builds at the top. We might want to filter out GISAID entirely (as this is just.. minus useful headings and stuff), and maybe ZH (I'm not sure randomly directing to a Chinese site is useful for most!)

Anyway, not a big deal at all, but just a thought!

jameshadfield:
@emmahodcroft thanks for the comments! I've filtered out the `_gisaid.json` and `_zh.json` as they were duplicates in terms of samples included. I haven't changed the sort order because (a) even though I only display YYYY-MM-DD they're actually in strict chronological order as their upload times are all different and (b) I'm hesitant to emphasise the non-datestamped builds as they're actually the ones that are most likely to be wrong, in the sense that the dataset JSON has been updated and the search functionality hasn't stayed up-to-date.

Related to this, there's a commented-out link on the splash page highlighting this functionality. Once we've tested it a bit we can show this!

> What is the script that we should be running when we deploy? We could maybe make this a 'nextstrain' snakemake rule so that it's easy for build shepards to run it after deploying?

The script is within the nextstrain.org repo and can be run via `./scripts/collect-search-results.js --pathogen sars-cov-2` which takes about 5 minutes and then will print out the command to upload the database to the correct place (`nextstrain remote upload s3://nextstrain-scratch ./data/search_sars-cov-2.json`). It would be great to see how hard it is to run this as we upload the nCoV datasets and perhaps automate it into the build.

---

I'm going to merge (once the CI checks & review app are working) because this approach (separating the database creation from the deployment) is better than what we currently have.

PR #193, opened 2020-06-24
Group overview

joverlee521:
Allow Nextstrain groups to customize the content of their splash page using `group-overview.md`.
~Expects `group-overview.md` to have YAML front matter with `title` and `byline`.~
`group-overview.md` is required to have `title` in the YAML front matter, 
but `byline`, `showDatasets`, `showNarratvies` are optional variables. 

The content is displayed using `MarkdownDisplay` from Auspice. 

Note: The markdown display currently only uses the default styling in Auspice

### Testing
I've added a `group-overview.md` for the "blab-private" group.


src/sources.js
```diff
@@ -323,6 +325,16 @@ class S3Source extends Source {
     }
     return object.Body;
   }
+  parseOverviewMarkdown(overviewMarkdown) {
+    const frontMatter = yamlFront.loadFront(overviewMarkdown);
+    if (!frontMatter.title || !frontMatter.byline) {
+      throw new Error("Incorrectly formatted frontmatter in overview file");
```
jameshadfield, line 331: This error will never be displayed (the catch block following where `parseOverviewMarkdown` is called does nothing with the error object). Given that groups users won't have access to the server logs, in addition (or instead of) logging it server-side we should forward a helpful message as a value of the JSON response and then log this in the browser console.

src/sources.js
```diff
@@ -323,6 +325,16 @@ class S3Source extends Source {
     }
     return object.Body;
   }
+  parseOverviewMarkdown(overviewMarkdown) {
+    const frontMatter = yamlFront.loadFront(overviewMarkdown);
+    if (!frontMatter.title || !frontMatter.byline) {
```
jameshadfield, line 330: It's probably ok to make `title` compulsory but `byline` should be optional. If the markdown exists but `byline` is missing then we shouldn't display one.

src/sources.js
```diff
@@ -323,6 +325,16 @@ class S3Source extends Source {
...
+      throw new Error("Incorrectly formatted frontmatter in overview file");
+    }
+    // handle files with CRLF endings (windows)
+    const content = frontMatter.__content.replace(/\r\n/g, "\n");
+
+    return [frontMatter.title, frontMatter.byline, content];
```
jameshadfield, line 336: Can we also include the following fields in the frontmatter (precise naming can be changed if desired):
* `displayAvailableDatasets` (bool)
* `displayAvailableNarratives` (bool)

And update the customised splash page accordingly.

trvrb:
Very cool! https://nextstrain-s-group-over-00betr.herokuapp.com/groups/blab works as expected. Small aesthetic comment. Could you move `byline` directly under `title`? There's a bunch of whitespace now. Perhaps making the text slightly gray. Take a look at: https://github.com/blab for example styling that I think works well.

trvrb:
^ I added `website` as an option in the YAML header. I thought it was important to push groups to surface this link. But it is still optional.

trvrb:
Okay. I've made some styling adjustments on top of this PR. You can check things out here:

* https://nextstrain-s-group-over-00betr.herokuapp.com/groups/blab
* https://nextstrain-s-group-over-00betr.herokuapp.com/groups/seattleflu
* https://nextstrain-s-group-over-00betr.herokuapp.com/groups/nextspain
* https://nextstrain-s-group-over-00betr.herokuapp.com/community/grubaughlab/CT-SARS-CoV-2/

to see how pages look when populated with `overview.md`, when not and how community pages look as well.

I'm very happy with the state of the PR at this point and think it can be merged. Thank you @joverlee521 for putting this together. 🙏🏻

tsibley:
This looks great! I did notice that the subtitle/description and website link are _really_ low contrast (grey on grey). I think increasing their contrast would improve their appearance and readability, especially for folks with less than perfect vision.

PR #197, opened 2020-06-25
update auspice to 2.17.1

jameshadfield:


jameshadfield:
Note that the travisCI builds did pass -- https://travis-ci.com/github/nextstrain/nextstrain.org/builds/172948199

PR #202, opened 2020-07-09
Send error messages in the response body, not the status line

### LINKED ISSUE #201: Stop using response.statusMessage for error messages
Context in https://github.com/nextstrain/auspice/issues/1180. nextstrain.org copied the `handleError` function from Auspice, so it also needs updating here.

tsibley:
The HTTP status line message is not intended for arbitrary messages and
not suitable for freeform text due to character restrictions.

Using statusMessage this way has caused nextstrain.org server crashes
when the field is assigned an error containing a newline, as newlines
are not allowed.

Resolves #201.

Related to <https://github.com/nextstrain/auspice/issues/1180>.

---

A draft PR for now as one more commit is still needed to upgrade Auspice to the next released version that includes PR https://github.com/nextstrain/auspice/pull/1181. Will wait for that to merge and release before adding the final commit here.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,9 +14,9 @@
       }
     },
     "@babel/compat-data": {
-      "version": "7.10.3",
-      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.10.3.tgz",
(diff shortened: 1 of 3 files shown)


tsibley:
I tested this a variety of ways, most notably by loading http://localhost:5000/community/narratives/EduanWilkinson/ncov_cpt@bb2833e7 which uses the original trigger @trvrb troubleshooted in Slack.

PR #206, opened 2020-08-04
SARS-CoV-2 splash page

jameshadfield:
This PR starts the work making a main nextstrain.org/sars-cov-2 splash page by combining the previous work making a list of builds and a list of situation reports (both led by @eharkins). There are three main areas needing some more work bef release, but the overall idea is here I think. Commits would be welcomed 👍 

- content additions for the abstract, a section about nextclades etc
- improvements to styling
- a system to keep the search db and builds db (both flat-files) up-to-date


rneher:
I just added a few words as an abstract and fixed a link. Would be good if the abstract also had a link to Nextclade, but not sure how the static page works... Also, just noticed my opinionated editor made a bunch of extra changes -- sorry.

jameshadfield:
I think this is ready to go -- thanks @rneher for the abstract! I've added some subtext to the entries in the contents list which I think works nicely and allows an explanation of Nextclade. I've also added some more entries to that list -- we have a lot of really amazing resources here, and it'll be great to surface them in this way.

This PR modifies the SARS-CoV-2 section of the main splash page to steer people towards the SARS-CoV-2 splash page, and the URLs /ncov and /ncov-sit-reps will redirect to the SARS-CoV-2 splash page (previously they redirected to the global build and the separate page listing the sit-reps, respectively). cc @trvrb & @eharkins. 

e035218 updates the location of the search databases to use the canonical nextstrain-data bucket (the message printed by the script indicating where to upload the file to has been updated also). cc @joverlee521 for seasonal flu. I'll try to update this frequently for SARS-CoV-2 but automation would be preferable long-term.

trvrb:
This is fantastic James! I don't think I have anything to add here. Really nice to have everything compiled.

PR #195, opened 2020-06-25
External docs links

jameshadfield:
This is a small PR to improve duplicated (or overly similar) documentation pages across different sites using the ability for nextstrain.org/docs to have external links in the sidebar (see #191, on which this PR is build). I also add a few more external links to pages which seemed to be immediately useful for users of nextstrain.org/docs. 
I realise that a major shake-up of how documentation will be orgainsed is under discussion (this is the not-yet-publicised but google indexed docs.nextstrain.org), and that questions such as whether tutorials should live in the main nextstrain.org docs (currently nextstrain.org/docs) or within augur are currently unresolved.

# Changes to sections within nextstrain.org/docs

## Getting Started

No changes

#### Augur commands

nextstrain.org/docs/bioinformatics/augur-commands was a stub with a link to https://docs.nextstrain.org/projects/augur/en/stable/usage/cli/cli.html.
I've therefore replaced it with an external link straight from the sidebar.

#### FAQ

Added an external link to the RTD hosted FAQ.

## Tutorials.

A general observation is that our tutorials are actually a combination of a tutorial and a bunch of how-to documentation combined.
Separating those out is a project for another day!

### Zika: (nextstrain.org/docs/tutorials/zika vs https://docs.nextstrain.org/projects/augur/en/stable/tutorials/zika_tutorial.html)
These tutorials are very diverged (perhaps they don't share a common ancestor?!?).
I find the nextstrain.org version easier to follow, as the RTD version spends a lot of time introducing the concept of augur commands, and have therefore left it untouched.

### TB (nextstrain.org/docs/tutorials/tb vs https://docs.nextstrain.org/projects/augur/en/stable/tutorials/tb_tutorial.html)

These are very similar; the main differences being that the RTD version assumes you have installed augur, but also seems to be the most up-to-date (for instance using `--output-node-data` instead of `--output`).
I've removed nextstrain.org/docs/tutorials/tb and added a sidebar entry (and redirect) to the RTD version.

#### Defining clades

Untouched. I am in favor of the changes in https://github.com/nextstrain/nextstrain.org/pull/170/files but think they should be moved to the augur docs.
Will cross-comment on that PR.

## Contributing

Unchanged

## Narratives

This content is more up-to-date (and better) in the auspice docs.
I've therefore removed this from nextstrain.org/docs and added two external links, one within the Visualisation section (an introduction to narratives) and one in the tutorials section (a step-by-step on how to write).
A redirect is added from the now removed page to the narratives introduction.

### Download data & interpreting the map

[...]

PR #209, opened 2020-08-20
move interpretation doc to /help

eharkins:
Move the guide on interpreting the map to /help for now since it seems to fit better under that top level heading - it's more about understanding nextstrain than running nextstrain.

### Related issue(s)  
#195 

### Testing
Locally with gatsby


eharkins:
No longer relevant; see https://github.com/nextstrain/docs.nextstrain.org/pull/26

PR #211, opened 2020-08-23
Add pagination for S3 listings to allow buckets with >1000 keys

### LINKED ISSUE #186: Use pagination to list items in S3 buckets
The `S3.listObjectsV2` API, which we use [here](https://github.com/nextstrain/nextstrain.org/blob/master/src/sources.js#L299) to list the items in S3 buckets, is returning truncated lists of objects for certain buckets (I noticed this playing around with `nextstrain-data` for another purpose). There's a comment in the code indicating that we "should iterate over the whole bucket contents using the S3 client's pagination support".

I don't actually think we use the `_listObjects` method for the `nextstrain-data` bucket, so there may not be a hurry on this, but it's plausible that some Nextstrain Groups buckets will hit the limit at some point.

jameshadfield:
We now have Nextstrain Groups with buckets with over 1000 keys. The previous
implementation only returned 1000 keys from a bucket resulting in us missing some
datasets / narratives when listing the bucket, which meant they weren't displayed
on the group's splash page.

Better solutions are more than welcome, and it would be valuable to know how others would have implemented this.

Closes #186

### DIFF
--- a/src/sources.js
+++ b/src/sources.js
@@ -294,13 +294,19 @@ class S3Source extends Source {
     return `https://${this.bucket}.s3.amazonaws.com`;
   }
   async _listObjects() {
-    // XXX TODO: This will only return the first 1000 objects.  That's fine for
-    // now (for comparison, nextstrain-data only has ~500), but we really
-    // should iterate over the whole bucket contents using the S3 client's
-    // pagination support.
-    //   -trs, 30 Aug 2019
-    const list = await S3.listObjectsV2({Bucket: this.bucket}).promise();
-    return list.Contents;
+    return new Promise((resolve, reject) => {
+      let contents = [];
+      S3.listObjectsV2({Bucket: this.bucket}).eachPage((err, data, done) => {
+        if (err) {
+          return reject(err);
+        }
+        if (data===null) { // no more data
+          return resolve(contents);
+        }
+        contents = contents.concat(data.Contents);
+        return done();
+      });
+    });
   }
   async availableDatasets() {
     const objects = await this._listObjects();


src/sources.js
```diff
@@ -294,13 +294,18 @@ class S3Source extends Source {
...
   }
   async _listObjects() {
-    // XXX TODO: This will only return the first 1000 objects.  That's fine for
-    // now (for comparison, nextstrain-data only has ~500), but we really
-    // should iterate over the whole bucket contents using the S3 client's
-    // pagination support.
-    //   -trs, 30 Aug 2019
-    const list = await S3.listObjectsV2({Bucket: this.bucket}).promise();
-    return list.Contents;
+    const s3Params = {Bucket: this.bucket, MaxKeys: 1000};
```
kairstenfay, line 297: The [AWS SDK](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3.html#listObjectsV2-property) documents that the default `MaxKeys` is 1000, so this change is not necessary.

src/sources.js
```diff
@@ -294,13 +294,18 @@ class S3Source extends Source {
...
+    let contents = [];
+    let existMoreKeys = true;
+    let lastKeySeen = false;
+    while (existMoreKeys) {
+      // eslint-disable-next-line no-await-in-loop
+      const list = await S3.listObjectsV2(lastKeySeen ? {...s3Params, StartAfter: lastKeySeen} : s3Params).promise();
+      existMoreKeys = list.IsTruncated;
+      contents = [...contents, ...list.Contents];
```
kairstenfay, line 305: we can avoid iterating over both `contents` and `list.Contents` here by instead writing
```
contents = contents.concat(list.Contents)
```

src/sources.js
```diff
@@ -294,13 +294,18 @@ class S3Source extends Source {
...
-    // now (for comparison, nextstrain-data only has ~500), but we really
-    // should iterate over the whole bucket contents using the S3 client's
-    // pagination support.
-    //   -trs, 30 Aug 2019
-    const list = await S3.listObjectsV2({Bucket: this.bucket}).promise();
-    return list.Contents;
+    const s3Params = {Bucket: this.bucket, MaxKeys: 1000};
+    let contents = [];
+    let existMoreKeys = true;
+    let lastKeySeen = false;
```
kairstenfay, line 300: You don't need to assign a value here; the initialized variable `lastKeySeen` will evaluate as a falsy boolean.
e.g.
```
let lastKeySeen
!!lastKeySeen
> false
```

joverlee521:
Instead of the while loop, you can iterate over objects with [eachPage()](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Request.html#eachPage-property)

kairstenfay:
> Instead of the while loop, you can iterate over objects with [eachPage()](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Request.html#eachPage-property)

Oh nice, I missed this.

kairstenfay:
> > Instead of the while loop, you can iterate over objects with [eachPage()](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Request.html#eachPage-property)
> 
> Oh nice, I missed this.

I think @joverlee521 `eachPage()` find is even better than any suggestion I made.

jameshadfield:
@joverlee521 / @kairstenfay could you please re-review this? I now use `eachPage()` as recommended by @joverlee521  which is nicer than a while loop but requires wrapping with a promise as the `promise` method is no longer available.

src/sources.js
jameshadfield: Done. Thanks!

src/sources.js
jameshadfield: I have removed this on your suggestion, however I originally chose to include it as I preferred the explicitness this provided at the small cost of verbosity.

src/sources.js
jameshadfield: This is no longer needed in the new implementation using `eachPage` but I did this on purpose as I preferred the explicitness of it :)

src/sources.js
kairstenfay: Fair enough. This is why I called it a nitpick. I accept a difference of opinion here, and I would not make something like that a blocker on this PR :slightly_smiling_face:

src/sources.js
kairstenfay: Ditto re: my calling it a nitpick. I completely understand the desire to be more explicit.

jameshadfield:
Thanks both!

tsibley:
Thanks everyone. I think it'd be nice to use async iterators/generators in the future for this sort of thing, which makes the code more imperative and easier to follow. It looks like the AWS JS SDK doesn't have built-in support for async iterators (unlike the Python SDK, which does), but it'd be possible with a small wrapper function.

PR #216, opened 2020-09-17
Allow dataset source to be user-defined via /fetch URLs

jameshadfield:
This PR implements functionality discussed internally by @sidneymbell, @trvrb and myself. Namely, it allows a publicly accessible URL to a dataset (or narrative) to be displayed within nextstrain.org using a nextstrain.org/fetch URL syntax.

### Examples:
* https://nextstrain-s-fetch-sour-8nfiar.herokuapp.com/fetch/https://data.nextstrain.org/zika.json
* https://nextstrain-s-fetch-sour-8nfiar.herokuapp.com/fetch/https://raw.githubusercontent.com/nextstrain/narratives/master/intro-to-narratives.md
* https://nextstrain-s-fetch-sour-8nfiar.herokuapp.com/fetch/https://data.nextstrain.org/flu_seasonal_yam_na_2y.json:fetch/https://data.nextstrain.org/flu_seasonal_yam_ha_2y.json

### Things blocking this PR

* This PR requires a slight modification to auspice (regarding how it interprets `:` within the URL, as `:` is also the separator character for defining multiple datasets to display). That PR must be merged into auspice and this PR updated accordingly.
* A review of the implementation and direction taken here
* (perhaps not blocking, but highly desired) a simple test script which accesses a number of `/charon/` API calls and checks the response. This will give us confidence that PRs such as this one do not break other functionality unintentionally. ~I will create a separate issue detailing this.~ See https://github.com/nextstrain/nextstrain.org/issues/217

### Notes

1. Sidecar files (e.g. tip-frequences, root-sequence files) are not fetched currently. We could attempt to fetch them by attempting to modify the URL (e.g. if given https://data.nextstrain.org/zika.json, try to load https://data.nextstrain.org/zika_tip-frequencies.json). 

2. We currently pipe responses through the nextstrain.org server (as we do with all other datasets). An alternative implementation is to return a redirect and have auspice fetch this directly. This will require changes to auspice I believe, and may have CORS issues. Related to this approach is the observation that we maintain backwards compatibility with previous auspice dataset schemas by converting them as they pass through the nextstrain.org server.

PR #225, opened 2020-10-07
Next release

jameshadfield:
This PR represents a version of the nextstrain.org website which is up for testing before going live (which will happen automatically upon merge of this PR). This branch has been deployed to dev.nextstrain.org for testing purposes (the heroku app should work for most purposes, however you can't login to nextstrain using review apps).

### Changes compared to current nextstrain.org:

* Auspice upgraded to 2.19.0 (see https://nextstrain.github.io/auspice/releases/changelog) which includes the ability to fetch the root-sequence JSON, as well as a polyfill for IE.
* Smoke testing of server responses to /charon GET requests (see https://github.com/nextstrain/nextstrain.org/pull/221) 
* Allow dataset source to be user-defined via /fetch URLs (see https://github.com/nextstrain/nextstrain.org/pull/216)

### Example URL to test

The best way to test some of the new functionality is via the following URL

https://dev.nextstrain.org/fetch/staging.nextstrain.org/flu_seasonal_h3n2_ha_2y.json?c=gt-HA2_10

Which will utilise the "/fetch" functionality (see new docs [here](https://dev.nextstrain.org/docs/contributing/fetch-data-from-custom-urls)) to pull down a user defined JSON (in this case from https://staging.nextstrain.org/flu_seasonal_h3n2_ha_2y.json), and auspice will request the root-sequence sidecar JSON (https://staging.nextstrain.org/flu_seasonal_h3n2_ha_2y_root-sequence.json) which allows coloring by codon 10 in HA2 (you'll see it transition from grey to colored after the sidecar file arrives). Compare this to https://dev.nextstrain.org/fetch/data.nextstrain.org/flu_seasonal_h3n2_ha_2y.json?c=gt-HA2_10 where the sidecar file request fails (we don't currently have root-sequence JSONs on the nextstrain-data bucket) and thus we don't have the information to color the tree / know which AA the position has.

### Linting is currently failing

_update: this has now been fixed_

Tests are failing because of a strange eslint error on numerous files (most of which haven't been modified):
```
npm run lint
...
/Users/naboo/github/nextstrain/nextstrain.org/static-site/src/util/structureEdges.js
  0:0  error  Parsing error: Cannot read property 'Literal' of undefined
```
This will have to be fixed before merge to master (failing tests mean an automatic rebuild won't happen!)


jameshadfield:
The linting errors are due to this change in the package-lock file where the `eslint-visitor-keys` package (a dependency of `eslint`, among others) somehow switched to a completely different package called`camelize` (or at least that's what it looks like!). Diff: https://github.com/nextstrain/nextstrain.org/pull/225/commits/04b7da89d57941676742bcb9a88c3982def89d07#diff-32607347f8126e6534ebc7ebaec4853dR7503-R7506 introduced in 04b7da8. Switching back in the lockfile fixes the problem, but i'll find a way to change things so that we don't encounter this again the next time we update something...

PR #226, opened 2020-10-29
Redirect /docs and /help to docs.nextstrain.org

eharkins:
As per https://github.com/nextstrain/docs.nextstrain.org/issues/13, we need to send what used to be nextstrain.org/docs and nextstrain.org/help both over to docs.nextstrain.org since we are migrating our documentation there. This does that in three steps:
- replace all the links (e.g. in the navigation bar at the top of nextstrain.org) which previously when to these pages to external links that send you to docs.nextstrain.org
- redirect all current nextstrain.org/docs/* and nextstrain.org/help/* urls to the corresponding pages at docs.nextstrain.org
- remove the (mostly markdown) content that we have migrated to docs.nextstrain.org from this repository, so that Gatsby no longer builds it into the static site.

### Related issue(s)  
https://github.com/nextstrain/docs.nextstrain.org/issues/13
https://github.com/nextstrain/docs.nextstrain.org/issues/35 (in case we want to adjust redirects for clade labeling page)

### Testing
I've just been testing this locally with gatsby (`npm run dev` in `static-site` and `npm run server` from the root directory), checking that redirects work and clicking on external links to docs.nextstrain.org. I will deploy this on heroku to test today.


static-site/content/blog/2019-10-21-auspice-v2.md
```diff
@@ -2,6 +2,7 @@
 author: "James Hadfield"
 date: "2019-10-21"
 title: "Auspice v2 released"
+anchorText: "Auspice v2"
```
eharkins, line 5: It's necessary to have at least one instance of `anchorText` or else gatsby will complain.

eharkins:
Upon or before merging, we'll want to remove the banner from the docs.nextstrain.org home page (https://github.com/nextstrain/docs.nextstrain.org/blob/master/src/index.rst) that says:
> This site is currently only a stub, to show what's possible with Read The Docs for an umbrella documentation project.
> For the real documentation entry point, please go to nextstrain.org/docs.

Since going to nextstrain.org/docs will redirect back to docs.nextstrain.org.

auspice-client/customisations/navbar.js
```diff
@@ -123,8 +124,8 @@ const NavBar = ({sidebar, narrativeTitle, width}) => {
           null : (
             <div style={{...styles.flexColumns, paddingRight: "12px"}}>
               <div style={{flex: 5}}/>
-              {renderLink("DOCS", "/docs", styles.link)}
-              {renderLink("HELP", "/help", styles.link)}
+              <Link href="https://docs.nextstrain.org/en/latest/index.html" style={styles.link} target="_blank" rel="noopener noreferrer" text={<React.Fragment>DOCS <FaExternalLinkAlt/></React.Fragment>}/>
```
jameshadfield, line 127: These links to a different subdomain of nextstrain.org are best thought of as [internal links](https://www.seroundtable.com/subdomains-google-webmaster-tools-13960.html) and thus there should be no change in how they're rendered (the addresses themselves do need changing as you've done here).

redirects.js
```diff
@@ -52,6 +52,44 @@ const setup = (app) => {
...
+    "/docs/visualisation/download-data": `${mainReadTheDocs}/guides/share/download-data.html`,
+    "/docs/visualisation/map-interpretation": `${mainReadTheDocs}/learn/interpret/map-interpretation.html`,
+    "/help/general/about-nextstrain": `${mainReadTheDocs}/learn/about-nextstrain.html`,
+    "/help/general/how-to-read-a-tree": `${mainReadTheDocs}/learn/interpret/how-to-read-a-tree.html`,
+    "/help/general/interacting-with-nextstrain": `${mainReadTheDocs}/learn/interpret/interacting-with-nextstrain.html`,
+    "/help/coronavirus/FAQ": `${mainReadTheDocs}/learn/pathogens/coronavirus/FAQ.html`,
+    "/help/coronavirus/SARS-CoV-2": `${mainReadTheDocs}/learn/pathogens/coronavirus/SARS-CoV-2.html`,
+    "/help/coronavirus/Technical-FAQ": `${mainReadTheDocs}/learn/pathogens/coronavirus/Technical-FAQ.html`,
+    "/help/coronavirus/human-CoV": `${mainReadTheDocs}/learn/pathogens/coronavirus/human-CoV.html`
+  };
```
jameshadfield, line 86: The redirects from https://github.com/nextstrain/nextstrain.org/blob/master/static-site/redirects.json should be shifted to here, which is consistent with our plan to remove all "docs" and "help" content from the Gatsby section of the site.

static-site/src/components/nav-bar/index.jsx
```diff
@@ -120,14 +121,8 @@ class NavBar extends React.Component {
...
         <div style={{flex: 5}}/>
-        {this.selectedClass("help") ?
-          <NavLinkInactive minified={minified}>HELP</NavLinkInactive> :
-          <NavLinkToBeHandledByGatsby minified={minified} to="/help">HELP</NavLinkToBeHandledByGatsby>
-        }
-        {this.selectedClass("docs") ?
-          <NavLinkInactive minified={minified}>DOCS</NavLinkInactive> :
-          <NavLinkToBeHandledByGatsby minified={minified} to="/docs">DOCS</NavLinkToBeHandledByGatsby>
-        }
+        <NavLinkToGoToServer minified={minified} href="https://docs.nextstrain.org/en/latest/learn/about-nextstrain.html" target="_blank" rel="noopener noreferrer">HELP <FaExternalLinkAlt/></NavLinkToGoToServer>
```
jameshadfield, line 124: Same comment as with the auspice nav-bar - these are still considered internal links and thus shouldn't have a `<FaExternalLinkAlt/>` icon.

static-site/src/components/sars-cov-2/builds.jsx
```diff
@@ -75,7 +74,7 @@ class Index extends React.Component {
         <splashStyles.FocusParagraph>
           This section is an index of public Nextstrain builds (datasets) for SARS-CoV-2, organized by geography.
           Some of these builds are maintained by the nextstrain team and others are maintained by independent research groups.
-          (<Link to="docs/bioinformatics/what-is-a-build">See here</Link> for more information on what a build is, and see <a href="https://nextstrain.github.io/ncov/">this tutorial</a> for a walkthrough of running your own phylogenetic analysis of SARS-CoV-2 data!)
+          (<a href="https://docs.nextstrain.org/projects/augur/en/migrate-docs/faq/what-is-a-build.html" target="_blank" rel="noopener noreferrer">See here <FaExternalLinkAlt/></a> for more information on what a build is, and see <a href="https://nextstrain.github.io/ncov/">this tutorial</a> for a walkthrough of running your own phylogenetic analysis of SARS-CoV-2 data!)
```
jameshadfield, line 77: Same comment as above -- no need for `<FaExternalLinkAlt/>`

static-site/src/components/splash/index.jsx
```diff
@@ -94,8 +95,8 @@ class Splash extends React.Component {
 
         <FlexCenter>
           <Styles.CenteredFocusParagraph>
-            Analyses by independent groups <Link to="/docs/contributing/community-builds">stored and
-            accessed via public GitHub repos</Link>
+            Analyses by independent groups <a href="https://docs.nextstrain.org/en/latest/guides/share/community-builds.html" target="_blank" rel="noopener noreferrer">stored and
+            accessed via public GitHub repos <FaExternalLinkAlt/></a>
```
jameshadfield, line 99: Same as above, no need for `<FaExternalLinkAlt/>`

static-site/src/components/splash/index.jsx
```diff
@@ -203,8 +204,8 @@ class Splash extends React.Component {
         <BigSpacer/>
 
         <FlexCenter>
-          <Styles.Button to="docs">
-            Read the documentation
+          <Styles.Button to="https://docs.nextstrain.org/en/latest/index.html" target="_blank" rel="noopener noreferrer">
+            Read the documentation <FaExternalLinkAlt/>
```
jameshadfield, line 208: Same as above, non need for `<FaExternalLinkAlt/>`

static-site/src/components/splash/index.jsx
```diff
@@ -203,8 +204,8 @@ class Splash extends React.Component {
         <BigSpacer/>
 
         <FlexCenter>
-          <Styles.Button to="docs">
-            Read the documentation
+          <Styles.Button to="https://docs.nextstrain.org/en/latest/index.html" target="_blank" rel="noopener noreferrer">
```
jameshadfield, line 207: I don't think we want to change this to open this in a new tab - but perhaps there is a specific reason for changing this?

redirects.js
```diff
@@ -52,6 +52,44 @@ const setup = (app) => {
   conflates a language part ("zh") with a region ("europe") so we want to keep zh out of the manifest JSON */
```
jameshadfield, line 52: There's one more change for this file:

```js
  /* send auspice to the auspice docs (currently hosted as github pages) */
  app.route("/auspice")
    .get((req, res) => res.redirect('https://nextstrain.github.io/auspice/'));
```

Should now point to an appropriate section in docs.nextstrain.org :)

jameshadfield (changes requested):
Looks good @eharkins - have left specific line-comments for minor changes. Thanks!

auspice-client/customisations/navbar.js
eharkins: Ah, thanks -  I didn't know that! Considering we are treating them as internal links, I think staying in the same tab is more in line with that. As I was considering them external, i figured a separate tab would be appreciated by users but that is also based mostly in my own preference. 

To be clear, when such links are rendered in the static-site we still want to move away from them being gatsby-server-handled links right?

static-site/src/components/nav-bar/index.jsx
eharkins: Same question as with auspice nav-bar: We still want to use NavLinkToGoToServer instead of NavLinkToBeHandledByGatsby to be
> consistent with our plan to remove all "docs" and "help" content from the Gatsby section of the site.

correct?

static-site/src/components/splash/index.jsx
eharkins: No, you are right, I only think that would make sense if we were considering it an external link.

redirects.js
eharkins: I would send to the auspice subproject: https://docs.nextstrain.org/projects/auspice/en/migrate-docs/

Do you agree?

redirects.js
eharkins: I am adding entries for all those redirects trying to match values from the gatsby mapping with keys in the new mapping I have here. e.g. for the entry in the gatsby redirects:
>   "/about/overview/introduction": "/docs/getting-started/introduction",

I look for any existing main redirect with the key "/docs/getting-started/introduction":
> "/docs/getting-started/introduction": `${mainReadTheDocs}/learn/about-nextstrain.html`,

and add the following entry to the main redirects list (with the goal of reducing double redirects):
> "/about/overview/introduction": `${mainReadTheDocs}/learn/about-nextstrain.html`,



I deviated from this in some cases where it seems like it gets folks to a more relevant place in docs.nextstrain.org.

eharkins:
@jameshadfield 4d30a24 should resolve all of the above, let me know if it's not the case, thanks for the review!

auspice-client/customisations/navbar.js
jameshadfield: > when such links are rendered in the static-site we still want to move away from them being gatsby-server-handled links right?

Yes -- I believe those are only used for links which remain internal to the Gatsby-produced section of the site

redirects.js
jameshadfield: Nice work!

redirects.js
jameshadfield: Leaving a placeholder "Todo" here as this probably wants to be https://docs.nextstrain.org/projects/auspice/en/latest/ but this requries the merge of Auspice branch `migrate-docs`

redirects.js
```diff
@@ -52,6 +52,67 @@ const setup = (app) => {
...
   app.route("/ncov/zh").get((req, res) => res.redirect("/ncov/global/zh"));
 
+  // DOCS, HELP Redirects for docs migration:
+  const mainReadTheDocs = "https://docs.nextstrain.org/en/latest";
+  const augurReadTheDocs = "https://docs.nextstrain.org/projects/augur/en/migrate-docs";
```
jameshadfield, line 57: Leaving a placeholder "Todo" here as this should be /latest but this requires the Augur PR to be merged

static-site/src/components/sars-cov-2/builds.jsx
```diff
@@ -75,7 +74,7 @@ class Index extends React.Component {
         <splashStyles.FocusParagraph>
           This section is an index of public Nextstrain builds (datasets) for SARS-CoV-2, organized by geography.
           Some of these builds are maintained by the nextstrain team and others are maintained by independent research groups.
-          (<Link to="docs/bioinformatics/what-is-a-build">See here</Link> for more information on what a build is, and see <a href="https://nextstrain.github.io/ncov/">this tutorial</a> for a walkthrough of running your own phylogenetic analysis of SARS-CoV-2 data!)
+          (<a href="https://docs.nextstrain.org/projects/augur/en/migrate-docs/faq/what-is-a-build.html" >See here</a> for more information on what a build is, and see <a href="https://nextstrain.github.io/ncov/">this tutorial</a> for a walkthrough of running your own phylogenetic analysis of SARS-CoV-2 data!)
```
jameshadfield, line 77: Leaving a placeholder "Todo" here as this should be /latest but this requires the Augur PR to be merged

static-site/src/components/nav-bar/index.jsx
jameshadfield: Yup - `<NavLinkToGoToServer>` is simply a styled `<a>` element, whereas (from memory) `<NavLinkToBeHandledByGatsby>` acts via an event handler which causes problems when you want to go to a non-Gatsby page

eharkins (commented):
Merging https://github.com/nextstrain/docs.nextstrain.org/pull/36 would / will mean some updates are needed here. Since there were only a few changes I just added them in comment suggestions which we can add, instead of creating yet another PR into this one.

PR #227, opened 2020-11-04
Refactor secondTreeOptions

kairstenfay:
Refactor the computation of second tree options data by moving it into
the getAvailableDatasets() method of the Source classes.

Note that it still uses globally computed data for now. While stopping
or centralizing the usage of global variables is a desirable direction
to take this codebase, this commit is just moving the functionality to a
place it's more expected.

### DIFF
--- a/src/getAvailable.js
+++ b/src/getAvailable.js
@@ -24,17 +24,10 @@ const getAvailable = async (req, res) => {
     utils.verbose(`No narratives available for ${source.name}`);
   }
 
-  // XXX TODO: This is bad and should go away by refactoring second tree
-  // enumeration into our source classes, potentially as part of the
-  // availableDatasets() method.
-  //   -trs, 3 Oct 2019
-  const secondTreeOptions = (dataset) =>
-    (global.availableDatasets.secondTreeOptions[source.name] || {})[dataset] || [];
-
   return res.json({
     datasets: datasets.map((path) => ({
       request: joinPartsIntoPrefix({source, prefixParts: [path]}),
-      secondTreeOptions: secondTreeOptions(path),
+      secondTreeOptions: source.secondTreeOptions(path),
       buildUrl: source.name === "community"
         ? `https://github.com/${source.repo}`
         : null
--- a/src/sources.js
+++ b/src/sources.js
@@ -37,6 +37,12 @@ class Source {
   narrative(pathParts) {
     return new Narrative(this, pathParts);
   }
+
+  // The computation of these globals should move here.
+  secondTreeOptions(path) {
+    return (global.availableDatasets.secondTreeOptions[this.name] || {})[path] || [];
+  }
+
   availableDatasets() {
     return [];
   }


jameshadfield (approved):
Looks good to me & tested things out locally

src/getAvailable.js
```diff
@@ -14,7 +14,7 @@ const getAvailable = async (req, res) => {
     return unauthorized(req, res);
   }
 
-  const datasets = await source.availableDatasets() || [];
+  const [datasets, secondTreeOptions] = await source.availableDatasets() || [];
```
tsibley, line 17: In the fallback case, this changes behaviour from `datasets = []` to `datasets = undefined`, which will break the subsequent call to `datasets.map(…)` below and thus throw a server error.

src/getAvailable.js
```diff
@@ -24,17 +24,10 @@ const getAvailable = async (req, res) => {
...
-  //   -trs, 3 Oct 2019
-  const secondTreeOptions = (dataset) =>
-    (global.availableDatasets.secondTreeOptions[source.name] || {})[dataset] || [];
-
   return res.json({
-    datasets: datasets.map((path) => ({
+    datasets: datasets.map((path, idx) => ({
       request: joinPartsIntoPrefix({source, prefixParts: [path]}),
-      secondTreeOptions: secondTreeOptions(path),
+      secondTreeOptions: secondTreeOptions[idx],
```
tsibley, line 30: Shared indexing into dual arrays generally indicates that a better data representation/structure is warranted. It is all too easy for such dual arrays to become out of sync and produce bugs. Two thoughts:

1. Instead of `secondTreeOptions` switching from a function that takes a path to an array which takes an index, it could be a dictionary/object which takes a path.

2. Instead of `secondTreeOptions` switching from a function at all, how about making it a method that takes a path on the source object? The change to make `availableDatasets` return two lists would go away and it can be left simpler.

tsibley (changes requested):

src/getAvailable.js
kairstenfay: I went with option 2 in a new, fixup commit.

kairstenfay:
Thank you @tsibley for your feedback. I just changed the code to do the more sensible thing -- not changing the `getAvailable` method (or what it returns), and instead just creating a new method, `secondTreeOptions`, that takes a dataset i.e. path. 

I just tested it locally and it appears to work the same as before.

src/sources.js
```diff
@@ -37,8 +37,15 @@ class Source {
...
+
+  // The computation of these globals should move here.
+  secondTreeOptions(path) {
+    return (global.availableDatasets.secondTreeOptions[this.name] || {})[path] || [];
+  }
+
   availableDatasets() {
-    return [];
+    const datasets = [];
+    return datasets;
```
kairstenfay, line 48: Oops, I'll push a fixup to revert this change

kairstenfay:
I'll update the commit message when rebasing before merging to master.

tsibley (approved):
Thanks, @kairstenfay. This moving around looks fine to me. In local testing, it produces no change in the response of `/charon/getAvailable`.

PR #228, opened 2020-11-10
Error handling

kairstenfay:
Change the error handling for client errors.

Previously, we were treating incorrectly typed (or non-parseable) URLs as 500 errors, when in my opinion, we should be returning a response of a 400 error, indicating that the request was bad because the query was malformed. 

~This branch is currently based off of #227 to minimize the differences shown here between it and the master branch.~

### DIFF
--- a/src/exceptions.js
+++ b/src/exceptions.js
@@ -7,6 +7,12 @@ class NextstrainError extends Error {
   }
 }
 
+class ResourceNotFoundError extends NextstrainError {
+  constructor(msg = "The requested URLs do not exist", ...rest) {
+    super(msg, ...rest);
+  }
+}
+
 /* Thrown when a valid Source object is asked to create a new Dataset object
  * without a dataset path.
  */
@@ -28,5 +34,6 @@ class InvalidSourceImplementation extends NextstrainError {
 
 module.exports = {
   InvalidSourceImplementation,
+  ResourceNotFoundError,
   NoDatasetPathError
 };
--- a/src/getDataset.js
+++ b/src/getDataset.js
@@ -3,7 +3,7 @@ const assert = require('assert').strict;
 
 const utils = require("./utils");
 const helpers = require("./getDatasetHelpers");
-const {NoDatasetPathError} = require("./exceptions");
+const {ResourceNotFoundError, NoDatasetPathError} = require("./exceptions");
 const auspice = require("auspice");
 const request = require('request');
 
@@ -73,6 +73,9 @@ const requestMainDataset = async (res, fetchUrls) => {
           res.send(dataToReturn);
           return resolve();
         }
+        /* Check for a 404 status code after we've had the opportunity to do a
+        follow-up request for a V1 dataset. */
+        if (response.statusCode === 404) reject(new ResourceNotFoundError());
         return reject(dataToReturn);
       });
   });
@@ -152,6 +155,9 @@ const getDataset = async (req, res) => {
     try {
       await requestCertainFileType(res, req, fetchUrls.additional, query);
     } catch (err) {
+      if (err instanceof ResourceNotFoundError) {
+        return res.status(404).send("The requested dataset does not exist.");
+      }
       return helpers.handle500Error(res, `Couldn't fetch JSON: ${fetchUrls.additional}`, err.message);
     }
   } else {
@@ -162,6 +168,11 @@ const getDataset = async (req, res) => {
         utils.verbose("Request is valid, but no dataset available. Returning 204.");
         return res.status(204).end();
       }
+
+      if (err instanceof ResourceNotFoundError) {
+        return res.status(404).send("The requested URL does not exist.");
+      }
+
       return helpers.handle500Error(res, `Couldn't fetch JSONs`, err.message);
     }
   }
--- a/src/getDataset.js
+++ b/src/getDataset.js
@@ -92,7 +92,7 @@ const requestMainDataset = async (res, fetchUrls) => {
 const getDataset = async (req, res) => {
   const query = queryString.parse(req.url.split('?')[1]);
   if (!query.prefix) {
-    return helpers.handleError(res, `getDataset request must define a prefix`);
+    return helpers.handle500Error(res, `getDataset request must define a prefix`);
   }
 
   /*
@@ -122,7 +122,7 @@ const getDataset = async (req, res) => {
       utils.verbose(err.message);
       return res.status(204).end();
     }
-    return helpers.handleError(res, `Couldn't parse the url "${query.prefix}"`, err.message);
+    return helpers.handle500Error(res, `Couldn't parse the url "${query.prefix}"`, err.message);
   }
 
   const {source, dataset, fetchUrls, auspiceDisplayUrl} = datasetInfo;
@@ -148,7 +148,7 @@ const getDataset = async (req, res) => {
     try {
       await requestCertainFileType(res, req, fetchUrls.additional, query);
     } catch (err) {
-      return helpers.handleError(res, `Couldn't fetch JSON: ${fetchUrls.additional}`, err.message);
+      return helpers.handle500Error(res, `Couldn't fetch JSON: ${fetchUrls.additional}`, err.message);
     }
   } else {
     try {
@@ -158,7 +158,7 @@ const getDataset = async (req, res) => {
         utils.verbose("Request is valid, but no dataset available. Returning 204.");
         return res.status(204).end();
       }
-      return helpers.handleError(res, `Couldn't fetch JSONs`, err.message);
+      return helpers.handle500Error(res, `Couldn't fetch JSONs`, err.message);
     }
   }
   return undefined;
--- a/src/getDataset.js
+++ b/src/getDataset.js
@@ -1,4 +1,6 @@
 const queryString = require("query-string");
+const assert = require('assert').strict;
+
 const utils = require("./utils");
 const helpers = require("./getDatasetHelpers");
 const {NoDatasetPathError} = require("./exceptions");
@@ -91,8 +93,10 @@ const requestMainDataset = async (res, fetchUrls) => {
  */
 const getDataset = async (req, res) => {
   const query = queryString.parse(req.url.split('?')[1]);
-  if (!query.prefix) {
-    return helpers.handle500Error(res, `getDataset request must define a prefix`);
+  try {
+    assert(query.prefix);
+  } catch {
+    return res.status(400).send('getDataset request must define a prefix');
   }
 
   /*
@@ -122,7 +126,7 @@ const getDataset = async (req, res) => {
       utils.verbose(err.message);
       return res.status(204).end();
     }
-    return helpers.handle500Error(res, `Couldn't parse the url "${query.prefix}"`, err.message);
+    return res.status(400).send(`Couldn't parse the url "${query.prefix}"`);
   }
 
   const {source, dataset, fetchUrls, auspiceDisplayUrl} = datasetInfo;
--- a/src/getDatasetHelpers.js
+++ b/src/getDatasetHelpers.js
@@ -24,9 +24,6 @@ const unauthorized = (req, res) => {
  * roundtrip losslessly.
  */
 const splitPrefixIntoParts = (prefix) => {
-  /* This could be const, but use let to signal to the reader that we use
-   * modifying methods like shift() and splice().
-   */
   const prefixParts = prefix
     .replace(/^\//, '')
     .replace(/\/$/, '')
--- a/src/getDatasetHelpers.js
+++ b/src/getDatasetHelpers.js
@@ -1,7 +1,7 @@
 const utils = require("./utils");
 const sources = require("./sources");
 
-const handleError = (res, clientMsg, serverMsg="") => {
+const handle500Error = (res, clientMsg, serverMsg="") => {
   utils.warn(`${clientMsg} -- ${serverMsg}`);
   return res.status(500).type("text/plain").send(clientMsg);
 };
@@ -230,7 +230,7 @@ const parsePrefix = (prefix, otherQueries) => {
 module.exports = {
   splitPrefixIntoParts,
   joinPartsIntoPrefix,
-  handleError,
+  handle500Error,
   unauthorized,
   parsePrefix
 };
--- a/src/getNarrative.js
+++ b/src/getNarrative.js
@@ -6,10 +6,10 @@ const getNarrative = async (req, res) => {
   const query = req.query;
   const prefix = query.prefix;
   if (!prefix) {
-    return helpers.handleError(res, "No prefix in getNarrative URL query");
+    return helpers.handle500Error(res, "No prefix in getNarrative URL query");
   }
   if (!query.type || !["markdown", "md"].includes(query.type.toLowerCase())) {
-    return helpers.handleError(res, "The nextstrain.org server only serves getNarrative requests in markdown format. Please specify `?type=md`");
+    return helpers.handle500Error(res, "The nextstrain.org server only serves getNarrative requests in markdown format. Please specify `?type=md`");
   }
 
   /*
@@ -49,7 +49,7 @@ const getNarrative = async (req, res) => {
     res.set("Content-Type", "text/markdown");
     return response.body.pipe(res);
   } catch (err) {
-    return helpers.handleError(res, `Narratives couldn't be served -- ${err.message}`);
+    return helpers.handle500Error(res, `Narratives couldn't be served -- ${err.message}`);
   }
 };
 
--- a/src/getNarrative.js
+++ b/src/getNarrative.js
@@ -1,15 +1,24 @@
 const fetch = require('node-fetch');
+const assert = require('assert').strict;
+
 const utils = require("./utils");
 const helpers = require("./getDatasetHelpers");
 
 const getNarrative = async (req, res) => {
   const query = req.query;
   const prefix = query.prefix;
-  if (!prefix) {
-    return helpers.handle500Error(res, "No prefix in getNarrative URL query");
+
+  try {
+    assert(prefix);
+  } catch {
+    return res.status(400).send("No prefix in getNarrative URL query");
   }
-  if (!query.type || !["markdown", "md"].includes(query.type.toLowerCase())) {
-    return helpers.handle500Error(res, "The nextstrain.org server only serves getNarrative requests in markdown format. Please specify `?type=md`");
+
+  try {
+    assert(query.type);
+    assert(["markdown", "md"].includes(query.type.toLowerCase()));
+  } catch {
+    return res.status(400).send("The nextstrain.org server only serves getNarrative requests in markdown format. Please specify `?type=md`");
   }
 
   /*
--- a/src/getNarrative.js
+++ b/src/getNarrative.js
@@ -52,7 +52,9 @@ const getNarrative = async (req, res) => {
   try {
     utils.log(`Fetching narrative ${fetchURL} and streaming to client for parsing`);
     const response = await fetch(fetchURL);
-    if (!(response.status === 200 || response.status === 304)) {
+    if (response.status === 404) {
+      return res.status(404).send("The requested URL does not exist.");
+    } else if (!(response.status === 200 || response.status === 304)) {
       throw new Error(`Failed to fetch ${fetchURL}: ${response.status} ${response.statusText}`);
     }
     res.set("Content-Type", "text/markdown");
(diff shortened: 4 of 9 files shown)


src/getNarrative.js
```diff
@@ -43,6 +51,10 @@ const getNarrative = async (req, res) => {
   try {
     utils.log(`Fetching narrative ${fetchURL} and streaming to client for parsing`);
     const response = await fetch(fetchURL);
+    if (response.status === 404) {
+      return res.status(404).send("The requested narrative does not exist.");
+    }
+
```
kairstenfay, line 57: This might belong in a new commit since it's an added feature rather than a conversion from 500 -> 4XX.

src/getSourceInfo.js
```diff
@@ -1,21 +1,25 @@
...
   }
+
   const {source} = helpers.splitPrefixIntoParts(query.prefix);
 
   let sourceInfo;
   try {
     sourceInfo = await source.getInfo();
   } catch (err) {
-    return helpers.handleError(res, `No source info available`, err.message);
+    return helpers.handle500Error(res, `No source info available`, err.message);
```
jameshadfield, line 22: If we're tidying up the HTTP status codes, then this should be

```
return helpers.handle500Error(res, `Error processing source info`, err.message);
```

(I initially thought `204` (no content) would be better, but reading the code of the `getInfo()` methods, they all seem to return an object if there's no available source info, rather than throwing an error.)

src/getNarrative.js
jameshadfield: Looks good. Does this change auspice behavior at all? (I wouldn't think so, but haven't tested)

src/getDataset.js
```diff
@@ -148,7 +151,7 @@ const getDataset = async (req, res) => {
     try {
       await requestCertainFileType(res, req, fetchUrls.additional, query);
     } catch (err) {
-      return helpers.handleError(res, `Couldn't fetch JSON: ${fetchUrls.additional}`, err.message);
+      return helpers.handle500Error(res, `Couldn't fetch JSON: ${fetchUrls.additional}`, err.message);
```
jameshadfield, line 154: For narratives, you now return `404` if the file doesn't exist. This should ideally be the case here as well.

jameshadfield: Or maybe 204 is more appropriate in this case? I'm not an expert on what's best here

src/getNarrative.js
kairstenfay: I have only been testing Nextstrain.org and so far everything works as expected. Do you have any recommendations for testing Auspice, @jameshadfield ?

jameshadfield: If you're running (nextstrain.org's implementation of) auspice as part of testing nextstrain.org, then that should be enough. I'm not sure if you've just been using curl or similar.

src/getDataset.js
kairstenfay: Hmm, I'm not sure either. I've been assuming 404, but after looking up 204, I can see either one as acceptable. Do you have any thoughts here, @tsibley ?

kairstenfay: In the meantime, I made it a 404 error.

kairstenfay:
Thanks for your feedback, @jameshadfield. When you get a chance, please check out the last two commits which incorporated your suggestions. Please let me know if I didn't interpret your comments correctly.

src/getDataset.js
jameshadfield: The code you've added doesn't work as expected because `requestCertainFileType` throws if the request 404s so we jump ahead to the `catch` block and therefore skip your `if (response.status === 404) {` check. You can test this by going to http://localhost:5000/zika which will attempt to fetch http://data.nextstrain.org/zika_root-sequence.json (which doesn't exist).

I think you'll have to change `utils -> fetchJSON()` to throw a custom `Error` (or similar) upon 404, and then check the `err` object in this catch block.

I tend to think this should be a 204, as we designed things around a "make lots of fetches and some of them are allowed to fail", but @tsibley will know the best code to use here.

jameshadfield: P.S. https://github.com/nextstrain/auspice/pull/1226/files is an example of using custom `Error`s. (Comments on that PR welcome!)

kairstenfay: Ok @jameshadfield , I'm finally getting back to this.

I rewrote the 404 errors as 204 errors and created a custom Error class like in the example PR from Auspice you linked.

I also did some extra error handling for unauthorized (403) errors, rewriting them as 404 errors to the client rather than 500. 

Please see the final two commits for these changes. I would appreciate any feedback, especially since I haven't been thinking in the Nextstrain-space for the past few weeks.

jameshadfield (changes requested):
Thanks @kairsten. I think efforts here will be rewarded -- I wrote most of this code a long time ago and it's definitely in need of tightening up now that Nextstrain has grown so much. From the [point of view of the auspice client](https://docs.nextstrain.org/projects/auspice/en/latest/server/api.html), we never really considered how status codes should be interpreted, with the exception of `getDataset -> 204` which was a quick solution to a pressing problem. It may be a good idea to specify our general approach to status codes here (and this could be consistent between charon v1 & v2 syntax). 

To test this PR I added the following tests which describe the status codes I think best for different scenarios. Some of these fail, and I'm happy to discuss further / be convinced that something else is better! See comments in JSON for notes on each one, and please add these (or similar) tests to the PR.

```json
  {
    "name": "Check that getAvailable API call for a non-existent community source 404s",
    "url": "/charon/getAvailable?prefix=/community/jameshadfield/doesntexist",
    "__PR_228_comment": "Kairsten you may consider may be beyond the scope of this PR? I'm very surprised it returns 200!",
    "expectStatusCode": 404
  },
  {
    "name": "Fetch (core) narrative",
    "url": "/charon/getNarrative?prefix=/narratives/ncov/sit-rep/2020-05-15&type=md",
    "expectStatusCode": 200
  },
  {
    "name": "Fetch narrative without specifying a type (invalid request)",
    "url": "/charon/getNarrative?prefix=/narratives/ncov/sit-rep/2020-05-15",
    "expectStatusCode": 400
  },
  {
    "name": "Fetch dataset with a malformed API call",
    "url": "/charon/getDataset",
    "expectStatusCode": 400
  },
  {
    "name": "Attempt to access a private Nextstrain Group when not logged in",
    "url": "/charon/getAvailable?prefix=/groups/blab-private",
    "expectStatusCode": 404
  },
  {
    "name": "Fetch narrative which doesn't exist",
    "url": "/charon/getNarrative?prefix=/narratives/ncov/sit-rep/XXXX-XX-XX",
    "__PR_228_comment": "response in this PR is 400 (bad request) but request is well formed",
    "expectStatusCode": 404
  },
  {
    "name": "Fetch dataset which doesn't exist",
    "url": "/charon/getDataset?prefix=/ncov/XXXX-XX-XX",
    "__PR_228_comment": "response in this PR is 400 (bad request) but request is well formed",
    "expectStatusCode": 404
  },
  {
    "name": "Check getDataset API for non-existing a sidecar file tip-frequencies which doesn't exist",
    "__comment": "Auspice only makes such a request if the dataset JSON indicates that the frequency panel is to be displayed",
    "__PR_228_comment": "response in this PR is still 500 (Internal Server Error), as per before, but should be 404",
    "url": "/charon/getDataset?prefix=/ncov/2020-01-23&type=tip-frequencies",
    "expectStatusCode": 404,
    "responseIsJson": true
  }
```

src/getDataset.js
```diff
@@ -4,6 +4,7 @@ const helpers = require("./getDatasetHelpers");
 const {NoDatasetPathError} = require("./exceptions");
 const auspice = require("auspice");
 const request = require('request');
+const assert = require('assert');
```
tsibley, line 7: Ah, nice, I didn't know about this core NodeJS library. When I looked it up (thinking it was a third-party library), I noticed that the official docs suggest using strict equality by default. This makes sense, as we also prefer `===` vs. `==`. Looks like it should be a small change to use strict equality, by doing:

    const assert = require('assert').strict;

https://nodejs.org/docs/latest-v13.x/api/assert.html#assert_strict_assertion_mode

kairstenfay: Thanks, Tom. I added that into my commits.

kairstenfay:
> Thanks @kairsten. I think efforts here will be rewarded -- I wrote most of this code a long time ago and it's definitely in need of tightening up now that Nextstrain has grown so much. From the [point of view of the auspice client](https://docs.nextstrain.org/projects/auspice/en/latest/server/api.html), we never really considered how status codes should be interpreted, with the exception of `getDataset -> 204` which was a quick solution to a pressing problem. It may be a good idea to specify our general approach to status codes here (and this could be consistent between charon v1 & v2 syntax).
> 
> To test this PR I added the following [tests](https://github.com/nextstrain/nextstrain.org/blob/master/test/smoke-test/auspice_client_requests.json) which describe the status codes I think best for different scenarios. Some of these fail, and I'm happy to discuss further / be convinced that something else is better! See comments in JSON for notes on each one, and please add these (or similar) tests to the PR.
> 
> ```json
>   {
>     "name": "Check that getAvailable API call for a non-existent community source 404s",
>     "url": "/charon/getAvailable?prefix=/community/jameshadfield/doesntexist",
>     "__PR_228_comment": "Kairsten you may consider may be beyond the scope of this PR? I'm very surprised it returns 200!",
>     "expectStatusCode": 404
>   },
>   {
>     "name": "Fetch (core) narrative",
>     "url": "/charon/getNarrative?prefix=/narratives/ncov/sit-rep/2020-05-15&type=md",
>     "expectStatusCode": 200,
>     "responseIsJson": false
>   },
>   {
>     "name": "Fetch narrative without specifying a type (invalid request)",
>     "url": "/charon/getNarrative?prefix=/narratives/ncov/sit-rep/2020-05-15",
>     "expectStatusCode": 400
>   },
>   {
>     "name": "Fetch dataset with a malformed API call",
>     "url": "/charon/getDataset",
>     "expectStatusCode": 400
>   },
>   {
>     "name": "Attempt to access a private Nextstrain Group when not logged in",
>     "url": "/charon/getAvailable?prefix=/groups/blab-private",
>     "expectStatusCode": 404
>   },
>   {
>     "name": "Fetch narrative which doesn't exist",
>     "url": "/charon/getNarrative?prefix=/narratives/ncov/sit-rep/XXXX-XX-XX&type=md",
>     "expectStatusCode": 404
>   },
>   {
>     "name": "Fetch dataset which doesn't exist",
>     "url": "/charon/getDataset?prefix=/ncov/XXXX-XX-XX",
>     "__PR_228_comment": "response in this PR is 400 (bad request) but request is well formed",
>     "expectStatusCode": 404
>   },
>   {
>     "name": "Check getDataset API for non-existing a sidecar file tip-frequencies which doesn't exist",
>     "__comment": "Auspice only makes such a request if the dataset JSON indicates that the frequency panel is to be displayed",
>     "__PR_228_comment": "response in this PR is still 500 (Internal Server Error), as per before, but should be 404",
>     "url": "/charon/getDataset?prefix=/ncov/2020-01-23&type=tip-frequencies",
>     "expectStatusCode": 404
>   }
> ```
> 
> > I rewrote the 404 errors as 204 errors
> 
> I realize I suggested the possibility of using `204` responses more frequently (i.e. in more places than the one documented scenario), but after re-reading the wiki page on this, I think 404 makes sense for most dataset/narrative requests where the underlying file doesn't exist: "404: The requested resource could not be found but may be available in the future". Having said this, my testing didn't expose changes in this PR which mean the server now returns `204`, but I probably missed them.

Thanks so much for writing these tests, @jameshadfield . I added them in a new commit where you're tagged as a co-author.

I wanted to wrap up our 204 vs. 404 discussion -- the consensus seems to be to only return a 204 status code when the requested dataset is a valid Community source dataset which does not exist. All other requests for missing files should return a 404 status code. The smoke tests now reflect this.

src/getDataset.js
kairstenfay: See [my comment further down](https://github.com/nextstrain/nextstrain.org/pull/228#issuecomment-745713851) for the conclusion on this thread re: 404 vs. 204.

src/getDataset.js
```diff
@@ -18,6 +18,7 @@ const requestCertainFileType = async (res, req, additional, query) => {
   if (query.type === "tree") {
     res.json({tree: jsonData});
   }
+  // TODO why does this work without a return?
   res.json(jsonData);
```
kairstenfay, line 22: I didn't have time to dig into this, but if anyone has some insight onto why sometimes we don't have to return a Response object, I'd love to learn more.

kairstenfay:
@jameshadfield @tsibley My PR is ready for re-review.

I've done a lot of rework on the error code / response handling. In some places, I inspect the error code itself (e.g. 404) and return a response with a status, and other times I throw/catch a custom Nextstrain `NoContentError`. I don't like how mixed my code currently is between these two approaches. I wonder if there are some changes I could make to standardize across one approach versus the other? Maybe some uniformity within the implementation of Sources vs. getDataset/getNarrative/etc. could work here? I would love to know your thoughts.

src/getDataset.js
jameshadfield: The short answer is that `res.json(jsonData)` on its own will send the data to the client --  see the [express docs](https://expressjs.com/en/5x/api.html#res.json).

Digging one level deeper, `requestMainDataset` is called by `getDataset`, which doesn't store the return value, so what `requestMainDataset` returns doesn't matter. (In this case the return value is `undefined`.)

Digging another layer deeper, we see that `getDataset` returns a variety of things depending on the code path taken -- `undefined`, [`ServerResponse`](https://nodejs.org/api/http.html#http_class_http_serverresponse), and maybe others. That's probably not ideal in terms of reusing that code in other servers etc.

jameshadfield: If you want to dig even deeper, `getDataset` is the callback [we use to handle GET requests to `"/charon/getDataset"`](https://github.com/nextstrain/nextstrain.org/blob/master/server.js#L92-L93). From [the express docs](https://expressjs.com/en/api.html#app.route) it's not clear if the return value of this callback matters -- the examples there don't explicitly return anything -- but perhaps it is?!?

src/getDataset.js
```diff
@@ -158,7 +166,12 @@ const getDataset = async (req, res) => {
         utils.verbose("Request is valid, but no dataset available. Returning 204.");
```
jameshadfield, line 166: > In some places, I inspect the error code itself (e.g. 404) and return a response with a status, and other times I throw/catch a custom Nextstrain NoContentError. I don't like how mixed my code currently is between these two approaches.

This seems like a good example of this. The code chooses to call either `requestCertainFileType` or `requestMainDataset`. The former throws custom exceptions, which we examine. The latter doesn't, so we examine the status code here. I believe you can modify the latter function to do something like
```js
Promise.reject(NoContentError);
```
but I haven't double checked. This would let you use consistent error handling within the `getDataset` handler.

jameshadfield (changes requested):
Thanks @kairstenfay -- these smoke tests look great and nicely set out how we expect the API to behave. I'm going to re-review tomorrow with fresh eyes but my current thinking is that `UnauthorizedError` is almost never being triggered in the code and is thus largely unnecessary (smoke testing often doesn't pick things like this up, because it's not possible to distinguish between code paths which catch this error and send `404` and other code path that also send `404`).

src/getNarrative.js
```diff
@@ -43,13 +53,17 @@ const getNarrative = async (req, res) => {
   try {
     utils.log(`Fetching narrative ${fetchURL} and streaming to client for parsing`);
     const response = await fetch(fetchURL);
-    if (!(response.status === 200 || response.status === 304)) {
+    if (response.status === 404) throw new NoContentError();
```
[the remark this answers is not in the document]
kairstenfay: Thanks, I appreciate this suggestion. I was getting bogged down in all the status codes and error handling that it was getting increasingly harder for me to write clean code.

src/getDataset.js
kairstenfay: Thank you!

src/getDatasetHelpers.js
```diff
@@ -1,7 +1,7 @@
 const utils = require("./utils");
 const sources = require("./sources");
 
-const handleError = (res, clientMsg, serverMsg="") => {
+const handle500Error = (res, clientMsg, serverMsg="") => {
```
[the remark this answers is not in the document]
kairstenfay: I would be open to `handleServerError` as an alternate naming scheme.

src/getSourceInfo.js
```diff
@@ -1,21 +1,31 @@
...
   const query = queryString.parse(req.url.split('?')[1]);
-  if (!query.prefix) {
-    throw new Error("No prefix defined");
+  try {
+    assert(query.prefix);
+  } catch {
+    return res.status(400).send("No prefix defined");
   }
+
   const {source} = helpers.splitPrefixIntoParts(query.prefix);
```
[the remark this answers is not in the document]
kairstenfay: Sure. Which status code should be returned given that example URL? 404? 400?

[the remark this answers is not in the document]
kairstenfay: I'll also add this URL to the smoke tests.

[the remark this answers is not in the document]
kairstenfay: Actually, given that this is outside of the scope of this PR, I'll just expect a 500 response code for now.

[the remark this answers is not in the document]
jameshadfield: > Which status code should be returned given that example URL? 404? 400?

Not `400`, as the request is properly formatted. `404` would be ok (the resource doesn't exist). `500` would be the most accurate as  "The server has encountered a situation it doesn't know how to handle" and a separate PR could debug the `TypeError: Source is not a constructor` error.

src/sources.js
```diff
@@ -206,7 +207,9 @@ class CommunitySource extends Source {
     const qs = queryString.stringify({ref: this.branch});
     const response = await fetch(`https://api.github.com/repos/${this.repo}/contents/auspice?${qs}`);
 
-    if (response.status !== 200 && response.status !== 304) {
+    if (response.status === 403) throw new UnauthorizedError();
```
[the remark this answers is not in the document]
kairstenfay: Oh, good point. I didn't test this bit of the code when i was trying to cover all bases. GitHub would mask a private repo as a 404. I'll remove this bit.

src/sources.js
```diff
@@ -380,9 +383,13 @@ class S3Source extends Source {
...
   }
   async availableDatasets() {
-    const objects = await this._listObjects();
-    const pathnames = utils.getDatasetsFromListOfFilenames(objects.map((object) => object.Key));
-    return pathnames;
+    let objects;
+    try {
+      objects = await this._listObjects();
+    } catch (err) {
+      if (err.statusCode === 403) throw new UnauthorizedError();
```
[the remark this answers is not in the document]
kairstenfay: I deleted this bit of the code

src/sources.js
```diff
@@ -472,6 +479,8 @@ class S3Source extends Source {
       };
 
     } catch (err) {
+      if (err.statusCode === 403) throw new NoContentError();
```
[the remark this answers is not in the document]
kairstenfay: I deleted this bit of the code

kairstenfay:
Thanks so much for your feedback again, @jameshadfield . I believe I've addressed every PR comment you left. And you were right about the `UnauthorizedError` never being triggered. I dropped the related commit for catching those errors from this PR.

src/getDataset.js
kairstenfay: Thanks!

src/getSourceInfo.js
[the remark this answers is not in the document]
kairstenfay: As suggested, it raises a 500 error for now.

jameshadfield (changes requested):
Thanks for the changes @kairstenfay -- the smoke tests look good (and pass) and the code looks good in my view. I think all that's needed is to change the name of `NoContentError` and fix the linting errors (which is why the build is currently failing).

src/exceptions.js
```diff
@@ -7,6 +7,12 @@ class NextstrainError extends Error {
   }
 }
 
+class NoContentError extends NextstrainError {
```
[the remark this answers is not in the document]
kairstenfay: Great point. I originally named it `NoContentError` when I was under the impression we were returning 204 codes instead of 404. I'll update that.

kairstenfay:
> Thanks for the changes @kairstenfay -- the smoke tests look good (and pass) and the code looks good in my view. I think all that's needed is to change the name of `NoContentError` and fix the linting errors (which is why the build is currently failing).

Error class renamed and linter errors removed (sorry I missed that)!

jameshadfield (approved):

PR #229, opened 2020-11-18
update sars-build registry

rneher:
update sars splash -- might be too much...


rneher:
looking at the page, I think this is actually still manageable and a useful addition

jameshadfield (approved):
Agree -- this is still very usable and it's better that we surface these builds. cc @eharkins who has been thinking about ways to improve the overall page layout / UI here.

Issue #196, opened 2020-06-25
Allow access to S3 versioned datasets


jameshadfield:
This issue is the expansion on a comment by @joverlee521 in https://github.com/nextstrain/nextstrain.org/pull/192#issuecomment-649132688

> Would it help to take advantage of the versioning enabled in the nextstrain-data bucket? I believe you can [retrieve specific versions of an object from S3](https://docs.aws.amazon.com/AmazonS3/latest/dev/list-obj-version-enabled-bucket.html)

This would be interesting for nextstrain.org to explore as an alternate way in which we could access "old" datasets instead of making separate datestamped files. Note that `nextstrain-data` is already using versioning. The best / first usecase for this would be the seasonal flu builds, especially when combined with the search functionality of #192. 

It's unclear whether this approach would be better than simply keeping datestamped files on the bucket, but if we want to pursue this we could implement a complete solution within the nextstrain.org server with no auspice modifications necessary.

### Example

The h1n1-ha-3y dataset from 2020-01-01 could be accessed via https://nextstrain-data.s3.amazonaws.com/flu_seasonal_h1n1pdm_ha_3y.json?versionId=1HEJjqgcCardojwksJF1wIyDyGD3HE_s were the object to be public, or we can do it server side (with credentials) via 

```js
S3.getObject({
  Bucket: "nextstrain-data",
  Key: "flu_seasonal_h1n1pdm_ha_3y.json",
  VersionId: "1HEJjqgcCardojwksJF1wIyDyGD3HE_s"
}).promise()
```

Leveraging similar syntax which we already have for accessing datasets on a particular github branch -- e.g. [/community/jameshadfield/scratch@test-branch/placentalia](https://nextstrain.org/community/jameshadfield/scratch@test-branch/placentalia) we could process URLs such as `/flu/seasonal/h1n1pdm/ha/3y@1HEJjqgcCardojwksJF1wIyDyGD3HE_s` or go one step further and have the server keep track of the upload date of each version to allow URLs such as `/flu/seasonal/h1n1pdm/ha/3y@2020-01-01`. Since the nextstrain.org server essentially acts as a middleman between S3 and the client it is possible to interpret such a `getDataset` API call and return the correct version of the object (file). There may be some cloudfront modifications needed for this to work. 

[...]

Issue #182, opened 2020-06-16
SARS-CoV-2 tools


rneher:
There is a growing need for people to assign their sequences to clades and do some simple QC. In the longer term, we also want to support running augur on SARS-CoV-2 data in the cloud. I imagine a collection of tools that grows over time, but there are two distinct categories:

1) full builds that are run in the cloud with the results hosted permanently. This probably wants to be integrated with groups and requires more infrastructure

2) simple tools that run on the client like clade assignment, QC (mutations relative to reference etc). We have one prototype of such a tool at https://github.com/neherlab/webclades

We are planning to have a site that lists all SARS-CoV-2 builds (https://github.com/nextstrain/nextstrain.org/issues/180). This Issue is meant to start a discussion of how to make builds and tools as discoverable as possible.

jameshadfield:
My idea (not particularly innovative) would be to transform https://nextstrain.org/sars-cov-2/ into a general landing page for all things SARS-CoV-2. This would include (either within the page, or via links)

* General introduction
* https://clades.nextstrain.org/
* List of all SARS-CoV-2 datasets that we are aware of (what's currently on https://nextstrain.org/sars-cov-2/). We talked previously about making a geo-based viz for this to better convey the coverage of builds.
* List of situation reports (https://nextstrain.org/ncov-sit-reps)
* Link to tutorial to run your own build https://nextstrain.github.io/ncov/

emmahodcroft:
I like James' idea above 👍

eharkins:
First version of this was done in https://github.com/nextstrain/nextstrain.org/pull/206 and can be seen at https://nextstrain.org/sars-cov-2/. Today on a call we talked about some improvements, here are some notes / proposals from the meeting notes:

> 1. Can we make it so that people can update their build inclusion in our ‘list’ via a PR somehow?
> 2. Can we get some easy wins initially by some visual adjustments to our current list - 2 columns, smaller gaps, smarter wrapping?
> 3. Long term - can we make it more generic that can be used for any Nextstrain builds lists. Should it be searchable? Map? Images? 
> 4. Can we move things like the search database updates to Github Actions so that they’re easier to run (either Build Shepherds or scheduled) once a month or so, and keep up to date?

I can help make simple UI improvements right now (number 2) if we are clear on how we want this to look.

trvrb:
[...]

PR #230, opened 2020-11-19
Update sars-cov-2 search regularly

eharkins:
This is a Github action which runs `scripts/collect-search-results.js` and uploads this data to s3 in order to update https://nextstrain.org/search/sars-cov-2. It is scheduled to happen every weekday as close as possible to the time by which we have usually published new sars-cov-2 (ncov) builds to s3; as I wrote in a comment in the workflow itself, the timing of this is imperfect for sars-cov-2 (ncov) builds because they are run in different time zones on alternating days. More jobs in this workflow or separate ones with separate timelines can be added for other pathogens we want to maintain search pages for. If we want to make this 'smarter' we can use something like aws lambdas which could automatically run upon updating of certain files in an s3 bucket, but that doesn't seem to be such a huge improvement over this to necessitate the time it would take (at least for me with my lack of lambdas experience).
 
### Related issue(s)  
[Slack thread](https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1605660746030400)
#187 
#192 
#196 
#182 

### Testing
I will trigger this manually now to test that it works and also monitor it to see that the cron schedule is as desired (I think it will apply even though it's in a branch since I didn't say anything about which branch it runs on, but I need to check the default there.

### DIFF
--- a/.github/workflows/update-search.yml
+++ b/.github/workflows/update-search.yml
@@ -1,10 +1,6 @@
 name: 'update nextstrain.org/search from latest datasets in s3'
 
 on:
-  push:
-    branches:
-      - update-search
-
   schedule:
     # * is a special character in YAML so you have to quote this string
     # "At 01:00 UTC on every day-of-week from Tuesday through Saturday. 
@@ -15,10 +11,6 @@ on:
     # so this update will lag behind that by ~9 hours. It can also be
     # triggered from the command line at any time with a respository_dispatch (below)."
     - cron:  '0 1 * * TUE-SAT'
-
-  # Manually triggered using `./bin/trigger update-search`
-  repository_dispatch:
-    types: update-search
 
 jobs:
   update-search:
--- a/.github/workflows/update-search.yml
+++ b/.github/workflows/update-search.yml
@@ -1,6 +1,10 @@
 name: 'update nextstrain.org/search from latest datasets in s3'
 
 on:
+  push:
+    branches:
+      - update-search
+
   schedule:
     # * is a special character in YAML so you have to quote this string
     # "At 01:00 UTC on every day-of-week from Tuesday through Saturday. 
--- a/.github/workflows/update-search.yml
+++ b/.github/workflows/update-search.yml
@@ -33,6 +33,7 @@ jobs:
         npm ci
         python3 -m pip install --upgrade pip setuptools
         python3 -m pip install nextstrain-cli
+        PATH="$HOME/.local/bin:$PATH"
         ./scripts/collect-search-results.js --pathogen sars-cov-2
         nextstrain remote upload s3://nextstrain-data ./data/search_sars-cov-2.json
       env:
--- a/.github/workflows/update-search.yml
+++ b/.github/workflows/update-search.yml
@@ -21,20 +21,18 @@ on:
     types: update-search
 
 jobs:
-  ingest:
+  update-search:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/update-search.yml
```diff
@@ -0,0 +1,39 @@
...
+      run: |
+        npm ci
+        python3 -m pip install --upgrade pip setuptools
+        python3 -m pip install nextstrain-cli
+    - name: update sars-cov-2 search
+      run: |
+        ./scripts/collect-search-results.js --pathogen sars-cov-2
+        nextstrain remote upload s3://nextstrain-data ./data/search_sars-cov-2.json
+      env:
+        AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
```
eharkins, line 37: I don't know if this is necessary, I just added it since other AWS github actions we have specify it.

eharkins: Which reminds me, could @jameshadfield or someone else with permissions add the AWS secrets in this workflow to this repository? Thanks!

emmahodcroft:
I'm not really qualified to review for this, but the functionality sounds great, and I think this seems like a good solution to me, as far as being (relatively) easy to get up and running, and will greatly improve having that search function stay up-to-date!

.github/workflows/update-search.yml
jameshadfield: @eharkins I have created an appropriate IAM user & sent you the credentials to add to the GitHub setup 👍

eharkins:
This basically works now. Still todo:
- allow triggering manually from the command line using a copy of https://github.com/nextstrain/ncov-ingest/blob/master/bin/trigger (or maybe we can share this across the organization if we think it can be generalized)
- test cron timing
- add other pathogens in addition to sars-cov-2
- Fetching dataset script improvements (separate PR)

eharkins:
> allow triggering manually from the command line using a copy of https://github.com/nextstrain/ncov-ingest/blob/master/bin/trigger (or maybe we can share this across the organization if we think it can be generalized)

> test cron timing

These only work on default branch for a repository so won't be able to test until we merge.

How often would we like to update https://nextstrain.org/search/seasonal-flu and are there other search pages we have or would like to have that we should include here?

Once we get those added, I think it's ok to merge and then begin testing the different ways to run this workflow mentioned above.

jameshadfield (approved):
Looks good Eli! Thanks for making those recent changes. Let's merge at a time when we can monitor that the `schedule` works as expected.

eharkins:
Looking like the [schedule is working now](https://github.com/nextstrain/nextstrain.org/runs/1477408994?check_suite_focus=true).

I let Jover know this exists in case we want to use it for seasonal-flu.

I also realized the only difference between 
> allow triggering manually from the command line

and running this locally as we have done in the past would be that it happens on gh actions, which has pros and cons. I would imagine 
> Fetching dataset script improvements (separate PR)

is a bigger priority, @jameshadfield did you have specific improvements in mind or did you just want to generally fetch things more quickly/efficiently?

jameshadfield:
> specific improvements in mind or did you just want to generally fetch things more quickly/efficiently

The script as written was amost a proof-of-principle approach. Each time it runs it downloads every JSON; a better approach would only download new/changed JSONs and store the results somewhere for access next time. Not a huge issue right now as S3 access is cheap, but something to consider in the future as needed. I would place other things - SARS front-page UI etc - as a higher priority.

Issue #62, opened 2019-12-13
Revamp dataset selection UI


trvrb:
The current "tile" UI to navigate to datasets on the splash page works for well for a few items; it's attractive and approachable. However, I can imagine aiming for ~20 core datasets alongside as about as many non-core datasets. I'd propose a general UI for dataset selection. This would get used on the splash page, but could also be used on say nextstrain.org/groups/blab/ to display `blab` specific datasets.

I imagine this working by crawling the S3 bucket at some interval and collecting datasets and the meta portion of the JSON (primarily `updated`).

I wanted to combine datasets that begin with the same filestem the way we do now, otherwise the 40 seasonal flu datasets will overwhelm the single Zika dataset. After playing around for a little while, here's what I came up with:

![dataset-selection-closed](https://user-images.githubusercontent.com/1176109/70769017-a43d6c00-1d1c-11ea-8c0b-12d0dc97b2bb.png)

The filestem is used to collect datasets under a single umbrella. The dataset owner is displayed alongside logo as well as date updated for the most recent dataset. The toggle on the left can be used to display a list of all datasets collected under the same umbrella:

![dataset-selection-open](https://user-images.githubusercontent.com/1176109/70769168-48271780-1d1d-11ea-9f1e-8b20e1cc6236.png)

This reuses the "list" styling in Auspice built for filters.

This is related to issue #48 (manifest). Sketch mock can be found on [Google Drive here](https://drive.google.com/drive/folders/1e97lJE8ONR9TV-kDd9XVM-S8S05mUr17).

tsibley:
Looks very nice! I think this is a good direction.

Instead of using this table on the splash page, I do think it'd be nicer to keep a small number of selected tiles followed by a link to "see more…" that goes to another page which uses this table view to show all builds/datasets (nextstrain.org/core or something).

trvrb:
I'd be happy with something like this. I had thought about something like nextstrain.org/pathogens or nextstrain.org/datasets to give full listing of core and non-core, but that nextstrain.org/core could just show full listing of core datasets, the same way that nextstrain.org/groups/grubaughlab would show listing of grubaughlab datasets.

I imagine also allowing fields to be sorted and having an elastic search box.

jameshadfield:
Seems good to be thinking about this.

jameshadfield:
[...]

PR #232, opened 2020-11-25
Build listing UI improvements

eharkins:
As discussed in various issues listed below, on slack, and in nextstrain meetings, the list of sars-cov-2 builds is getting rather long and would benefit from changes to the interface to make it take up less space and allow you to get the build you are looking for (or just browse) more effectively and faster. 

This is just a start which steals the collapsing-menu code from the sit-reps listing directly below the builds listing on the same page, but already saves a decent chunk of space (I added some duplicate builds to fill out a row):
Screen Shot 2020-11-24 at 4 53 09 PM

I wanted to just create this PR so that we can get some 👍  or 👎  before taking it further in any particular direction. 

I also thought it would be worth mentioning that if all we wanted was a collapsible listing like this, read the docs does this pretty well as part of their table of contents, so we could just move such listings into a top-level docs heading at docs.nextstrain.org or a separate read the docs project (you could imagine builds.nextstrain.org or something like that).

### Related issue(s)  
#62 
#180 
#182 
#206 

### Testing
Locally with gatsby


eharkins:
Also want to note that the code used to do this can be abstracted out better so that it may be reused for listing builds for different pathogens. Just waiting to do that until we were closer to choosing a particular approach (although these things can probably happen at the same time).

emmahodcroft:
I like how this is developing - the layout is more compact and being able to collapse regions really helps to get an overview and just focus on whatever part you want to focus on!

Just a couple of suggestions off the top of my head: The expand/collapse animation is a bit too slow for me - if you are clicking through a number of EU countries or US states, it gets a little irritating. 

I also wonder if an 'expand all' for some categories - like Europe or USA states might be useful? If I imagine I'm going to be clicking between many different EU countries for example, I may want to just expand them all so I can click around more easily. We probably wouldn't want that for, say, all builds in all regions, though - just some 'subsections' (I hope that makes sense!).

eharkins:
@emmahodcroft thanks for the feedback!

9551913 addresses:
1. the animation time; this should be much faster
2. the collapsible menu problem; since we use a javascript library to make these menus, it would be non-trivial to make a collapse / expand all feature. If this is important, we can certainly make it happen, but instead I am trying a simple approach of just auto-expanding menus when it seems reasonable to do so (i.e. when there are fewer than 5 dropdowns at that same level - since e.g. expanding all countries in Europe might be overkill). Hopefully between this and reducing the animation time, the overall experience is improved to make an expand-all less necessary; there are definitely other improvements to make that might help with this general goal of more quickly navigating these menus both in cases where you are looking for something particular and just browsing.

eharkins:
Some styling improvements still to come here as well.

emmahodcroft:
>the collapsible menu problem; since we use a javascript library to make these menus, it would be non-trivial to make a collapse / expand all feature. If this is important, we can certainly make it happen, but instead I am trying a simple approach of just auto-expanding menus when it seems reasonable to do so (i.e. when there are fewer than 5 dropdowns at that same level - since e.g. expanding all countries in Europe might be overkill). Hopefully between this and reducing the animation time, the overall experience is improved to make an expand-all less necessary; there are definitely other improvements to make that might help with this general goal of more quickly navigating these menus both in cases where you are looking for something particular and just browsing.

Thanks for clarifying Eli, that makes sense! I hadn't thought about this from a computational point of view!

PR #234, opened 2020-12-05
map sars-cov-2 builds

eharkins:
_just creating this for feedback before I go too far with styling and function_

known bugs at this point:
map scrolling vertical limit
same coordinate builds
mobile view
general styling improvements

### Related issue(s)  

Fixes #  
Related to #  

### Testing


jameshadfield:
@eharkins I added some commits which allow this to build under Gatsby (see commit messages) & removed a few React key errors. Feel free to squash them into your commit as you progress with this. Please double check that this hasn't changed any of the appearance of the map!

jameshadfield:
This is a good start Eli -- thanks for posting 🙏 Here are my initial comments, in no particular order!

### Map sizes

Giving `MapContainer` `width: 100%` is problematic on mobile as there is 25px of padding either side. For bigger screens we also want it to appear in a wider format to make use of available space. The following styling worked well for me but need further testing:

```
const MapContainer = styled.div`
  max-width: 1080px;
  width: 100%;
```

Note also that the `+` `-` icons are off-screen for smaller resolutions. I suggest moving them to the lower-right, as in Auspice.

### Tooltips & Zoom

Currently we have behavior where clicking on tooltips zoom into the map, and to open a build one has to click on the link in the on-hover info-box. This necessitates keeping the info-box open when the mouse leaves the marker. This is tricky to get right -- for instance, when you hover over another marker you must ensure the previous info-box disappears, else you can get lots shown simultaneously (as is the case currently). 

I think a better approach is to have the info-box only appear while hovering over the marker, and clicking on the marker jumps to the build. We may need to explore what to show at different zoom levels -- for instance, when fully zoomed out we don't show all the European builds but rather display text / icons along the lines of "43 builds available, zoom in to see more".


### Legend

I would change the legend elements to be "Nextstrain build" and "Community Build" (with a black? circle for the latter). As we go further we may have differnet heirachies here (Public Health Lab, research group etc). In general I think the legend can be made more subtle, and on-hover info boxes used to explain more as needed. 

### Map tiles

We should choose wireframe / greyscale map tiles so that the focus is on the markers (and not the terrain).

### Future ideas to keep in mind (beyond this PR)

* It would be worth exploring ways to map points to areas (as a build is rarely a "point"). For instance we could associate shapes to certain builds, so that when hovering over their marker you get an indication of what's in the build -- this may be as simple as conveying whether a build focuses on a country or a region.

* If the list of builds grows & is maintained (as we hope) then adding search functionality to auto-filter the displayed builds will be valuable. Here i'm thinking similar to Auspice's sidebar filtering.

eharkins:
Thanks @jameshadfield! Just pushed some commits addressing this feedback in pieces, but more to come. Something to note about the new map marker implementation: since you can't hover on mobile you cant get information about builds on the map via tooltip, you can only click them as links.


next up:
- changing map tiles
- dealing with overlap of markers
- making sure map sizing is good for all screen sizes
- legend tooltip info

eharkins:
Remaining todos:

- some map styling
- clustering builds / map markers with the same or similar coordinates (like this http://alex3165.github.io/react-mapbox-gl/demos):

I have some code taken from the example mostly working, the thing is they don't actually ever separate when things have exact same coords. One solution is to give them slightly different coords such that when you zoom in far enough they separate. A better solution i think is to leave the coords the same, and implement the clickable tooltip (like in the example above) that lists all the children of that cluster and allow you to click on them as links inside that tooltip so that they never have to separate at all (faster UX than clicking to zoom many times).

jameshadfield:
This is looking really good Eli!

emmahodcroft:
Hooray, I was happy to catch this while it was deployed! (I missed the last few) This is looking awesome guys! 

I know it's still under development, but I found I can't drag the map around (like, after zooming in) - you probably already are aware of that. 

I think the idea of one location that pops up with a list is good - that way people can easily see all 'versions' (if there are multiple) of a build equally, rather than just happening to click on one first because it's more 'North' or similar. 

I love the differentiating of nextstrain & community builds!

I can't zoom in & drag to North America so I might just not be able to see it, but was wondering if 'regional' that's below Nextstrain regional should also have a circle of different colour (South-Central build, for example) to tell apart from states or countries. Anyway - that's a random thought.

This is very, very cool!

eharkins:
> I know it's still under development, but I found I can't drag the map around (like, after zooming in) - you probably already are aware of that.

I was not aware of this, thanks @emmahodcroft. What browser are you using and did you zoom by double clicking the map or using the zoom buttons? Assuming you are on desktop not mobile? Any other details to help reproduce are greatly appreciated!

emmahodcroft:
I'm on Windows 10 pro, using Chrome Browser version 87.0.4280.88 (haven't tried on mobile)

Ah - I hadn't tried zooming by double-clicking! So I can do this and it lets me see more of the map - since it also re-centers to wherever you clicked. But I can't use the hand to like 'hold and drag' like you can with Nextstrain maps or Google maps. Not sure why... I can 'close the hand' by clicking, but when I move it the map does not move.

The double-clicking did let me zoom in on the US though, so I can see the South-Central build - I wonder if maybe we should just make these red, to show it's a step up in 'level' from the blue builds? Otherwise it might look at a glance like it's a build for... east Texas or something.

I've done a truly terrible job of coloring it in with a red pointer-pen to illustrate 😆 
image

eharkins:
@emmahodcroft does your computer have a touchscreen? That would explain the issue since dragging to pan the map was configured only if a touch device is not detected: https://github.com/nextstrain/nextstrain.org/pull/234/files#diff-0f173667f13492ac340f86da7913e9ca9b1ba4dd31e96264c5b7f99f2d0eb4baR110

I think I will just remove this line either way, since it was just inherited from some code I started with, and I'd rather have touch behavior not working perfectly in general than have clicking / mouse behavior only work on devices that don't have touch capabilities.

Regarding the regional builds, I agree it makes sense to differentiate them visually. We could do this by color or by making them larger, or in a different way. We have no explicit coding of "level" of the builds in the [build yaml file](https://github.com/nextstrain/nextstrain.org/blob/69990ede01df6a9701b8bd5cc2c76af1adc46975/static-site/content/allSARS-CoV-2Builds.yaml) but could either:
- infer that from the `parentGeo`, `geo` hierarchy or
- include a field specifying the level at which we'd like to recognize each build visually on the map. 

I might go for the latter since it's easier and more flexible. Should be simple to do either way, so I'll push a first attempt next week and we can go from there.

emmahodcroft:
Ah - good catch! Sorry, I never use my touchscreen (find it totally annoying) so tend to forget it's there. Yes - I do have a touch screen (which I wish I could disable to be honest). 

Weirdly, though I can double-tap and 'pinch out' with the touch screen to zoom, I still can't pan/move the map. Either with one finger or two (sometimes you have to use 2 for embedded maps to avoid dragging while scrolling).

So yes, I'd probably vote to take this out - esp given it doesn't seem to work for touchscreens! 

For the regions - yep, sounds like a good proposal for me! I'd also agree that the latter may be better - in case we end up with more levels or even just more 'things' that we want to specify to be different somehow. This would give us nice future flexibility to be able to have multiple 'types' of markers for different settings, and that might be very useful!

trvrb:
This is really nice work Eli! Just a few quick thoughts:

1. I think I'd go with brand color green https://github.com/nextstrain/auspice/blob/master/src/globalStyles.js#L12 for the "Nextstrain" dots and make them a little larger than the blue community dots. The flower logos are cute, but I think a bit too much when there's 7 of them.
2. It would be nice if the tooltip gave a bit more information on the contents of the build. I'm thinking of things like total strains in dataset and last updated date. However, if JSON doesn't have these this functionality can certainly wait.
3. `/ncov/global` looks a bit strange hanging out in middle of the Atlantic. I don't have a great place to put this. Maybe best to leave out?
4. The map should probably have some latitude bounds. Right now you can drag past the north pole and past the south pole (not hugely important).
5. The starting map lat/long bounds are slightly off. Greenland should move up out of view and Australia should be better in view. But starting zoom level is great.
5. Place a bit of text in the tooltip in italics to warn the user that clicking will jump them over to a detailed dataset view. Just like in Auspice each tooltip uses italics text to say what a click will do.

tsibley:
@eharkins Really awesome to see this coming together!

While taking a gander, I noticed that some breakpoints for the width of the map seem to be missing (at least relative to the body text above it). Best demonstrated:

sars-cov-2-map-breakpoints

eharkins:
^ These commits should address most concerns discussed except for:
@trvrb's:
> It would be nice if the tooltip gave a bit more information on the contents of the build. I'm thinking of things like total strains in dataset and last updated date. However, if JSON doesn't have these this functionality can certainly wait.

This seems like it would require requesting some info about these datasets. As @jameshadfield pointed out to me, that's something we do already for https://nextstrain.org/search/sars-cov-2 (albeit for nextstrain datasets only, not community ones) so maybe we can unify the approach for fetching those datasets to give us what we need for both resources on nextstrain.org (search and map). Probably something for a separate PR.

and

https://github.com/nextstrain/nextstrain.org/pull/234#issuecomment-747146059 (map marker clustering); just in case something about this doesn't work, it's in a PR to this branch: https://github.com/nextstrain/nextstrain.org/pull/242, since I will be on vacation for a bit in case @jameshadfield is keen to get the map merged with or without this feature in the meantime.

PR #235, opened 2020-12-08
Charon v2

kairstenfay:
Take a first pass at the Charon V2 API for GET requests only. I'm interested in your feedback on what I've written so far. I did the bulk of this a few weeks ago, but I _think_ this is a fairly complete first pass. Apologies if I've forgotten where I stood here -- if so, I will push additional commits.

### Related issue(s)  
Incomplete prerequisite for https://github.com/nextstrain/cli/issues/90.

I am interested in feedback on what I've written so far. Next steps for this repo include:
* Documenting the Charon V1 API with a flowchart I've created
* Adding POST/PUT and DELETE endpoints 
* ...?

### DIFF
--- a/server.js
+++ b/server.js
@@ -83,8 +83,39 @@ app.use(express.static(gatsbyAssetPath()));
 app.route("/dist/*")
   .all(expressStaticGzip(auspiceAssetPath(), {maxAge: '30d'}));
 
-
 /* Charon API used by Auspice.
+ */
+app.route("/charon/v2/dataset/")
+  .get([(req, res, next) => {
+    // Add empty prefix
+    req.prefix = '';
+    next();
+  }, auspiceServerHandlers.getAvailable]);
+
+app.route("/charon/v2/dataset/:path*/:type$")
+  .get([(req, res, next) => {
+    // Construct URL from parameters
+    req.query.prefix = path.join(req.params.path, req.params['0'], req.params.type);
+    next();
+  }, auspiceServerHandlers.getDataset]);
+
+app.route("/charon/v2/narrative/:name*")
+  .get([(req, res, next) => {
+    // Construct query prefix, type from parameters
+    [req.params['0'], req.query.type] = req.params['0'].split('.');
+    req.query.prefix = path.join('narratives', req.params.name, req.params['0']);
+    next();
+  }, auspiceServerHandlers.getNarrative]);
+
+app.route("/charon/v2/sourceInfo/:path*")
+.get([(req, res, next) => {
+  // Parse query
+  req.prefix = path.join(req.params.path, req.params['0']);
+  next();
+}, auspiceServerHandlers.getSourceInfo]);
+
+
+/* Preserve legacy (v1) charon API
  */
 app.route("/charon/getAvailable")
   .all(cors({origin: 'http://localhost:8000'})) // allow cross-origin from the gatsby dev server
--- a/server.js
+++ b/server.js
@@ -9,6 +9,7 @@ const compression = require('compression');
 const argparse = require('argparse');
 const utils = require("./src/utils");
 const cors = require('cors');
+const queryString = require("query-string");
 
 const production = process.env.NODE_ENV === "production";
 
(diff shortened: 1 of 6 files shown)


server.js
```diff
@@ -82,21 +83,65 @@ app.use(express.static(gatsbyAssetPath()));
...
+
+
+/* Preserve legacy (v1) charon API
+ */
 app.route("/charon/getAvailable")
   .all(cors({origin: 'http://localhost:8000'})) // allow cross-origin from the gatsby dev server
-  .get(auspiceServerHandlers.getAvailable);
+  .get([(req, res, next) => {
+    // Parse prefix
+    req.prefix = queryString.parse(req.url.split('?')[1]).prefix || "";
```
tsibley, line 124: This (and similar chunks in other route handlers) could be simplified to `req.query.prefix || ""`.

That said, since the handler has direct access to `req` itself, it begs the question of why it's necessary to do even that.

I think it'd be clearer and more maintainable to refactor the various `auspiceServerHandlers.*` into plain functions that take a `prefix` (or `dataset` or `narrative`) instead of chaining route handlers and passing what amount to function args via non-standard properties on `req`. Then both v1 and v2 route handlers could call those plain functions.

src/getDataset.js
```diff
@@ -90,9 +91,12 @@ const requestMainDataset = async (res, fetchUrls) => {
...
 const getDataset = async (req, res) => {
-  const query = queryString.parse(req.url.split('?')[1]);
-  if (!query.prefix) {
-    return helpers.handleError(res, `getDataset request must define a prefix`);
+  const query = req.query;
+
+  try {
+    assert(query.prefix);
+  } catch {
+    return res.status(400).send('getDataset request must define a prefix');
```
tsibley, line 99: Why `try/catch` with `assert()` instead of just keeping the `if`?

src/getDataset.js
```diff
@@ -133,7 +137,7 @@ const getDataset = async (req, res) => {
   }
 
   const baseUrl = req.url.split(query.prefix)[0];
-  let redirectUrl = baseUrl + '/' + auspiceDisplayUrl;
+  let redirectUrl = path.join(baseUrl, auspiceDisplayUrl);
   if (query.type) {
     redirectUrl += `&type=${query.type}`;
   }
```
tsibley, line 143: `path.join()` will vary across operating systems.  This could be addressed by using `require("path/posix")` or `path.posix.join()`, but that said, the semantics of filesystem paths aren't identical to those of URL paths, despite being very similar.

Looking more broadly at this chunk of code, the existing form is already problematic for several reasons, including that whether your other query params get passed along or not with the baseUrl depends on where they are relative to the prefix param. (Also, while unlikely in this case, if your prefix value appears elsewhere in the URL earlier than the query param, it all breaks.) A better approach is to replace the `prefix` query value in the request URL with the desired new value using URL handling libs instead of string manip. Something like:

    const redirectUrl = new URL(req.url);
    redirectUrl.searchParams.set("prefix", auspiceDisplayUrl);

tsibley (commented):
Thanks for starting this work. I left some line-specific comments, but other general impressions are below.

- The new `getSourceName()`, `getIsNarrative()`, and `getSource()` functions _modify_ the arguments passed to them and _must_ be called in a specific order because of that.  This makes the code dangerously hard to reason about.

- `getIsNarrative()` seems like it should return a value, but it doesn't at the point its refactored.  (A return is added in the next commit.)

- It's not clear to me why `getSourceName()` and `getSource()` are separate at all. That is, why not just the latter?

- `getSourceName()` and `getIsNarrative()` would match the rest of their sibling functions if they kept the `prefixParts` naming for their args.  I'd intentionally standardized the `prefix` and `prefixParts` names across these functions (and their callers) to make flow easier to follow.  They weren't before, and it made the code less clear.

- Stepping back, by the time I made it to the end of the branch, I wasn't sure why the functions above were introduced at all since no new callsites of them were introduced.

kairstenfay:
Thanks for your review @tsibley 

> * The new `getSourceName()`, `getIsNarrative()`, and `getSource()` functions _modify_ the arguments passed to them and _must_ be called in a specific order because of that.  This makes the code dangerously hard to reason about.

This is a good point.

> * `getIsNarrative()` seems like it should return a value, but it doesn't at the point its refactored.  (A return is added in the next commit.)

I think that was a mistake on my end. The return should have been added in the same commit. 

> * It's not clear to me why `getSourceName()` and `getSource()` are separate at all. That is, why not just the latter?

There was an initial motivation for this (that I'll get into below), but ultimately I just prefer more, smaller functions that do less, but not at the cost of readability or reasonableness.

> * `getSourceName()` and `getIsNarrative()` would match the rest of their sibling functions if they kept the `prefixParts` naming for their args.  I'd intentionally standardized the `prefix` and `prefixParts` names across these functions (and their callers) to make flow easier to follow.  They weren't before, and it made the code less clear.

This is a good point. I guess I was trying to abstract the function to be a bit more generic, but probably only because I found the  "prefix" and "prefixParts" a bit confusing when I first started looking at the code base.
 
 > * Stepping back, by the time I made it to the end of the branch, I wasn't sure why the functions above were introduced at all since no new callsites of them were introduced.
 
My initial approach for implementing the v2 API required more control and the desire to call the various functions I refactored / broke down from the larger `getSource()` function. However, I abandoned that approach in favor of what you currently see in the v2 introduction commit. I left the refactoring in there because, as I mentioned, I tend to prefer multiple, smaller functions that are named well and do less over a 20+ line long function that does a lot and is harder to read. But reading over your concerns, it's clear to me now that these refactoring contributions did not enhance the code base, so I will drop them from this PR.

jameshadfield:
Thanks @kairstenfay, i'm excited with the possibilities this API will open up! Following on from this week's meeting, we agreed that best place to start for the proposed API is with a spec proposal for internal discussion, so I won't comment on the commits just yet. This proposal may include some examples of smoke tests if you'd like, but they're by no means necessary. I know there's a couple of gists and some issues that discuss aspects of the API, but I'd like to see a detailed description before starting implementation work. None of this is to say the spec can't change during implementation -- I suspect it will! I've been writing some notes about the API but will wait to see the spec before commenting.

kairstenfay:
I dropped the refactor commits and rebased this branch onto `error-handling`.

src/getDataset.js
kairstenfay: I think this is a better question for #228 where this commit was introduced; the fact that it showed up in this PR is just an artifact of my not having rebased this branch onto `error-handling` after making changes over there. 

But to answer your question here -- no strong reason except that I like the forcefulness of "assert" in defensive programming, and I was trying to corral the sending of non-200 status codes into `catch` blocks as much as possible in the `getDataset` (etc) files. I'm happy to change it back to `if` if you'd prefer that.

PR #242, opened 2020-12-24
[build-map] cluster community builds on the map;

eharkins:
This uses a react-mapbox-gl component to cluster builds on the map. Only community builds are clustered in this way. Clicking a cluster will zoom to the bounds of the builds it contains. Cluster tooltips don't show the contents of the
cluster (this is possible, just adds complexity to implementation). Important to note that clusters will never separate if two markers (builds) have the same coordinates. Since programmatically adding small differences to identical coordinates is non-trivial, we do this manually in the yaml file for now. We should add a console warning so that this doesn't happen by mistake.

### Related issue(s)  

Fixes #  
Related to #  

### Testing

### DIFF
--- a/static-site/content/allSARS-CoV-2Builds.yaml
+++ b/static-site/content/allSARS-CoV-2Builds.yaml
@@ -205,7 +205,7 @@ builds:
     region: North America
     level: division
     coords:
-      - -119.64
+      - -118.0
       - 36.357039
     org:
       name: CZ BioHub
@@ -217,8 +217,8 @@ builds:
     region: North America
     level: division
     coords:
-      - -119.690182
-      - 36.357039
+      - -121.0
+      - 37.357039
     org:
       name: SEARCH, San Diego
       url: https://searchcovid.info/
@@ -229,8 +229,8 @@ builds:
     region: North America
     level: location
     coords:
-      - -117.143358
-      - 32.733024
+      - -118.143358
+      - 33.733024
     org:
       name: SEARCH, San Diego
       url: https://searchcovid.info/
@@ -301,8 +301,8 @@ builds:
     region: North America
     level: location
     coords:
-      - -71.0589
-      - 42.3601
+      - -71.18
+      - 42.45
     org:
       name: Broad Institute
       url: https://auspice.broadinstitute.org
@@ -313,8 +313,8 @@ builds:
     region: North America
     level: location
     coords:
-      - -71.3
-      - 42.3601
+      - -71
+      - 42.25
     org:
       name: Broad Institute
       url: https://auspice.broadinstitute.org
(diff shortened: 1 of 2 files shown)


emmahodcroft:
I really like this Eli! Good job!

Couple small points:
- Can we make the icons a little bigger so that the default zoom-in is a little less? I found it a bit weird how zoomed in I got when clicking, for example, on the 'Texas cluster' or the 'New England cluster' (dots). However, this isn't really a biggie. 
- I also am really struggling to tell the differences in sizes of dots - I think 4 blue sizes may be too much... Especially if they aren't next to each other on the map. I'd vote for maybe introducing more colours, or shades of colours?

I know Trevor didn't like the Nextstrain Icons for the Nextstrain builds, but again, I wonder if we might use a broader range of colours to help these stand out a bit from the local build. Right now it's a very subtle difference! 

Anyway, I think the mouse-over and the clustering is working super well, and this is looking really great!! Thanks for all the hard work on this!

trvrb:
I love this! Super clean. I don't have much to add at all, maybe just a couple very small aesthetic items:

1. The "local" build dot is pretty tiny. If you click on Boston you'll see what I mean. Maybe increase the radius like 1 pixel?
2. Might need more jitter when two builds have the same lat/long location. This can be seen in the 2 California builds (BioHub and Andersen lab) where clicking the "2" zooms into two dots like 4 blocks apart.

Awesome work.

trvrb:
Oh, and I think Emma's points here are good:

>I know Trevor didn't like the Nextstrain Icons for the Nextstrain builds, but again, I wonder if we might use a broader range of colours to help these stand out a bit from the local build. Right now it's a very subtle difference!

Maybe just use our brand colors for the 5? A more vibrant ramp might be:
- #4C90C0
- #75B681
- #B2BD4D
- #E1A03A
- #E04929

spectrum-1

A less vibrant ramp might be:
- #4C90C0
- #75B681
- #92BC63
- #B2BD4D
- #CEB541

spectrum-2

emmahodcroft:
I agree with Trevor's two comments above about the size & the two 'close' builds (it ends up taking me down to street level if I click which seems a bit 'intimate'!)

I like the more vibrant ramp myself, as I think it'll be really easy to see the difference!

emmahodcroft:
But yes, this is so cool Eli - Thanks for all your work on this!

eharkins:
Thanks @emmahodcroft and @trvrb for your feedback - I agreed with everything so tried to implement it all in the above commits. For details on "jitter" aka space between map markers with identical coords / solving the problem of identical coords in general see the notes of 6e83375. Let me know if you have other edits / suggestions!

trvrb:
Fantastic! Thank you Eli. This PR is great to merge from my perspective. (And I think that with these additions `build-map` can also be merged into `master`).

jameshadfield:
Thanks for chasing this up @eharkins! Merging now :)

PR #246, opened 2021-01-07
Adds WA build to map

cassiawag:
This PR adds the [Washington State SARS-Cov-2 build](https://nextstrain.org/groups/blab/ncov/wa/4m) to the global map.


trvrb:
@cassiawag: This failed because of improper indentation. I just fixed this in the above commit.

cassiawag:
Thanks @trvrb!

PR #247, opened 2021-01-07
Add BE build to map

barneypotter24:
This PR adds the Baele Lab's Belgium-focused build to the map.

In addition to adding the new build (centered Leuven), coordinates for the Neher Lab's currently existing build have been shifted slightly to Brussels center so that the two builds can be differentiated on the map.


trvrb:
Looks great! Thanks for the PR @barneypotter24.

PR #248, opened 2021-01-09
Sars-cov-2 builds yaml clean up

eharkins:
Removing some unnecessary `null` fields in this yaml listing of sars-cov-2 builds and adjusting the code to allow them to be undefined instead. Also moves all the hierarchy defining builds to the top for easier reading (I think, though can revert this if others disagree).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3659,6 +3659,16 @@
             "path-exists": "^4.0.0"
           }
         },
+        "js-yaml": {
+          "version": "3.14.1",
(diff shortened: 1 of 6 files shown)


eharkins:
Worth noting that we anticipate potentially more changes to the yaml schema as we extend the page to be able to fetch detailed information about each build like number of sequences and date updated.

corneliusroemer:
Shouldn't clash with #250 
#250 should probably merged first to allow fast forward.

jameshadfield:
Thanks @eharkins! I've rebased onto master and added
* brief comments in the YAML
* basic script to sanity check YAML

PR #257, opened 2021-01-13
Build updates

jameshadfield:
Updates package-lock files, node versions and switches to `npm ci` to remedy hard-to-debug dependency-related errors in the built auspice client.

It's somewhat unclear whether the [heroku-buildpack-nodejs](https://github.com/heroku/heroku-buildpack-nodejs) uses `npm ci` or `npm install`, so this PR adds an extra explicit `npm ci` step. At worst, this should simply unnecessarily increse the build time. This should be investigated, as the ideal situation is that heroku runs `npm ci`, thus ensuring the dependency tree matches the (committed) lock file. Looking at the buildpack itself, `npm ci` seems to have become default behavior in the past week or so, but what's happening for our builds is not clear from the logs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ language: node_js
 sudo: false
 dist: bionic
 node_js:
-  - "13.6"
+  - "14.15"
 install:
   - npm ci
 script:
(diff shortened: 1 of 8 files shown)


jameshadfield:
Closing this as it was a failed attempt to address a bug which turned out to be in auspice, however there are some commits here which we should cherry-pick into a new PR as they improve the build process more generally.

PR #258, opened 2021-01-13
Add public canada and ontario builds to canadian public nextstrain in…

fmaguire:
Adds public and actively updated augur builds for Canada and Ontario to the list of available builds.


jameshadfield:
Superseeded by #260

PR #260, opened 2021-01-14
Add public canada nCoV builds

jameshadfield:
This is PR #258 with slight change to the build names so that they are clearer when hovering on the map and the addition of ontario as a heirachy block in the YAML which is necessary for inclusion in the table.

Awesome to see another auspice server being run!


fmaguire:
Cheers @jameshadfield I really appreciate all the work you have done on both the code and the documentation side! Made my life very easy.

PR #261, opened 2021-01-18
Add the GCC build to the global map

alswajiab:
This PR adds the [GCC](https://nextstrain.org/community/alghoribi-lab/ncov/gcc) build to the global map.


eharkins:
Thanks @alswajiab! I've moved this to #264 and added the full name of the GCC and reversed the lat, longs, which originally placed the build in Europe (the way they are listed for builds in this file may be the reverse of convention, in which case we should change the order!)

PR #276, opened 2021-02-18
Hierarchical dataset select UI

eharkins:
starting to try to make @jameshadfield's sketch (https://github.com/nextstrain/nextstrain.org/issues/62#issuecomment-777191703) a reality


eharkins:
Will want to allow custom variable names if we go with this implementation and we want to display the key and value in a filtering component like in auspice:
Screen Shot 2021-02-18 at 4 42 35 PM

since having `variable1 -> h3n2` would be better as `pathogen -> h3n2` or something more descriptive.

PR #271, opened 2021-02-10
Influenza page

eharkins:
see commit messages. WIP

many commits here should go in separately as #266 but this is a separate concept so I used `master` as the base branch in anticipation of that getting merged first.

PR #278, opened 2021-02-27
Flat dataset listing interface with filters

eharkins:
Note: base branch means this would merge into #271, assuming we still want to stand that page up. There was some suggestion that a more flexible approach like this would mean pages like /influenza might not be necessary if their only purpose was to list datasets for that pathogen - that is as opposed to providing more bespoke resources beyond datasets like /sars-cov-2 does. Either way seems like we'll need to decide/implement how exactly we want this to fit into the site in general (on the splash page?) beyond /influenza - some of that discussion has already taken place in #62.

This is a flatter alternative to #276 as discussed in https://github.com/nextstrain/nextstrain.org/issues/62#issuecomment-781732697

The general approach was to take auspice components which do this same thing, and change them so they fit into the nextstrain.org context. One such change is using react component state to manage the filters rather than redux. I chose to do this since no redux state exists in this app and it seemed beyond the scope of this feature to set one up.

Filters are generated in the `FilterBuilds` component from the list of builds passed as the `builds` prop, using a set of fields and corresponding display names passed as the `filterPropertyMappings` prop:

```
<FilterBuilds builds={this.state.catalogueBuilds}
  filterPropertyMappings={[
    ["org.name", "Organization name"],
    ["grouping", "Flu type"],
    ["pathogen", "Pathogen name"],
    ["segment", "Segment"],
    ["duration", "Duration"]
  ]}
/>
```

The `filterPropertyMappings` requires these properties to exist in the builds as listed in the YAML they come from, but you could imagine those properties being generated in the script used to generate the YAML (latest version in #266: https://github.com/nextstrain/nextstrain.org/blob/b50b224e8d1aa36fc4fb2ae45bd25f4ab3b70daf/scripts/collect-pathogen-resources.js).


eharkins:
Filter component is still missing some styling which I should be able to get from auspice as well.

eharkins:
Added some styling improvements. Another small fix is to sort the filter options.

trvrb:
Thanks for starting in on this Eli. Having usable implementation is really helpful to start getting a feel for how the dataset UI would operate.

My main issue here is that your manually curated YAML data won't be generally available and I don't want to build an interface around manually curated data. For influenza you've manually curated a mapping that says that `6m`, `2y`, etc... are "Duration", `HA`, `NA`, etc... are "Segment", etc...  via your `all-influenza-builds.yaml`:

```
  - url: https://nextstrain.org/flu/seasonal/h3n2/ha/12y
    name: h3n2/ha/12y
    grouping: Seasonal
    pathogen: A/H3N2
    segment: HA
    duration: 12y
    org:
      name: Nextstrain Team
```

However, we don't have this curation for the vast majority of datasets and I don't want to sign ourselves up to have to invest in this curation in order to have datasets listed in the UI. I'd urge that the UI used to show datasets on `/influenza` would be the same UI that's used to show datasets on `/groups/neherlab`.

Thus, I'd recommend the "bag-of-words" approach that I detailed in the comment: https://github.com/nextstrain/nextstrain.org/issues/62#issuecomment-787195636, where you can filter to `h3n2` without knowing that this is categorized as "Pathogen name".

I want to be able to directly scrape an S3 bucket to gather a JSON (or YAML would be fine) with entries that look like:
```
name: "flu_seasonal_h3n2_ha_2y",
url: "https://nextstrain.org/flu/seasonal/h3n2/ha/2y",
title: "Real-time tracking of influenza A/H3N2 evolution using data from GISAID",
contributor: "Nextstrain",
last_updated: "2021-02-22",
maintained_by: "Jover Lee, Richard Neher and Trevor Bedford",
sequence_count: 1526
```

This would require no semantic decision making to assign "Duration", "Segment", etc... and could be used as a direct basis for the proposed flat dataset selection UI in https://github.com/nextstrain/nextstrain.org/issues/62#issuecomment-787195636.

trvrb:
I've attempted to address these issues in PR https://github.com/nextstrain/nextstrain.org/pull/279.

eharkins:
0153f96 removes the UI element originally added here in favor of the one added in #279 which is more flexible. I also remove the manually-curated flu YAML in this commit since post- #279 we depend on the output of `scripts/collect-datasets.js` being pushed to s3 in a github action. 

cdfe76c was sorting the filter options in the old implementation, I forgot I hadn't pushed this, I will implement the same thing in the new component.

PR #279, opened 2021-03-03
Select influenza datasets via keywords

trvrb:
@eharkins ---

This commit streamlines the dataset selection UI to use generic keywords from filenames. Principally, it:
1. Creates a new light weight `collect-datasets.js` script from `collect-pathogen-resources.js` that provisions `datasets_influenza.json`
2. Modifies `influenza.jsx` to use this dataset listing
3. Creates a new `DatasetSelect` component from `FilterBuilds` that filters by keyword and that implements some row styling

Some notes:

I chose a very simple light weight metadata format of the form:
```
 {
  "updated": "2021-01-26",
  "filename": "flu_avian_h5n1_mp.json",
  "url": "https://nextstrain.org/flu/avian/h5n1/mp",
  "contributor": "Nextstrain"
 },
```
(the full file is available at http://data.nextstrain.org/datasets_influenza.json). This format does not require actually fetching the JSON documents, only querying via `S3.listObjectsV2`. I figure that we'll want to be either producing this data structure on demand for calls to endpoints like https://nextstrain.org/groups/blab or reprovisioning a single `datasets.json` file for the across-Nextstrain dataset listing at https://nextstrain.org/datasets. For either dynamic call for a Groups page or compiling across many dataset JSONs we want something that can be really fast and so not pulling down a bunch of JSON data is important.

I didn't make many changes to the functioning of the dataset filtering you set up. I really liked the basic approach. I just moved from a manually curated list of `grouping`, `pathogen`, `segment`, `duration` to a generic "keyword" approach where keywords are generated from splitting filenames on underscores. Filtering intersects on keywords so that filtering by `h3n2` and `ha` gives:

filter

There are some styling fixes that I didn't get around to here:
- The 2nd column is not properly left-aligned (need to a larger fix to the use of flexbox for this)
- This should have column names ("Dataset", "Contributor", "Last Updated")

My idea here would be to get this dataset select UI working to our liking for the `/influenza` page and then merge this work into `master` branch and from there repurpose this UI for `/pathogens` (with all the core datasets), `/groups` (with datasets across Groups) and `/datasets` (with both together), etc... I suspect we'll want a chron job to regenerate `datasets.json` every 5 minutes and separately to generate the same data structure on the fly when retrieving individual Groups pages, but this can be tackled later on.


trvrb:
@eharkins --- Okay. I'm done working on this PR now. There's currently no bells and whistles like sortable columns, but I think the basic interface is there. It works on tablet / laptop to display dataset name, contributor (logo + test) and date updated:

desktop

And it works on mobile to display dataset name and contributor logo (I wanted to keep it to a single line):

mobile

I'm really liking working with https://github.com/LoicMahieu/react-styled-flexboxgrid for layout.

static-site/src/pages/influenza.jsx
```diff
@@ -1,6 +1,5 @@
 import React from "react";
 import Helmet from "react-helmet";
-import yaml from "js-yaml";
 import ScrollableAnchor, { configureAnchors } from "react-scrollable-anchor";
```
jameshadfield, line 3: 🐛 there's something going on with the setup of scroll anchors at the moment. In dev mode this crashes the page:
```
Uncaught TypeError: element is null
    getElementOffset scroll.js:14
    doesElementContainScrollTop scroll.js:29
    getBestAnchorGivenScrollLocation scroll.js:88
    getBestAnchorGivenScrollLocation scroll.js:86
...
```

In production mode this is limited to a console error & the page continues to work.

eharkins: I get this too but doesn't crash the page for me. It looks familiar so I it could be that it wasn't introduced here and I managed to ignore it before in the build-select branch if it wasn't causing crashes but just warnings in the console. Will see what's causing it / where it is introduced.

eharkins: Introduced in https://github.com/nextstrain/nextstrain.org/pull/278/commits/bafb08b0d68ea30d10fca9d49fb0d711cd5444c2
Fix is https://github.com/gabergg/react-scrollable-anchor/issues/43#issuecomment-538661917

eharkins: Will push the fix soon

eharkins: resolved in https://github.com/nextstrain/nextstrain.org/pull/279/commits/3fcb3d6b06290ee9a65d6554cd083674e3b95892

eharkins:
> I'm really liking working with https://github.com/LoicMahieu/react-styled-flexboxgrid for layout.

Just noting that https://react-table.tanstack.com/ looks promising if we want to refactor this component in a subsequent PR with extra bells and whistles such as sortable columns. Would replace https://github.com/LoicMahieu/react-styled-flexboxgrid.

eharkins (approved):
I added a cron job to update the JSON every 5 minutes and changed the field `updated` to `date_uploaded` (to avoid confusion between date auspice was run and date uploaded to s3 - we use the latter here) as discussed. Will merge soon.

PR #286, opened 2021-03-18
filterable dataset list on /sars-cov-2

eharkins:
Instead of dropdowns, use the filterable dataset-select component.  Details in commit message.

### DIFF
--- a/static-site/src/components/build-pages/dataset-select.jsx
+++ b/static-site/src/components/build-pages/dataset-select.jsx
@@ -60,7 +60,7 @@ const LogoContainer = styled.a`
   cursor: pointer;
 `;
 
-const renderDatasets = (datasets) => {
+const renderDatasets = (datasets, showDates) => {
   return (
     <>
       <Grid fluid>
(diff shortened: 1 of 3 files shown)


eharkins:
I find the difference between the black-colored links of the datasets themselves contrasting with the blue of the contributor links to be a bit odd. I can change the contributor links to match the black, or I could remove the inline style that makes the dataset links black, so that they're both blue. I prefer blue to convey a link, but Trevor made them black explicitly so maybe feels differently.

trvrb:
Awesome! Just a couple small comments:

1. I had played with link colors when doing design work for `/influenza` and I believe that the black text looks signficantly cleaner (I like calling out links when embedded in a paragraph, but if everything is a link I don't see it as helpful to highlight). Take a look at GitHub interface for example. Files [here](https://github.com/nextstrain/auspice) are still black text but highlight to blue underline. Here I'd suggest to behave similarly. Keep dataset name as bold, black text, move contributor name to non-bold black text. Make both hover into blue underlined text.
2. I'd sort dataset list to be alphabetized.

trvrb:
This looks great Eli! Thanks for making these styling changes. This can be merged from my perspective.

trvrb (approved):

eharkins:
Sorry for the last minute changes after review but just noticed a few mistakes.

PR #287, opened 2021-03-24
dataset-select filter by contributor

eharkins:
This adds the dataset contributor as a keyword for the filtering of datasets in the dataset-select component. In some cases this results in duplicate or near-duplicate keywords if the contributor name is also in the dataset name.

It doesn't add contributor keywords if the datasets are all from the same contributor.

### DIFF
--- a/static-site/src/components/Datasets/dataset-select.jsx
+++ b/static-site/src/components/Datasets/dataset-select.jsx
@@ -2,11 +2,10 @@ import React from "react";
 import PropTypes from 'prop-types';
 import "react-select/dist/react-select.css";
 import "react-virtualized-select/styles.css";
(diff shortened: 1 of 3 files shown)


eharkins:
Second commit makes contributor filters distinct from keyword ones. They show up like `contributor → Nextstrain Team` in the dropdown and use the intersection with any keyword filters when applied.

jameshadfield:
@eharkins I rebased this onto #302, and extended the matching logic to require datasets to
match the union of filterValues for each filterType specified. This means that we can filter to (e.g.) `3y, 2y` and get all `3y` and all `2y` datasets (previously we got none).

jameshadfield:

PR #288, opened 2021-03-26
dataset-select mobile view contributor links;

eharkins:
this was previously hardcoded to be a
nextstrain logo. I replace it here with a
person icon for non-Nextstrain contributors

will merge right away since its a 1 commit quick fix to #286 and people might be mad if they view /sars-cov-2 on mobile and it looks like we are crediting ourselves for their datasets.


eharkins:
For whatever reason when I tried to apply a css selector to multiple element types it didn't work so c026823 fixes that by separating the two

PR #289, opened 2021-03-26
flexible column prop for dataset-select listing

eharkins:
This doesn't change how things are rendering right now but creates a more flexible react pattern that will allow us to change how things render more easily in the future if we want. 

This allows you to specify the columns you would like to display when rendering datasets below the filter bar in the scrollable table. They render in the order you pass them. You must choose from the options in the object: "datasetListingColumn" in the dataset-select component. There's no error handling for if you choose one that isn't there since the options you choose are static / baked into the code and not determined dynamically.


jameshadfield:
Superseded by #289, which takes this commit and applies it (with changes) onto another PR.

PR #291, opened 2021-03-31
Filter dataset viz

eharkins:
On top of #289. As discussed with @jameshadfield, this allows the filtering to act on the map or on a different component passed to the dataset-select component as the "viz" prop.

Need to make some styling fixes to allow the width of the viz to not be pinched in by the styles meant for the filter component.


jameshadfield:
These commits have been applied onto PR #299 (which refactored things quite a bit) and are now represented by #301 and #302. The design choices of @eharkins have been largely maintained, although there have been some changes.

PR #295, opened 2021-04-02
Cookie security

tsibley:


jameshadfield (approved):
Tested on localhost & can confirm that existing cookies remain valid, and the login/logout flow still works as expected.

tsibley:
Deployed. Thanks for the review, @jameshadfield!

Issue #218, opened 2020-09-18
The nextstrain.org server no longer needs to identify dataset requests for multiple trees


jameshadfield:
The nextstrain.org server parses the dataset requests originating from the Auspice client largely within the `parsePrefix` function. This function contains logic for identifying single requests for multiple datasets (i.e. tangle-trees), which is an artifact from a time where the server would fetch all relevant JSONs and combine them server-side.

Auspice now always makes separate requests for each dataset. E.g. http://localhost:5000/ncov/global:ncov/oceania results in two GET requests, http://localhost:5000/charon/getDataset?prefix=ncov/oceania and http://localhost:5000/charon/getDataset?prefix=/ncov/oceania. 

As we move to simplify the `parsePrefix` function and delegate more responsibility to the `Source` class / subclasses, this should be removed.

(There are lots of little bits which are no longer accurate, such as `auspiceDisplayUrl` now being used as a redirect rather than handed to Auspice. I'm unsure whether to make individual issues for all of these.)

tsibley:
Closed by #296.

PR #296, opened 2021-04-02
Grab bag of cleanups to /charon/… handlers

tsibley:

These are all things I started tidying up while in the code for other purposes.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -14,7 +14,7 @@ Each handler is defined in an file of the same name within `src`.
 
 ### Autorization
 
-Each handler is responsable for checking authorization by calling a `Source` class method like so:
+Each handler is responsible for checking authorization by calling a `Source` class method like so:
 ```js
 if (!source.visibleToUser(req.user)) {
   return helpers.unauthorized(req, res);
--- a/src/getDataset.js
+++ b/src/getDataset.js
@@ -144,15 +144,18 @@ const getDataset = async (req, res) => {
     return helpers.unauthorized(req, res);
   }
 
-  const baseUrl = req.url.split(query.prefix)[0];
-  let redirectUrl = baseUrl + '/' + resolvedPrefix;
-  if (query.type) {
-    redirectUrl += `&type=${query.type}`;
-  }
+  /* If we got a partial prefix and resolved it into a full one, redirect to
+   * that.  Auspice will notice and update its displayed URL appropriately.
+   */
+  if (resolvedPrefix !== helpers.canonicalizePrefix(query.prefix)) {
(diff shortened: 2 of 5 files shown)


jameshadfield:
5a9199c will close #218

jameshadfield:
These changes all seem sensible to me - the code is now more consistent across different sources and removal of unused code is always nice. I couldn't detect any regressions during manual testing (mainly of `UrlDefinedSource`). 

Unfortunately smoke-testing is currently failing on master (it should really be part of CI, but that's another issue), which makes it harder to check any regressions from this PR. Comparing against current master, I believe this PR does break [this test](https://github.com/nextstrain/nextstrain.org/blob/004667ed0ad811e4ba55c3c5a51765bab579d12d/test/smoke-test/auspice_client_requests.json#L9-L15):

```
    Expected: "http://localhost:5000/charon/getDataset?prefix=/flu/seasonal/h3n2/ha/2y"
    Received: "http://localhost:5000/charon/getDataset?prefix=flu%2Fseasonal%2Fh3n2%2Fha%2F2y"
```

The redirect still works as expected, so this may be simply a matter of updating the test. I haven't read enough technical docs to have an opinion of `/` vs `%2F`. 

On the topic of testing, it would be great for this repo to include unit tests, which not only identify bugs but also help conceptual understanding of the code. Functions such as `parsePrefix`, `splitPrefixIntoParts`, `joinPartsIntoPrefix` would be great places to start.

tsibley:
Ah, thanks for flagging that failing test! I've repushed to update the test in the commit responsible. Both `/` and `%2F` are acceptable in the query part of a URL; `%2F` is the more conservative choice made by the standard `URL` APIs here. In human-facing queries, I'd support using `/` over `%2F` for better readability, but as this redirect is entirely machine-facing, I'm not concerned by it.

> it should really be part of CI, but that's another issue

Yeah, I would have caught these myself if these tests were part of CI. What's the blocker to doing so?

> On the topic of testing, it would be great for this repo to include unit tests, which not only identify bugs but also help conceptual understanding of the code. Functions such as `parsePrefix`, `splitPrefixIntoParts`, `joinPartsIntoPrefix` would be great places to start.

Agreed! As we don't yet have any framework for such tests in this repo, I'm hesitant to bundle addition of one into this PR. Small, pureish functions like these would be great starting points for addition of a unit testing framework and integration with CI in another PR, though.

jameshadfield:
> What's the blocker to doing so?

Someone just needs to set it up 😉 

> I'm hesitant to bundle addition of one into this PR. 

Yup, no worries with this being separate. 

LGTM!

PR #297, opened 2021-04-02
authn: Use claims from Cognito tokens instead of making additional requests

### LINKED ISSUE #277: Use claims from Cognito access token to obtain group membership list
The current authentication code makes a second request to the Cognito admin API to list the group memberships of the just-authenticated user:

https://github.com/nextstrain/nextstrain.org/blob/e9497235386cc96e5fbca8090416c7202fbdee7e/authn.js#L63-L81

Instead of making this call (which adds overhead like extra permissions, latency, API quota usage), we could instead verify the `accessToken` and use the `cognito:groups` claim within the token payload.

This would likely mean using jose's [`jwtVerify()`](https://github.com/panva/jose/blob/HEAD/docs/functions/_jwt_verify_.jwtverify.md#readme). JWTs are notoriously easy to shoot yourself in the foot with, so extra care must be taken and attention to detail paid to ensure the information is appropriately verified before using it, [per guidance from AWS](https://docs.aws.amazon.com/cognito/latest/developerguide/amazon-cognito-user-pools-using-tokens-verifying-a-jwt.html).

tsibley:
Cuts out two requests on every login: one to the user info endpoint and
another to the Cognito admin list user groups endpoint.  This also
avoids the server needing AWS credentials with admin permissions in
Cognito.

The verifyToken() function will be useful when we start accepting tokens
directly for authorizing API requests (e.g. from the CLI).

Resolves #277.

### DIFF
--- a/authn.js
+++ b/authn.js
@@ -7,9 +7,10 @@ const Redis = require("ioredis");
 const RedisStore = require("connect-redis")(session);
 const FileStore = require("session-file-store")(session);
 const passport = require("passport");
-const fetch = require("node-fetch");
-const AWS = require("aws-sdk");
 const OAuth2Strategy = require("passport-oauth2").Strategy;
+const {jwtVerify} = require('jose/jwt/verify');                   // eslint-disable-line import/no-unresolved
+const {createRemoteJWKSet} = require('jose/jwks/remote');         // eslint-disable-line import/no-unresolved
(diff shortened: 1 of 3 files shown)


jameshadfield (approved):
Don't know enough about cognito to review the specifics, but I've tested on localhost & can confirm that existing cookies remain valid, and the login/logout flow still works as expected.

tsibley:
Deployed. Thanks for the review, @jameshadfield!

PR #299, opened 2021-04-09
Updates to dataset display

jameshadfield:
_UPDATED 2021-04-14_

This PR is a mixture of commits by @eharkins and myself from different PRs. Here is a summary of the changes:

**refactoring of `<DatasetSelect>`**

Refactored to isolate its functionality into three children components with separate concerns: `<DatasetFilteringSelection>`, `<FilterDisplay>` and `<ListDatasets>`.

**dynamic routing in gatsby**

Allows pages such as `/influenza/a/b/c` to be rendered by the "parent" page (in this case `/influenza`), with their URLs updating to reflect this. If `a`, `b` or `c` (from the example) are valid keyword filters then they are applied to the dataset listing. We may end up handling this by a custom router higher up, and if so these changes are minor and easily reverted / removed.

**Visualisations now update as filters are applied**

The interface prop passed to <DatasetSelect> (originally implemented in 9733985) now take any number of custom react components as props. Currently this is used by the geographic map in sars-cov-2, which updates as filters are applied.

**Pages define which columns to show in the table**

Adds a more flexible react pattern that allows pages
(e.g. /influenza, /sars-cov-2) to decide which columns are rendered,
as well as how they look (e.g. logo, mobile text, links etc).
Currently the number of columns must be 2 or 3, and the 3rd will
not be rendered for small screen sizes. This could be improved in
future as needs arise.

**Filtering selection improved**

As well as "keywords" (`/` separated dataset names), we now add all the information which is rendered in the tables.


jameshadfield:
I wasn't are that due to the way Gatsby [automagically creates static assets for files in `./src/pages`](https://www.gatsbyjs.com/docs/creating-and-modifying-pages/), and the way we [serve these static assets](https://nextstrain.org/sequence-search-list/), we have unintentially added URLs which we don't want to.

For instance, this PR adds `pages/influenza-page.jsx` which will be visible at nextstrain.org/infleunza-page. An example which currently exists in master is https://nextstrain.org/sequence-search-list/.

eharkins:
@jameshadfield I tried both using named exports and putting files in a separate directory next to or within `static-site/src/pages` to solve the issue you point out above, and only the latter worked. Here it is: 
https://github.com/nextstrain/nextstrain.org/compare/url-filter-params...gatsby-meta-pages

jameshadfield:
Thanks @eharkins! Merged into master locally & re-tested before pushing. This sets the scene nicely for lots of future pages 👍

PR #300, opened 2021-04-09
authn: Accept Bearer tokens in Authorization headers

tsibley:
Series of related commits leading up to acceptance of `Authorization: Bearer <token>` headers for authentication. See commit messages for full details.

### DIFF
--- a/authn.js
+++ b/authn.js
@@ -10,7 +10,7 @@ const passport = require("passport");
 const OAuth2Strategy = require("passport-oauth2").Strategy;
 const {jwtVerify} = require('jose/jwt/verify');                   // eslint-disable-line import/no-unresolved
 const {createRemoteJWKSet} = require('jose/jwks/remote');         // eslint-disable-line import/no-unresolved
-const {JWTClaimValidationFailed} = require('jose/util/errors');   // eslint-disable-line import/no-unresolved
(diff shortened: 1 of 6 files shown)


authn.js
```diff
@@ -257,6 +311,24 @@ function setup(app) {
   });
 }
 
+
+/**
+ * Creates a user record from the given `idToken` after verifying it.
+ *
+ * @param {String} idToken
+ * @param {String} client. Optional. Passed to `verifyToken()`.
```
joverlee521, line 319: I was initially confused by the type hinting here since the array `BEARER_COGNITO_CLIENT_IDS` is being passed in as `client`. I think it would be clearer for the type hint to be `{String|String[]}`.

joverlee521 (approved):
This looks reasonable to me. The custom `passIfMissing` option is very handy ⭐ 
Glad I looked through this since it prompted me to look through the old login PR for the cli and connect the dots.

authn.js
tsibley: Good catch! Updated.

tsibley:
Had a good walkthrough of this with @joverlee521, @huddlej, and @jameshadfield this afternoon to build some more shared understanding.

tsibley:
Deployed.

PR #301, opened 2021-04-10
Allow pages to define and inject dataset-related components

jameshadfield:
This PR is based off commits 9733985, a8db4cb and 491682e (all by @eharkins, see PR #291). The main changes are due to the refactoring previously done in #299.

The `interface` prop passed to `<DatasetSelect>` (originally implemented in 9733985) has been generalised to take any number of (react) components as props, although this is only currently used by the geographic map in sars-cov-2.


jameshadfield:

PR #302, opened 2021-04-10
Pages can define which columns are rendered

jameshadfield:
This commit creates a more flexible react pattern that allows pages
(e.g. /influenza, /sars-cov-2) to decide which columns are rendered,
as well as how they look (e.g. logo, mobile text, links etc).

In turn, this will give us the ability for different pages to use
different dataset structures and display bespoke information in the
table.

Currently the number of columns must be 2 or 3, and the 3rd will
not be rendered for small screen sizes. This could be improved in
future as needs arise.

Based on work by @eharkins in PR #289. @eharkins has already reviewed a previous iteration of this PR, and this has been updated accordingly.


jameshadfield:

PR #303, opened 2021-04-10
WIP: list all datasets at path "/all"

jameshadfield:
Proof of principle page which lists all (known) datasets at the URL path "/all". This provides the third page listing datasets (after /influenza and /sars-cov-2).

* Nextstrain core + staging via listing of the S3 buckets
* Nextstrain public groups via the same technique
* Community via [@tsibley's community-search](https://github.com/nextstrain/community-search/)

Currently there is no automation of dataset collection, it is run manually and uploaded to the staging bucket.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14647,6 +14647,11 @@
       "integrity": "sha1-Sr6/7tdUHywnrPspvbvRXI1bpPc=",
       "dev": true
     },
+    "ndjson-parse": {
+      "version": "1.0.4",
+      "resolved": "https://registry.npmjs.org/ndjson-parse/-/ndjson-parse-1.0.4.tgz",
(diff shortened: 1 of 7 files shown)


jameshadfield:
I've created #305 and #306 to track implementation of the frontend and backend to get to a production-ready version of this functionality.

eharkins:
Nice! I'm not sure about this but I would think we wouldn't want to list staging datasets on a public interface; I think about that bucket as being private and data in it as being private until it's pushed live.

jameshadfield:
> I'm not sure about this but I would think we wouldn't want to list staging datasets on a public interface; I think about that bucket as being private and data in it as being private until it's pushed live.

Yes, this is probably sensible!

I don't expect this PR to ever be merged, it was a quick way to confirm to myself that such a listing is simple to create and helped me realize where the complexity lies. I've written about this in #306 so let's continue the discussion there and I hope that some of the code here may help with the implementation we end up with!

jameshadfield:
We're still interested in pursuing this direction, but I don't think this direction is quite right.

PR #307, opened 2021-04-12
Serve community builds from default GitHub branch

jameshadfield:
When Nextstrain community builds were implemented, the default GitHub branch was "master", and this assumption was build into our server. The ability to explicitly define which GitHub branch should be used was possible (by specifying it in the URL via "@branch`), but in the absence of this "master" was always used.

It is now common for the default / only branch to be "main", and it is surprising for users when Nextstrain community builds result in an error in these cases.

Here we query GitHub for the default branch and use that. If the branch is made explicit via the use of `@branch` in the URLs, then that branch is used (unchanged behaviour). Such a query necessitates that rather a lot of methods & functions become async, and some getters have become (normal) functions.

@tsibley could you review this? It necessitated a cascade of synchronous functions / getters becoming async functions, but in most cases these functions are returning (promise-fulfilled) values. There are other ways of achieving this, but I didn't want to change the data flow of the app too drastically. 

URLs for testing: 
* /community/trvrb/nextstrain-community-build/20210401 (master branch)
* /community/jameshadfield/nextstrain-community-build/20210401 (main branch)

[This docs page](https://github.com/nextstrain/docs.nextstrain.org/blob/master/src/guides/share/community-builds.md) should be updated after merge.

### DIFF
--- a/src/getAvailable.js
+++ b/src/getAvailable.js
@@ -15,37 +15,38 @@ const getAvailable = async (req, res) => {
     return unauthorized(req, res);
   }
 
-  let datasets;
-  let narratives;
+  let datasetPaths;
+  let narrativePaths;
 
   try {
-    datasets = await source.availableDatasets() || [];
-    narratives = await source.availableNarratives() || [];
+    datasetPaths = await source.availableDatasets() || [];
+    narrativePaths = await source.availableNarratives() || [];
   } catch (err) {
     if (err instanceof ResourceNotFoundError) {
       return res.status(404).send("The requested URL does not exist.");
     }
   }
 
-  if (!datasets || !datasets.length) {
+  if (!datasetPaths || !datasetPaths.length) {
     utils.verbose(`No datasets available for ${source.name}`);
   }
-  if (!narratives || !narratives.length) {
+  if (!narrativePaths || !narrativePaths.length) {
     utils.verbose(`No narratives available for ${source.name}`);
   }
 
-  return res.json({
-    datasets: datasets.map((path) => ({
-      request: joinPartsIntoPrefix({source, prefixParts: [path]}),
-      secondTreeOptions: source.secondTreeOptions(path),
-      buildUrl: source.name === "community"
-        ? `https://github.com/${source.repo}`
-        : null
-    })),
-    narratives: narratives.map((path) => ({
-      request: joinPartsIntoPrefix({source, prefixParts: [path], isNarrative: true})
-    }))
-  });
+  const datasets = [];
+  const narratives = [];
+  for (let i=0; i<datasetPaths.length; i++) {
+    const request = await joinPartsIntoPrefix({source, prefixParts: [datasetPaths[i]]}); // eslint-disable-line no-await-in-loop
+    const secondTreeOptions = source.secondTreeOptions(datasetPaths[i]);
+    const buildUrl = source.name === "community" ? `https://github.com/${source.repo}` : null;
+    datasets.push({request, secondTreeOptions, buildUrl});
+  }
+  for (let i=0; i<narrativePaths.length; i++) {
+    const request = await joinPartsIntoPrefix({source, prefixParts: [narrativePaths[i]], isNarrative: true}); // eslint-disable-line no-await-in-loop
+    narratives.push({request});
+  }
+  return res.json({datasets, narratives});
 };
 
 module.exports = {
--- a/src/sources.js
+++ b/src/sources.js
@@ -25,8 +25,8 @@ class Source {
   get name() {
     return this.constructor._name;
   }
-  get baseUrl() {
-    throw new InvalidSourceImplementation("baseUrl() must be implemented by subclasses");
+  async baseUrl() {
+    throw new InvalidSourceImplementation("async baseUrl() must be implemented by subclasses");
   }
   static isGroup() { /* is the source a "nextstrain group"? */
     return false;
@@ -90,8 +90,8 @@ class Dataset {
       ? `${baseName}.json`
       : `${baseName}_${type}.json`;
   }
-  urlFor(type) {
-    const url = new URL(this.baseNameFor(type), this.source.baseUrl);
+  async urlFor(type) {
+    const url = new URL(this.baseNameFor(type), await this.source.baseUrl());
     return url.toString();
   }
   get isRequestValidWithoutDataset() {
@@ -111,15 +111,15 @@ class Narrative {
     const baseName = this.baseParts.join("_");
     return `${baseName}.md`;
   }
-  url() {
-    const url = new URL(this.baseName, this.source.baseUrl);
+  async url() {
+    const url = new URL(this.baseName, await this.source.baseUrl());
     return url.toString();
   }
 }
 
 class CoreSource extends Source {
   static get _name() { return "core"; }
-  get baseUrl() { return "http://data.nextstrain.org/"; }
+  async baseUrl() { return "http://data.nextstrain.org/"; }
   get repo() { return "nextstrain/narratives"; }
   get branch() { return "master"; }
 
@@ -160,14 +160,15 @@ class CoreSource extends Source {
 
 class CoreStagingSource extends CoreSource {
   static get _name() { return "staging"; }
-  get baseUrl() { return "http://staging.nextstrain.org/"; }
+  async baseUrl() { return "http://staging.nextstrain.org/"; }
   get repo() { return "nextstrain/narratives"; }
   get branch() { return "staging"; }
 }
 
 class CoreNarrative extends Narrative {
-  url() {
-    const repoBaseUrl = `https://raw.githubusercontent.com/${this.source.repo}/${this.source.branch}/`;
+  async url() {
+    const branch = await this.source.branch;
+    const repoBaseUrl = `https://raw.githubusercontent.com/${this.source.repo}/${branch}/`;
     const url = new URL(this.baseName, repoBaseUrl);
     return url.toString();
   }
@@ -186,18 +187,28 @@ class CommunitySource extends Source {
 
     if (!this.repoName) throw new Error(`Cannot construct a ${this.constructor.name} without a repoName after splitting on /@/`);
     if (!this.branch) {
-      this.branch = "master";
+      this.branch = fetch(`https://api.github.com/repos/${this.owner}/${this.repoName}`)
+        .then((res) => res.json())
+        .then((data) => data.default_branch || "master")
+        .catch(() => {
+          console.log(`Error interpreting the default branch of ${this.constructor.name}`);
+          return "master";
+        });
     }
   }
 
   static get _name() { return "community"; }
   get repo() { return `${this.owner}/${this.repoName}`; }
-  get baseUrl() { return `https://github.com/${this.repo}/raw/${this.branch}/`; }
+  async baseUrl() {
+    const branch = await this.branch;
+    return `https://github.com/${this.repo}/raw/${branch}/`;
+  }
 
-  get repoNameWithBranch() {
+  async getRepoNameWithBranch() {
+    const branch = await this.branch;
     return this.branch === "master"
       ? this.repoName
-      : `${this.repoName}@${this.branch}`;
+      : `${this.repoName}@${branch}`;
   }
 
   dataset(pathParts) {
@@ -208,7 +219,8 @@ class CommunitySource extends Source {
   }
 
   async availableDatasets() {
-    const qs = queryString.stringify({ref: this.branch});
+    const branch = await this.branch;
+    const qs = queryString.stringify({ref: branch});
     const response = await fetch(`https://api.github.com/repos/${this.repo}/contents/auspice?${qs}`);
 
     if (response.status === 404) throw new ResourceNotFoundError();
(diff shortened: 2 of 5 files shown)


jameshadfield:
This work makes me realise that it's also possible to relax our naming structure for community builds by quering the GitHub API in a similar fashion. For instance, `/community/O/R/D` currently requires a dataset named `auspice/R_D.json` but we could allow this to be named `auspice/D.json` which is probably more intuitive for users (whilst still maintaining backwards compatibility).

emmahodcroft:
I'd be in favour of considering allowing datasets _not_ to share the name of the repository at the beginning, if you think that's possible. That's probably the most-often-asked question/problem I get about community builds, so I do think this might remove a low but common stumbling-barrier, if doing so would work & maintain backwards compatibility!

tsibley:
I agree it'd be nice to shed the requirement to repeat the repo name in the filename. That's always seemed unintuitive to me. However, it will raise some questions we will need to answer:

1. If both filenames are present, which is used? This could be quite confusing to users.

2. Do we document both as equal alternatives? Or preferentially document the new way and play down the old way (but still mention it for posterity/existing users)?

It's also worth noting that the more querying we have to do to GitHub, the slower each community dataset load becomes.

tsibley (changes requested):
Overall this looks reasonable to me! I left a few specific comments/suggestions below for consideration. It'll be good to support arbitrary default branches instead of hardcoding `master`. :pray:

src/getAvailable.js
```diff
@@ -15,37 +15,38 @@ const getAvailable = async (req, res) => {
...
+    const request = await joinPartsIntoPrefix({source, prefixParts: [datasetPaths[i]]}); // eslint-disable-line no-await-in-loop
+    const secondTreeOptions = source.secondTreeOptions(datasetPaths[i]);
+    const buildUrl = source.name === "community" ? `https://github.com/${source.repo}` : null;
+    datasets.push({request, secondTreeOptions, buildUrl});
+  }
+  for (let i=0; i<narrativePaths.length; i++) {
+    const request = await joinPartsIntoPrefix({source, prefixParts: [narrativePaths[i]], isNarrative: true}); // eslint-disable-line no-await-in-loop
+    narratives.push({request});
+  }
+  return res.json({datasets, narratives});
```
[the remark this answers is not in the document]
jameshadfield: Thanks for the review! I'll take a look at implementing this, however my understanding was that something like:
```js
const data = arr.map((item) => ({
   x: (await someAsyncFunction())
})
```
was problematic because `Array.map()` is a synchronous function which results in the `data` array being created with entries containing (keys with values which are) potentially/probably/definately unfulfilled promises. I know one can get around this via
```js
const data = await Promise.all(arr.map(async (item) => ({
   x: (await someAsyncFunction())
})
```
but I preferred loops over this form.

P.S. Why are three clause loops an "anti-pattern"? Can completely understand that people might not like the style, but I've used them a lot (in various languages) and would be interested in learning why they are discouraged.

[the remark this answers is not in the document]
tsibley: Ah, right. Indeed, the async declaration and `Promise.all` are necessary. I prefer that functional approach to an imperative loop that builds up an array because the structure of the code ends up matching the data structure produced.

Either way, I'm realizing just now that we should think about parallelization of requests here, since that's what's being repeatedly awaited by this code. (More on that in another comment.)

Re: C-style for loops: It's really easy to get the start/stop condition subtly wrong in a C-style for loop, resulting in off-by-one errors (for example). JavaScript, like Python and many other langs, have explicit syntax for iterating over arrays (and many other iterables, like generators) precisely so the intent of the code can be expressed more directly and correctly than index calculations and bounds testing. It's syntax that aims to design away a whole class of bugs and avoid a small form of _accidental complexity_. Like any pattern/anti-pattern, any one instance of it is not a big deal in isolation; it's in aggregate that the gains are made over time. (The same thing that's true of linting, for example.)

tsibley:
Some additional notes after I tested behaviour:

1. The additional request to the GitHub API to get the branch name reliably takes half a second (sometimes longer). This blocks any further processing until a response is received.

2. Multiple independent requests are made for the branch name, as different bits of code don't use the same `CommunitySource` object instance (such as across requests to Charon).

3. Successful responses are cached, however, at the server process-level, which makes additional requests (from the same server process) much faster after the first load for a given repo.

4. Failed requests (such as 404s for non-existent repos) _aren't_ cached and as such incur a bit of a penalty that noticeably slows down the error page. We may not consider this an issue.

src/getAvailable.js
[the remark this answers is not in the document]
jameshadfield: >  prefer that functional approach to an imperative loop

Changed to the functional approach

src/sources.js
```diff
@@ -160,14 +160,15 @@ class CoreSource extends Source {
...
   get repo() { return "nextstrain/narratives"; }
   get branch() { return "staging"; }
 }
 
 class CoreNarrative extends Narrative {
-  url() {
-    const repoBaseUrl = `https://raw.githubusercontent.com/${this.source.repo}/${this.source.branch}/`;
+  async url() {
+    const branch = await this.source.branch;
+    const repoBaseUrl = `https://raw.githubusercontent.com/${this.source.repo}/${branch}/`;
```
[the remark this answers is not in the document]
jameshadfield: Done

src/sources.js
```diff
@@ -186,18 +187,28 @@ class CommunitySource extends Source {
...
   static get _name() { return "community"; }
   get repo() { return `${this.owner}/${this.repoName}`; }
-  get baseUrl() { return `https://github.com/${this.repo}/raw/${this.branch}/`; }
+  async baseUrl() {
+    const branch = await this.branch;
+    return `https://github.com/${this.repo}/raw/${branch}/`;
+  }
 
-  get repoNameWithBranch() {
+  async getRepoNameWithBranch() {
```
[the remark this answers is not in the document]
jameshadfield: Done

src/getAvailable.js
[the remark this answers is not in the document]
jameshadfield: > Regardless of approach here, I'm realizing just now that we should think about parallelization of requests here, since that's what's being repeatedly awaited by this code. (More on that in another comment.)

My understanding is that each callback here is async because of `joinPartsIntoPrefix` → `[source.]repoNameWithBranch` → `[source.]branch`, the last of which is a single promise set up in the source's constructor. This promise will only run once per-source-instance, no matter how many methods on that source (e.g. `source.repoNameWithBranch`) access it. 

Related to this, is that each different API handler (getAvailable, getDataset etc) instantiates its own source object, which means that the `[source.]branch` promise (fetch) will run multiple times. A similar situation applies to non-community sources.

jameshadfield:
Update 2021-05-17: This PR has been rebased onto #327.

jameshadfield (commented):
Update after multiple rebases and some more testing. The functionality works well, however there is a small bug: 

The `getAvailable` API response includes the `@branch` syntax, even for the default branch. For instance, 

* `/charon/getAvailable?prefix=/community/jameshadfield/nextstrain-community-build` returns results such as `community/jameshadfield/nextstrain-community-build@main/20210401`, where the `@main` shouldn't be specified (it's the default). 
* `/charon/getAvailable?prefix=/community/jameshadfield/scratch` includes results such as `community/jameshadfield/scratch@master/ncov-demo`

These are reflected in the splash pages such as /community/jameshadfield/nextstrain-community-build and /community/jameshadfield/scratch

jameshadfield:
Rebased onto master and adjusted to accommodate @joverlee521's changes from 1a25e2cc95c619c3e1187fb53a3c97ca91459568

Added 895738f to encourage the `@branch` syntax to be dropped for the default branch.

URLs for testing (using the notation from the commit message of 895738f):
- (1) https://nextstrain-s-flexible-c-qf0flk.herokuapp.com/community/jameshadfield/scratch
- (1) https://nextstrain-s-flexible-c-qf0flk.herokuapp.com/community/jameshadfield/nextstrain-community-build (default branch: "main")
- (2) https://nextstrain-s-flexible-c-qf0flk.herokuapp.com/community/jameshadfield/scratch@master
- (2) https://nextstrain-s-flexible-c-qf0flk.herokuapp.com/community/jameshadfield/nextstrain-community-build@main
- (3) https://nextstrain-s-flexible-c-qf0flk.herokuapp.com/community/jameshadfield/scratch@test-branch

@eharkins / @joverlee521 / @tsibley would you mind re-reviewing?

src/sources.js
```diff
@@ -195,18 +195,27 @@ class CommunitySource extends Source {
...
+    return `https://github.com/${this.repo}/raw/${await this.branch}/`;
+  }
 
-  get repoNameWithBranch() {
+  async repoNameWithBranch() {
+    const branch = await this.branch;
     return this.branch === "master"
       ? this.repoName
-      : `${this.repoName}@${this.branch}`;
+      : `${this.repoName}@${branch}`;
```
tsibley: This change (in the first commit on the branch) forgot to update the `this.branch === "master"` comparison to use `branch` instead (the local variable after `await`-ing), which was the source of

> however there is a small bug:
> 
> The `getAvailable` API response includes the `@branch` syntax, even for the default branch.

src/sources.js
```diff
@@ -191,31 +191,35 @@ class CommunitySource extends Source {
...
-          return "master";
-        });
-    }
+
+    this.branch = fetch(`https://api.github.com/repos/${this.owner}/${this.repoName}`)
+      .then((res) => res.json())
+      .then((data) => ({
+        default: data.default_branch,
+        name: urlDefinedBranch || data.default_branch
+      }))
```
tsibley, line 204: Rather than modify the type of `this.branch` from a string to an object containing two values, I think the internal API might be simpler by keeping `this.branch` as the branch from the URL and adding `this.defaultBranch` as a new property. Thoughts?

tsibley (commented):
> Case (2) URLs will work, however the `getAvailable` results will not
include the branch. This results in dataset listings pointing to URLs
in format (1), with the downside that the dataset dropdown isn't
working if a dataset of type (2) is loaded.

Is there a reason we can't support correct `getAvailable` paths for type (2) as well? It seems to me that we should support it to be friendlier to a) users and b) any programmatic generation of these URLs which then won't have to worry about doing its own default branch canonicalization.

src/sources.js
jameshadfield: That's where the bug was, but since we didn't keep track of the default branch we couldn't check whether the  branch (`await this.branch`) was the default one.

tsibley: Yep, I understand that. IMO, the bug in the first commit should be fixed, regardless of the subsequent change to stop assuming `master` is the default branch for a repo.

jameshadfield: Resolved in a2ecf7c

src/sources.js
jameshadfield: I see advantages to both (encapsulating branch-related information into `this.branch`, which I chose, or using properties such as `this.branch`, `this.defaultBranch`).

The only way I could achieve the latter is via code the following code, which I don't think is that nice. Do you have a better way?

```js
    let branchResolve, defaultBranchResolve;
    this.branch = new Promise((resolve) => {
      branchResolve = resolve;
    });
    this.branch = new Promise((resolve) => {
      defaultBranchResolve = resolve;
    });
    fetch(`https://api.github.com/repos/${this.owner}/${this.repoName}`)
      .then((res) => res.json())
      .then((data) => {
        branchResolve(urlDefinedBranch || data.default_branch);
        defaultBranchResolve(data.default_branch);
      })
      .catch(() => {
        console.log(`Error interpreting the default branch of ${this.constructor.name}`);
        branchResolve("master");
        defaultBranchResolve("master");
      });
```

tsibley: > Do you have a better way?

I believe something like the below would work just fine. Diff shown against the clean slate of `master` to more clearly demonstrate the changes:

```diff
--- a/src/sources.js
+++ b/src/sources.js
@@ -194,19 +194,31 @@ class CommunitySource extends Source {
     [this.repoName, this.branch] = repoName.split(/@/, 2);
 
     if (!this.repoName) throw new Error(`Cannot construct a ${this.constructor.name} without a repoName after splitting on /@/`);
+
+    this.defaultBranch = fetch(`https://api.github.com/repos/${this.owner}/${this.repoName}`)
+      .then((res) => res.json())
+      .then((data) => data.default_branch || "master")
+      .catch(() => {
+        console.log(`Error interpreting the default branch of ${this.constructor.name}`);
+        return "master";
+      });
+
     if (!this.branch) {
-      this.branch = "master";
+      this.branch = this.defaultBranch;
     }
   }
 
@@ -205,6 +216,7 @@ class CommunitySource extends Source {
 
-  get repoNameWithBranch() {
-    return this.branch === "master"
+  async repoNameWithBranch() {
+    const branch = await this.branch;
+    return branch === await this.defaultBranch
       ? this.repoName
-      : `${this.repoName}@${this.branch}`;
+      : `${this.repoName}@${branch}`;
   }
 
```

jameshadfield: Resolved in 
d81afdc

jameshadfield:
> It seems to me that we should support [Case (2) URLs] to be friendlier to a) users and b) any programmatic generation of these URLs which then won't have to worry about doing its own default branch canonicalization.

Resolved in d81afdc

PR #308, opened 2021-04-13
WIP redirect datasets which dont exist to parent page

jameshadfield:
We are beginning to add pages listing datasets / narratives into nextstrain.org such as /influenza and /sars-cov-2. (Note the former may be moved to "/flu" shortly.) Where possible, requests to "child" pages which don't have a dataset should redirect to these "parent" pages. 

**What happens currently**
Most URLs end up in Auspice (see flowchart) where a request is made for the corresponding dataset. If this dataset doesn't exist, Auspice shows a 404 page and a listing of datasets/narratives for that Source (core, staging etc). 

**What we would like to happen**
If the corresponding dataset does not exist, then we should redirect the client to the closest parent page, which is rendered from Gatsby. For instance,

* requests to /flu/no-dataset should redirect to /influenza
* requests to /flu/h3n2/no-dataset should redirect to /influenza/seasonal/h3n2 (were this to exist)
* requests to /ncov/antarctica should redirect to /sars-cov-2
* requests to /community/$org/$repo (assuming no default dataset exists) should result in a Gatsby page similar to /influenza which lists the available datasets & narratives for that repo. Similarly for /community/$org/$repo/$x where $x doesn't exist.
* requests for /staging/no-dataset should redirect to /staging, which lists all datasets/narratives from the staging source
* requests for /groups/X (and groups/X/Y if dataset Y doesn't exist) should display a page listing those 

**Complexities**
There are two major ones that I am aware of:
1. We currently don't know if a dataset exists until we try to access it. If it 404s then we can assume it doesn't currently exist. We must recheck this each time as new datasets may have been added (or removed) since the last check.
2. The /community/$org/$repo address should display different pages (one from Gatsby, one from Auspice) depending on whether a default dataset exists.

**This implementation**
This PR begins to implement the above by allowing Auspice to redirect to the desired nextstrain.org page if the dataset does not exist (see orange section in flow chart). Currently only URLs starting with `/flu` are handled, which redirect to `/influenza` if the dataset doesn't exist.

image
 
**Alternative implementation**

We could have the initial URL request the client makes to the server attempt to access the corresponding dataset. If this exists, we return Auspice (which then requests the dataset the server's just accessed). If the dataset doesn't exist we can return the Gatsby HTML (or a 302 to the correct page). I didn't attempt this as I believe it will result in long response times, as the server must go fetch the dataset, however there may be clever shortcuts available here.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,9 +14,9 @@
       }
     },
     "@babel/compat-data": {
-      "version": "7.13.12",
-      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.13.12.tgz",
-      "integrity": "sha512-3eJJ841uKxeV8dcN/2yGEUy+RfgQspPEgQat85umsE1rotuquQ2AbIub4S6j7c50a2d+4myc+zSlnXeIHrOnhQ=="
+      "version": "7.13.15",
+      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.13.15.tgz",
+      "integrity": "sha512-ltnibHKR1VnrU4ymHyQ/CXtNXI6yZC0oJThyW78Hft8XndANwi+9H+UIklBDraIjFEJzw8wmcM427oDd9KS5wA=="
     },
     "@babel/core": {
       "version": "7.11.6",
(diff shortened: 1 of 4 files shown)


jameshadfield:
Closing this, as I'll make an issue for discussion

Issue #293, opened 2021-04-02
/users/:name response has a 404 Not Found status code instead of 200 Ok


tsibley:
**Current Behavior**  
https://nextstrain.org/whoami responds with a 302 Redirection to https://nextstrain.org/users/trs (my username is _trs_), which responds with a 404 Not Found. The page content is as expected, just the status code is misleading.

**Expected behavior**  
/users/:name endpoint returns a 200 Ok response.

jameshadfield:
I encontered this as well while doing some work for a forthcoming PR. It occurs as none of our handlers in `server.js` handle this route, so it falls through to 
```js
app.get("*", (req, res) => {
  utils.verbose(`Sending Gatsby entrypoint for ${req.originalUrl}`);
  res.sendFile(gatsbyAssetPath(""), {}, (err) => {
    // This error callback is used to display the 404 page
    if (err.code === 'EISDIR') res.status(404).sendFile(gatsbyAssetPath("404.html"));
  });
});
```
Which, as far as I can understand, will _always_ use the error callback since `gatsbyAssetPath("")` doesn't exist (it's a directory). Gatsby's generated pages have client-side code which will end up rendering the users page, but how exactly this happens is unclear to me. 

This handling style, including the `404` was introduced via dad77d80ed180e2757e5a8d58ba36e811f8fbdc4. Before that it was simpler but, from memory, didn't display 404 pages as expected.
```js
app.get("*", (req, res) => {
  utils.verbose(`Sending Gatsby entrypoint for ${req.originalUrl}`);
  res.sendFile(gatsbyAssetPath("index.html"));
```

PR #311, opened 2021-04-16
List available datasets in Gatsby (not Auspice)

### LINKED ISSUE #310: List available datasets in Gatsby (not Auspice)
We have begun building out pages and components which render a filterable list of datasets / narratives, such as `/influenza` and `/sars-cov-2`. (Note the former may be moved to "/flu" shortly.) These will be expanded in due course to cover different “sources” such as nextstrain community, nextstrain core, nextstrain groups. This is detailed in #305. 

As part of this, we wish for URLs which are not backed by a dataset to render / redirect a corresponding “parent” page. For instance,  `/ncov/antarctica` should redirect to `/sars-cov-2`. (This issue use the word “redirect” but it doesn’t have to be client-redirects, it would also be possible to maintain the URL and simply render the page of `/sars-cov-2`.) 

**What happens currently**

Most URLs end up in Auspice (see flowchart) where a request is made for the corresponding dataset. If this dataset doesn't exist, Auspice shows a 404 page and a listing of datasets/narratives for that Source (core, staging etc). Note that this page is actually defined in nextstrain.org as a build-time customisation to auspice.

**What is desired**

Given a path X without an associated dataset, we wish to redirect to the closest “parent” page. Examples:

1. We maintain a list of [core dataset prefixes](https://github.com/nextstrain/nextstrain.org/blob/master/auspicePaths.js#L5). Any path matching these, but not backed by a dataset, should be redirected to `/pathogens` (or `/core`) which lists all core datasets.
2. Similarly the “staging” datasets, for which we don’t maintain a list of prefixes, should redirect to `/staging`.
3. Pages starting with `/flu` or `ncov`, `/influenza`, `/sars-cov-2` etc should redirect to `/influenza` or `/sars-cov-2` as appropriate. In the future there may be more pages such as these.
4. Groups URLs which don’t define a valid group (or aren’t authorised for that group) go to `/groups`. These should display all groups datasets, including private ones (see #306). 
5. Groups URLs for which a dataset doesn’t exist go to `/groups/name` which is rendered from within Gatsby, not Auspice.
6. Community URLs for which a dataset doesn’t exist go to `/community/org/name` which is rendered from Gatsby, irregardless of whether a default dataset exists.
7. Similarly to (6), `/community/org/name` is a Gatsby page if no default dataset is present. This allows us to remove the “special cased” `204` response from the `getDataset` API.
8. These pages are all rendered by Gatsby using reusable components, and we should expose the fact that the URL here wasn’t 

**Complexities**
There are two major ones that I am aware of:
* We currently don't know if a dataset exists until we try to access it. If it 404s then we can assume it doesn't currently exist. We must recheck this each time as new datasets may have been added (or removed) since the last check.
* The /community/$org/$repo address should display a different page (one from Gatsby, one from Auspice) depending on whether a default community dataset exists. (See 6 and 7 above)

image

jameshadfield:
Begins an implementation which would close #310. See commit messages for more information.

Currently works with core and staging URLs. All other sources (groups, community etc) are still handled by Auspice. Here are some trial URLs:
* https://nextstrain-s-redirects--ouuvbt.herokuapp.com/flu/vic/yam Invalid flu URL, shows main influenza page instead. 
* https://nextstrain-s-redirects--ouuvbt.herokuapp.com/ncov/antartica ncov dataset which doesn't exist. Shows SARS-CoV-2 main page.
* https://nextstrain-s-redirects--ouuvbt.herokuapp.com/ebola/equateur dataset doesn't exist. Shows the splash page, as we don't yet have a page listing all core datasets.
* https://nextstrain-s-redirects--ouuvbt.herokuapp.com/staging/no-dataset shows new staging page listing datasets & narratives.

(Note that the URL often updates after loading, but that shouldn't happen.)

Next steps:
- [ ] Stop the URL updating (it's happening client side)
- [ ] Add a reusable "Dataset didn't exist, here's the parent page" banner to pages.
- [ ] More unit tests
- [ ] Explore what caching is being used for the server-side dataset fetching (if any)
- [ ] Measure loading times 

Comments welcomed on this direction!

### DIFF
--- a/auspicePaths.js
+++ b/auspicePaths.js
@@ -39,7 +39,6 @@ const potentialAuspiceRoutes = [
   ...coreBuilds.map(url => url + "/*"),
   "/narratives",
   "/narratives/*",
-  "/staging",
   "/staging/*",
   ...groups.map((group) => `/groups/${group}`),
   ...groups.map((group) => `/groups/${group}/*`),
@@ -80,7 +79,6 @@ const isRequestBackedByAuspiceDataset = async (req, res, next) => {
     if (
       pathParts[0]==="status" ||
       pathParts[0]==="narratives" ||
-      pathParts[0]==="staging" ||
       pathParts[0]==="groups" ||
       pathParts[0]==="community" || // note that /community itself is redirected earlier
       pathParts[0]==="fetch"
--- a/auspicePaths.js
+++ b/auspicePaths.js
@@ -1,6 +1,11 @@
+const fetch = require("node-fetch");
 const sources = require("./src/sources");
+const { splitPrefixIntoParts, parsePrefix } = require("./src/getDatasetHelpers");
 
-/* Dataset and narrative paths hit Auspice's entrypoint.
+/**
+ * Auspice core builds have URLs where the first (`/`-separated) field matches the following list.
+ * New core builds must be added here else they will be routed to Gatsby.
+ * This could be dynamically generated at server start time via a S3 API call if we wish.
  */
 const coreBuilds = [
   "/dengue",
@@ -20,7 +25,13 @@ const coreBuilds = [
 
 const groups = [...sources.keys()].filter((k) => !["core", "staging", "community"].includes(k));
 
-const auspicePaths = [
+
+/**
+ * The following routes _may_ be handled by Auspice.
+ * Note: as we add more functionality to Gatsby, this list will shrink
+ * (e.g. /staging will be a gatsby page, not an auspice one)
+ */
+const potentialAuspiceRoutes = [
   "/status",
   "/status/*",
   ...coreBuilds,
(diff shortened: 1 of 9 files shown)


server.js
```diff
@@ -101,20 +103,25 @@ app.route("/charon/getSourceInfo")
 app.route("/charon/*")
   .all((req, res) => res.status(404).end());
 
-app.route(auspicePaths).get((req, res) => {
-  utils.verbose(`Sending Auspice entrypoint for ${req.originalUrl}`);
-  res.sendFile(auspiceAssetPath("dist", "index.html"), {headers: {"Cache-Control": "no-cache, no-store, must-revalidate"}});
-});
+/* Specific routes to be handled by Gatsby client-side routing
```
jameshadfield, line 106: Closes #293

jameshadfield:
I'm going to merge this into the `next` branch in order for @eharkins PR's to be simpler and move forward in this direction. We've discussed this in passing with @tsibley who agrees with the general idea, however given the sweeping changes made to how we handle requests I feel more review / testing / improvements are needed. Hence the merge into `next` rather than `master`, as the latter would be automatically deployed to nextstrain.org.

I'll create issues for the to-do items listed in this PR.

auspicePaths.js
```diff
@@ -45,4 +56,85 @@ const auspicePaths = [
...
+/**
+ * Does the main dataset file exist?
+ * There is no attempt here to examine the validity or completeness of such a file.
+ * Note: as we don't care about the file contents, there should be a
+ * way to avoid fetching the data here and thus speed up this function.
+ * @returns {bool}
+ */
+async function datasetExists(dataset) {
+  try {
+    if ((await fetch(dataset.urlFor("main"))).status===200) {
```
joverlee521, line 122: Changing to the [HEAD method](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD) can speed up the request since it doesn't return the data.

jameshadfield: Thanks @joverlee521 -- implementation up at https://github.com/nextstrain/nextstrain.org/pull/325

PR #314, opened 2021-04-21
Gatsby handle individual groups pages

eharkins:
on top of #313. Second commit message has the details. WIP

### DIFF
--- a/static-site/gatsby-node.js
+++ b/static-site/gatsby-node.js
@@ -199,6 +199,13 @@ exports.createPages = ({graphql, actions}) => {
           component: path.resolve("src/pages/groups.jsx")
         });
 
+        // Create individual groups pages
+        createPage({
+          path: "/groups/:groupName",
+          matchPath: "/groups/:groupName",
+          component: path.resolve("src/sections/individual-group-page.jsx")
+        });
+
         // Create page detailing all things SARS-CoV-2
         // Note that this is in src/sections, not src/pages. This is because we don't
         // want to render anything at the exact url of nextstrain.org/sars-cov-2-page
--- a/static-site/src/components/splash/groupHeading.jsx
+++ b/static-site/src/components/splash/groupHeading.jsx
@@ -23,13 +23,11 @@ function Title({avatarSrc, children}) {
     object-fit: contain;
   `;
   const TitleDiv = styled.div`
-    && {
-      font-weight: 500;
-      font-size: 26px;
-      display: flex;
-      flex-direction: column;
-      justify-content: center;
-    }
+    font-weight: 500;
+    font-size: 26px;
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
   `;
   return (
     <div style={{display: "flex", justifyContent: "start", padding: "50px 0px 20px 0px"}}>
@@ -47,12 +45,10 @@ function Title({avatarSrc, children}) {
 function Byline({children}) {
   if (!children) return null;
   const Div = styled.div`
-    && {
-      font-size: 18px;
-      font-weight: 400;
-      line-height: 1.428;
-      color: #A9ADB1;
-    }
+    font-size: 18px;
+    font-weight: 400;
+    line-height: 1.428;
+    color: #A9ADB1;
   `;
   return (<Div>{children}</Div>);
 }
--- a/static-site/src/components/splash/groupHeading.jsx
+++ b/static-site/src/components/splash/groupHeading.jsx
@@ -1,0 +1,81 @@
+import React from "react";
+import styled from 'styled-components';
+import { StyledLink } from "../Datasets/list-datasets";
+import { FlexCenter } from "../../layouts/generalComponents";
+
+const OverviewContainer = styled.div`
+  text-align: justify;
+  font-size: 16px;
+  margin-top: 5px;
+  margin-bottom: 5px;
+  font-weight: 300;
+  color: var(--darkGrey);
+  line-height: 1.42857143;
+  margin: 0px auto 0px auto;
+  max-width: 900px;
+`;
+
+function Title({avatarSrc, children}) {
+  if (!children) return null;
+  const AvatarImg = styled.img`
+    width: 140px;
+    margin-right: 20px;
+    object-fit: contain;
+  `;
+  const TitleDiv = styled.div`
+    && {
+      font-weight: 500;
+      font-size: 26px;
+      display: flex;
+      flex-direction: column;
+      justify-content: center;
+    }
+  `;
+  return (
+    <div style={{display: "flex", justifyContent: "start", padding: "50px 0px 20px 0px"}}>
+      {avatarSrc ?
+        <AvatarImg alt="avatar" src={avatarSrc}/> :
+        null
+      }
+      <TitleDiv>
+        {children}
+      </TitleDiv>
+    </div>
+  );
+}
+
+function Byline({children}) {
+  if (!children) return null;
+  const Div = styled.div`
+    && {
+      font-size: 18px;
+      font-weight: 400;
+      line-height: 1.428;
+      color: #A9ADB1;
+    }
+  `;
+  return (<Div>{children}</Div>);
+}
+
+export default function GroupHeading({sourceInfo}) {
+  return (
+    <>
+      <FlexCenter>
+        <Title avatarSrc={sourceInfo.avatar}>
+          {sourceInfo.title}
+          <Byline>{sourceInfo.byline}</Byline>
+          {sourceInfo.website &&
+            <div style={{fontSize: "18px"}}>
+              <StyledLink href={sourceInfo.website}>{sourceInfo.website.replace(/(^\w+:|^)\/\//, '')}</StyledLink>
+            </div>}
+        </Title>
+      </FlexCenter>
+      {sourceInfo.overview &&
+        <FlexCenter>
+          <OverviewContainer>
+            {sourceInfo.overview}
+          </OverviewContainer>
+        </FlexCenter>
+      }
+    </>);
+}
--- a/static-site/src/sections/individual-group-page.jsx
+++ b/static-site/src/sections/individual-group-page.jsx
@@ -42,7 +42,6 @@ class Index extends React.Component {
   async componentDidMount() {
     const groupName = this.props["groupName"];
     try {
-      console.log("Start", groupName);
       const [sourceInfo, availableData] = await Promise.all([
         fetchAndParseJSON(`/charon/getSourceInfo?prefix=/groups/${groupName}/`),
         fetchAndParseJSON(`/charon/getAvailable?prefix=/groups/${groupName}/`)
--- a/static-site/src/sections/individual-group-page.jsx
+++ b/static-site/src/sections/individual-group-page.jsx
@@ -1,0 +1,129 @@
+import React from "react";
+import ScrollableAnchor, { configureAnchors } from "react-scrollable-anchor";
+import {
+  HugeSpacer,
+  FlexCenter
+} from "../layouts/generalComponents";
+import * as splashStyles from "../components/splash/styles";
+import DatasetSelect from "../components/Datasets/dataset-select";
+import GenericPage from "../layouts/generic-page";
+import { fetchAndParseJSON } from "../util/datasetsHelpers";
+import GroupHeading from "../components/splash/groupHeading";
+
+const GroupNotFound = ({groupName}) => (
+  <FlexCenter>
+    <splashStyles.CenteredFocusParagraph>
+      {`The Nextstrain Group "${groupName}" doesn't exist yet, or there was an error getting data for that group. `}
+      Please <a href="mailto:hello@nextstrain.org">contact us at hello@nextstrain.org </a>
+      if you believe this to be an error.</splashStyles.CenteredFocusParagraph>
+  </FlexCenter>
+);
+
+class Index extends React.Component {
+  constructor(props) {
+    super(props);
+    configureAnchors({ offset: -10 });
+    this.state = {
+      groupNotFound: false
+    };
+  }
+
+  // parse getAvailable listing into one that dataset-select component accepts
+  createDatasetListing = (list, groupName) => {
+    return list.map((d) => {
+      return {
+        filename: d.request.replace(`groups/${groupName}/`, '').replace('narratives/', ''),
+        url: `https://nextstrain.org/${d.request}`,
+        contributor: groupName
+      };
+    });
+  };
+
+  async componentDidMount() {
+    const groupName = this.props["groupName"];
+    try {
+      console.log("Start", groupName);
+      const [sourceInfo, availableData] = await Promise.all([
+        fetchAndParseJSON(`/charon/getSourceInfo?prefix=/groups/${groupName}/`),
+        fetchAndParseJSON(`/charon/getAvailable?prefix=/groups/${groupName}/`)
+      ]);
+      this.setState({
+        sourceInfo,
+        groupName,
+        datasets: this.createDatasetListing(availableData.datasets, groupName),
+        narratives: this.createDatasetListing(availableData.narratives, groupName)
+      });
+    } catch (err) {
+      console.error("Cannot find group.", err.message);
+      this.setState({groupName, groupNotFound: true});
+    }
+  }
+
+  render() {
+    const groupName = this.props["groupName"];
+    if (this.state.groupNotFound) {
+      return (
+        <GenericPage location={this.props.location}>
+          <GroupNotFound groupName={groupName}/>
+        </GenericPage>
+      );
+    }
+    if (!this.state.sourceInfo) {
+      return (
+        <GenericPage location={this.props.location}>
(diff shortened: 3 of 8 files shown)


eharkins:
Seems like the issues with requesting available datasets and source info for groups were only on the local server as e.g. https://nextstrain-s-individual-hdwuev.herokuapp.com/groups/ngs-sa seems to be successfully requesting both.

eharkins:
Marking as ready for review since to-do items I had are mostly covered by recent commits. Noting to myself that this will probably want to be rebased onto #311.

eharkins:
#315 separates the concern of parsing the markdown groups intro so that we can review here with a working dev app. For now we display the markdown content without parsing and formatting it.

static-site/src/sections/individual-group-page.jsx
```diff
@@ -0,0 +1,211 @@
...
+                    value: (dataset) => dataset.filename.replace(/_/g, ' / ').replace('.json', ''),
+                    url: (dataset) => dataset.url
+                  }
+                ]}
+                unit="narrative"
+              />
+            </div>
+          </ScrollableAnchor>
+        )}
+        { this.state.errorFetchingData &&
```
jameshadfield, line 192: `errorFetchingData` is set if the `getAvailable` call fails. In this case, we should display the group page as normal, but swap out the `DatasetSelect` component(s) for an error message saying "Listing of datasets failed, please contact us at ..."

static-site/src/sections/individual-group-page.jsx
```diff
@@ -0,0 +1,211 @@
...
+          {this.state.sourceInfo.overview &&
+            <FlexCenter>
+              <OverviewContainer>
+                {this.state.sourceInfo.overview}
+              </OverviewContainer>
+            </FlexCenter>
+          }
+        </>}
+        <HugeSpacer />
+        {this.state.dataLoaded && this.state.sourceInfo && this.state.sourceInfo.showDatasets && this.state.datasets.length > 0 && (
```
jameshadfield, line 156: In the case of a successful `getAvailable` API call, but zero datasets, we wish to display a single line such as "No datasets are present in this group". Ditto for narratives.

static-site/src/sections/individual-group-page.jsx
```diff
@@ -0,0 +1,211 @@
...
+          <FlexCenter>
+            <splashStyles.CenteredFocusParagraph>
+                  Something went wrong getting data.
+                  Please <a href="mailto:hello@nextstrain.org">contact us at hello@nextstrain.org </a>
+                  if this continues to happen.</splashStyles.CenteredFocusParagraph>
+          </FlexCenter>}
+        { this.state.groupNotFound &&
+        <FlexCenter>
+          <splashStyles.CenteredFocusParagraph>
+            {`The Nextstrain Group "${this.state.groupName}" doesn't exist yet. `}
```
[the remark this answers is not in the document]
jameshadfield: I would also suggest refactoring away from 
```js
  render() {
    return (
      <GenericPage location={this.props.location}>
        {this.state.sourceInfo && lotsOfContent}
        {this.state.hasError && displayError}
     </GenericPage>
```

towards an early-return control flow like

```js
  render() {
    if (state.groupNotFound) {
         return (
                 <GroupDoesntExist groupName={groupName}>
          )
    }
    return (
      <GenericPage location={this.props.location}>
        // normal content
     </GenericPage>
```

jameshadfield (changes requested):
Thanks @eharkins -- this is working well. Most of my comments are with an eye to future improvements which will expand the capabilities here. I think some of the commits could also be squashed together to improve the readability of history.

eharkins:
Thanks for the detailed review @jameshadfield! I tried to squash down most of the existing work into fewer commits concerning distinct matters. I've added a couple commits on top of this that hopefully address all of your comments, except https://github.com/nextstrain/nextstrain.org/pull/314#discussion_r627855773, which I feel more confident about trying out having used `visibleGroups` in https://github.com/nextstrain/nextstrain.org/pull/316/commits/781039e3465c16056781f70a8b7a1e75ea3bda55. 

Still need to test my last two commits more.

static-site/src/sections/individual-group-page.jsx
```diff
@@ -0,0 +1,223 @@
...
+        url: `https://nextstrain.org/${d.request}`,
+        contributor: groupName
+      };
+    });
+  };
+
+  async componentDidMount() {
+    const groupName = this.props["groupName"];
+    let sourceInfoPromise, availableDataPromise;
+    try {
```
jameshadfield, line 114: `Promise.allSettled` won't throw if the promises reject, so the error handling here will actually be triggered by trying to access `availableDataPromise.value.datasets` which I don't think is intentional. My view is that error handling here can be much simplified if we accept that the two API calls should either both work (if the group exists), or both fail (if the group doesn't); any other combination can be considered a server error and it's OK for the frontend to fall back to a "group doesn't exist" display.

With this assumption we can simplify the code here to use `Promise.all` in a try/catch block, remove `state.dataLoaded` and `state.errorFetchingData`, which allows for simplified rendering logic. For instance,

```diff
diff --git a/static-site/src/sections/individual-group-page.jsx b/static-site/src/sections/individual-group-page.jsx
index cb8853e..444c8f4 100644
--- a/static-site/src/sections/individual-group-page.jsx
+++ b/static-site/src/sections/individual-group-page.jsx
@@ -89,11 +89,7 @@ class Index extends React.Component {
     super(props);
     configureAnchors({ offset: -10 });
     this.state = {
-      dataLoaded: false,
-      errorFetchingData: false,
-      groupNotFound: false,
-      groupName: undefined,
-      sourceInfo: undefined
+      groupNotFound: false
     };
   }
 
@@ -110,31 +106,21 @@ class Index extends React.Component {
 
   async componentDidMount() {
     const groupName = this.props["groupName"];
-    let sourceInfoPromise, availableDataPromise;
     try {
-      [sourceInfoPromise, availableDataPromise] = await Promise.allSettled([
+      console.log("Start", groupName);
+      const [sourceInfo, availableData] = await Promise.all([
         fetchAndParseJSON(`/charon/getSourceInfo?prefix=/groups/${groupName}/`),
         fetchAndParseJSON(`/charon/getAvailable?prefix=/groups/${groupName}/`)
       ]);
       this.setState({
-        sourceInfo: sourceInfoPromise.value,
+        sourceInfo,
         groupName,
-        datasets: this.createDatasetListing(availableDataPromise.value.datasets, groupName),
-        narratives: this.createDatasetListing(availableDataPromise.value.narratives, groupName),
-        dataLoaded: true
+        datasets: this.createDatasetListing(availableData.datasets, groupName),
+        narratives: this.createDatasetListing(availableData.narratives, groupName)
       });
     } catch (err) {
-      if (sourceInfoPromise.status === "rejected") {
-        console.error("Cannot find group.", err.message);
-        this.setState({groupName, groupNotFound: true});
-      } else {
-        console.error("Error fetching / parsing data.", err.message);
-        this.setState({
-          groupName,
-          sourceInfo: sourceInfoPromise.value,
-          errorFetchingData: true
-        });
-      }
+      console.error("Cannot find group.", err.message);
+      this.setState({groupName, groupNotFound: true});
     }
   }
 
@@ -146,27 +132,31 @@ class Index extends React.Component {
         </GenericPage>
       );
     }
+    if (!this.state.sourceInfo) {
+      return (
+        <GenericPage location={this.props.location}>
+          <OverviewContainer>Data loading...</OverviewContainer>
+        </GenericPage>
+      );
+    }
     return (
       <GenericPage location={this.props.location}>
-        {this.state.sourceInfo &&
-        <>
+        <FlexCenter>
+          <Title avatarSrc={this.state.sourceInfo.avatar}>
+            {this.state.sourceInfo.title}
+            <Byline>{this.state.sourceInfo.byline}</Byline>
+            <Website>{this.state.sourceInfo.website}</Website>
+          </Title>
+        </FlexCenter>
+        {this.state.sourceInfo.overview &&
           <FlexCenter>
-            <Title avatarSrc={this.state.sourceInfo.avatar}>
-              {this.state.sourceInfo.title}
-              <Byline>{this.state.sourceInfo.byline}</Byline>
-              <Website>{this.state.sourceInfo.website}</Website>
-            </Title>
+            <OverviewContainer>
+              {this.state.sourceInfo.overview}
+            </OverviewContainer>
           </FlexCenter>
-          {this.state.sourceInfo.overview &&
-            <FlexCenter>
-              <OverviewContainer>
-                {this.state.sourceInfo.overview}
-              </OverviewContainer>
-            </FlexCenter>
-          }
-        </>}
+        }
         <HugeSpacer />
-        {this.state.dataLoaded && this.state.sourceInfo && this.state.sourceInfo.showDatasets && (
+        {this.state.sourceInfo.showDatasets && (
           <ScrollableAnchor id={"datasets"}>
             <div>
               <splashStyles.H3>Available datasets</splashStyles.H3>
@@ -187,7 +177,7 @@ class Index extends React.Component {
           </ScrollableAnchor>
         )}
         <HugeSpacer />
-        {this.state.dataLoaded && this.state.sourceInfo && this.state.sourceInfo.showNarratives && (
+        {this.state.sourceInfo.showNarratives && (
           <ScrollableAnchor id={"narratives"}>
             <div>
               <splashStyles.H3>Available narratives</splashStyles.H3>
@@ -208,13 +198,6 @@ class Index extends React.Component {
             </div>
           </ScrollableAnchor>
         )}
-        { this.state.errorFetchingData &&
-          <FlexCenter>
-            <splashStyles.CenteredFocusParagraph>
-                  Something went wrong getting data.
-                  Please <a href="mailto:hello@nextstrain.org">contact us at hello@nextstrain.org </a>
-                  if this continues to happen.</splashStyles.CenteredFocusParagraph>
-          </FlexCenter>}
       </GenericPage>
     );
   }
```

static-site/src/util/datasetsHelpers.js
```diff
@@ -0,0 +1,2 @@
+
+export const fetchAndParseJSON = async jsonUrl => fetch(jsonUrl).then((res) => res.json());
```
jameshadfield, line 2: As this is a helper function, brevity isn't critical and we could follow best practices here by enforcing headers:
```js
export const fetchAndParseJSON = async (jsonUrl) => {
  return fetch(jsonUrl, { headers: { Accept: 'application/json' }})
    .then((res) => res.json());
};
```

static-site/src/sections/individual-group-page.jsx
```diff
@@ -0,0 +1,223 @@
...
+
+class Index extends React.Component {
+  constructor(props) {
+    super(props);
+    configureAnchors({ offset: -10 });
+    this.state = {
+      dataLoaded: false,
+      errorFetchingData: false,
+      groupNotFound: false,
+      groupName: undefined,
```
jameshadfield, line 95: Is there an advantage to having `groupName` in state, as opposed to `this.props.groupName`? (E.g. may we modify this variable in the future?)

static-site/src/sections/individual-group-page.jsx
```diff
@@ -0,0 +1,223 @@
...
+      font-size: 18px;
+      font-weight: 400;
+      line-height: 1.428;
+      color: #A9ADB1;
+    }
+  `;
+  return (<Div>{children}</Div>);
+}
+
+function Website({children}) {
```
jameshadfield, line 55: I would use `StyledLink` from `components/Datasets/list-datasets.jsx` instead of this for two reasons:
1. The function signature implies that multiple children can be supplied, but this will be problematic
2. We often want the displayed text to be subtly different to the `href` (e.g. we often leave out the `https://`)

static-site/src/sections/individual-group-page.jsx
jameshadfield: Related to data fetching, we also need to display a "Data Loading" message while the fetches are happening. The above diff contains a very simple implementation of this.

jameshadfield (changes requested):

static-site/gatsby-node.js
```diff
@@ -193,6 +193,13 @@ exports.createPages = ({graphql, actions}) => {
           component: path.resolve("src/sections/users.jsx")
         });
 
+        // Create individual groups pages
+        createPage({
+          path: "/groups/:groupName",
+          matchPath: "/groups/:groupName",
```
jameshadfield, line 199: Is there any advantage of having this here, when it's the same as `path`? (Gatsby docs [here](https://www.gatsbyjs.com/docs/gatsby-internals-terminology/#matchpath)).

static-site/src/sections/individual-group-page.jsx
```diff
@@ -0,0 +1,200 @@
...
+  HugeSpacer,
+  FlexCenter
+} from "../layouts/generalComponents";
+import * as splashStyles from "../components/splash/styles";
+import DatasetSelect from "../components/Datasets/dataset-select";
+import GenericPage from "../layouts/generic-page";
+import { fetchAndParseJSON } from "../util/datasetsHelpers";
+import { StyledLink } from "../components/Datasets/list-datasets";
+
+function Title({avatarSrc, children}) {
```
jameshadfield, line 14: Refactoring recommendation, I would advocate pursuing one of  two styles here:

1. Make a number of styled components, with each of them doing one thing, and then call them all from the main render loop. E.g. `HeadingContainer`, `Title`, `Byline`, `GroupWebsite`, `Avatar`, `GroupDescription` etc. This is nice and declarative, and allows new developers who are reading the `render()` method for the first time to easily understand what's happening. 

2. Make an separate, higher level component such as `GroupHeading`, which takes in `this.props.sourceInfo` as its own prop and handles all the rendering. This can help comprehension of the render method if the solution in (1) becomes overly long. (Within this components file, you will most probably want to use the components described in (1).)

At the moment we've got a bit of both. I favour (2) as the complexity of this header section will only grow as we add in markdown parsing, but (1) would also work!

jameshadfield (changes requested):
Looks good @eharkins. 

Regarding the comment above about using `context` to access `visibleGroups`: I would leave this as a future todo. The current implementation is working well, and I don't think it's worth it to change it currently. The speed improvements would only be encountered on groups pages which don't exist, which we don't need to prioritise. Note that if you are not logged in and try to access a private group, you are simply redirected to a login page ([code here](https://github.com/nextstrain/nextstrain.org/blob/master/authn.js#L296-L312)).

I would advise refactoring the components used to render the heading in preparation for (a) markdown parsing and (b) to set the stage for future dev work when we expand the capabilities of this page.

eharkins:
- [ ] to-do item upon merging this into `next`: set [this same value](https://github.com/nextstrain/nextstrain.org/pull/314/commits/bbbb508b35b360842e80e2114c633368aa738dd4#diff-b3b844eb5004be685a00570e96581a69eb0e92852593a76b5c11f46358cd579fR205) for the /groups page's narrative listing so that it says "filter narratives" not "filter datasets."

static-site/gatsby-node.js
eharkins: I think it's vital; commenting it out like:
```
        createPage({
          path: "/groups/:groupName",
          // matchPath: "/groups/:groupName",
          component: path.resolve("src/sections/individual-group-page.jsx")
        });
```
, I get redirected to http://localhost:5000/ immediately upon visiting http://localhost:5000/groups/seattleflu

eharkins: Meanwhile I'm not sure `path` being exactly `/groups/:groupName` allows this to work in any way other than making a distinction from the `/groups` page for which we want to return a different page:
```
        createPage({
          path: "/groups",
          component: path.resolve("src/pages/groups.jsx")
        });
```

eharkins:
@jameshadfield I think this is good to go. #336 is a start on some of the behavior we discussed for when various groups pages don't exist.

static-site/src/sections/individual-group-page.jsx
```diff
@@ -0,0 +1,129 @@
...
+        url: `https://nextstrain.org/${d.request}`,
+        contributor: groupName
+      };
+    });
+  };
+
+  async componentDidMount() {
+    const groupName = this.props["groupName"];
+    try {
+      console.log("Start", groupName);
```
jameshadfield, line 45: Logging statement to remove 😉

static-site/src/components/splash/groupHeading.jsx
```diff
@@ -0,0 +1,81 @@
...
+        {children}
+      </TitleDiv>
+    </div>
+  );
+}
+
+function Byline({children}) {
+  if (!children) return null;
+  const Div = styled.div`
+    && {
```
jameshadfield, line 50: What is the `&&` syntax of styled components doing? Is this a CSS specificity thing?

jameshadfield (changes requested):
Looks good @eharkins -- one minor `console.log` to remove and this can be merged!

static-site/src/components/splash/groupHeading.jsx
eharkins: It resolves styling conflicts by setting specificity; https://styled-components.com/docs/basics#pseudoelements-pseudoselectors-and-nesting. Doesn't seem necessary in this case.

jameshadfield (approved):

PR #315, opened 2021-04-28
gatsby groups page parse markdown overview;

eharkins:
Copies the approach of auspice,
previously done by
auspice-client/customisations/splash.js.
That page will be retired once gatsby
performs all the necessary function to
handle routing to either gatsby landing
pages or to auspice.

* not working due to issue with `dompurify` which i am trying to sort out.


eharkins:
Superseded by #345

PR #317, opened 2021-05-06
ResourceNotFoundError if source not found;

eharkins:
If we can't parse a request prefix into
a source that we recognize, we should
throw some official error. Not sure this
is how we want to do it but before it
was just a TypeError:
"Source is not a constructor"
happening when trying to call the Source function.

I could be wrong about this behavior but I was confused by a 500 server error when requesting a nonexistent group.

### DIFF
--- a/src/getDatasetHelpers.js
+++ b/src/getDatasetHelpers.js
@@ -1,5 +1,6 @@
 const utils = require("./utils");
 const sources = require("./sources");
+const {ResourceNotFoundError} = require("./exceptions");
 
 const handle500Error = (res, clientMsg, serverMsg="") => {
   utils.warn(`${clientMsg} -- ${serverMsg}`);
@@ -56,6 +57,8 @@ const splitPrefixIntoParts = (prefix) => {
   }
 
   const Source = sources.get(sourceName);
+  // Source not found error
+  if (typeof Source !== "function") throw new ResourceNotFoundError();
   let source;
 
   switch (sourceName) {


src/getDatasetHelpers.js
```diff
@@ -56,6 +57,8 @@ const splitPrefixIntoParts = (prefix) => {
   }
 
   const Source = sources.get(sourceName);
+  // Source not found error
```
jameshadfield, line 60: A cleaner and more descriptive syntax than type checking here would be
```js
if (!sources.has(sourceName)) {
    throw new ResourceNotFoundError();
}
const Source = sources.get(sourceName)
```

src/getDatasetHelpers.js
```diff
@@ -56,6 +57,8 @@ const splitPrefixIntoParts = (prefix) => {
...
+  if (typeof Source !== "function") throw new ResourceNotFoundError();
```
jameshadfield, line 61: I think it's valid to use `ResourceNotFoundError` in this case, as we have examples where this is used internally to end up with the client being sent a `404`. 

Here we're within `splitPrefixIntoParts` which is called from ~7 different locations. Each of these should be wrapped in a `try/catch` block, including a specific handler for `ResourceNotFoundError` where appropriate. Looking at the code this happens in only one or two cases!

jameshadfield (changes requested):
Good find @eharkins, this failure mode should be handled much better than at the moment, where we get `UnhandledPromiseRejectionWarning`s on the server. Throwing a specific exception is part of the solution, but we also have to handle the exceptions (specific and general).

tsibley:
Reviewing this PR reminded me I had an unpushed alternate approach that fixed the same bug and also introduced what I think is more sustainable error handling for async code in our routes. I've opened https://github.com/nextstrain/nextstrain.org/pull/320 with it. Take a look?

jameshadfield:
Superseded by #320 - thanks @eharkins for identifying the error 💯

PR #320, opened 2021-05-11
Improve async error handling in routes

tsibley:
Switches to async-friendly Express routers. Async-friendly routers eliminate the need for most of our own try/catch blocks around internal awaits, as they automatically catch exceptions and send an appropriate error response.  This also lets us start throwing exceptions for HTTP error codes instead passing around the Express response object.

Few immediate functional changes, but this enables future simplifications as we touch other code paths. As an example, the 404 handling is now much simpler.

Reviewing https://github.com/nextstrain/nextstrain.org/pull/317 made me recall that on a local work-in-progress branch I had a similar fix for the 500 error on unknown sources that included these broader error-handling changes. I've rebased them off of that WIP, which was focused on entirely different changes, and onto the current `master` branch for this PR.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,11 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@awaitjs/express": {
+      "version": "0.6.3",
+      "resolved": "https://registry.npmjs.org/@awaitjs/express/-/express-0.6.3.tgz",
+      "integrity": "sha512-x7N92+Rmy/WHEaXXunV2ArVa/AqzkJlouc8pokuVF9h6DeUtbkPaEKVYKMDEFs2u5Hbw2XfwGVyLf69yr2UG5g=="
+    },
     "@babel/code-frame": {
       "version": "7.5.5",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.5.5.tgz",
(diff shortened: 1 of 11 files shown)


jameshadfield (changes requested):
Generally a much nicer approach here, definitely preferable to having to wrap everything in try/catch blocks and manage the exceptions in each handler.

I looked at the docs for `@awaitjs/express` (and `http-errors`), but it's not clear to me _where_ errors are being handled. For instance, in the following case the correct response is sent to the client, but the error handling code, wherever it is, re-prints the error which isn't desirable for us:
```
curl -I  http://localhost:5000/charon/getAvailable?prefix=/groups/none`
```
```
NotFoundError: Not Found
    at splitPrefixIntoParts (/Users/enigma/projects/nextstrain/nextstrain.org/src/getDatasetHelpers.js:61:22)
    at getAvailable (/Users/enigma/projects/nextstrain/nextstrain.org/src/getAvailable.js:10:20)
...
```

tsibley:
@jameshadfield [Express' default error handler](https://expressjs.com/en/guide/error-handling.html#the-default-error-handler) is responsible for handling the exceptions, which notes that

> The stack trace is not included in the production environment.

You can see this for yourself if you visit the review app (which is in production mode), e.g.

https://nextstrain-s-async-erro-d2w1ro.herokuapp.com/charon/getAvailable?prefix=/groups/none

says just "Not Found". We can provide our own error handler if we'd like, but I think the default is fine for now.

jameshadfield (approved):
Thanks for the docs link, PR looks good to me. @eharkins?

eharkins (approved):
Looking at this, it makes sense to me based on my still-developing understanding of the nextstrain.org server and this package. Testing it on #314, it does what I was going for with #317, and clearly addresses the same problem more comprehensively. Thanks!

PR #316, opened 2021-05-06
/groups page

eharkins:
to do:
- GH action to update static dataset listing
- detailed introduction to what groups are and links to docs
- list of all groups similar to cards on splash page

PR #321, opened 2021-05-13
use react context for userDataWrapper;

eharkins:
Instead of requiring that children components be nested directly
beneath the userDataWrapper component in order to use its state
(by cloning all children of the wrapper and passing props to each),
this uses react's Context (https://reactjs.org/docs/context.html)
to allow children at any nesting level within userDataWrapper to
access its state.

Needs testing. Useful for #316 and #314


jameshadfield (changes requested):
Looks great @eharkins - this will be really beneficial going forward. Tested locally and it's working well. Two small changes:
(1) Could you rebase / cherry-pick the relevant commit onto the `next` branch (I changed the base branch here, but that wasn't enough unfortunately)
(2) Could you update `generic-page.jsx` so that all the children there are within the `<UserDataWrapper>`. This will allow the groups pages to access them 😄

Issue #306, opened 2021-04-12
Routinely collect shared datasets & narratives


jameshadfield:
**Problem summary:**

Datasets & narratives are able to be visualised in Nextstrain through various mechanisms: core, staging, groups (public & private) and community. In order to achieve #305 we need to produce a file listing these datasets & narratives which the client can fetch. In the context of private groups, authentication is required for the client to view these. 

**Prior art**

We currently have various ways of collecting these data:
* The dataset behind https://nextstrain.org/influenza is regenerated ever 5 minutes (!) via [a GitHub action](https://github.com/nextstrain/nextstrain.org/blob/master/.github/workflows/update-flu-catalogue.yml) in the nextstrain/nextstrain.org repo which iterates over the objects in our main S3 bucket.
* There are [scripts available](github.com/nextstrain/community-search/) which use the GitHub API to search all public repos for datasets which match the required syntax for community sharing functionality. 
* The dataset behind nextstrain.org/sars-cov-2 is [stored in the GitHub repo](https://github.com/nextstrain/nextstrain.org/blob/master/static-site/content/SARS-CoV-2-Datasets.yaml) and is modified via PRs.
* Proof-of-principle script & UI to display all known public datasets: https://github.com/nextstrain/nextstrain.org/pull/303. JSON is 22kb (gzipped). Takes c. 20 seconds to scan all public S3 buckets, and this time isn't blocking.

**Augmented data**

The above examples limit the (meta)data we know about each dataset/narrative to filename, source, date the file was uploaded. By reading the file itself we can gather further information such as maintainer/author, number of samples in dataset, date updated etc. This task is treated as an extension of that described by this issue, which doesn’t propose reading the file contents themselves.

**Data freshness**

How frequently we update the database (flat file) here is important, but not crucially so. My gut feeling is that regenerating this file every 12 hours is good enough for the medium-term future. (Important to note that a dataset/narrative can be visualised independent of work in this issue, assuming one knows the URL!) 

Dynamically generated the database server-side each time a client requests the data (i.e. the server calls S3 / GitHub APIs) is not recommended as this would unnecessarily increase the required work server-side. Subsetting a cached file based on user authz per-client-request is probably fine.
 
For sources where we control the S3 bucket (all sources except community), it would be possible to regenerate these files whenever the underlying bucket is updated by using [S3 event notifications](https://docs.aws.amazon.com/AmazonS3/latest/userguide/NotificationHowTo.html). 

[...]

PR #324, opened 2021-05-13
Collect group assets

jameshadfield:
My attempt to solve #306 (for groups only) in a simple manner. @tsibley could you check out the first commit & how it relates to your implementation of sources. See commit messages for more details. Review App will be disabled for now.

### DIFF
--- a/src/getAvailable.js
+++ b/src/getAvailable.js
@@ -2,11 +2,17 @@ const utils = require("./utils");
 const queryString = require("query-string");
 const {splitPrefixIntoParts, joinPartsIntoPrefix, unauthorized} = require("./getDatasetHelpers");
 const {ResourceNotFoundError} = require("./exceptions");
+const metaSources = require("./metaSources");
 
 /* handler for /charon/getAvailable requests */
 const getAvailable = async (req, res) => {
   const prefix = queryString.parse(req.url.split('?')[1]).prefix || "";
   utils.verbose(`getAvailable prefix: "${prefix}"`);
+
+  // `prefix=/groups` is special-cased as it is not backed by a single Source
+  if (prefix.replace(/\//g, '')==="groups") {
+    return res.json(await collectAllAvailableGroups(req.user));
+  }
 
   const {source} = splitPrefixIntoParts(prefix);
 
@@ -47,6 +53,16 @@ const getAvailable = async (req, res) => {
     }))
   });
 };
+
+async function collectAllAvailableGroups(user) {
+  const source = new (metaSources.get("groups"))(user);
+  const datasets = await source.availableDatasets();
+  const narratives = await source.availableNarratives();
+  return {
+    datasets: datasets.map((request) => ({request})),
+    narratives: narratives.map((request) => ({request}))
+  };
+}
 
 module.exports = {
   default: getAvailable
--- a/src/metaSources.js
+++ b/src/metaSources.js
@@ -1,0 +1,86 @@
+const sources = require("./sources");
+const utils = require("./utils");
+
+const PRODUCTION = process.env.NODE_ENV === "production";
+
+/**
+ * A meta-source representing all known groups.
+ * Created using a IIFE to close around state representing the
+ * available datasets & narratives for all groups. This allows
+ * us to update this state on our own timeframe, independent
+ * of API requests.
+ */
+const Groups = (() => {
+
+  const groupSources = new Map( // instances of each group Source
+    Array.from((sources))
+      .filter(([name, Source]) => Source.isGroup())  // eslint-disable-line no-unused-vars
+      // restrict development server's private groups to "blab-private"
+      .filter(([name, Source]) => PRODUCTION ? true : (Source.visibleToUser() || name==="blab-private"))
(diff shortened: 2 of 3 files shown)


src/getDatasetHelpers.js
```diff
@@ -29,6 +30,11 @@ const splitPrefixIntoParts = (prefix) => {
     .replace(/\/$/, '')
     .split("/");
 
+  /* /groups is backed by a meta Source */
```
jameshadfield, line 33: An alternative approach to modifying these functions would be to modify the `getAvailable` handler and special case this API request, e.g.

```js
  if (prefix==="/groups") {
    const source = metaSources.get("groups");
    return res.json({
      datasets: await source.availableDatasets(),
      narratives: await source.availableNarratives()
    });
  }
```

jameshadfield: (I've updated this PR to use this approach)

jameshadfield:
Given @tsibley being away for the next ~10 days, I've switched the implementation here to a more conservative approach whereby the invocation of the class only happens within the `getAvailable` API handler, thus avoiding any unintended effects for our other code paths. @eharkins could you review? Thanks

src/getAvailable.js
```diff
@@ -48,6 +54,16 @@ const getAvailable = async (req, res) => {
   });
 };
 
+async function collectAllAvailableGroups(user) {
+  const source = new (metaSources.get("groups"))(user);
```
eharkins, line 58: Does this mean we create a new instance of the metasource every time `/charon/getAvailable?prefix=/groups` is requested? Would that then restart the scheduling of `updateKnownDatasetsAndNarratives` if the `Groups` IIFE is executed each time?

clarification: I guess the IIFE part only is executed when the expression is parsed at server start time, but my question stands about whether `updateKnownDatasetsAndNarratives` is rerun if we are creating a new instance of the `Groups` metasource for every request.

eharkins (commented):
This makes sense to me and seems to work at least in an immediate sense to serve the groups assets. I guess looking at heroku logs for the dev app will be the way to see if the updating is working?

src/getAvailable.js
jameshadfield: Good question - the IIFE runs only once (when the server starts) and since it's an anonymous function it's impossible to call it again (insofar as as I understand how JS referencing works). This function will initialise `datasetsPerGroup` etc,  call `updateKnownDatasetsAndNarratives` and scheduling it to run every ~6 hours via the `setTimeout`. 

The returned value (`MetaGroupSource`) is closed over the variables `datasetsPerGroup` etc. When we create an instance of this source, as we do each time a `/charon/getAvailable?prefix=/groups` request is handled, the constructor (etc) will run, but we won't run the anonymous function and therefore won't rerun `updateKnownDatasetsAndNarratives` etc.

(As an aside, the `MetaGroupSource` has access to the `updateKnownDatasetsAndNarratives` function, so it _could_ call it, but it doesn't. This is a common way to create private variables.)

I find adding a few `console.log` statements is a good way to test out understanding here (as well as reading docs etc of course!).

src/metaSources.js
```diff
@@ -0,0 +1,86 @@
...
+
+  const datasetsPerGroup = new Map();
+  const narrativesPerGroup = new Map();
+
+  const updateKnownDatasetsAndNarratives = async () => {
+    // following loop is intentionally sequential to reduce load on server
+    for await (const [groupName, source] of groupSources) {
+      datasetsPerGroup.set(
+        groupName,
+        (await source.availableDatasets()).map((p) => `/groups/${groupName}/${p}`)
```
jameshadfield, line 31: These calls should be wrapped into `try/catch` block as stochastic network timeouts, running the server without appropriate credentials etc will will cause `availableDatasets` to throw. Similarly for `availableNarratives` a few lines later.

eharkins: I thought that we wouldn't need to due to https://github.com/nextstrain/nextstrain.org/pull/320?

eharkins: Is the difference here that the async requests for availableDatasets are actually happening outside of the scope of express handlers due to the details of the IIFE you explained above?

jameshadfield: > Is the difference here that the async requests for availableDatasets are actually happening outside of the scope of express handlers due to the details of the IIFE you explained above?

Yup!

jameshadfield: Just tested this to confirm, and since it's happening inside a promise (`updateKnownDatasetsAndNarratives` is an `async` function) which isn't being `await`ed, then any errors are `UnhandledPromiseRejectionWarning`; while these aren't fatal, we should be catching them.

If a `availableDatasets` method fails, we have some options for what to do:
1. If there were previous data set for this source, then keep the previous data. Data listing for certain groups will now be more out-of-date than others (but never more than ~24 hours, as [the Heroku server restarts every day](https://devcenter.heroku.com/articles/dynos#automatic-dyno-restarts)).
2. If there were previous data set for this source, then clear the previous data.

My preference is for (1) here. @eharkins?

jameshadfield: Fixed by @eharkins in 613ed72 👍

PR #326, opened 2021-05-14
Allow server to work without AWS credentials

jameshadfield:
Accessing datasets and narratives in public & private groups relies
on using the AWS-SDK to access objects in the underlying S3 buckets.
The SDK silently loads credentials, if available [1].

If credentials are _not_ supplied, then access to private groups fail
with a suitable warning printed; however calls into public groups,
backed by publicly accessible buckets, simply hung (AWS method
`S3.listObjectsV2`) until the request timed out (`ETIMEDOUT`).
We now check to see if credentials are supplied, and if they are not,
skip the calls into S3 and log a warning instead.

If credentials were supplied, but they were invalid, then a fatal
`InvalidAccessKeyId` error was thrown. We now catch this and print a
warning.

In both cases the Source class method (`_listObjects`) will return an
empty array, allowing our charon API to return valid responses.

[1] https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/setting-credentials-node.html

### DIFF
--- a/src/sources.js
+++ b/src/sources.js
@@ -368,6 +368,7 @@ class S3Source extends Source {
       let contents = [];
       S3.listObjectsV2({Bucket: this.bucket}).eachPage((err, data, done) => {
         if (err) {
+          utils.warn(`Could not list S3 objects for group '${this.name}'\n${err.message}`);
           return reject(err);
         }
         if (data===null) { // no more data
--- a/package-lock.json
+++ b/package-lock.json
@@ -6164,13 +6164,13 @@
       }
     },
     "aws-sdk": {
-      "version": "2.520.0",
-      "resolved": "https://registry.npmjs.org/aws-sdk/-/aws-sdk-2.520.0.tgz",
-      "integrity": "sha512-I0BnCLyoqJ17bW48LYMpBcyhSpMW+c+H0GPnFoB7T8Xi2uype3sNV+4qzmHMVhVYv+2Xba9CMenyiH6qeYSwvQ==",
+      "version": "2.908.0",
+      "resolved": "https://registry.npmjs.org/aws-sdk/-/aws-sdk-2.908.0.tgz",
+      "integrity": "sha512-+UtrKOlwjFhGRRrtf3zl5iwFcAnvuh9m63gBnFj9aA+scbP4K2qOukJxPqXCBDeFPqLGH+ojmMJE/54oSlOfmQ==",
       "requires": {
-        "buffer": "4.9.1",
+        "buffer": "4.9.2",
         "events": "1.1.1",
-        "ieee754": "1.1.8",
+        "ieee754": "1.1.13",
         "jmespath": "0.15.0",
         "querystring": "0.2.0",
         "sax": "1.2.1",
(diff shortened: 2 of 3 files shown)


src/sources.js
```diff
@@ -356,11 +357,16 @@ class S3Source extends Source {
...
+      utils.warn(`Cannot list S3 objects for group '${this.name}' as no AWS credentials supplied.`);
+      return [];
+    }
+    return new Promise((resolve) => {
       let contents = [];
       S3.listObjectsV2({Bucket: this.bucket}).eachPage((err, data, done) => {
         if (err) {
-          return reject(err);
+          utils.warn(`Could not list S3 objects for group '${this.name}'\n${err.message}`);
+          return resolve([]);
```
ivan-aksamentov, line 369: I am not very familiar with the codebase, but am trying to be useful :)

Where will these warnings be printed and who will be reading them in the end?

Shouldn't this and the above check `reject()` the promise, and the error will be handled in the caller function instead? Or, even better, allow the exceptions to propagate all the way to the request handler of the server or the main function of the CLI and handle all the errors there? (depending on how much we want to unwind the stack)

jameshadfield: Always useful comments @ivan-aksamentov ! 

These changes are primarily for development purposes, where one may forget to supply (valid) AWS credentials. The messages will be printed in terminal running the server. We could indeed `reject` here which would, after the recent updates to our handers, result in a 404 or similar response. Would you think this is preferable to the current implementation, which returns 200 but with no datasets/narratives?

P.S. Our "live" instances of the server (nextstrain.org, dev.nextstrain.org, Heroku Review Apps) all have AWS credentials set as config variables.

ivan-aksamentov: Oh, I see. I misunderstood the feature. I thought it's user-facing. If for dev only, then it's fine, I guess.

The advantage of replying with 4xx (and I would carry over the whatever code AWS replies with, probably 403) is that the client can then show a specific message, as opposed to an empty page. Depending on want we want here. My understanding is that we want an empty result in dev, right?

joverlee521 (commented):
> If credentials are _not_ supplied, then access to private groups fail
> with a suitable warning printed; however calls into public groups,
> backed by publicly accessible buckets, simply hung (AWS method
> `S3.listObjectsV2`) until the request timed out (`ETIMEDOUT`).

This struck me as a bug of the aws-sdk. It seems other users had [similar issues](https://github.com/aws/aws-sdk-js/issues/2983) that have been resolved in later versions. We are using version `2.520.0` while the latest version is `2.908.0`, so we could try upgrading the package to resolve this bug. That said, I can see the custom error messaging being useful in development.

src/sources.js
jameshadfield: > My understanding is that we want an empty result in dev, right?

Initially, I was going to make the server die with an error if credentials weren't supplied (at start time), as for nextstrain devs it indicates "you forgot to set the appropriate env!", but that seemed overly nasty for people who have forked the repo etc.

I will rebase this PR upon our new server handling (once the fixes in #327 have gone in) and test what happens if we reject here, as that seems like the most consistent approach.

jameshadfield:
Many thanks @joverlee521 and @ivan-aksamentov, i've implemented both of your suggestions and it's much improved 👍 I've also rebased this onto current master, which includes better handling of errors thrown by our request handlers. Here's the current behaviour:


_r: response code, sMsg: server message, sometimes printed as a warning, sometimes as a caught error_

Credentials    |    Public Group      | Private group w. access |  Private group w/out access | Non-existent group 
    ---        |   ---                | ---                     | ---                         | ---
Not supplied   | sMsg: "Missing credentials". r: `500`) | N/A   | sMsg: "Denying an anonymous user". r: `404` | sMsg: "Not Found", r: `404`
Supplied       | r: `200`             | r: `200`                | sMsg: "Denying an anonymous user". r: `404` | sMsg: "Not Found", r: `404`
Invalid        | sMsg: "The AWS Access Key Id you provided does not exist in our records." r: `403` | sMsg: "Denying an anonymous user". r: `404` | sMsg: "Denying an anonymous user". r: `404` | sMsg: "Not Found", r: `404`

PR #327, opened 2021-05-15
Improve charon API handlers

jameshadfield:
When running this server daa8ff64348920f8a804728808a8a6de1503a2d6 both live (nextstrain.org) and locally, I was experiencing hard-to-reproduce errors when viewing community splash pages (e.g. /community/jameshadfield/scratch). This is related to the recently deployed #320, however it only occurred when the server requested data from GitHub. Hitting GitHub's API limits during testing didn't help either.

Investigating this led to a grab-bag of server improvements - see commit messages for more. 

One big improvement is the integration of smoke-testing in our CI runs, which ~may have~ would have identified this before deployment.

### DIFF
--- a/src/getDataset.js
+++ b/src/getDataset.js
@@ -6,7 +6,7 @@ const helpers = require("./getDatasetHelpers");
 const {NoDatasetPathError} = require("./exceptions");
 const auspice = require("auspice");
 const request = require('request');
-const {NotFound} = require("http-errors");
+const {NotFound, InternalServerError} = require("http-errors");
 
 /**
  *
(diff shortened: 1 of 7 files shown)


jameshadfield:
Currently this PR is incomplete -- streamed narrative responses, when they come from GitHub addresses [1], are not working. The handler (`getNarrative`) is returning `return response.body.pipe(res);` but this doesn't "end" the request/response loop [2], and thus express goes to the next handler, `app.routeAsync("/charon/*").all()`, which is designed to 404 all requests. I'll update this message when that's fixed.

[1] What's different about streaming results from raw.githubusercontent than, say, data.nextstrain.org is unknown to me. Please enlighten me if you know!
[2] Ditto, would be good to know _why_ this happens. I have some hunches, but no concrete understanding.

jameshadfield:
@eharkins @joverlee521 could you review if you have time? thanks

eharkins:
👍 This implementation is easier to read as well as addressing the issues you identified. 

> The handler (getNarrative) is returning return response.body.pipe(res); but this doesn't "end" express handling stack [2]

I'm not sure of this either (even after skimming through https://nodesource.com/blog/understanding-streams-in-nodejs/) but maybe you need to [explicitly end](http://expressjs.com/en/api.html#res.end) the response?

What does https://github.com/nextstrain/nextstrain.org/pull/327/commits/e8c05e62d77b7d71d1944aac8b0929a4918b6b82 (i.e. not using `response.body.pipe(res)`) mean in terms of how loading narratives will perform in practice if anything?

jameshadfield:
> I'm not sure of this either (even after skimming through https://nodesource.com/blog/understanding-streams-in-nodejs/) but maybe you need to explicitly end the response?

My hunch is that the stream does end the response, but that this happens some time in the future. In the meantime, the code continues and we end up in the next middleware/handler. Why this happened consistently when streaming from GitHub and not from other sources is still a mystery to me. In 4f737fb I moved to listening to the "end" event of the server-initiated request, and in e8c05e6 I moved away from piping the response as the narrative file had already finished fetching by this point, so it felt natural to `.send` the data rather than stream it. I don't expect any performance change here. P.S. In 4f737fb I also moved from `res.status(204).end();` to `res.status(204).send();` as the former would fall through to the next handler, which is still surprising to me.

src/getDataset.js
```diff
@@ -24,67 +24,67 @@ const requestCertainFileType = async (res, req, additional, query) => {
...
-  return new Promise((resolve, reject) => {
-    /* try to stream the (v2+) dataset JSON as the response */
-    const req = request
-      .get(main)
-      .on('error', (err) => reject(err))
-      .on("response", async (response) => { // eslint-disable-line consistent-return
-        if (response.statusCode === 200) {
-          utils.verbose(`Successfully streaming ${main}.`);
-          req.pipe(res);
-          return resolve();
```
joverlee521, line 71: I believe the `resolve()` is getting called before the stream has completed. 
I was able to get `community/jameshadfield/scratch/ncov-demo` to load locally by waiting for the stream to end: 
```javascript
const stream = req.pipe(res);
return stream.on('end', () => resolve());
```

src/getNarrative.js
```diff
@@ -52,15 +52,11 @@ const getNarrative = async (req, res) => {
   try {
     utils.log(`Fetching narrative ${fetchURL} and streaming to client for parsing`);
     const response = await fetch(fetchURL);
-    if (response.status === 404) {
-      return res.status(404).send("The requested URL does not exist.");
-    } else if (!(response.status === 200 || response.status === 304)) {
-      throw new Error(`Failed to fetch ${fetchURL}: ${response.status} ${response.statusText}`);
-    }
-    res.set("Content-Type", "text/markdown");
-    return response.body.pipe(res);
```
joverlee521, line 61: Like in `requestMainDataset` this final return needs to be wrapped in a new Promise. I was able to get `community/narratives/jameshadfield/scratch/example-tree-map-zoom` to load locally with: 
```javascript
const stream = response.body.pipe(res);
return new Promise((resolve, reject) => stream.on('close', () => resolve()));
```
I don't fully understand this either....I got to this by going through the [node-fetch docs on streams](https://github.com/node-fetch/node-fetch#streams)

src/getDataset.js
jameshadfield: Yup. That's my understanding too, and the race condition which resulted would explain the stochastic behaviour I initially saw. My solution in this PR also listens to the `'end'` event before resolving - are there any changes you are recommending for this PR?

src/getNarrative.js
jameshadfield: I avoided this by no longer streaming the response (using `.send` instead), as the server had already fetched the entire file before we started the stream. Do you have an approach you'd recommend / changes you'd recommend to my implementation?

joverlee521 (approved):

PR #339, opened 2021-05-25
ErrorBanner component

eharkins:
creates a generally applicable
ErrorBanner component to be passed
as the `banner` prop to GenericPage.
ErrorBanner takes `title` and `contents`
props.

Current examples can be seen on the
individual-group-page.jsx page, e.g.
/groups/not-a-group
/groups/not-a-group/not-a-dataset
/groups/blab/not-a-dataset

### DIFF
--- a/static-site/src/components/splash/errorMessages.jsx
+++ b/static-site/src/components/splash/errorMessages.jsx
@@ -1,29 +1,29 @@
 import React from 'react';
-import { FlexCenter } from "../../layouts/generalComponents";
+import styled from 'styled-components';
 import * as splashStyles from "./styles";
 
+const FixedBanner = styled.div`
+  left: 0px;
+  width: 100%;
+  height: 10%;
+  background-color: #c2c1be;
+  font-size: 18px;
+  padding: 5% 5%;
+  margin: 25px 0px;
+  text-align: center;
+`;
+
 const PleaseEmailUsIfYouBelieveThisToBeAnError = () => (
-  <p>Please <a href="mailto:hello@nextstrain.org">contact us at hello@nextstrain.org </a> if you believe this to be an error.</p>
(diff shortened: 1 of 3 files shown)


eharkins:
Just noting https://www.nytimes.com/404 as a good example of this type of UI pointed out by James.

jameshadfield (changes requested):
There are a few inconsistencies in the component design here, where some error messages use the new UI, and some don't (see https://nextstrain-s-generic-pa-o1n5ib.herokuapp.com/groups/not-a-group for an example of the latter). I'd suggest exposing a component along the lines of `<ErrorBanner>` which can be reused all across the static content, rather than individual components such as `GroupsDatasetNotFound`. This generic error banner could take (e.g.) a title & content prop to facilitate consistency. 

Other UI suggestions:

* If this component were to have a fixed height, we could update the props supplied without causing UI flickering. (For instance, when loading /groups/not-a-group/not-a-dataset, we can immediately display an error banner indicating the dataset doesn't exist, but it takes some time before we can update the banner indicating that the group doesn't exist either!)
* The banner should have more vertical padding above it, across all screen sizes I tested
* I don't think grey is the right colour for the background, but I don't have the perfect colour to suggest at the moment!

eharkins:
Thanks @jameshadfield for the review! I exposed `ErrorBanner` at your suggestion and tweaked the padding a bit and set the height. Regarding the color, I couldn't come up with anything better looking in my opinion than grey. Maybe we don't need to set the background color? Or were you hoping it would be even more contrasting / louder than the grey I chose?

jameshadfield:
@eharkins I'm going to merge this into `next` as-is, ao that when `next` is deployed to the live site the UI for individual groups pages can take advantage of the work here. I'll create a separate issue to discuss the next steps for the error UI started in this PR.

PR #345, opened 2021-06-03
parse /groups/:groupName markdown;

eharkins:
Copies auspice approach:
https://github.com/nextstrain/auspice/blob/master/src/util/parseMarkdown.js
but uses "sanitize-html" library:
https://github.com/apostrophecms/sanitize-html

A couple of config options from the auspice approach
didn't seem to have analogous options in the library I
use here. Also I don't add a hook to set the <a> tags
to open links in a new tab as is done in auspice. There's
probably a way to set this with the "sanitize-html" library
but it wasn't obvious to me and it wasn't working anyway
on nextstrain.org groups pages (which was using the auspice
code to parse markdown until recently for groups pages).

### DIFF
--- a/static-site/src/util/parseMarkdown.js
+++ b/static-site/src/util/parseMarkdown.js
@@ -47,11 +47,17 @@ export const parseMarkdown = (mdString) => {
   const sanitizerConfig = {
     allowedTags,
     allowedAttributes: {"*": allowedAttributes},
-    // These two config settings come from https://github.com/cure53/DOMPurify
-    // and it's not clear how to set them in https://github.com/apostrophecms/sanitize-html
-    // which we are using here.
-    // KEEP_CONTENT: false,
-    // ALLOW_DATA_ATTR: false
+    nonTextTags: ['style', 'script', 'textarea', 'option'],
+    transformTags: {
+      'a': function(tagName, attribs) { // eslint-disable-line
+        const url = new URL(attribs.href); // URL is not supported on Internet Explorer
+        if (url.hostname !== location.hostname) {
+          attribs.target = '_blank';
+          attribs.rel = 'noreferrer nofollow';
+        }
+        return {tagName, attribs};
+      }
+    }
   };
   const rawDescription = marked(mdString);
   const cleanDescription = sanitizeHtml(rawDescription, sanitizerConfig);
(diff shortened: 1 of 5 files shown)


joverlee521:
I'm curious what's the motivation for switching to the "sanitize-html" library?

eharkins:
@joverlee521 sorry, I meant to say. When adding DOMPurify to the package.json, I couldn't get the static-site to build. I think the reason has something to do with https://github.com/cure53/DOMPurify/issues/29#issuecomment-466626162. There seemed to be a solution for that problem in general and someone even packaged it up here https://github.com/kkomelin/isomorphic-dompurify but that didn't work for me either; I ran into this issue https://github.com/kkomelin/isomorphic-dompurify/issues/54 which doesn't seem to have a clear solution (yet).

eharkins:
`sanitize-html` seemed like the best alternative I could find.

joverlee521 (approved):
I left a couple of small notes, but overall looks good.

> sanitize-html seemed like the best alternative I could find.

Thanks for the clarification here, I forget this is now in Gatsby. 

> Also I don't add a hook to set the tags to open links in a new tab as is done in auspice. 

The [transformTags option](https://github.com/apostrophecms/sanitize-html#transformations) might be a good for this? (I think it's definitely just a nice to have)

eharkins:
Thanks @joverlee521! I've tried to address all your feedback in https://github.com/nextstrain/nextstrain.org/pull/345/commits/e7b59e32c920e85502676268169ef37d08e8c481.

joverlee521 (approved):

static-site/src/util/parseMarkdown.js
```diff
@@ -0,0 +1,65 @@
...
+  const sanitizerConfig = {
+    allowedTags,
+    allowedAttributes: {"*": allowedAttributes},
+    nonTextTags: ['style', 'script', 'textarea', 'option'],
+    transformTags: {
+      'a': function(tagName, attribs) { // eslint-disable-line
+        const url = new URL(attribs.href); // URL is not supported on Internet Explorer
+        if (url.hostname !== location.hostname) {
+          attribs.target = '_blank';
+          attribs.rel = 'noreferrer nofollow';
```
eharkins, line 56: @joverlee521 @jameshadfield I just got the following es-lint message in another context:
```
Using target="_blank" without rel="noopener noreferrer" is a security risk: see https://mathiasbynens.github.io/rel-noopenereslintreact/jsx-no-target-blank
```

Should we also have `noopener` here in addition to `noreferrer nofollow`?

PR #350, opened 2021-06-16
Redirect core (GISAID) nCoV URLs

jameshadfield:
Currently all core nCoV builds have used GISAID data, and have been
available at URLS /ncov/global, ncov/asia, etc. The nextstrain/ncov
PR #628 [1] updates our workflow to produce builds for both GISAID and
GenBank data sources, which will result in URLs /ncov/gisaid/:region and
/ncov/genbank/:region, respectively.

Links such as /ncov/:region are heavily used and are assumed to mean
"the latest analysis for :region using GISAID data". This commit
adds redirects so that /ncov/:region now goes to /ncov/gisaid/:region.

Note that this commit should only be merged after [1] has been merged
and a build successfully completed, which will produce the files
s3://nextstrain-data/ncov_{:region}.json.

[1] https://github.com/nextstrain/ncov/pull/628

### DIFF
--- a/redirects.js
+++ b/redirects.js
@@ -24,6 +24,18 @@ const setup = (app) => {
     res.redirect(`/groups${req.originalUrl}`);
   });
 
+  /** prior to June 2021 our core nCoV builds were available at
+   * /ncov/global, ncov/asia etc. These used GISAID data exclusively.
+   * We now have GISAID builds and GenBank builds, and so the URLs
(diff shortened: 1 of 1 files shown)


trvrb:
I think agree not to redirect old date-stamped URLs, eg https://nextstrain.org/ncov/global/2021-06-02, to new `gisaid/` URLs (https://nextstrain.org/ncov/global/2021-06-02) as currently implemented in PR.

PR #357, opened 2021-06-29
use GET request in datasetExists;

eharkins:
this is a temporary fix to allow
existence checking of private group
datasets which previously was failing
due to signed s3 requests not working
with the HEAD action (which was being
used to improve performance of this
existence check request). It is a temporary
fix in place of something that allows
making a HEAD or analogous request for
private group datasets.

See https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1624921294208400?thread_ts=1624913762.204500&cid=C01LCTT7JNN


jameshadfield:
Superseded by #358

PR #359, opened 2021-06-29
remove matchPath for /pathogens and /staging;

eharkins:
we were attempting to prefilter
listings on these pages according
to the * in e.g. /pathogens/* but
this was causing issues with linking
to auspice datasets from these pages.
This is not a high priority feature
and was not advertised yet so this
removes it (for now) to make sure
linking to auspice datasets from these
pages functions correctly.

### DIFF
--- a/static-site/gatsby-node.js
+++ b/static-site/gatsby-node.js
@@ -225,13 +225,11 @@ exports.createPages = ({graphql, actions}) => {
 
         createPage({
           path: "/staging",
-          matchPath: "/staging/*",
           component: path.resolve("src/sections/staging.jsx")
         });
 
         createPage({
           path: "/pathogens",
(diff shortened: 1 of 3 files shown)


eharkins:
Example of the issue is if you go to https://nextstrain.org/staging and click on the first dataset link it takes you to https://nextstrain.org/staging/staging/dengue/all in gatsby instead of https://nextstrain.org/staging/dengue/all in auspice.

jameshadfield (approved):
This is a bug we've run into the past (I remember discussing it with @huddlej & @tsibley, and perhaps others) where Gatsby would handle links client-side rather than letting the server handle them.

Fix looks good, and I think we are moving away from `urlDefinedFilterPath` anyway.

PR #362, opened 2021-06-29
Available staging data reflect bucket contents

jameshadfield:
This moves away from a hardcoded list of staging datasets (via the
manifest on s3://nextstrain-staging), which was out-of-date and
rarely updated. We now crawl the bucket so that all available datasets
are reported.

This introduces a significant speed penalty and the performance of the
server may not scale if many requests are made to list bucket contents.
Subsequent work should add a caching layer to `availableDatasets()`, or
similar.

The second tree options are now empty, which is an improvement on the
previous version which reported the options for the core (not staging)
source. This has the effect that the "Second Tree" dropdown is empty
when viewing staging datasets in Auspice.

### DIFF
--- a/src/setAvailableDatasets.js
+++ b/src/setAvailableDatasets.js
@@ -119,40 +119,24 @@ const convertManifestJsonToAvailableDatasetList = (old) => {
  * SIDE EFFECT: sets global.availableDatasets
  */
 const setAvailableDatasetsFromManifest = async () => {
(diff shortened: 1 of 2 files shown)


eharkins:
Superseded by #367. We may still try to implement a similar approach to this down the road if the benefit of dynamically listing datasets straight from s3 is big enough and we can address issues of performance etc.

Issue #361, opened 2021-06-29
Groups markdown parsing broken


jameshadfield:
**Current Behavior**  

The markdown parsing for certain groups is broken. 

**How to reproduce**  
Steps to reproduce the current behavior:  
1. Markdown parsing is working for https://nextstrain.org/groups/neherlab, https://nextstrain.org/groups/spheres and others
2. Markdown parsing is not working for https://nextstrain.org/groups/swiss/, https://nextstrain.org/groups/blab-private/

**Possible solution**  

~The example I've found is a private group, so this may be auth-related.~ _update: also broken for the `swiss` group, which is public_

Following error is in the (client) console:

```
Error parsing markdown: TypeError: URL constructor: www.viollier.ch is not a valid URL.
```

PR #363, opened 2021-06-29
Improve markdown parsing of URLs

jameshadfield:
Valid-looking URLs which we expect to be commonly used in markdown
(and, indeed, are being used in at least two group overviews) can
result in `new URL()` throwing a `TypeError`. Instead of failing
to parse the markdown, we instead allow the URLs but add the
suitable tags to the `<a>` element.

Closes https://github.com/nextstrain/nextstrain.org/issues/361

### DIFF
--- a/static-site/src/util/parseMarkdown.js
+++ b/static-site/src/util/parseMarkdown.js
@@ -50,8 +50,15 @@ export const parseMarkdown = (mdString) => {
     nonTextTags: ['style', 'script', 'textarea', 'option'],
     transformTags: {
       'a': function(tagName, attribs) { // eslint-disable-line
-        const url = new URL(attribs.href); // URL is not supported on Internet Explorer
-        if (url.hostname !== location.hostname) {
+        try {
+          const url = new URL(attribs.href); // URL is not supported on Internet Explorer
+          if (url.hostname !== location.hostname) {
+            attribs.target = '_blank';
+            attribs.rel = 'noreferrer nofollow';
+          }
+        } catch (err) {
+          // some valid-looking (and commonly used) URLs will cause `new URL` to
+          // throw a TypeError
           attribs.target = '_blank';
           attribs.rel = 'noreferrer nofollow';
         }


static-site/src/util/parseMarkdown.js
```diff
@@ -50,8 +50,15 @@ export const parseMarkdown = (mdString) => {
...
-        const url = new URL(attribs.href); // URL is not supported on Internet Explorer
-        if (url.hostname !== location.hostname) {
+        try {
+          const url = new URL(attribs.href); // URL is not supported on Internet Explorer
+          if (url.hostname !== location.hostname) {
+            attribs.target = '_blank';
+            attribs.rel = 'noreferrer nofollow';
+          }
+        } catch (err) {
+          // some valid-looking (and commonly used) URLs will cause `new URL` to
```
eharkins, line 60: @jameshadfield just for my own curiosity/learning, how did you figure out which URLs caused this? I had trouble logging from this function since its getting run inside the sanitize-html source code somewhere presumably.

eharkins: I guess it's easy enough to run it in a sandbox /separate context with some examples from Nextstrain Groups.

eharkins (approved):

static-site/src/util/parseMarkdown.js
jameshadfield: There were helpful errors printed in the client console -- e.g. see the one in https://github.com/nextstrain/nextstrain.org/issues/361#issue-932134209.

tsibley:
@jameshadfield Hmm, so this lets through `[example](example.com)` or `<a href="example.com">example</a>`, even though the link will be bad because it doesn't contain a scheme (i.e. isn't `https://example.com`). Letting it through is better than erroring! :upside_down_face: 

I wonder if we should treat these URLs as same-origin though for the purposes of `target` and `rel` attributes? Otherwise relative links to other pages on nextstrain.org will be treated like external links, but absolute links to other nextstrain.org pages won't be. To do that, we can use the second arg to `URL`'s constructor:

```
const url = new URL(attribs.href, location.href);
```

eharkins:
@tsibley So if I understand correctly the only thing your suggestion would change (when added to the existing code) is for links without a scheme and relative links, these would be treated as same origin (since we would add the base of https://nextstrain.org and therefore not set `target` and `rel`)? I get wanting to treat relative links to other pages on nextstrain.org as same-origin, but doesn't this also mean for `example.com` (external link without scheme)we would not set `target` and `rel` and so would visit the (bad) link in the same tab?

PR #364, opened 2021-06-29
Fix URL parsing of tanglterees

jameshadfield:
The recent code which checked for dataset existence (in order to route to Auspice or Gatsby) failed to take into account tangletree URL syntax, resulting in the inability to load any tangletrees from a URL. Here we implement a simple fix by only considering the first dataset of a tangle-tree-like URL. Note that Auspice will display a 404-like page if either of the datasets in a tangle-tree URL don't exist, so the solution implemented here is les-than-ideal.

### DIFF
--- a/auspicePaths.js
+++ b/auspicePaths.js
@@ -96,7 +96,8 @@ const isRequestBackedByAuspiceDataset = async (req, res, next) => {
     }
 
     /* Extract dataset information & check if it exists, storing the result as `req.sendToAuspice` */
-    const { dataset } = parsePrefix(req.path);
+    /* (If a tangletree URL is requested, we only consider the first dataset here) */
+    const { dataset } = parsePrefix(req.path.split(":")[0]);
     if (await datasetExists(dataset)) {
       req.sendToAuspice = true;
     }


eharkins (approved):
@jameshadfield this fixes issue in this case so good to merge in terms of fixing something currently not working at all.

Taking a step back, I'm trying to think of other places (e.g. other code bits introduced in 26d241fef70269d5245dfe7f7f40b688c1f9b088 but also more broadly) where tangletree-style paths might cause similar issues. Would it make sense for `parsePrefix` to be have some awareness of the potential for tangletree paths? Not sure what exactly the best behavior in general would be in terms of what it would return.

jameshadfield:
> Taking a step back, I'm trying to think of other places ... where tangletree-style paths might cause similar issues.

Sometime in the past, we shifted the tangle-tree related logic to the Auspice (client) side, which simplified things a lot. As an example, an auspice client loading `/A:B` will make two requests to the server (`/charon/getDataset?prefix=A` & `/charon/getDataset?prefix=B`); this is nice, as it allows `A` & `B` to be different "sources" on the server side. In general,  I think the less the server has to deal with tangle-trees the better. This code (`isRequestBackedByAuspiceDataset`) is a bit of an exception, and I'm not sure what's best here. I think this simple fix will serve us well for most cases. 

P.S. Force-pushed to fix spelling errors in the commit message.

PR #367, opened 2021-07-02
collect staging datasets instead of use manifest;

eharkins:
Currently the /staging page uses a /charon request
to source its dataset listing via the manifest. Here we
change to doing the following instead:
- Collect staging datasets with scripts/collect-datasets.js
  script and publish this list to s3 in a github action
- On /staging we request this listing from s3 to show a more
  frequently updated listing than with the manifest

This strategy is a comprimise between existing attempts:
(1)  using the manifest via a /charon request (current)
(2) crawling the s3 bucket via a /charon request (see https://github.com/nextstrain/nextstrain.org/pull/362)

in that it crawls the s3 bucket but on a regular interval
instead of with each request. Requests are made to a static
listing which is regenerated every 5 minutes (at no cost to the
nextstrain.org server). This should allow us to have caching
benefits that would be more complex to implement in (2) but
improves the freshness/accuracy of the listing compared to (1).

Some notes:
- currently the page is requesting the listing from https://staging.nextstrain.org/datasets_staging.json
  for testing purposes but we'll probably want this on
  data.nextstrain.org for production.
- we aren't getting narratives from the staging bucket.
  There aren't many there but we could do this if desired.
- I tacked the running of the collection script onto an existing
  GH action but there could be benefits to not coupling these udpates.
- I haven't implemented or configured any caching.
  There are many levels at which this can happen, and I'm
  not sure the degree we need for this case.

### DIFF
--- a/.github/workflows/update-flu-catalogue.yml
+++ b/.github/workflows/update-flu-catalogue.yml
@@ -1,4 +1,4 @@
-name: 'update nextstrain.org/influenza with latest flu datasets in s3'
+name: 'update nextstrain.org dataset listing resources on s3'
 
 on:
   schedule:
@@ -14,14 +14,15 @@ jobs:
     - uses: actions/setup-node@v1
       with:
         node-version: 14
-    - name: update flu catalogue
+    - name: update dataset listings # currently just /staging and /influenza use these
       run: |
         npm ci
         python3 -m pip install --upgrade pip setuptools
         python3 -m pip install nextstrain-cli
         PATH="$HOME/.local/bin:$PATH"
-        ./scripts/collect-datasets.js --pathogen flu
-        nextstrain remote upload s3://nextstrain-data data/datasets_influenza.json
+        ./scripts/collect-datasets.js --keyword flu
+        ./scripts/collect-datasets.js --keyword staging
+        nextstrain remote upload s3://nextstrain-data data/datasets_influenza.json data/datasets_staging.json
       env:
         AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
         AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
--- a/scripts/collect-datasets.js
+++ b/scripts/collect-datasets.js
@@ -25,14 +25,14 @@ const parser = new argparse.ArgumentParser({
   also hosted on the ${bucket} bucket.
   `
 });
-parser.addArgument('--pathogen', {action: "store", required: true, metavar: "NAME", help: "what set of datasets to scan, ie flu"});
+parser.addArgument('--keyword', {action: "store", required: true, metavar: "NAME", help: "what set of datasets to scan, ie flu"});
 main({args: parser.parseArgs()});
 
 // -------------------------------------------------------------------------------- //
 
 async function main({args}) {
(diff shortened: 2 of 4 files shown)


eharkins:
Also worth noting is that datasets do not display as e.g. `staging / flu / avian / h5nx / ha` in the listing but rather just `flu / avian / h5nx / ha`. I figured this was ok being on the /staging page, assuming no one would conflate `flu / avian / h5nx / ha` on /staging with the production version on /pathogens.

scripts/collect-datasets.js
```diff
@@ -114,10 +123,15 @@ function filenameLooksLikeDataset(filename) {
   if (!filename.endsWith(".json")) return false;
   if (filename.endsWith("_meta.json") || filename.endsWith("_tree.json")) return false;
   if (filename.endsWith("_frequencies.json")) return false;
+  if (filename.endsWith("_titers.json")) return false;
+  if (filename.endsWith("_frequencies.json")) return false;
```
jameshadfield, line 127: Duplicate of two lines previous

scripts/collect-datasets.js
```diff
@@ -114,10 +123,15 @@ function filenameLooksLikeDataset(filename) {
...
   if (filename.endsWith("_tip-frequencies.json")) return false;
   if (filename.endsWith("_aa-mutation-frequencies.json")) return false;
   if (filename.endsWith("_sequences.json")) return false;
   if (filename.endsWith("_entropy.json")) return false;
   if (filename.endsWith("_root-sequence.json")) return false;
+  if (filename.includes("/")) return false;
```
jameshadfield, line 133: Could you add a comment to this line indicating that S3 objects (files) in folders cannot be accessed by Auspice, and therefore are ignored by this script.

.github/workflows/update-flu-catalogue.yml
```diff
@@ -1,4 +1,4 @@
-name: 'update nextstrain.org/influenza with latest flu datasets in s3'
+name: 'update nextstrain.org dataset listing resources on s3'
 
 on:
   schedule:
```
jameshadfield, line 4: I think as part of this PR we can reduce the frequency of updates to 1 hour here. We should also add a line to the `/staging` page indicating that it may take an hour for updates to show (similar to the message at `/groups`).

jameshadfield (changes requested):
Thanks @eharkins - some minor changes needed but this looks good & will make the staging page immediately useful.

>  currently the page is requesting the listing from https://staging.nextstrain.org/datasets_staging.json  for testing purposes but we'll probably want this on  data.nextstrain.org for production.

Storing on the staging server, alongside the staging data is the right approach I think.

> we aren't getting narratives from the staging bucket.

Narratives are actually stored on the [staging branch of the narratives repo](https://github.com/nextstrain/narratives/tree/staging). As you point out, there aren't many currently, and we can leave their collection for a later date. Could you update the text on the `/staging` page abstract to make it clear that narratives aren't (yet) displayed.

>  I tacked the running of the collection script onto an existing GH action but there could be benefits to not coupling these udpates...  I haven't implemented or configured any caching.

This is fine for now. In the future, as we develop this approach, we can expand the `collect-datasets` script to employ caching.

eharkins:
Thanks @jameshadfield. Tried to address all your comments in 2877b89.

Issue #355, opened 2021-06-28
Inconsistencies in UI when a dataset does not exist


jameshadfield:
Given a nextstrain.org URL where a dataset does not exist, we use many different UI designs to across the site. These should be unified as much as possible. This issue is indended to highlight the different UIs in play, and to prompt  discussion for the desired approach. 

Currently, for URLs not backed by a dataset or narrative, we have the following range of designs (please add more if you find them!). All rendering is via Gatsby, unless noted otherwise. 

* URLs beginning with `/ncov`, '/influenza` or `/flu` are redirected to either `/sars-cov-2` or `/influenza` splash pages. There is no indication that the requested dataset doesn't exist.
    * Examples: nextstrain.org/flu/h3n1, nextstrain.org/ncov/all
* URLs which appear to be "core" (i.e. are not preceded by `/staging`, `/community`, `/groups` etc) are not redirected, they have the main splash page with a 404 banner above it saying "that page doesn't exist".
    * Examples: nextstrain.org/zika/none, nextstrain.org/none
* URLs beginning with `/staging` display the staging splash page. There is no error banner indicating the dataset doesn't exist. If any "parts" of the URL matches staging datasets, the dataset table on the page is filtered by those keywords.
    * Examples: nextstrain.org/staging/none, nextstrain.org/staging/zika/none, nextstrain.org/staging/flu/h5n1/none
* URLs for a group which doesn't exist render a plain page with a grey error banner (from #339) indicating the group doesn't yet exist. A similar UI is employed if we are asking for a dataset within this group.
    * Examples: nextstrain.org/groups/none, /groups/none/none
* URLs for a dataset in a group which exists, have the grey error banner atop the page for that group
    * Example: nextstrain.org/groups/blab/none
* Community URLs without 2 parts (which correspond to the GitHub org & repo) render the main splash page with a 404 banner.
    * Example: nextstrain.org/community/none
* Community URLs where the repo doesn't exist are rendered within Auspice. Instead, these should have an error banner above the main /community splash page.
    * Example: nextstrain.org/community/jameshadfield/none/
* Community URLs where the repo exists, but not the dataset, are currently rendered in Auspice. These should be moved to Gatsby, with an error banner atop a listing of the datasets / narratives available for this source.
    * Example: nextstrain.org/community/jameshadfield/scratch/none
 
    
**Next steps**

[...]

PR #368, opened 2021-07-03
error behavior in gatsby for non-existent dataset urls

eharkins:
these changes allow you to go to
- `/ncov/*`
- `/sars-cov-2/*`
and see the sars-cov-2 page.

If you go to `/ncov/*` it's assumed
you want a dataset but if you land on
the sars-cov-2 gatsby page then that
dataset doesn't exist (otherwise you
would get routed to auspice) so we
display a banner letting you know that
the dataset doesn't exist but don't
wipe the url in case you want to subtly
change it and try again.

If you go to `/sars-cov-2/*`, any value of
`*` is assumed to be intended as a filter
you'd like to apply on the dataset listing.
E.g. `/sars-cov-2/neherlab` filters to
datasets matching the keyword `neherlab`.

This is just one part of a broader attempt
to behave consistently when a dataset
doesn't exist on nextstrain.org.
For more, see https://github.com/nextstrain/nextstrain.org/issues/355


static-site/src/sections/sars-cov-2-page.jsx
```diff
@@ -132,6 +133,21 @@ class Index extends React.Component {
       catalogueDatasets: sarscov2Catalogue.datasets,
       filterParsed: false
     };
+    if (this.props.uri === "/sars-cov-2") {
```
jameshadfield, line 136: [notes from zoom call yesterday] Let's remove the URL-based filtering for the time being, so this added code can be removed, and thus we only need `nonExistentDatasetName = this.props["*"]` set as component state here in the constructor.

static-site/src/sections/sars-cov-2-page.jsx
```diff
@@ -183,8 +217,8 @@ class Index extends React.Component {
                   "FilterDisplay",
                   "ListDatasets"
                 ]}
-                urlDefinedFilterPath={this.props["*"]}
-                intendedUri={this.props.uri}
+                urlDefinedFilterPath={this.state.urlDefinedFilterPath}
```
jameshadfield, line 220: [notes from zoom call yesterday] these props should be removed

static-site/src/sections/sars-cov-2-page.jsx
```diff
@@ -143,9 +159,27 @@ class Index extends React.Component {
     }
   }
 
+  banner() {
+    let bannerTitle, bannerContents;
+    if (this.state.nonExistentDatasetName) {
+      bannerTitle = `The dataset "nextstrain.org/ncov/${this.state.nonExistentDatasetName}" doesn't exist.`;
```
jameshadfield, line 165: [notes from zoom call yesterday] banner message:

The dataset "nextstrain.org${this.props.uri}/${this.state.nonExistentDatasetName}" doesn't seem to exist. Please <a href="https://discussion.nextstrain.org/">post on the discussion board</a> if you believe there has been an error here.

Below is our main SARS-CoV-2 page which lists all current datasets & narratives which we are aware of.

jameshadfield (changes requested):

eharkins:
@jameshadfield 669c953 is my first attempt at what we discussed this week. Let me know what you think!

static-site/src/sections/sars-cov-2-page.jsx
```diff
@@ -135,17 +135,42 @@ class Index extends React.Component {
...
+  }
+
+  banner() {
+    if (this.state.nonExistentDatasetName && (this.state.nonExistentDatasetName.length > 0)) {
+      const bannerTitle = `The dataset "nextstrain.org${this.props.location.pathname}" doesn't exist.`;
+      const bannerContents = `Here is the SARS-CoV-2 page, where we have listed featured datasets,
+      narratives, and resources related to SARS-CoV-2. Note that some SARS-CoV-2
+      datasets may not be listed here. For a more comprehensive list of
+      Nextstrain-maintained (including SARS-CoV-2) datasets,
+      check out ${<a href="/pathogens">nextstrain.org/pathogens</a>}.`;
```
jameshadfield, line 164: This is being rendered as 
```
check out [object Object].
```

eharkins: Thanks @jameshadfield! Should be fixed in 8a2dc68.

jameshadfield (approved):

PR #377, opened 2021-07-12
preserve query params in gisaid builds redirects

eharkins:
Fixes a bug introduced by #350 where we lose query params when redirecting from e.g. https://nextstrain.org/ncov/africa?f_region=Africa and get https://nextstrain.org/ncov/gisaid/africa without the region filter query param.

Needs more testing and review as I'm not an expert in express.js. Also we redirect many other URLs, so might be worth having a look at doing something similar where appropriate among other existing redirects for nextstrain.org.

### DIFF
--- a/redirects.js
+++ b/redirects.js
@@ -1,3 +1,4 @@
+const url = require('url');
 const helpers = require("./src/getDatasetHelpers");
 const { parseNarrativeLanguage } = require("./src/utils");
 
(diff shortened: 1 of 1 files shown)


eharkins:
Seems like most important other places to apply similar logic would be when the URL (possibly) represents a narrative or auspice URL of some sort. Preserving params for static pages on nextstrain.org is presumably lower priority.

trvrb (approved):

Issue #388, opened 2021-07-26
Redesign splash page to highlight other pages


jameshadfield:
**Context**  
The [nextstrain.org](https://nextstrain.org) landing (splash) page consists of a number of sections highlighting different data collections: groups, sars-cov-2, (core) pathogens, community, narratives. Note that "groups" were only displayed when a user was logged in, despite the presence of public groups. 

We now have specific pages for all of these sections except narratives:
- https://nextstrain.org/groups/ (which works for non-authenticated users as well)
- https://nextstrain.org/sars-cov-2/
- https://nextstrain.org/pathogens
- https://nextstrain.org/community/

Those individual pages typically contain more complete data listings than those on the splash page, which was designed in a time when far fewer datasets were shared through nextstrain. Furthermore, there are no links from the splash page to these pages. 

**Description** 

Redesign the splash page to reduce the content in each section and pointing to the respective page. Using similar mocks to #305:

![image](https://user-images.githubusercontent.com/8350992/126939080-88ea238e-2094-4c17-a596-c59f49920f09.png)

- Similar to the layout of the "Philosophy" section, the sections should display as 2 columns when screen size allows. Breakpoints should be the same across these sections. 
- I think section abstracts should be centred, as they currently are (although the Philosophy section is left aligned).
- The number of cards displayed should be explored across different screen sizes. I think it's best to have only one row (maybe two rows on mobile view, or maybe no cards at all on mobile view). We will never provide an exhaustive listing of data with the cards as we once could, the idea is to direct users to the page for that section which does list everything.
    - There's plenty of things to explore with the display of cards - carousels, random selections, etc - but for expediency they should stay hardcoded.
- the Narrative "Find out more" button link is [to the docs](https://nextstrain.github.io/auspice/narratives/introduction)

**Examples**

#305 proposed a similar end result, but with a focus on reusable components. This issue focuses should be on the user-facing changes rather than component design, which can come later.

trvrb:
I back this. I don't have major suggestions, but just a couple small thoughts:
1. I would like to keep an obviously clickable image to bring users directly to https://nextstrain.org/ncov/gisaid/global. Though this could easily be accomplished in above design with two tiles under SARS-CoV-2: one for `ncov/gisaid/global` and the other for `ncov/open/global` and otherwise people clicking on the "see all resources" text.
[...]

Issue #305, opened 2021-04-12
Mocks for pages listing shared datsets & narratives


jameshadfield:
Recent work (including #287 #299 #201 #302) now makes it feasible to create pages listing datasets & narratives shared through nextstrain.org via methods such as community, groups etc. This issue focuses on mocks for the frontend work. Here I don’t consider the problem of collecting this data and keeping it in sync, this issue presumes that such a file is able to be fetched by the client. I also assume that such a file does not include parsing of the dataset file itself, which limits the metadata we are aware of for a given dataset / narrative. 

For most of the sources I believe we can use a single template page, with different content (both text & fetched data file). These would be sourced from the following pages:
* `/pathogens` (`/core` would redirect here) - representing our core datasets. Table columns would be dataset & date updated.
* `/staging` Same table columns as above.
* `/groups` Private groups should also be shown, implying that the datasets must be dynamically generated by our server. Table columns: “Dataset”, “date updated”, “Group name”.
* `/community` Consideration should be taken on whether all datasets are shown, or whether people can opt-out. Table columns: “Dataset”, “date updated”, “GitHub organisation”.

![image](https://user-images.githubusercontent.com/8350992/114330937-f43d7480-9b96-11eb-9a5a-18acb612d57e.png)

We may also wish to have an `/sharing` page, which fetches data from all the above sources. Such a page would imply the others have been implemented!

* `/fetch` would work somewhat differently, but it would be nice to surface a page for this functionality.

![image](https://user-images.githubusercontent.com/8350992/114330940-f8699200-9b96-11eb-9ed1-0dbc17a62414.png)

jameshadfield:
@eharkins's issue #343 on how to surface groups from the nextstrain.org frontpage made me revisit this issue. In general, I think we need a small set (6 or so) of components that we can reuse across all of the pages we are now creating (/groups, /community etc). These components will handle the styling, so that page authors can focus on the content rather than the aesthetics. It also provides continuity across the site.

**Six reusable components that will allow the functionality we want for _most_ pages**

(Colours here identify the components, I'm not proposing we make each component a different colour 😂 )

![image](https://user-images.githubusercontent.com/8350992/120091897-2a44b100-c163-11eb-82ba-c26e23dffb41.png)

**How a selection of pages would look*

One benefit is that the source code for each of these pages would now have a much higher ratio of content to styling. Right now the two are very intertwined.

[...]

PR #390, opened 2021-07-28
splash page redesign; #388;

eharkins:
temp solution for cards is to allow the Cards
component to use flexbox instead of bootstrap
for a more compact, two-column layout. This
will be refactored when we address the design
and flexibility of these components in #305, #347.

TODO:
- might need to unsquash ncov and narratives cards
to be able to read them on mobile
- allowing page sections to tile horizontally,
e.g. groups alongside sars cov 2 as in james' sketch
- more testing

### DIFF
--- a/static-site/src/components/splash/index.jsx
+++ b/static-site/src/components/splash/index.jsx
@@ -107,7 +107,7 @@ class Splash extends React.Component {
               Analyses by independent groups <a href="https://docs.nextstrain.org/en/latest/guides/share/community-builds.html">stored and
               accessed via public GitHub repos</a>
             </>)}
-            cards={communityCards}
+            cards={communityCards.filter((c) => c.frontpage)}
             buttonText="Learn more"
             buttonLink="/community"
           />
--- a/static-site/src/components/splash/index.jsx
+++ b/static-site/src/components/splash/index.jsx
@@ -108,7 +108,7 @@ class Splash extends React.Component {
               accessed via public GitHub repos</a>
             </>)}
             cards={communityCards}
-            buttonText="See featured community analyses"
+            buttonText="Learn more"
             buttonLink="/community"
           />
           <Section
--- a/static-site/src/components/splash/index.jsx
+++ b/static-site/src/components/splash/index.jsx
@@ -12,6 +12,29 @@ import Footer from "../Footer";
 import { createGroupCards } from "./userGroups";
 import { UserContext } from "../../layouts/userDataWrapper";
 
+const Section = ({title, abstract, cards, buttonText, buttonLink}) => (
+  <div className="col-md-6" style={{paddingBottom: "40px"}}>
+    <div style={{display: "flex", flexDirection: "column", alignItems: "center", height: "100%"}}>
+      <Styles.H1>{title}</Styles.H1>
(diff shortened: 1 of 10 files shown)


static-site/src/components/splash/index.jsx
```diff
@@ -54,86 +53,144 @@ class Splash extends React.Component {
...
-
-        <FlexCenter>
-          <Styles.CenteredFocusParagraph>
-            Genomic analyses of specific pathogens kept up-to-date by the Nextstrain team
-          </Styles.CenteredFocusParagraph>
-        </FlexCenter>
+        <div className="row">
+          {/* SARSCOV2 */}
+          <div className="col-md-6">
+            <ScrollableAnchor id={'ncov'}>
```
jameshadfield: We don't need scrollable anchors now that we have individual pages for each, but as they're already here there's no problem to keep them

static-site/src/components/splash/index.jsx
```diff
@@ -54,86 +53,144 @@ class Splash extends React.Component {
...
+              <Cards
+                squashed
+                compactColumns
+                cards={communityCards}
+              />
+            </FlexCenter>
+            <BigSpacer/>
+            <FlexCenter>
+              <Styles.Button to="/community">
+                See featured community analyses
```
jameshadfield: Perhaps "Learn more" is more appropriate, while we don't actually have any featured analyses on that page!

jameshadfield (changes requested):
Looks really good @eharkins. Given that the five sections are identical, I think we'd benefit from making these a component, if only to make styling adjustments simpler in the short term. I played with this approach as I wanted to improve the aesthetics so that the cards & buttons were always lined up horizontally when the sections were side-by-side. I've added this up as a commit, but please modify as necessary. 

Suggested card label changes (for brevity):
* WNV can become "WNV in the Americas"
* DRC Ebola can become "DRC Ebola Sit Rep"
* ncov/open card, replace "public" with "open"

The only non-trivial to-do is the rendering of cards on mobile - the best case here would be to maintain the nextstrain.org behavior, i.e. keep the 2 cards but stack them on top of each other.

static-site/src/components/splash/index.jsx
```diff
@@ -54,86 +53,144 @@ class Splash extends React.Component {
...
+            <BigSpacer/>
+            <FlexCenter>
+              <Styles.Button to="/groups">
+                See all groups
+              </Styles.Button>
+            </FlexCenter>
+          </div>
+        </div>
 
         <HugeSpacer/>
```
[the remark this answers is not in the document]
jameshadfield: (Fixed in 0393cda)

static-site/src/components/splash/index.jsx
```diff
@@ -54,86 +53,144 @@ class Splash extends React.Component {
         </FlexCenter>
 
         <HugeSpacer/>
```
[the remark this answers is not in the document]
jameshadfield: (Fixed in 0393cda)

static-site/src/components/splash/index.jsx
eharkins: resolved in 8919324

jameshadfield (approved):

PR #396, opened 2021-08-26
Add coreCards back to /pathogens;

eharkins:
Perhaps misleading to exclude the two influenza
cards which are on the splash page from this selected
list of pathogens on the /pathogen page? Could imagine
someone clicking "see all pathogens" underneath the
two influenza cards wanting to see a list including those
two influenza cards. My reason for not including them
was the niceness of the 3x3 grid without them.


jameshadfield (approved):
Thanks @eharkins - looks a good improvement

PR #407, opened 2021-09-21
update open build resource links

eharkins:
update https://nextstrain.org/blog/2021-07-08-ncov-open-announcement:
- typo in Emma's name
- fix link to https://docs.nextstrain.org/projects/ncov/en/latest/reference/remote_inputs.html

update to https://nextstrain.org/sars-cov-2/:
- change it to say we update builds "regularly" instead of "daily"
- add link to https://docs.nextstrain.org/projects/ncov/en/latest/reference/remote_inputs.html in the list of resources
- add all the same links to `open` builds as we have for `gisaid` and label them as such - this maybe is getting crowded so could remove regional build links or otherwise thin the text there.


emmahodcroft:
Thanks Eli, this is fantastic. And also brilliant how I didn't catch the typo in my own name 😆

tsibley:
Equivalent or similar changes have been made already by other commits, except for:

- The second instance of a typo in Emma's name, which I've rectified with 943f92e.
- The link to https://docs.nextstrain.org/projects/ncov/en/latest/reference/remote_inputs.html from /sars-cov-2/, which I think we're ok omitting for now since other ncov docs mention it.

Issue #330, opened 2021-05-19
Improve server error logging


jameshadfield:
Improvements in our server handling (e.g. #320, #327) have moved us in a direction where we throw suitable errors to be caught by express' default error handling layer. Each time such an error is caught, for instance when a dataset is not found, the server logs the stack trace such as:

![image](https://user-images.githubusercontent.com/8350992/118756084-41abb080-b8be-11eb-8390-266feb3c26b2.png)

This happens irregardless of what NODE_ENV is set to. (The response content the client receives _does_ change according to this variable, but not the server logging.)

We should only print stack traces if specifically requested, e.g. via a command line arg, as most of the time when developing these traces are more hindrance than help.

jesum44:
Hey @jameshadfield, we are starting to work on this issue and wondering if you could answer a few questions.

We are trying to recreate the server logs that you showed in the example, but are unsure on where to find them. Is there a log file that shows all the logs, or do we need to have server access and the logs will just show up on the console?

Also, should there be a separate command line argument that specifies whether devs should show the server error logging or should we include the functionality as part of the Verbose command line argument that already exists?
Thanks!

jameshadfield:
Hey @jesum44 

> We are trying to recreate the server logs that you showed in the example, but are unsure on where to find them. 

They are printed into your terminal whenever you run the server locally. They won't have identical formatting as the above screenshot, but that doesn't matter. This issue is around [node's development vs production mode](https://nodejs.dev/learn/nodejs-the-difference-between-development-and-production), so the two incantations are:
```sh
NODE_ENV="dev" npm run server --verbose # don't technically need to set NODE_ENV here, but it's clearer
NODE_ENV="production" npm run server --verbose
```
Note that to run in production mode locally you'll have to comment out / remove [this line](https://github.com/nextstrain/nextstrain.org/blob/master/src/app.js#L60) so that the server doesn't try to use https on localhost. 

> should there be a separate command line argument that specifies whether devs should show the server error logging

If `NODE_ENV="production"` then the stack traces should not be shown. Whether they should be toggle-able in "dev" mode is a separate issue, and can be addressed once this is sorted.

Thanks in advance!
james

tsibley:
FWIW, I think this is likely to be subsumed by bigger improvements proposed in https://github.com/nextstrain/nextstrain.org/issues/511 instead.

genehack:
Closing this older issue out in favor of the aforementioned #511.

PR #411, opened 2021-10-01
Grab bag of bug fixes, small improvements, and file reorganization

tsibley:

### Testing
- `npm run test:ci` and `lint` pass locally for me.
- No obvious regressions while clicking around a local instance.


server.js
```diff
@@ -33,139 +21,33 @@ parser.addArgument('--verbose', {action: "storeTrue", help: "verbose server logg
...
-    {},
-    (err) => {
-      // callback runs when the transfer is complete or when an error occurs
-      if (err) res.status(404).sendFile(gatsbyAssetPath("404.html"));
+const port = process.env.PORT || 5000;
+
+/* Import app after setting global.verbose so that calls to utils.verbose()
+ * respect our --verbose option as expected.
+ */
+const app = require("./src/app");
```
jameshadfield, line 29: [no changes requested] Could you comment on the decision to use this form, rather than creating a function in `./src/app.js` which you'd call here along the lines of `const app = setupExpressServer()`? This construction feels strange to me, but it works and it seems like lots of people use it.

jameshadfield (commented):
Most of these seem like solid little improvements, thanks Tom!

Re: error handling. It'd be great to improve our handling of thrown errors (e.g. from `http-errors`) so that they don't result in a stack trace every time (see https://github.com/nextstrain/nextstrain.org/issues/330); perhaps this is on the cards in a future commit?

Re: file naming structure. Could you speak more about your intentions here around naming of endpoints? There was some discussion at https://github.com/nextstrain/docs.nextstrain.org/pull/39, but this is a big topic. The actual reorganisation in this PR (`src/endpoints/charon/`) implies there will be other endpoints 😉 

Finally, it'd be great to rethink `splitPrefixIntoParts` and `parsePrefix` - these functions were written a _looong_ time ago, when the server was far more limited in scope than it has now grown to be.

server.js
tsibley: Sure. IIUC, you're asking why I used

```
const app = require("./src/app");
```

instead of something like:

```
const {setupExpressApp} = require("./src/app");
const app = setupExpressApp();
```

If so, the basic reason is that it was less churn/change in the `server.js` → `src/app.js` move to keep the existing `app` variable and simply export it than to wrap everything in a function, and thus the diff is easier to read.

Functionally, the two forms are ~identical; the difference is the extra level of indirection that the second form provides, which would definitely be warranted if we ever have a need to instantiate multiple copies of the Express app object. Without that need right now, I kept it slightly simpler.

tsibley:
> Re: error handling. It'd be great to improve our handling of thrown errors (e.g. from `http-errors`) so that they don't result in a stack trace every time (see #330); perhaps this is on the cards in a future commit?

Agreed! Thrown errors will enable us to customize this systematically using app-level error handlers. For example, I have an app-level error handler in another (WIP) branch to consistently format errors as JSON for API clients. A similar error handler can be added to log differently than Express' default error handler.

Do you think that's important to implement before merging this PR, or can I get to it in a later PR?

> Re: file naming structure. Could you speak more about your intentions here around naming of endpoints? There was some discussion at [nextstrain/docs.nextstrain.org#39](https://github.com/nextstrain/docs.nextstrain.org/pull/39), but this is a big topic. The actual reorganisation in this PR (`src/endpoints/charon/`) implies that there will be separate namespaces of endpoints :wink:

Here's my basic organizational principles, which I should totally document somewhere for reference and as the basis for commentary/discussion/iteration:

- Routes (e.g. `/charon/getDataset`) all stay in `src/app.js` (at least for now) so we can see all (server-side) routes in one place.
- Endpoints ("route handlers", e.g. `charon.getDataset(req, res, next)`) live in `src/endpoints/`, grouped into arbitrary directories/modules as seems fit.
- The names in route paths and the endpoint source file paths do not necessarily need to match, although they may (as in the case of Charon).
- There does not need to be a 1:1 correspondence between a single route and a single endpoint source file, although there may be (as in the case of Charon). For example, in my other WIP, I have a single file, `src/endpoints/groups.js`, which defines several endpoints used for several routes under `/groups/…`.

Thoughts on where this sort of doc best lives in this repo? A block of source code comments are probably my default in this case, but in `app.js` or elsewhere?

> Finally, it'd be great to rethink `splitPrefixIntoParts` and `parsePrefix` - these functions were written a _looong_ time ago, when the server was far more limited in scope than it has now grown to be.

Yes, absolutely. The WIP I have on another branch (built off this one) has progress towards making those functions obsolete by implementing the prefix handling as real server-side routes instead, as I noted should be done when I wrote [this comment 2 years ago](https://github.com/nextstrain/nextstrain.org/blob/682aa80a1f7f3be878bd44c11c61c155e37243ee/src/utils/prefix.js#L6-L9). :no_mouth: (However, there will still be some need for those functions as long as the Charon API remains the same, unless we can express it (har har) in terms of other API routes by proxying to them.)

jameshadfield:
> Do you think that's important to implement before merging this PR, or can I get to it in a later PR?

Later PR is fine.

> Routes (e.g. /charon/getDataset) all stay in src/app.js (at least for now) 

Thanks - I find this super useful to get an overview of what routes the server expects. One day it may outgrow a single file, but the longer we can keep it as such the better, in my opinion.

> Thoughts on where this sort of doc best lives in this repo?

Thanks for sharing your thoughts here - those organisational principles seem pretty sensible. I'd suggest adding them to https://github.com/nextstrain/nextstrain.org/blob/master/docs/api.md.

Stepping back (e.g. seeing that `nextstrain.org/groups/*` endpoints are on the cards), perhaps a RFC would be a great next step? There was a lot of work behind some of the comments in https://github.com/nextstrain/docs.nextstrain.org/pull/39 which I think we want to keep in mind for any future dev work.

P.S. I want to avoid proxying routes.

jameshadfield:
Fine to merge from my point of view 👍

tsibley:
(Well, I didn't get back to this in time, so will wait until next week.)

PR #420, opened 2021-10-29
Declarative routing

### LINKED ISSUE #294: /users/:name endpoint always responds with current user's information
**Current Behavior**
The `/users/:name` endpoint responds with the current user's information to any request under `/users/**`. For example, you can visit https://nextstrain.org/users/hello/world/how/are/you.

**Expected behavior**  
`/users/:name` only works for the currently logged in username.

In the future, `/users/:name` may return a user's public profile for any valid user, but that is a separate concern.

tsibley:
The top-line change is overhauling the routing to be declarative and more explicit about routes. Some other changes follow from this and a couple others piggyback along. Details in each commit message, and it's probably easiest to review commit-by-commit rather than as one big diff.

This work sets a better foundation for adding more server-side routes, such as the work-in-progress API endpoints which motivated the changes.

### Related issue(s)  
Fixes #294.
Related to the changes in #411, which I split out of this work.

### Testing
`npm run test:ci` and `npm run lint` pass for me. I've done extensive manual testing in the browser (both clicking around and typing in URLs) as well as via `curl` and don't believe there are any behaviour regressions. More 👀 and QA definitely appreciated!


src/app.js
```diff
@@ -91,48 +90,220 @@ app.routeAsync("/charon/*")
...
+    return next(err);
+  }
+
+  if (err instanceof NotFound) {
+    /* A note about routing: if the current URL path (i.e. req.path) matches a
+     * a page known to Gatsby (e.g. via the Gatsby page's "path" or "matchPath"
+     * properties), then Gatsby will perform client-side routing to load that
+     * page even though we're serving a static page (404.html) here.
+     */
+    utils.verbose(`Sending Gatsby 404`);
```
jameshadfield, line 300: The verbose logging is perhaps a little too verbose - this statement will end up followed by another in `sendGatsbyPage`. I'd think that this could be removed.

src/app.js
```diff
@@ -91,48 +90,220 @@ app.routeAsync("/charon/*")
...
+
+/* Everything else gets 404ed.
+ */
+app.use((req, res, next) => next(new NotFound()));
+
+
+/* Error handler
+ */
+app.useAsync(async (err, req, res, next) => {
+  /* XXX TODO: Replace calls to Express' next() with calls to our own custom
```
jameshadfield, line 284: This would be great. At least with these changes we don't have traces for `NotFound` which (anecdotally) is the most common by far.

jameshadfield (changes requested):
It's nice to see expanded use of express middleware which attaches extra information to the request, as `isRequestBackedByAuspiceDataset` did, I really like this pattern. 

Tested on Heroku Review app & locally (`node server.js --verbose`). I didn't review the gatsby dev server changes, but tested it via `npm run dev` and it works as expected on `localhost:5000`. However when running locally _after_ running `npm run dev` [1], `node server.js --verbose` no longer works - I get a `200` response with no content, and no logging from the server. (I can restore behavior via `npm run build` first, but this hasn't been necessary before.)

[1] I'm hesitant to use words like "development" here, as we cannot run production mode on localhost - `NODE_ENV=production node server.js --verbose` - because of our `sslRedirect`.

src/app.js
tsibley: Yeah, definitely want to come back to this. (but wanted to stop bundling up changes in order to get this out the door.)

src/app.js
tsibley: Sure, done.

Relatedly, I think it'd be great to adopt a more structured logging framework so we can include more logging in the code and tune output via config/environ instead of adjusting code. It's impossible to strike the right balance for all situations, so defaulting to "log it" and having knobs is key.

tsibley:
> Tested on Heroku Review app & locally (`node server.js --verbose`). I didn't review the gatsby dev server changes, but tested it via `npm run dev` and it works as expected on `localhost:5000`. However when running locally _after_ running `npm run dev` [1], `node server.js --verbose` no longer works - I get a `200` response with no content, and no logging from the server. (I can restore behavior via `npm run build` first, but this hasn't been necessary before.)

Hmm. I think this is because `gatsby develop` writes files to `static-site/public/` that assume it's presence. I'll dig a little.

> [1] I'm hesitant to use words like "development" here, as we cannot run production mode on localhost - `NODE_ENV=production node server.js --verbose` - because of our `sslRedirect`.

Nod. The SSL issue is surmountable and I'd like to add local SSL support (so we can test the redirects, for instance), but there still remain other reasons of production safety to not support running actual full production mode outside of production (not using production client ids, production external resources like Redis, etc). That said, we can move to making the local server mode (`npm run dev`) more like production in other ways (SSL, local Redis, etc) if we want.

tsibley:
> Hmm. I think this is because `gatsby develop` writes files to `static-site/public/` that assume it's presence. I'll dig a little.

Indeed, it modifies `static-site/public/index.html` to use the `gatsby develop` websocket and other resources. From reading the source, it looks like `gatsby develop` expects to own `public/`, which means a subsequent `gatsby build` (via `npm run build` or `./build.sh static`) will always be necessary to switch back to the static files.

As a workaround, I thought we might be able to instruct `gatsby develop` to put its files elsewhere, but there's no apparent way to do so (which some Googling of GitHub seems to confirm). Instead, our `./develop.sh` could relocate `public/` upon start and restore it before exit, although this might be more confusing if it goes wrong?

> (I can restore behavior via `npm run build` first, but this hasn't been necessary before.)

FWIW, I can reproduce this failure mode on current `master`, so I don't think this is a regression in `./develop.sh`'s behaviour.

jameshadfield (approved):
I should have checked that this wasn't also the case on master (from memory it wasn't, but it's been a while). Another vote to remove Gatsby 😉

PR #422, opened 2021-11-04
Fix community narratives routes

tsibley:


tsibley:
Since this is a hotfix, if CI passes and the test deployment looks good, I'll merge it without waiting on review.

PR #438, opened 2021-11-23
Use node version 14

victorlin:
The currently used node version `13` [has already reached end-of-life status](https://github.com/nodejs/Release#end-of-life-releases) and is preventing auspice npm updates from working on nextstrain.org (#437, https://github.com/nextstrain/auspice/commit/bec41d6ad42cbb500da81d2a240af7f349d7db42)

Choice of `14` is mostly due to consistency as it is already used in GitHub actions:

- [update-dataset-listings](https://github.com/nextstrain/nextstrain.org/blob/04acaaffe766274ded930b6cafc55b75df8c1311/.github/workflows/update-dataset-listings.yml#L16)
- [update-search](https://github.com/nextstrain/nextstrain.org/blob/04acaaffe766274ded930b6cafc55b75df8c1311/.github/workflows/update-search.yml#L22)

Also updating npm version to 6.14.x since it corresponds with node=14 per https://nodejs.org/en/download/releases/.

[Syntax reference for `^14`, `^6.14`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#dependencies)

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
   "repository": "github:nextstrain/nextstrain.org",
   "homepage": "https://nextstrain.org",
   "engines": {
-    "node": "13.6.x",
-    "npm": "6.13.x"
+    "node": "^14",
+    "npm": "^6.14"
   },
   "scripts": {
     "build": "./build.sh",


jameshadfield (approved):
Looks good to me. Here's the build log from the Heroku review app:

```
-----> Installing binaries
       engines.node (package.json):  ^14
       engines.npm (package.json):   ^6.14

       Resolving node version ^14...
       Downloading and installing node 14.18.1...
       Bootstrapping npm ^6.14 (replacing 6.14.15)...
       npm ^6.14 installed
```

jameshadfield:
P.S. Consistency is good, however if there's reasons to use node 16 over node 14 then I'd be be happy to do that across the board here. I know that performance improved in nodejs 16, but whether that would be noticeable for us I'm not sure.

victorlin:
@jameshadfield I think we should definitely move to node 16 at some point, the earlier the better. We can first merge this then have another PR to upgrade package.json + GitHub actions simultaneously.

Also, this is specifically for nextstrain.org. For auspice, we should maintain compatibility with a set of actively supported node versions for more consumption flexibility. That's what I'm hoping to do with nextstrain/auspice#1436.

**Summary**: node versions in package.json and GitHub actions should be consistent per repo.

tsibley:
:tada:

victorlin:
Hmm looks like I missed out on a couple references:

- [.travis.yml](https://github.com/nextstrain/nextstrain.org/blob/d71c4625ab5bdb62b0468a4c15460b240ba8eceb/.travis.yml#L5)
- [static-site/package.json](https://github.com/nextstrain/nextstrain.org/blob/d71c4625ab5bdb62b0468a4c15460b240ba8eceb/static-site/package.json#L14)

I'll make another PR for those.

PR #440, opened 2021-12-04
Explain the design and intent of our source model classes

tsibley:
Prompted by @jameshadfield's comments during review of PR #435.

### DIFF
--- a/src/sources/models.js
+++ b/src/sources/models.js
@@ -85,6 +85,8 @@ const utils = require("../utils");
  *      optimization might be to dynamically choose to redirect a downstream
  *      client to a subresource URL if the upstream source supports CORs
  *      (instead of always proxying the data through us).
+ *
+ *  -trs, Dec 2021
  */
 
 class Source {
--- a/src/sources/models.js
+++ b/src/sources/models.js
@@ -64,6 +64,27 @@ const utils = require("../utils");
  * and URL structure.  Subclasses of these model classes define their specific
  * implementation details and override any base behaviour which doesn't apply
  * to them (ideally limited).
+ *
+ * The advantages of passing around URLs (i.e. passing data by reference)
+ * instead of directly returning the data or even an IO stream include:
+ *
+ *    - You can implement the latter (returning the data or IO streams) with
+ *      the former (URLs), but not vice versa.
+ *
+ *    - URLs and HTTP are a looser coupling between components, which for
+ *      example, makes it easier to change one component later without
+ *      affecting another.
+ *
+ *    - We can re-use the design decisions already made by the HTTP spec for
+ *      how to transmit metadata about some data stream (encoding, content
+ *      type, caching, last modified, etc) so we don't have to re-invent this.
+ *
+ *    - HTTP for the upstream sources aligns with our most common context of
+ *      responding to downstream HTTP clients, and so the opportunities for
+ *      optimization through alignment are greater.  For example, one
+ *      optimization might be to dynamically choose to redirect a downstream
+ *      client to a subresource URL if the upstream source supports CORs
+ *      (instead of always proxying the data through us).
  */
 
 class Source {
--- a/src/sources/models.js
+++ b/src/sources/models.js
@@ -3,14 +3,67 @@ const {fetch} = require("../fetch");
 const {NoResourcePathError} = require("../exceptions");
 const utils = require("../utils");
 
-
-/* These Source, Dataset, and Narrative classes contain information to map an
- * array of dataset/narrative path parts onto a URL.  Source selection and
- * dataset path aliasing (/flu → /flu/seasonal/h3n2/ha/3y) is handled in
- * utils/prefix.parsePrefix().
+/* The model classes here are the base classes for the classes defined in
+ * ./core.js, ./community.js, ./groups.js, etc.
  *
- * The class definitions would be a bit shorter/prettier if we were using Babel
- * to allow class properties on Node.
+ * Sources represent a remote HTTP data source.  Sources contain Datasets and
+ * Narratives (both Resources).  Datasets and Narratives have Subresources
+ * representing specific files/data that make up the conceptual Resource as a
+ * whole.
+ *
+ * Source
+ *   Dataset                (implements Resource interface)
+ *     DatasetSubresource   (implements Subresource interface)
+ *   Narrative              (implements Resource interface)
+ *     NarrativeSubresource (implements Subresource interface)
+ *
+ * Subresources are separate from Resources so that our code can more easily
+ * talk about, address, and pass around values representing both the conceptual
+ * whole of a dataset or narrative and the concrete individual files making up
+ * those wholes.
+ *
+ * A concrete example:
+ *
+ * CoreSource (in ./core.js) represents a Cloudfront distribution
+ * (https://data.nextstrain.org) in front of an S3 bucket
+ * (s3://nextstrain-data).
+ *
+ *   const coreSource = new CoreSource()
+ *
+ * The core dataset "flu/seasonal/h3n2/ha/2y" is represented by a Dataset
+ * instance you get by calling:
+ *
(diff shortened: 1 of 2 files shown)


src/sources/models.js
```diff
@@ -3,14 +3,67 @@ const {fetch} = require("../fetch");
...
-
-/* These Source, Dataset, and Narrative classes contain information to map an
- * array of dataset/narrative path parts onto a URL.  Source selection and
- * dataset path aliasing (/flu → /flu/seasonal/h3n2/ha/3y) is handled in
- * utils/prefix.parsePrefix().
+/* The model classes here are the base classes for the classes defined in
+ * ./core.js, ./community.js, ./groups.js, etc.
+ *
+ * Sources represent a remote HTTP data source.  Sources contain Datasets and
+ * Narratives (both Resources).  Datasets and Narratives have Subresources
```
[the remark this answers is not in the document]
jameshadfield: And from a slightly wider point of view, the current situation instantiates objects each time the endpoint is hit. This is nice as we can check autn each time etc. But there are other use-cases where a cache / stored state is necessary - e.g. we don't want to recompute the available datasets for a particular source each time a request is made. We currently get around this in different ways: using `globals` and a  `MetaGroupSource` which essentially stores the `getAvailable` data for each group. 

Is there a high level way where newly instantiated classes can have access to a cache (or similar) of data? Or could we cache at the level of the request itself (e.g. `fetch(source.available.subresource("???").url())` doesn't actually make a network response but we maintain a periodically updated cache).

src/sources/models.js
```diff
@@ -3,14 +3,67 @@ const {fetch} = require("../fetch");
...
+ *
+ *   const dataset = coreSource.dataset(["flu", "seasonal", "h3n2", "ha", "2y"])
+ *
+ * That Dataset has Subresources identified by the names "main" and
+ * "tip-frequencies", which you get by calling:
+ *
+ *   dataset.subresource("main")
+ *   dataset.subresource("tip-frequencies")
+ *
+ * These Subresources can be retrieved at the following URLs, which you obtain
```
jameshadfield, line 44: Do you foresee these always being HTTP endpoints? I am thinking in particular of things backed by S3 buckets, where we currently use `S3.getSignedUrl`. Unsure if there are speed penalties / limited functionality when we use signed URLs (vs, e.g. a `S3.getObject` call) 

Guaranteeing that it's always HTTP is nice in that we can run `fetch(source.dataset.subresource(type).url())` without knowing anything about how things are stored behind the scenes, but I could imagine another abstraction where we ran `source.dataset.getSubresouce(type)` and it handled how data is fetched.

src/sources/models.js
```diff
@@ -3,14 +3,67 @@ const {fetch} = require("../fetch");
...
+ * (s3://nextstrain-data).
+ *
+ *   const coreSource = new CoreSource()
+ *
+ * The core dataset "flu/seasonal/h3n2/ha/2y" is represented by a Dataset
+ * instance you get by calling:
+ *
+ *   const dataset = coreSource.dataset(["flu", "seasonal", "h3n2", "ha", "2y"])
+ *
+ * That Dataset has Subresources identified by the names "main" and
```
jameshadfield, line 38: I'm interested in bigger picture architecture and it's really nice to see the server evolve from a collection of functions which had to pass around lots of data to middeware which instantiates classes representing the requested data. 

With that in mind, what was the motivation to have a `Subresource` class rather than a method on (e.g.) the Resource object? I know that dataset's are complicated as one dataset can have multiple assets (files) which complicates methods such as `dataset.exists()` - was this part of the motivation?

jameshadfield (commented):
Thanks Tom! Asked questions in-line so that the conversation can be threaded

src/sources/models.js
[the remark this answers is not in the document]
tsibley: Currently the Charon endpoints `getSourceInfo` and `getAvailable` use methods on Source classes which directly return data instead of returning a reference to data (e.g. a URL), which is a design that still makes sense to me. I'm not sure this needs to change anytime soon, but if we had reasons to switch that information to static files for decoupling/caching reasons, then I can imagine using new Resource subclasses for them.

Newly instantiated instances of these classes can still access caches, and there are a number of different options for caching that can be mixed and matched as needed including per-process, in-memory caching like we do now, or external, shared caching in something like Redis, or expanding our network caching (with make-fetch-happen).

src/sources/models.js
```diff
@@ -3,14 +3,67 @@ const {fetch} = require("../fetch");
...
+ * representing specific files/data that make up the conceptual Resource as a
+ * whole.
+ *
+ * Source
+ *   Dataset                (implements Resource interface)
+ *     DatasetSubresource   (implements Subresource interface)
+ *   Narrative              (implements Resource interface)
+ *     NarrativeSubresource (implements Subresource interface)
+ *
+ * Subresources are separate from Resources so that our code can more easily
```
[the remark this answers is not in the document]
tsibley: Happy to field suggestions if you have them, as none of the alternatives I came up with during dev made more sense to me. "Subresource" to me makes sense as "something contained within or part of a Resource", like a subfield of a field of study.

My top contender was just `File`, but I didn't like how that played out in names in the code, the (false) implication that subresources had to be backed by a single file, and the confusion with local files vs `File`s which are actually URLs.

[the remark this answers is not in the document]
tsibley: Also, fortunately, none of that terminology is exposed externally for now, so we can still change it wholesale in the future should we decide on something better later.

src/sources/models.js
tsibley: I think they can likely remain HTTP endpoints ~indefinitely, and there are advantages to doing so. That said, we could add support for other URL schemes/wire protocols in the future if we needed and still handle them pretty transparently within this codebase.

Good to think about limitations of things, but I don't think we have much to worry about with S3 signed URLs. They're basically exposing what S3 clients are doing under the hood normally to make requests. While S3 client APIs present the facade of a "session", behind the facade it's all separately authenticated HTTP requests that look just like signed URLs. So the distinction of using signed URLs vs. not is mostly whether you want to talk plain HTTP or use a client library's API instead. There's a few things which if you need them are a little more ceremony without the benefit of a client library, like multi-part uploads, but nothing that's not doable.

The advantages of URLs over an already-loaded data structure or even an IO stream returned by something like `source.dataset.getSubresouce(type)` are:

- You can implement the latter (loaded data structure or IO streams) with the former (URLs), but not vice versa. It's a more basal concept.
- URLs and HTTP are a looser coupling between components (e.g. easier to change one component later without affecting the other).
- We can re-use the design decisions already made by the HTTP spec for how to transmit metadata about some data stream (i.e. so we don't have to re-invent this)
- The opportunities for optimization are greater, since we're mostly making these HTTP requests to service our own HTTP responses. For example, one optimization might be to dynamically choose to redirect to the subresource URL if the upstream Source supports CORs instead of always proxying the data through us. (I've had this roughly prototyped locally for a bit.)

src/sources/models.js
tsibley: The reason for the Resource and Subresource split is precisely because of the distinction between the conceptual whole ("a dataset" or "a narrative") and the concrete data/files/parts that comprise the whole. The codebase needs to be able to track/query information about both the whole and the parts and also needs to be able to pass representations of them around. I wrote this paragraph to try to cover just that point:

https://github.com/nextstrain/nextstrain.org/blob/85305017e5519ad37ced55c0fdefb7f79d23ede0/src/sources/models.js#L20-L23

The example of `dataset.exists()` is precisely an example of this distinction, but also things like `dataset.resolve()` and, in #443, things like `subresource.media_type`.

src/sources/models.js
tsibley: Added a slightly revised version of that last part to the comment.

tsibley:
I've merged this to stanch its languishing but am happy to address any further review comments/suggestions with a subsequent PR.

PR #442, opened 2021-12-08
Resolve our JSON Schema $id URLs by redirecting to the files on GitHub

tsibley:
We may want to publish these files elsewhere in the future, but this is
a easy place to start with barely any new overhead.

See also https://github.com/nextstrain/augur/pull/806.


jameshadfield (approved):

jameshadfield (changes requested):
Looks good - just requires minor changes to pass linting.

PR #462, opened 2022-02-03
Don't check existence of datasets and narratives for /fetch/…

### LINKED ISSUE #461: Nextstrain.org/fetch broke sometime between yesterday and today
**Current Behavior**  
We use long, AWS pre-signed URLs with nextstrain.org/fetch. 

These were working as of Tuesday 1/31 at ~2pm PST, but then we noticed this morning that they no longer work (we get a blank screen). I've posted an example URL / dataset in the blab slack. 

**Expected behavior**  
A clear and concise description of what you expected to happen instead.  

**How to reproduce**  
Steps to reproduce the current behavior:  
1. Navigate to URL described in this message: https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1643846043841639?thread_ts=1643844470.928079&cid=C7SDVPBLZ

**Possible solution**  
(optional)  

**Your environment: if browsing Nextstrain online**  
Chrome and safari

**Your environment: if running Nextstrain locally**  
 - Operating system:   
 - Browser:  
 - Version (e.g. `auspice 2.7.0`):  

**Additional context**

tsibley:
This fixes a key use case of /fetch where AWS S3 signed URLs are given.

These signed URLs dictate the permitted S3 action, and the same URL
can't support both a HEAD (headObject) and GET (getObject) request.
Prior to the merge and deploy of #443 this morning, the /fetch endpoints
by design didn't guard Auspice's entrypoint with a HEAD request for the
dataset to check existence.  This inadvertently changed in "Refactor
endpoints for GETing a dataset/narrative to make room for additional
media types" (86190b7).

Now once again for /fetch we'll always serve up Auspice, which will make
a request to Charon, which may return an error if the upstream request
fails, and an error from Charon will trigger Auspice's error page.
That's unlike everywhere else where we try to avoid Auspice's error
page, but it's ok.

### Related issues(s)
Resolves #461.

### Testing
Tested locally by generating a signed URL against `nextstrain-data` and watching it no longer make the HEAD request and succeed.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -190,15 +190,11 @@ app.use(["/fetch/narratives/:authority", "/fetch/:authority"],
 
 app.routeAsync("/fetch/narratives/:authority/*")
   .all(setNarrative(req => req.params[0]))
-
-  // Assume existence; little benefit to checking when we don't have a fallback page.
   .getAsync(getNarrative)
 ;
 
 app.routeAsync("/fetch/:authority/*")
   .all(setDataset(req => req.params[0]))
-
-  // Assume existence; little benefit to checking when we don't have a fallback page.
   .getAsync(getDataset)
 ;
 
--- a/src/sources/fetch.js
+++ b/src/sources/fetch.js
@@ -35,6 +35,16 @@ class UrlDefinedDataset extends Dataset {
   subresource(type) {
     return new UrlDefinedDatasetSubresource(this, type);
   }
+  async exists() {
+    /* Assume existence.  There's little benefit to checking with extra
+     * requests when we don't have a natural fallback page (e.g. the Group page
+     * or Community page) and checking means that AWS S3 signed URLs can't be
+     * used with /fetch since they dictate a single action (i.e. can't work for
+     * both HEAD and GET).
+     *   -trs, 2 Feb 2022
+     */
+    return true;
+  }
 }
 
 class UrlDefinedDatasetSubresource extends DatasetSubresource {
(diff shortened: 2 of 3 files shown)


tsibley:
Ah, bother, I didn't update tests to account for this. Will fix in morning.

jameshadfield:
Thanks Tom -- going to deploy this now. Feel free to change the tests in another PR as desired.

There may have also been an (unrelated?) bug where loading those (failing) fetch URLs didn't result in a gatsby 404 page but rather a blank page with console errors. I don't have any URLs around to actually recreate this now.

tsibley:
@jameshadfield Thanks for merging and for that test fix! LGTM.

tsibley:
> There may have also been an (unrelated?) bug where loading those (failing) fetch URLs didn't result in a gatsby 404 page but rather a blank page with console errors. I don't have any URLs around to actually recreate this now.

I believe we'd be able to recreate this by running a local server from before this was merged and accessing `/fetch/…` with an upstream URL that returned an error, e.g. `/fetch/httpbin.org/status/403`.

vincent-czi:
Thanks for the extremely speedy fix on this!

> There may have also been an (unrelated?) bug where loading those (failing) fetch URLs didn't result in a gatsby 404 page but rather a blank page with console errors. I don't have any URLs around to actually recreate this now.
I did encounter this error yesterday where the page was loading as totally blank. I think @tsibley's plan for re-creating the error would be effective, but trying it now with a known-expired pre-signed S3 link is giving me (what I assume is) the expected error page, "There seems to have been an error accessing that dataset.", etc

tsibley:
@vincent-czi Nod. Good to confirm you saw totally blank pages yesterday, which would be unexpected even in the presence of an issue with signed URLs. The error you're seeing now is Auspice's error page which is expected with this PR:

image

but we would expect a different error page (our 404, not from Auspice) with the nextstrain.org codebase just prior to the merge of this PR. I tried that out earlier this afternoon as I'd suggested this morning, but I was unable to reproduce the blank page and got the expected 404 page:

image

Trying again now with an actual S3 signed URL, I am able to reproduce the blank page. It appears to be a client-side issue with our 404 page, related to Gatsby trying to fetch its page data. [console-errors.txt](https://github.com/nextstrain/nextstrain.org/files/7999287/console-errors.txt)

tsibley:
Ah, simply putting a `%3f` (`?`) in the upstream URL to `/fetch/…` is enough to break the our Gatsby 404 page. I can reproduce the blank page and console errors with

```
https://localhost:5000/fetch/httpbin.org/status/403%3f
```

when running eec2250.

PR #478, opened 2022-02-23
Multi-tenant storage for Groups

tsibley:
Many changes leading up to and supporting storing Groups data in a multi-tenant bucket. Details in the commit messages.

There's some background for the choices here in [this design doc](https://docs.google.com/document/d/11iJWtVNmTDLiM8K58S6-HpvZ3ehXxxFSfR1StJ1UUmY/edit?usp=sharing), which I recently pulled together from much earlier notes and slides.

See also the [vastly simplified](https://nextstrain.atlassian.net/wiki/pages/diffpagesbyversion.action?pageId=140542102&selectedPageVersions=8&selectedPageVersions=3) process for [onboarding a new group](https://nextstrain.atlassian.net/wiki/spaces/NEXTSTRAIN/pages/140542102/Onboard+a+new+group).

### Testing
Lots of local testing during dev via clicking around and exercising the API finds no regressions. CI tests pass locally.

### Deploy
- [x] Update AWS IAM policy for nextstrain.org server to allow writes to `nextstrain-groups`


src/groups.js
```diff
@@ -0,0 +1,194 @@
...
+  GROUPS_DATA
+    /* Groups for dev/testing.  Might be nice to expose these in production too,
+     * but we'd need additional changes first to support "unlisted" Groups.
+     *   -trs, 24 Jan 2022
+     * */
+    .filter(groupRecord => PRODUCTION ? !groupRecord.isDevOnly : true)
+    .map(groupRecord => [normalizeGroupName(groupRecord.name), groupRecord])
+);
+
+assert(GROUP_RECORDS.size === GROUPS_DATA.length);
```
joverlee521, line 25: Would this assertion fail in production since we filter out the `isDevOnly` groups?

joverlee521 (changes requested):
This is great! The simplified group creation process is definitely worth all the work!
I left a few comments but the only blocking review I have is the assert check. 

Generally, I'm still a little confused about the separation of `Group` and `GroupSource`, so I want to confirm my mental model is correct: 

A `Group` represents the entity that we are referring to when we talk about groups in Nextstrain. This includes metadata about the group such as the group's name, pubic/private setting, and roles.

A `GroupSource` represents the data source of datasets and narratives managed by a `Group`. This includes the data location (S3 bucket/ S3 object prefix) and the policies/tags associated with the data.

So technically, a `Group` can have more than one `GroupSource`. Do we want to allow this and what would this look like in our new multi-tenant bucket? Do we separate each `GroupSource` with additional prefixes? e.g.:
```
s3://nextstrain-groups/${group_name}/{group_source_1}/...
s3://nextstrain-groups/${group_name}/{group_source_2}/...
```

src/groups.js
tsibley: Oh, it sure would! :grimacing: Nice catch! I'll fix that.

scripts/provision-group
```diff
@@ -0,0 +1,398 @@
...
+
+  const validationErrors = !Array.isArray(members)
+    ? ["Not an array"]
+    : [
+      ...members.filter(m => !(m.username && m.email))
+          .map(m => `Username and/or email missing for member ${JSON.stringify(m)}`),
+
+      ...members.filter(m => m.role && !group.membershipRoles.get(m.role))
+          .map(m => `Unrecognized role "${m.role}" for user ${m.username}`),
+    ];
```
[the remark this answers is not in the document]
tsibley: Good question. My view is to let Cognito handle it. Email validation is notoriously hard to get right. There are lot more valid forms in use than the basic `x@y.z`. We could check `email.includes("@")` with ~zero effort, but that doesn't seem worth it at that point. Cognito will DTRT and is the ultimate arbiter of what we will/can accept anyway.

src/sources/groups.js
```diff
@@ -152,23 +146,23 @@ class GroupSource extends Source {
   async getInfo() {
     try {
       /* attempt to fetch customisable information from S3 bucket */
-      const objects = await this._listObjects();
-      const objectKeys = objects.map((object) => object.Key);
+      const [logoResponse, overviewResponse] = await Promise.all([
+        fetch(await this.urlFor("group-logo.png"), {cache: "no-cache"}),
+        fetch(await this.urlFor("group-overview.md"), {cache: "no-cache"}),
+      ]);
```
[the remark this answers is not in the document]
tsibley: The behaviour here maintains the previous behaviour, which coupled them. That said, I think it'd be reasonable to uncouple them, but some questions arise with error reporting (although the current `error` key looks unsurfaced in the UI?). Related, I also noticed while looking into this that the React component for the group page couples the `/charon/getSourceInfo` request with the `/charon/getAvailable` request. I've created a new issue to track both: https://github.com/nextstrain/nextstrain.org/issues/483

src/sources/groups.js
```diff
@@ -1,10 +1,25 @@
...
-const PRODUCTION = process.env.NODE_ENV === "production";
+
+const GroupSource = (name) => {
+  const group = new Group(name);
+
+  const BaseSource = group.isPublic
+    ? PublicGroupSource
+    : PrivateGroupSource;
+
+  // eslint-disable-next-line no-shadow
```
jameshadfield: [General OOP question] `GroupSource` is a class factory, what advantage does this give as opposed to making it a class itself, so that each group is an object with `name` and `bucket` set in the constructor? We previously used classes for each group, so this implementation maintains this, but I can't remember why we did that now.

src/sources/models.js
```diff
@@ -140,6 +140,12 @@ class Source {
   get authzTagsToPropagate() {
     return new Set();
   }
+  static toString() {
```
jameshadfield, line 133: hah! Did not realise JS had this prototype which you could override. Just like python ;) 

But of course, JS being JS, `console.log(Source)` doesn't work as expected!

test/groupNames.test.js
```diff
@@ -0,0 +1,34 @@
...
+
+describe("validate", () => {
+  const repeat = (str, n) => Array(n+1).join(str);
+
+  /* eslint-disable no-multi-spaces */
+  const names = [
+    ["foo-bar", {ok: true}],
+    ["foo/bar", {ok: false, msg: expect.stringMatching(/disallowed char/)}],
+    ["foobar/", {ok: false, msg: expect.stringMatching(/disallowed char/)}],
+    ["blаb",    {ok: false, msg: expect.stringMatching(/disallowed char/)}],
```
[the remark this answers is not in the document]
tsibley: Repushed with a fix for this.

src/sources/groups.js
jameshadfield: Answered a few commits later in 4eac3516194992ebc92bd7660ebce716ae4d5289 :) 

> What was the
GroupSource function briefly is now a class so that we can properly
introspect it, e.g. with instanceof.  A parameterized class wasn't
possible before removing the static _name property, so public vs.
private had to be separate still and papered over with the previous impl
as a plain function.

src/sources/groups.js
```diff
@@ -144,6 +153,13 @@ class GroupSource extends Source {
    * codebase is that we can iterate on it after pushing live to nextstrain.org
    */
   async getInfo() {
+    const defaults = {
+      title: `"${this.group.name}" Nextstrain group`,
+      byline: `The available datasets and narratives in this group are listed below.`,
+      showDatasets: true,
+      showNarratives: true,
+    };
```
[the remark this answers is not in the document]
tsibley: Good point. The values themselves match the existing fallbacks, but they are applied more granularly now. Done in https://github.com/nextstrain/docs.nextstrain.org/pull/89.

This also prompted me to add 047cab4b7b32859ad93a2bccc37ee8d556bc8529, which moves the byline default to the default overview content.

src/sources/groups.js
```diff
@@ -152,23 +146,23 @@ class GroupSource extends Source {
...
+        fetch(await this.urlFor("group-overview.md"), {cache: "no-cache"}),
+      ]);
 
-      let logoSrc;
-      if (objectKeys.includes("group-logo.png")) {
-        // Use pre-signed URL to allow client to fetch from private S3 bucket
-        logoSrc = S3.getSignedUrl('getObject', {Bucket: this.bucket, Key: "group-logo.png"});
-      }
+      const logoSrc = logoResponse.status === 200
+        ? await asDataUrl(logoResponse)
```
jameshadfield, line 189: This is back to how we used to do it... or at least the original way we were going to do it! Nice to avoid an additional fetch by the client 👍

tsibley:
Your understanding of the separation of responsibilities of `Group` vs. `GroupSource` is correct (with maybe one slight exception).

> So technically, a `Group` can have more than one `GroupSource`. Do we want to allow this and what would this look like in our new multi-tenant bucket? Do we separate each `GroupSource` with additional prefixes? e.g.:

Hmm. The intention is for a 1:1 correspondence between Group and GroupSource. This is why GroupSource is parameterized by Group:

https://github.com/nextstrain/nextstrain.org/blob/047cab4b7b32859ad93a2bccc37ee8d556bc8529/src/sources/groups.js#L21-L23

One way to think about this is that a Group is the conceptual thing (e.g. the Nextstrain Group "blab") while a GroupSource implements, for any group, the common Source interface used internally by our web APIs (Charon endpoints, new RESTful endpoints). A GroupSource is what makes a Group legible to our dataset handlers.

tsibley:
Hmm. Maybe you're wondering about why Group and GroupSource aren't blended into the same class? They could be and then the Group class would be said to _implement_ the Source interface. But IMO this would mix separate concerns instead of keeping them separate. (And also, JS doesn't have a good "X implements Y" pattern, only limited single-parent subclassing.)

jameshadfield (commented):
Reviewed up to 4993d25d15b1cb4deae6561274b79fb500ab9966 and all looks good so far. Feel free to merge & I can review the rest during the testing phase.

src/sources/models.js
tsibley: Yeah, `console.log()` does some "unusual" formatting, despite MDN saying:

> A list of JavaScript objects to output. The string representations of each of these objects are appended together in the order listed and output.

You can force stringification by using the printf-like form of `console.log`:

```
> console.log("%s", Source)
[Source class]
```

PR #479, opened 2022-02-23
authn: Upgrade existing sessions which pre-date addition of authzRoles

tsibley:
Fixes a regression¹ where authz checks throw an assertion error because
authzRoles is missing on the user because many existing sessions lack
it.  This results in 500 errors for much of the web site if you have
such a session.

The user.groups array will also be updated to exclude any authz role
groups.

One downside of this approach is that the changes aren't saved back to
the session, so the existing session must be "upgraded" on the fly for
every request until the user logs out and logs back in (or the session
expires).  We could explicitly force a resave of the session when
needed, but I'm going to leave that for another commit to get a fix out
sooner.

¹ https://github.com/tsibley/blab-standup/blob/master/2022-02-23.md

---

Validated locally. Will merge and deploy once validated on dev.nextstrain.org too.

### DIFF
--- a/src/authn/index.js
+++ b/src/authn/index.js
@@ -130,6 +130,15 @@ function setup(app) {
       user.authzRoles = new Set(user.authzRoles);
     }
 
+    /* Update existing sessions that pre-date the presence of authzRoles.  When
+     * authzRoles is absent, "groups" is the unparsed set of Cognito groups.
+     */
+    if (!("authzRoles" in user)) {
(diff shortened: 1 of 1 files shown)


tsibley:
Fixes a regression introduced by https://github.com/nextstrain/nextstrain.org/pull/452.

jameshadfield:
Awesome! Thanks for the quick fix Tom. 

What was your approach for testing such a fix? Create a localhost session the "old" code, then interact with the new code?

PR #481, opened 2022-03-02
Measurements sidecar

joverlee521:
Adds "measurements" as a subresource type & sidecar file.
I searched for "tip-frequencies" within the repo and added "measurements" to the appropriate places.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -291,6 +291,7 @@ const schemaRoutes = [
   ["/schemas/dataset/v2",              "https://github.com/nextstrain/augur/raw/master/augur/data/schema-export-v2.json"],
   ["/schemas/dataset/root-sequence",   "https://github.com/nextstrain/augur/raw/master/augur/data/schema-export-root-sequence.json"],
   ["/schemas/dataset/tip-frequencies", "https://github.com/nextstrain/augur/raw/master/augur/data/schema-tip-frequencies.json"],
+  ["/schemas/dataset/measurements",    "https://github.com/nextstrain/augur/raw/master/augur/data/schema-measurements.json"],
 ];
 /* eslint-enable no-multi-spaces */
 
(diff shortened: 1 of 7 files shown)


src/app.js
```diff
@@ -291,6 +291,7 @@ const schemaRoutes = [
   ["/schemas/dataset/v2",              "https://github.com/nextstrain/augur/raw/master/augur/data/schema-export-v2.json"],
   ["/schemas/dataset/root-sequence",   "https://github.com/nextstrain/augur/raw/master/augur/data/schema-export-root-sequence.json"],
   ["/schemas/dataset/tip-frequencies", "https://github.com/nextstrain/augur/raw/master/augur/data/schema-tip-frequencies.json"],
+  ["/schemas/dataset/measurements",    "https://github.com/nextstrain/augur/raw/master/augur/data/schema-measurements.json"],
```
jameshadfield, line 294: This github URL doesn't exist (yet). There's https://github.com/nextstrain/augur/pull/827 but it's not yet ready for merge, and it doesn't name the file `schema-measurements.json` either.

jameshadfield (commented):
Tested the charon endpoints via http://localhost:5000/charon/getDataset?prefix=community/joverlee521/nextstrain-testing/flu/seasonal/h1n1pdm/ha/09-17&type=measurements ✅

src/app.js
jameshadfield: Ahhh, I now see your commit comment covers this: "Currently gets redirected to a 404 page, but will soon be added as a JSON schema in augur."

joverlee521:
Rebased onto the auspice release, so now the [testing app can displays the measurements panel](https://nextstrain-s-measuremen-tlovqh.herokuapp.com/community/joverlee521/nextstrain-testing/flu/seasonal/h1n1pdm/ha/09-17).

tsibley (approved):
:+1: FWIW, the first three commits feel like they should be lumped together to me since they're all dealing with the same thing: accounting for a new sidecar type by adding the "measurements" key places. But not a big deal at all and certainly not a must.

Would you also update our [data formats page](https://docs.nextstrain.org/en/latest/reference/data-formats.html)?

PR #493, opened 2022-03-11
Upgrade react select

joverlee521:
The [upgrade of the packages within Auspice](https://github.com/nextstrain/auspice/pull/1482) made the package versions incompatible, leading to failed builds (see https://github.com/nextstrain/nextstrain.org/pull/491).

Most of the code is taken straight from the upgrade in Auspice, with a slight modification of using an additional `AutoSizer` component since the `FilterSelection` component needs to be responsive. 

The last testing commit will be dropped before merging.


jameshadfield (changes requested):
Thanks @joverlee521 - tested and working well. 

I'd suggest we write a comment above `VirtualizedMenuList` (or even abstract it into it's own file) to indicate that this code is a direct duplicate of auspice's [src/components/controls/virtualizedMenuList.js](https://github.com/nextstrain/auspice/pull/1482/files#diff-0b1ceca215f56186dcdf3c9b4e9e4dcd0ef34f0faa674665c5f3437a5c6e5186). In an ideal world this duplication would be replaced by an import, but I think it's much preferable to keep the static-site (gatsby) and auspice siloed for now. 



Proposed next steps:
- [x] add comment / move `VirtualizedMenuList` into separate file
- [ ] drop [c3466a5](https://github.com/nextstrain/nextstrain.org/pull/493/commits/c3466a513ac3e5353b735266bc20f3955927d706)
- [ ] I'll release a new version of auspice now
- [ ] add a commit to this PR upgrading auspice
- [ ] merge this PR (once the test-app passes etc)

Issue #492, opened 2022-03-10
Automatic redirect from main site → canary for users who opt in


tsibley:
Make it easier to use our (upcoming) canary deployment (see #485) for day-to-day work by adding support for setting a flag on users (e.g. for internal team members) which automatically redirects their requests from main → canary (until the person flips the flag back or logs out (or uses incognito mode / different browser)). This lets people choose to use the canary without having to remember to do so every time. Persistence on the session is very easy, but persistence on the user is desired as more convenient/stable. It's only slightly more work.

Approaches for persistence on the user:

1. Use a Cognito group, e.g. `@canary`, `!flag/wants-next`, or something similar, to mark users. This works well with the intentional restrictions we've put on Nextstrain Group names and how they map to Cognito groups. (I've long assumed we'd use Cognito groups for internal things too, like designating nextstrain.org admins at some point.)
2. Store a per-user key in Redis (where sessions are stored) with the flag value, e.g. `flag:canary:user/${userId}` or something.

Both can be easily expanded to include other user flags in the future, like feature flags. Approach (2), however, ends up using lots of keys potentially as it scales and requires another query to Redis for each request.

Approaches for opting in and out:

- Initially this could be manual. That'd be fine.
- But eventually a button + current status on the website somewhere (e.g. on a `/flags` page) could be useful which could make a simple PUT request to set the value and/or POST to toggle for the current user.

Considerations:

- [x] Should this user flag affect non-browser API requests too or just web sessions? I'm inclined to think just web sessions, as there are other ways for an API client to opt-into canary usage (e.g. `NEXTSTRAIN_DOT_ORG` with the Nextstrain CLI). This would mean checking both the flag and the presence of `req.session` before issuing a redirect to the canary.

jameshadfield:
Nice one! 

One consideration is nextstrain team members using nextstrain.org inside talks / presentations. If there is a problem with canary (which there will be, that's kind of the point 😉 ), we want a way to access nextstrain.org (e.g. a different browser). So this might rule out approach (1)?

tsibley:
Ah, both approaches relate to user-level persistence, not session-level, as the former seems more desired from discussion in #485. Definitely agreed that there needs to be a way to switch away from canary. With user-level persistence, that means either a) logging out (several ways to do this) or b) toggling the bit on your user. If your talk requires your logged in user, then (b) is necessary and would definitely be quicker/easier if a button/endpoint existed to do it.

[...]

Issue #485, opened 2022-03-04
Revise deploy process


tsibley:
Our current deploy process looks like this:

- Push to `master` to auto-deploy to nextstrain.org (after CI passes)

I'd like to revise it with a canary (staging) site:

- Push to `master` to auto-deploy to next.nextstrain.org (after CI passes)
- Nothing auto-deploys to nextstrain.org
- After kicking the tires on next.nextstrain.org, [promote](https://devcenter.heroku.com/articles/pipelines#promoting) from that staging phrase of [our Heroku pipeline](https://dashboard.heroku.com/pipelines/38f67fc7-d93c-40c6-a182-501da2f89d9d) to the main site.

The main benefits:

- We always pass through a staging phase where the new version is running on the canary site in the Heroku production environment but doesn't impact the main site yet. This is very useful for catching prod-only bugs.
- Deploys to the main site are very fast, since promotion is basically just a big file copy for Heroku (it re-uses the slug).

Considerations:

- [x] We'll want to make the env/settings of next.nextstrain.org and nextstrain.org basically identical. They should share sessions (like dev.nextstrain.org currently does), despite this being a way that the canary site can impact the main site.
- [x] We'll need to make sure our build process doesn't bake in any env that's different between canary/main, since promotion is just a copy of the same slug.
- [x] Since dev.nextstrain.org has been historically used for ad-hoc testing, we may want to keep that and instead use a new domain for the staging phase. Perhaps canary.nextstrain.org or next.nextstrain.org (my fave :upside_down_face:).

tsibley:
Some [conversation is happening](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1646422166450619) on Slack.

jameshadfield:
Fully supportive of this. Nextstrain's grown a lot since we implemented the current deployment. I've pushed for the wider group to be in the habit of using dev.nextstrain.org day-to-day, but as it is only loosely coupled to the production site this never got much traction. With these changes to deployment I'd hope we can all get in the habit of using the dev site a lot more, otherwise "kicking the tires on dev.nextstrain.org" isn't much different to testing locally or in a review app (auth excepted). 

> Since dev.nextstrain.org has been historically used for ad-hoc testing, we may want to keep that and instead use a new domain for the staging phase.

I'd suggest just this. It's really valuable to have a dev server to play with, even if it's not used for most work.

tsibley:
[...]

PR #500, opened 2022-03-18
Canary deploys and user flag for auto-redirection

tsibley:
Implements #485 and #492.

The former was mostly done on the Heroku side, but is described in updated doc here in https://github.com/nextstrain/nextstrain.org/commit/2e78f0f9123552f5bda1c363d6f99c73cdb9f666.

The latter is https://github.com/nextstrain/nextstrain.org/commit/9ed028ce3214e1d795e24e958c9d5654d1f3b5f3 which makes use of the more general groundwork laid in the earlier commits.

### Testing

- [x] Tested locally by hand in many configurations and with old sessions.
- [x] Test deploy on dev.nextstrain.org with next.nextstrain.org
- [x] Appears to be some funkiness related to interaction with existing host-only cookies and new domain-wide cookies. Need to dig more next week.

### Deploy
- [x] Set `CANARY_ORIGIN` on nextstrain-server and nextstrain-canary
- [ ] Add some folks to `!flags/canary`
- [ ] Remove `CANARY_ORIGIN` from nextstrain-dev after testing this PR

### DIFF
--- a/docs/infrastructure.md
+++ b/docs/infrastructure.md
@@ -20,6 +20,10 @@ Deploys to the production app are performed by manually [promoting](https://devc
 
 ### Environment variables
 
+  - `NODE_ENV` is used to condition behaviour between production and non-production environments, following the widely-used convention set by Express.
+    Its value affects not just the explicit conditionals in this repo but also Express and other layers in our dependencies.
+    All our Heroku apps [run under `NODE_ENV=production`](https://devcenter.heroku.com/articles/nodejs-support#runtime-behavior).
+
   - `SESSION_SECRET` must be set to a long, securely generated string.
     It protects the session data stored in browser cookies.
     Changing this will invalidate all existing sessions and forcibly logout people.
--- a/docs/infrastructure.md
+++ b/docs/infrastructure.md
@@ -2,7 +2,7 @@
 
 ## Domains
 
-**nextstrain.org** and **dev.nextstrain.org** are [hosted on Heroku](#heroku).
+**nextstrain.org**, **next.nextstrain.org**, and **dev.nextstrain.org** are [hosted on Heroku](#heroku).
 
 **data.nextstrain.org** is an AWS CloudFronted S3 bucket, [`nextstrain-data`](#nextstrain-data).
 
@@ -12,8 +12,11 @@
 
 ## Heroku
 
-The production Heroku app is [`nextstrain-server`](https://dashboard.heroku.com/apps/nextstrain-server), which is part of a [Heroku app pipeline of the same name](https://dashboard.heroku.com/pipelines/38f67fc7-d93c-40c6-a182-501da2f89d9d).
-Deploys of `master` happen automatically after Travis CI tests are successful.
+We use a [Heroku pipeline named `nextstrain-server`](https://dashboard.heroku.com/pipelines/38f67fc7-d93c-40c6-a182-501da2f89d9d) to manage multiple related apps.
+The production app serving **nextstrain.org** is [`nextstrain-server`](https://dashboard.heroku.com/apps/nextstrain-server).
+The canary app serving **next.nextstrain.org** is [`nextstrain-canary`](https://dashboard.heroku.com/apps/nextstrain-canary).
+Deploys of `master` to the canary app happen automatically after Travis CI tests are successful.
+Deploys to the production app are performed by manually [promoting](https://devcenter.heroku.com/articles/pipelines#promoting) the canary's current release to production.
 
 ### Environment variables
 
--- a/docs/infrastructure.md
+++ b/docs/infrastructure.md
@@ -41,6 +41,10 @@ Deploys to the production app are performed by manually [promoting](https://devc
     If not provided, requests to the GitHub API are unauthenticated.
     The token we use in production is associated with the `nextstrain-bot` GitHub user.
 
+  - `CANARY_ORIGIN` is the origin of a canary deployment to redirect to for users who are opted-in to the `canary` flag (the `!flags/canary` Cognito group).
+    Redirection does not happen if this variable is not defined or if the request's origin already matches this variable's value.
+    In production we set this to `https://next.nextstrain.org`.
+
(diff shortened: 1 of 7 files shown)


tsibley:
Deploying this to dev.nextstrain.org so we can test it with next.nextstrain.org.

tsibley:
I've [tested](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1647892245229639) promotion from `nextstrain-canary` to `nextstrain-server` on Heroku with [a small change](https://github.com/nextstrain/nextstrain.org/commit/002f9ec0cca05aeff5f3e6725855ae6baa3575c4), and it worked fine. This just tells us that promotion as a mechanism works (as expected) with our app.

src/authn/index.js
```diff
@@ -119,32 +150,64 @@ function setup(app) {
...
+     * In almost all cases, flags will be an empty set.  However, if a user is
+     * added to a flags group (or other internal group) in Cognito *and* then
+     * also establishes a new session in production before this new flags
+     * support (i.e. internal groups support) support is deployed to
+     * production, then that session's user object will be wrong and we'll fix
+     * it here.  Such a wrong user object would have, for example, a "!flags"
+     * Nextstrain Group with role "!flags/canary".  This is likely to only
+     * impact internal team members.
+     */
+    if (!("flags" in user)) update([...user.authzRoles], "missing flags");
```
victorlin, line 205: What's the reasoning behind having `user.groups || []` above and `[...user.authzRoles]` here?

src/app.js
```diff
@@ -69,6 +70,16 @@ app.use(middleware.rejectParentTraversals);
...
+  // Set the app's origin centrally so other handlers can use it
+  //
+  // We can trust the HTTP Host header (req.hostname) because we're always
+  // running behind name-based virtual hosting on Heroku.  A forged Host header
+  // will be rejected by Heroku and never make it to us.
+  req.context.origin = PRODUCTION
+    ? `${req.protocol}://${req.hostname}`
+    : `${req.protocol}://${req.hostname}:${req.app.get("port")}`;
+
   next();
```
victorlin, line 83: Is there a reason for `next()` vs. `return next()` here?

victorlin, line 83: Is there a reason for `next()` vs. `return next()` here?

docs/infrastructure.md
```diff
@@ -38,6 +41,10 @@ Deploys of `master` happen automatically after Travis CI tests are successful.
     If not provided, requests to the GitHub API are unauthenticated.
     The token we use in production is associated with the `nextstrain-bot` GitHub user.
 
+  - `CANARY_ORIGIN` is the origin of a canary deployment to redirect to for users who are opted-in to the `canary` flag (the `!flags/canary` Cognito group).
+    Redirection does not happen if this variable is not defined or if the request's origin already matches this variable's value.
+    In production we set this to `https://next.nextstrain.org`.
+
```
victorlin, line 47: Should `NODE_ENV` also be documented here?

victorlin, line 47: Should `NODE_ENV` also be documented here?

victorlin (approved):
Your commit messages + comments were helpful since I know next to nothing about Cognito and Express. Just a few questions for my own understanding.

src/authn/index.js
tsibley: If you mean why the different syntax, the reason is the function takes an array and `user.groups` is an array already but `user.authzRoles` is a Set, so `[...user.authzRoles]` converts it to an array.

If you mean why the different properties, the reason is that these conditional branches act on different past versions/schemas of the session-persisted user object, so the property which contains the "original Cognito groups array" in each is different (as noted in the comments above each).

src/app.js
tsibley: Nope. It's equivalent in this context as the final statement, so just style. I would probably write `return next()` most times now, but the existing code omitted the `return`.

docs/infrastructure.md
tsibley: This is an Express/Node convention that we didn't set, but we do use it ourselves, so documented!

tsibley:
Before merging this I want to investigate the interaction of the existing host-only cookie with the new domain-wide cookie, as I was able to replicate the weirdness yesterday (unfortunately without pinpointing the precise conditions). Need to dig more to understand what's going on exactly and how to best address it.

tsibley:
Got back to this today and [resolved the duplicate cookie issue with this diff](https://github.com/nextstrain/nextstrain.org/compare/1e4a7ee0689345b1eebac41aeb1fcfd4c82e80be..53767af97fb8cbf26a3a6c34d16f35ce213d1e5a). Notably:

- The bulk of that diff can be removed after ~30 days post-deploy.
- I had to test on "real" domains/subdomains, as browsers (at least Firefox) always seem to treat cookies for `localhost` (or any `*.localhost` domain) as host-only, regardless of the cookie's `Domain`. This caused a lot of false starts and head scratching at first.
- This "use a new cookie name and copy the old cookie to the new one" solution was the simplest I came up with. There were a few alternates, but all were more complex/nuanced/harder to understand and implement.

tsibley:
Latest force-push is rebasing onto master, so I can manually deploy it out-of-order (somewhat ironically) to next.nextstrain.org for testing.

tsibley:
Ok, this is now deployed at both https://dev.nextstrain.org (standing in for https://nextstrain.org) and https://next.nextstrain.org. The two things you should observe if you test are:

1. If you log in on dev.nextstrain.org, your session is portable to next.nextstrain.org, and vice versa.
2. If you're a member of the `!flags/canary` Cognito group (just me atm), then accessing dev.nextstrain.org redirects you over to next.nextstrain.org.

jameshadfield (approved):
✅ logging into canary results in being logged into dev also.
✅ after adding myself to `!flags/canary`, loading dev redirects to canary (didn't even have to login/logout).

Nice one!

tsibley:
> (didn't even have to login/logout).

Hmm, that's unexpected…

PR #512, opened 2022-04-01
Community dataset + narrative listing

### LINKED ISSUE #486: Community datasets to add to community splash page
Copying details from an email request to add the following to the nextstrain.org/community page to better keep track of this request.

1) Hantaan virus (HTNV) L, M, S segments
https://nextstrain.org/community/modoree/htv/L
https://nextstrain.org/community/modoree/htv/M
https://nextstrain.org/community/modoree/htv/S

2) Influenza virus HA and NA
https://nextstrain.org/community/modoree/ifz/ha
https://nextstrain.org/community/modoree/ifz/na

3) Severe fever with thrombocytopenia syndrome virus (SFTSV) L, M, S segments
https://nextstrain.org/community/modoree/sftsv/L
https://nextstrain.org/community/modoree/sftsv/M
https://nextstrain.org/community/modoree/sftsv/S

jameshadfield:
This implements a listing of (known) community pages via a hardcoded
YAML file. This is the first implementation, and there are plenty
of improvements to be made when we systematically address how assets
are collected and displayed.

I am aware of plenty more datasets and narratives than these, however
we want to make this an opt-in listing of resources.

Closes #486


static-site/src/sections/community-page.jsx
```diff
@@ -63,11 +83,48 @@ class Index extends React.Component {
...
+      const d = {...entry};
+      d.url = url;
+      d.githubOrg = d.isNarrative ? urlFields[2] : urlFields[1];
+      d.githubRepo = d.isNarrative ? urlFields[3] : urlFields[2];
+      /* is the entry a dataset or a narrative?
+      NOTE that we cannot currently distinguish if a URL /community/org/repo
+      is a "default" dataset or loads the splash page. In hindsight I wish I hadn't
+      implemented "defaults" but that's what I did.             james 2022 */
+      d.isNarrative = urlFields[1]==="narratives";
+      return d;
```
tsibley, line 127: The URL and path handling here could be considerably more robust¹ and a bit simpler by constructing a [`URL` object](https://developer.mozilla.org/en-US/docs/Web/API/URL) and using pattern matching instead of doing replace/split-based string parsing. For example, something like this:

```js
const communityUrlPattern = new RegExp("^/community(?<narrative>/narratives)?/(?<org>[^/]+)/(?<repo>[^/]+)");

const url = new URL(entry.url);
const urlMatches = url.pathname.matches(communityUrlPattern);

if (!urlMatches) {
  console.warn(`Removing data entry "${entry.name}" as URL <${entry.url}> is not valid`);
  return;
}

return {
  ...entry,
  url: url.pathname,
  isNarrative: !!urlMatches.groups.narrative,
  githubOrg: urlMatches.groups.org,
  githubOrg: urlMatches.groups.repo,
};
```

If that looks better to you too, then I'll push a commit to the branch and test it.

¹ For example, a URL like https://nextstrain.org/community/tsibley/nextstrain.org/… would be handled incorrectly.

tsibley (changes requested):
+1 for this step forward!

One comment on the code below and a couple new visual nits on the Community page.

Flash of unsightly popover on page load:

flash of unsightly popover

Popover missing its background after page load:

popover missing its background after page load

victorlin (changes requested):

static-site/src/sections/community-page.jsx
jameshadfield: > If that looks better to you too, then I'll push a commit to the branch and test it.

👍

static-site/src/sections/community-page.jsx
```diff
@@ -8,8 +8,8 @@ import {
 import * as splashStyles from "../components/splash/styles";
 import GenericPage from "../layouts/generic-page";
 import { ErrorBanner } from "../components/splash/errorMessages";
-import communityCards from "../components/Cards/communityCards";
```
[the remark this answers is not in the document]
jameshadfield: It's still used on the main splash page ;)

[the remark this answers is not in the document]
victorlin: Could the main splash page be updated to use the new YAML as well? Seems like `community-datasets.yaml` already differs from `communityCards.js` so the link from main splash page to `/community` is a bit out of sync.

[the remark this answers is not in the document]
trvrb: We should have a handful of cards for highlighted datasets in addition to the more complete listing below via the datasets UI. This looks a bit like https://nextstrain.org/influenza/ or https://nextstrain.org/pathogens.

trvrb:
Really nice work here @jameshadfield ⭐ 

I think it would be good to assess consistency (always my hangup I know) with other dataset listings pages.

For https://nextstrain.org/pathogens we list:
- "Name" which is actually the dataset slug not the "title" in JSON, ala "enterovirus / d68 / genome", "flu / avian / h5n1 / ha", etc...
- "Type" between "Dataset" and "Narrative"
- "Contributor" 

For https://nextstrain.org/sars-cov-2/ we list:
- "Dataset" which is also the dataset slug as above ala "spheres / ncov / montana"
- "Contributor" which is taken from "maintained by" field in JSON ala "CDC/AMD"

For https://nextstrain.org/groups/ we list:
- "Dataset" which is the dataset slug as above
- "Group Name" which is the URL version of the Group name ala "ngs-sa", "africa-cdc", etc...

I'd think to bring this into more consistency by replacing "name" with "Name" and have this use the URL slug:
- "inrb-drc / ebola-nord-kivu"
- "ViennaRNA / CHIKV"
- "JianiC / rsv / A / G"

And then do "Contributor" the same we do it for the `/sars-cov-2` page:
- "E. Kinganda Lusamaki, INRB & C. Pratt"
- "Thomas Spicher, Markus Delitz, Adriano de Bernardi Schneider and Michael T. Wolfinger"
- "Jiani Chen"

This should probably have built-in ellipses to truncate to ~50 characters or some such.

What do you think? The long form titles are nice to read through, but the URL names were originally chosen because you can do things like filter to `ncov` based on the use of `/` delimiters.

static-site/src/sections/community-page.jsx
tsibley: Pushed [9095e41](https://github.com/nextstrain/nextstrain.org/pull/512/commits/9095e41412d480be14e671c4893c7a61ec933f95).

static-site/src/sections/community-page.jsx
```diff
@@ -63,11 +94,45 @@ class Index extends React.Component {
...
+  const communityUrlPattern = new RegExp("^/community(?<narrative>/narratives)?/(?<org>[^/]+)/(?<repo>[^/]+)");
+
+  return yamlData.data
+    .map((entry) => {
+      const url = new URL(entry.url, "https://nextstrain.org");
+      const urlMatches = url.pathname.match(communityUrlPattern);
+
+      if (!urlMatches) {
+        console.warn(`Removing data entry "${entry.name}" as URL <${entry.url}> is not valid`);
+        return undefined;
```
tsibley, line 122: When testing 9095e41, I noticed that this error-handling here (which I preserved) doesn't make a very loud noise during `gatsby build`. @jameshadfield Thoughts on throwing an exception here instead? If we were requesting the YAML file on every page load, then a warning + skip on the client would make sense to me as the YAML would change more independently, but since this code runs at static site build time, an exception will help ensure we keep the data correct.

jameshadfield: The code runs both at compile time and on the client each page load. At least, it does in my testing, and this is how I understand Gatsby to work. 

I _think_ you could condition on this via something like 👇  if you want to pursue this?

```js
if (typeof window === 'undefined') {
  throw new Error(...)
} else {
  console.warn(...);
}
```

tsibley: Ah, it does appear to run at both build and run time, despite the latter being completely unnecessary. I thought it was supposed to be compiled away given that it's all static and cannot change.

In any case, I'm not going to pursue this more. Will hopefully just ditch Gatsby sooner than later.

joverlee521 (changes requested):
I think there needs to be some slight tweaks in the css for `ListDatasets`. The long GitHub repos wrap if there are hyphens in the name, but overflow out of the row container when there are no hyphens:

Screen Shot 2022-04-07 at 4 36 41 PM

static-site/src/sections/community-page.jsx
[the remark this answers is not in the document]
jameshadfield: I've left these out of /community for now (they're still on the main splash page). I'd like to revisit this when we have more routing updating of metadata associated with each asset, and make the cards show (e.g.) the most recently updated 6 datasets.

I have moved the information about cards (as used on the splash page) into the main YAML & thus removed `communityCards.js`

jameshadfield:
Thanks for all the review comments!

Fixes made:
- [x] Improved consistency in column names & content. (Looking across the different tables, there is a lot of variation... I'm not going to modify other tables as this is part of a bigger push which includes collecting more detailed metadata about assets. See [this google doc](https://docs.google.com/document/d/1CvE05kzGlys0sDYoT8RUKl-DYp7xJ__04skrKiaafbk/edit) for more.)
- [x] CSS in our tables fixed so that the 2nd + 3rd columns are truncated with ellipses, and improved the case where only 2 columns were used. 
- [x] Added the maintainer/author information for each dataset via a helper script. As a bonus, I added a bunch more metadata but because we're not running the script automatically I chose not to display the data (e.g. updated date) as it'll fall out of date.
- [x] Shifted the community card information into the main community YAML file.

joverlee521 (commented):
I found a few datasets that no longer exist: 
- https://nextstrain.org/community/czbiohub/covidtracker/ca
- https://nextstrain.org/community/grubaughlab/CT-SARS-CoV-2/current
- https://nextstrain.org/community/alghoribi-lab/ncov/gcc

Is this something we need or want to address? There's no way for us to control this since users can just remove their data from GItHub at anytime, but we can do our best to catch these in `scripts/parse-community-data.js` and exclude them from the table.

jameshadfield:
Thanks @joverlee521 -- agree that we should remove these (done in efad614).

Unless there are objections I'm going to merge this into canary 🐦 tomorrow

victorlin (approved):

PR #514, opened 2022-04-08
Include "nextclade" as top-level path for Auspice

trvrb:
Including this allows posting nextclade reference trees to https://nextstrain.org/nextclade/sars-cov-2, etc...

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -157,6 +157,7 @@ const coreBuildPaths = [
   "/mers",
   "/mumps",
   "/ncov",
+  "/nextclade",
   "/tb",
   "/WNV",
   "/yellow-fever",


tsibley (approved):
Looks fine to me! I double checked that we don't have any redirects of `/nextclade…` in `src/redirects.js` which might preclude this from working, and we don't.

Please

- [x] test this in the PR review app
- [x] then merge and test on https://next.nextstrain.org after auto-deploy to there
- [ ] and then when satisfied, promote the canary to production and test it works there.

For reference, see our [brief infra doc section on deploying](https://docs.nextstrain.org/projects/nextstrain-dot-org/en/latest/infrastructure.html#heroku).

trvrb:
Thanks for outlining the steps here Tom. I'll work through them.

PR #517, opened 2022-04-15
Underscore

### LINKED ISSUE #432: URLs allow underscores in place of slashes in dataset paths
Some dataset (and presumably narrative) routes allow using underscores instead of slashes in the URL path, e.g. <https://nextstrain.org/groups/blab/ncov_early-outbreak_root-A> instead of <https://nextstrain.org/groups/blab/ncov/early-outbreak/root-A>.

Underscores work because of an implementation detail about how we're backing those files that is leaking through. The presence of underscores breaks many assumptions made by nextstrain.org and Auspice, a minor example of which is Auspice's dataset selector. Other broken assumptions could lead to more serious breakage, such as authz bugs (though I don't think this is the case currently).

We should prevent the unexpected behaviour of underscores working in order to make the system consistent to reason about.

**Possible solutions**  
- Change route patterns to exclude `_` (e.g. `:dataset([^_]*)` instead of `*`)
- Change Source/Dataset/etc model classes to reject paths with underscores

skorman:

This pull request addresses issue #432, to send an error when a user tries to request a dataset containing underscores. It also adds a test suite to test this issue.

### Related issue(s)

Fixes #432
Related to #

### Testing

To run the test, start the server and then separately do:

`npx jest --run-tests-by-path test/underscore.test.js`

Alternatively, these tests can be performed by running the entire test suite:

`npm run test:ci`

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -61,7 +61,7 @@ app.use(sslRedirect()); // redirect HTTP to HTTPS
 app.use(compression()); // send files (e.g. res.json()) using compression (if possible)
 app.use(nakedRedirect({reverse: true})); // redirect www.nextstrain.org to nextstrain.org
 app.use(middleware.rejectParentTraversals);
-
+app.use(middleware.rejectRoutesWithUnderscores);
 
 /* Setup a request-scoped context object for passing arbitrary request-local
  * data between route and param middleware and route handlers.  Akin to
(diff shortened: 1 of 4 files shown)


tsibley:
Thanks for this attempt @skorman and @schneid44. It's a start in the right direction—you identified the right spot in `Resource` to put the check—but wasn't suitable as-is (e.g. `/fetch/…` URLs need to allow underscores). Apologies I wasn't able to work with you on fixing it up at the time. I wrote a similar fix this afternoon as #611.

PR #521, opened 2022-04-26
Replace Travis CI with GitHub Actions

victorlin:

Context: for consistency, we are migrating .travis.yml to a GitHub Actions workflow.

- Use `aws-actions/configure-aws-credentials@v1` to authenticate as `user/nextstrain.org.dev`, necessary for `npm run test:ci`.
- Use `DEV_SERVER_*` Actions secrets to distinguish from existing secrets used in other workflows.

### Related issue(s)

_N/A_

### Testing

See PR checks which reflects this change.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,26 @@
+on:
+  push:
+    branches:
+      - master
+      - release
+  pull_request:
+name: ci
+jobs:
+  run:
```
tsibley, line 9: Maybe call this job `test` since it runs the CI tests (as opposed to the CI deploy, for example)?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,26 @@
+on:
+  push:
+    branches:
+      - master
+      - release
```
tsibley, line 5: Why just these branches? Also, there's no `release` branch. :upside_down_face: The current Travis setup runs on all branches + all PRs, and it seems like this should do the same...

tsibley (changes requested):

.github/workflows/ci.yml
victorlin: Whoops, this was copy/paste from another workflow. Fixed to be consistent with https://github.com/nextstrain/augur/blob/master/.github/workflows/ci.yaml.

.github/workflows/ci.yml
victorlin: Done.

jameshadfield (approved):
Looks good Victor. Really nice to have only 1 CI platform per repo 😄 

`aws sts get-caller-identity` leaks the (AWS) `UserId` into the logs, but I think only we can see those (and I don't think it's considered private anyways).

tsibley (approved):

PR #523, opened 2022-04-27
Switch Nextstrain CLI setup in the GitHub Actions workflows to our new shared action

tsibley:
Resolves issues with system Python described in
<https://github.com/nextstrain/cli/issues/171>, but also makes future
maintenance easier by standardizing and centralizing how `nextstrain` is
installed.

Before merging:

- [x] Merge https://github.com/nextstrain/.github/pull/17
- [ ] Drop tip `tmp!` commit on this branch

### DIFF
--- a/.github/workflows/update-dataset-listings.yml
+++ b/.github/workflows/update-dataset-listings.yml
@@ -14,12 +14,10 @@ jobs:
     - uses: actions/setup-node@v1
       with:
         node-version: 14
+    - uses: nextstrain/.github/actions/setup-nextstrain-cli@master
     - name: update dataset listings # currently just /staging and /influenza use these
       run: |
         npm ci
-        python3 -m pip install --upgrade pip setuptools
-        python3 -m pip install nextstrain-cli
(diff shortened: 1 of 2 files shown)


jameshadfield:
scheduled actions run on the [last commit on default branch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#schedule) so may want to add something like the following to test?

```yaml
on:
  push:
    branches:
      - 'trs/github-actions-workflow-setup'
```

tsibley:
@jameshadfield Yeah, good call. :-) I'd thought I could manually trigger them for this branch, but after pushing I realized they didn't have `on: workflow_dispatch`. I've added the `on: push` condition you suggested to see that they work here.

jameshadfield (approved):
[They work](https://github.com/nextstrain/nextstrain.org/actions/runs/2235800923) :) 

I'll let you merge after dropping the `!tmp` commit & after merging the appropriate nextstrain/.github PR. Thanks!

PR #527, opened 2022-04-29
Update GitHub Actions

victorlin:

- Prevent Actions from running on forks - this is because I kept getting failures over at https://github.com/victorlin/nextstrain.org/actions
- Update other Actions usages for consistency.

### Related issue(s)

_N/A_

### Testing

_N/A_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,6 +2,7 @@ name: CI
 on: [push, pull_request]
 jobs:
   test:
+    if: github.repository == 'nextstrain/nextstrain.org'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-node@v2
+      - uses: actions/setup-node@v3
         with:
           node-version: '14'
       - run: node --version
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,9 @@ jobs:
         with:
           node-version: '14'
       - run: node --version
+      - run: npm ci
+      - run: npm run build
+      - run: npm run lint
       # configure AWS to run dev server necessary for `npm run test:ci`
       - uses: aws-actions/configure-aws-credentials@v1
         with:
@@ -17,7 +20,4 @@ jobs:
           aws-access-key-id: ${{ secrets.DEV_SERVER_AWS_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.DEV_SERVER_AWS_SECRET_ACCESS_KEY }}
       - run: aws sts get-caller-identity
-      - run: npm ci
-      - run: npm run build
-      - run: npm run lint
       - run: npm run test:ci
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ jobs:
     if: github.repository == 'nextstrain/nextstrain.org'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: actions/setup-node@v2
         with:
           node-version: '14'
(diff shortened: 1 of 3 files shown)


.github/workflows/update-search.yml
```diff
@@ -21,13 +21,15 @@ jobs:
...
     - uses: nextstrain/.github/actions/setup-nextstrain-cli@master
     - name: update sars-cov-2 search
       run: |
         npm ci
         ./scripts/collect-search-results.js --pathogen sars-cov-2
         nextstrain remote upload s3://nextstrain-data ./data/search_sars-cov-2.json
-      env:
-        AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
-        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
tsibley, line 32: This change (and the corresponding one in the other workflow) increases the exposure of the AWS creds to more steps of the workflow. In particular, it exposes them to steps which install third-party software (via `setup-nextstrain-cli`), although they were already exposed to `npm ci`. I'd suggest splitting the steps and re-ordering so that setup/install stuff happens before the AWS creds are setup. More generally, I think that unlike `setup-node` and `setup-python` which we want to come early in a job, the AWS creds setup should come as late as possible.

tsibley (changes requested):
Good tidy ups!

I note that the latest versions of actions/checkout and setup-node are v3 but this only bumps to v2. Seems to me that we should just update both to v3? I don't think there's any backwards compat concerns.

.github/workflows/update-search.yml
victorlin: @tsibley good point, I've moved them.

.github/workflows/ci.yml
```diff
@@ -2,21 +2,22 @@ name: CI
 on: [push, pull_request]
 jobs:
   test:
+    if: github.repository == 'nextstrain/nextstrain.org'
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-node@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
```
victorlin, line 9: > @tsibley Seems to me that we should just update both to v3?

Done

.github/workflows/update-dataset-listings.yml
```diff
@@ -8,21 +8,25 @@ on:
...
+    # configure AWS necessary for `nextstrain remote upload`
+    - uses: aws-actions/configure-aws-credentials@v1
+      with:
+        aws-region: ${{ secrets.AWS_DEFAULT_REGION }}
+        aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+    - run: aws sts get-caller-identity
     - name: update dataset listings # currently just /staging and /influenza use these
       run: |
         npm ci
```
tsibley, line 28: Move `npm ci` up before `configure-aws-creds` here too?

.github/workflows/update-search.yml
```diff
@@ -14,19 +14,23 @@ on:
...
+    # configure AWS necessary for `nextstrain remote upload`
+    - uses: aws-actions/configure-aws-credentials@v1
+      with:
+        aws-region: ${{ secrets.AWS_DEFAULT_REGION }}
+        aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+    - run: aws sts get-caller-identity
     - name: update sars-cov-2 search
       run: |
         npm ci
```
tsibley, line 34: Ditto.

tsibley (changes requested):

.github/workflows/update-dataset-listings.yml
victorlin: Oh right, I didn't read your first comment closely enough and I see this was suggested. Done.

.github/workflows/update-search.yml
victorlin: Done.

tsibley (approved):

PR #541, opened 2022-05-16
Gracefully handle session deserialization errors

tsibley:
See individual commits for details. The primary goal here is to make session deserialization issues less fatal so that a) bugs we introduce don't have as much of an impact on users (e.g. they can still use public resources without having to wipe cookies or switch browsers) and that b) older versions of this codebase can co-exist with sessions created by newer versions. This last point is important, as I want to retain the ability to rollback from session-changing deploys gone wrong without new sessions poisoning the rollback.

To that end, I will rebase #537 on top of this and re-push it. My plan is then to deploy this first and then deploy #537.

### Testing
- Manually tested various deserialization error states (missing keys hitting the new asserts, not JSON at all, etc).
- Logged in and out via OAuth2
- Logged in via Bearer auth

### DIFF
--- a/src/authn/index.js
+++ b/src/authn/index.js
@@ -361,7 +361,9 @@ function setup(app) {
     (req, res) => {
       // We can trust this value from the session because we are the only ones
       // in control of it.
-      res.redirect(req.session.afterLoginReturnTo || "/");
+      const afterLoginReturnTo = req.session.afterLoginReturnTo;
+      delete req.session.afterLoginReturnTo;
+      res.redirect(afterLoginReturnTo || "/");
     }
   );
 
--- a/src/authn/index.js
+++ b/src/authn/index.js
@@ -407,7 +407,10 @@ function authnWithToken(req, res, next) {
  */
 function authnWithSession(req, res, next) {
   const authenticate = passport.authenticate(STRATEGY_SESSION, (err, user) => {
-    if (err) return next(err);
+    if (err) {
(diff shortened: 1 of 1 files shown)


joverlee521 (approved):
> "and then Passport does some stuff" hand waving.

I always thought this was the appeal of using Passport.js 😅 

I can't actually find details on what Passport does in `authenticate` from the [docs](https://www.passportjs.org/docs/). I assume you had to go through the source code to figure it out?

tsibley:
> I always thought this was the appeal of using Passport.js sweat_smile

Magic is always appealing… until it's not. :upside_down_face:

Less glibly, I think the auth stuff we need/want is complicated enough that it tipped the scales to doing a little more of the mechanics ourselves within the framework Passport provides. We also already called `req.login()` in some cases ourselves, and I thought it was a little confusing that it was called "somewhere in Passport" in other cases.

> I can't actually find details on what Passport does in `authenticate` from the [docs](https://www.passportjs.org/docs/). I assume you had to go through the source code to figure it out?

I definitely read a bit of the source, but several parts of it are documented here and there in JSDoc comments (although these don't seem to make it into the web docs, which are frustratingly all example based). There are hints in the web doc examples though, like https://www.passportjs.org/concepts/authentication/login/ which mentions:

> Note: `passport.authenticate()` middleware invokes `req.login()` automatically. This function is primarily used when users sign up, during which `req.login()` can be invoked to automatically log in the newly registered user.

PR #548, opened 2022-06-03
mpx docs

jameshadfield:


jameshadfield:
While I still think it's a good idea to have more pathogen-specific splash pages, this one never gained traction internally.

Issue #511, opened 2022-03-31
Improve logging


tsibley:
Logging in nextstrain.org is currently ad-hoc and sometimes useful, sometimes not enough.  The goal is to make what's logged more systematic and comprehensively useful while at the same time improving our ability to inspect and act on the logs.

Below are some proposed changes to make, spanning both the codebase itself as well as our deployment infra.  I expect to break each of these items out into separate issues in due course.

- [ ] Standardize on [debug](https://www.npmjs.com/package/debug) or [bunyan](https://www.npmjs.com/package/bunyan) for emitting logs.

  `debug` is already used by Express and several related deps (set `DEBUG='*'` to see these logs), which is a point in its favor, but it's relatively limited (e.g. no log levels, no structured logs, no child loggers).  `bunyan` provides a more traditional and modern set of logging features, including a focus on structured logs which greatly helps with analysis.

  Retire our existing logging functions (e.g. `utils.verbose()`) in favor of the above.

- [ ] Switch from ad-hoc messages that embed limited context to structured logs with increased context across the board (e.g. all logs include req context, incl. critical request ids).

  The ad-hoc messages don't all have to go away—they're still useful for skimming by eye, particularly if well-written—but they shouldn't be the _only_ thing produced by a log line.

- [ ] Archive raw logs to S3 for retrospective debugging/analysis.

  Archives should be raw, straight from Heroku's Logplex service.

  Two quick examples of things archives are useful for:
  
  - Investigating reports like "this weird thing happened last
  week"
  - Analyzing the impact of a bug over time, e.g. "how many requests
  were impacted by this over the last month?"

  Keep the retention period fairly short so we're not storing (potentially sensitive) data we don't need.  Maybe retain for 30 days?  (which is also the minimum duration you pay for with S3)

  Papertrail [supports S3 archiving](https://www.papertrail.com/help/automatic-s3-archive-export/), but we don't currently ingest raw logs into Papertrail due to account limitations (see [our incoming log filters](https://papertrailapp.com/destinations/19338442/filter)).

- [ ] Ingest logs into analysis system for real-time inspection.

  Currently this is Papertrail, a Heroku add-on.  It's got a nice straightforward interface.  However, we're using a pretty stringent free tier (10MB/day, 2d retention for search) so have to omit lots of useful logs (e.g. all GET requests!).  Retention period is also pretty short.  Paid tiers that would accommodate our usage largely unfiltered get pricey.

[...]

PR #552, opened 2022-06-13
sources: Bypass CloudFront for core and staging source S3 buckets

tsibley:
It seems that ostensibly the main reason for going through CloudFront
for requests from the nextstrain.org server → S3¹ is that it's cheaper
(and maybe faster).

While this is true for requests from the internet, our Heroku dynos run
on AWS EC2 in us-east-1 (the same region as our S3 buckets) and thus are
(AFAIK) treated as free internal requests for the purposes of pricing.
I don't believe CloudFront marks down data egress back into AWS
separately from other egress like S3, and CloudFront also has more
expensive per-request pricing than S3, esp. for our most common requests
(GET).  That said, CloudFront has higher free tier quotas for both
bandwidth and requests, so it's hard to know the net balance of the two
services without our request/egress numbers for S3 ⟷ Heroku (which we
don't have and can't easily acquire without a better story for logging
and metrics¹).

Re: speed, given that the nextstrain.org server is a fixed point in
network space rather than a moving set of users around the world (i.e.
the nextstrain.org *users*), it's hard to see how CloudFront could
improve response speed by *adding* a layer of indirection.

All of that said, the biggest actual win gained by removing CloudFront
is that it means removing a caching layer and the chance for bugs due to
stale data (particularly read-after-write inconsistency issues).  In
this regard, I don't think CloudFront is helping, and by itself this
makes the switch worth it to me.

¹ See https://github.com/nextstrain/nextstrain.org/issues/511

### Testing
- [x] CI passes
- [x] Poking around the review app works as expected
- [ ] A good followup would be for someone with billing access (that I don't have) to see if there's a noticeable change in S3/CloudFront charges before/after this change.

### DIFF
--- a/src/sources/core.js
+++ b/src/sources/core.js
@@ -12,7 +12,7 @@ const authorization = process.env.GITHUB_TOKEN
 
 class CoreSource extends Source {
   get name() { return "core"; }
-  async baseUrl() { return "http://data.nextstrain.org/"; }
+  async baseUrl() { return "https://nextstrain-data.s3.amazonaws.com/"; }
   get repo() { return "nextstrain/narratives"; }
   get branch() { return "master"; }
 
@@ -87,7 +87,7 @@ class CoreSource extends Source {
 
 class CoreStagingSource extends CoreSource {
   get name() { return "staging"; }
-  async baseUrl() { return "http://staging.nextstrain.org/"; }
+  async baseUrl() { return "https://nextstrain-staging.s3.amazonaws.com/"; }
   get repo() { return "nextstrain/narratives"; }
   get branch() { return "staging"; }
 }
--- a/src/sources/models.js
+++ b/src/sources/models.js
@@ -24,9 +24,7 @@ const {NoResourcePathError} = require("../exceptions");
  *
  * A concrete example:
  *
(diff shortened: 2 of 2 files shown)


joverlee521 (approved):
> our Heroku dynos run on AWS EC2 in us-east-1 (the same region as our S3 buckets) and thus are (AFAIK) treated as free internal requests for the purposes of pricing.

[Heroku's docs](https://devcenter.heroku.com/articles/s3#name-your-bucket) even suggests to set S3 buckets to the same region as the Heroku app for free data transfer rates.

> CloudFront has higher free tier quotas for both bandwidth and requests

Looking at [AWS Free Tier details](https://aws.amazon.com/free/?all-free-tier.sort-by=item.additionalFields.SortRank&all-free-tier.sort-order=asc&awsf.Free%20Tier%20Types=*all&awsf.Free%20Tier%20Categories=*all), seems like CloudFront _always_ has a free tier but S3 free tier only lasts for 12 months after initial sign-up. Wouldn't this mean we will have to pay for all requests to S3?

That said, the benefits of removing the CloudFront caching layer would outweigh the costs.

PR #560, opened 2022-06-21
Remove redundant `npm ci`

victorlin:

The first step in the local dev guide is to run `npm ci`. However, that also gets run in build.sh, which is used in the next step:

https://github.com/nextstrain/nextstrain.org/blob/8cee5e7b432d8d2e84c23595837ddce74ff39e88/build.sh#L37

This change removes the first step but keeps a short note on `npm ci` in the summary of build.sh.

### Related issue(s)

_N/A_

### Testing

_N/A_

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,21 +15,11 @@ This repository provides the tools you need to [build nextstrain.org locally](#b
 ---
 ## Build nextstrain.org locally
 
-### 1. Install prerequisites
-Install the node dependencies by running
-```
-npm ci
-```
-from this directory (the "nextstrain.org" directory).
-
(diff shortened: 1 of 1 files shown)


jameshadfield:
`build.sh` [runs npm ci only for the static directory](https://github.com/nextstrain/nextstrain.org/blob/master/build.sh#L37) (the gatsby stuff) but not for the root directory (auspice, server code etc). Unless I'm misunderstanding your issue here?

victorlin:
@jameshadfield oops good catch! I forgot that was also needed. This should be kept then.

PR #562, opened 2022-06-21
Adjust layout of multi-tenant groups bucket

tsibley:
### Description of proposed changes
The gist here is a behind-the-scenes layout change for the `nextstrain-groups` bucket to better separate objects within a group.

Currently:

```
nextstrain-groups/
  some-name/
    a_b.json
    x_y_z.json
    blah.md
    group-overview.md
```

With this:

```
nextstrain-groups/
  some-name/
    datasets/
      a_b.json
      x_y_z.json
    narratives/
      blah.md
    group-overview.md
```

This has been in the back of my mind for a while, but was recently reinvigorated/remotivated by discussion of issues with the current layout when describing https://github.com/nextstrain/nextstrain.org/issues/547.

See commit messages for full details.

### Related issue(s)
Related to #478.

### Testing
- [x] CI passes
- [x] Manual testing of the modified `migrate-group` script with `test-private` group (after "reverting" its migration)
- [x] Manual testing of the new `migrate-groups-layout` script against a private test bucket using objects copied from `s3://nextstrain-groups/test-private/`.

### Deploy

_to perform in order_

- [x] Deploy https://github.com/nextstrain/nextstrain.org/pull/563 all the way to production.
- [x] Run `./scripts/migrate-groups-layout --wet-run` to do the initial copy.
- [x] Check CI passes for this PR.
- [ ] Await review and address feedback.
- [ ] Merge, wait for canary deploy.
- [ ] Verify functionality on canary, then promote.
- [ ] Run `./scripts/migrate-groups-layout --wet-run --delete-after-copy` to pick up any intervening changes and do the final move.
- [ ] Remove the now-unnecessary `./scripts/migrate-groups-layout` file in a new commit.

PR #563, opened 2022-06-22
Prepare for adjusting the layout of the multi-tenant groups bucket

tsibley:
Ignore objects within each group's prefix starting with "datasets/" or
"narratives/".  These sub-prefixes are used in the new layout and during
a brief transition/testing period will have copies of the live data for
any groups already migrated to the multi-tenant bucket.  Ignoring them
avoids including them (as non-functional duplicate entries) in the
dataset and narrative listings (e.g. on an individual group's splash
page).

### Related issue(s)
Related to #562.

### Testing
- [x] Correctly ignores objects in the new layout for the `test-private` group, where I've manually adjusted the layout already. Other objects still correctly visible as datasets and narratives in listings.
- [x] Tests pass locally.

### DIFF
--- a/src/sources/groups.js
+++ b/src/sources/groups.js
@@ -108,6 +108,7 @@ class GroupSource extends Source {
             .map(object => object.Key)
             .filter(key => key.startsWith(this.prefix))
             .map(key => key.slice(this.prefix.length))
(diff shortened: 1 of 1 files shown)


tsibley:
I'm going to merge this pre-review, as it should have zero impact on the current behaviour (nothing of import exists under `datasets/` or `narratives/` currently) and merging will allow better review of #562.

PR #565, opened 2022-06-23
Resolve 503s from Heroku's routing layer

### LINKED ISSUE #564: Unexpected 503s from Heroku's routing layer masking our 401/403 responses on PUT
### Current Behavior
When we respond with a 401/403 on an unauthorized PUT (i.e. the client is sending/uploading data to us) it works locally but not when the request goes through Heroku's routing/proxy layer.

### Expected behavior
Our 401/403 responses should make it back to the client.

### Reproduction
<details>

I split up a real payload so I could insert a long delay in the middle to make sure it took some measurable time when transmitting to localhost.

```
$ split -dn2 zika.json zika.json
$ ls -l zika.json*
-rw-rw-r-- 1 tom tom 8404498 Jun 22 16:19 zika.json
-rw-rw-r-- 1 tom tom 4202249 Jun 22 16:22 zika.json00
-rw-rw-r-- 1 tom tom 4202249 Jun 22 16:22 zika.json01
```

Then tried uploading to a private group (as anonymous user) via a local server:

```
$ curl -sv localhost:5000/groups/test-private/zika -H'Content-Type: application/vnd.nextstrain.dataset.main+json' -T <(cat zika.json00; sleep 5; cat zika.json01)
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 5000 (#0)
> PUT /groups/test-private/zika HTTP/1.1
> Host: localhost:5000
> User-Agent: curl/7.58.0
> Accept: */*
> Transfer-Encoding: chunked
> Content-Type: application/vnd.nextstrain.dataset.main+json
> Expect: 100-continue
> 
< HTTP/1.1 100 Continue
} [16380 bytes data]
< HTTP/1.1 401 Unauthorized
< X-Powered-By: Express
< Vary: Accept, Accept-Encoding
< Content-Type: application/json; charset=utf-8
< Content-Length: 29
< ETag: W/"1d-ygKu0Kmhdv6dJRlYiPBiLwGfeoM"
< Date: Wed, 22 Jun 2022 23:40:02 GMT
< Connection: keep-alive
< Keep-Alive: timeout=5
* HTTP error before end of send, stop sending
< 
{ [29 bytes data]
* Closing connection 0
{"error":"UnauthorizedError"}
```

And again via nextstrain.org:

```
$ curl -sv https://nextstrain.org/groups/blab-private/zika -H'Content-Type: application/vnd.nextstrain.dataset.main+json' -T <(cat zika.json00; sleep 5; cat zika.json01)
*   Trying 54.91.6.89...
* TCP_NODELAY set
* Connected to nextstrain.org (54.91.6.89) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/certs/ca-certificates.crt
  CApath: /etc/ssl/certs
} [5 bytes data]
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
} [512 bytes data]
* TLSv1.3 (IN), TLS handshake, Server hello (2):
{ [122 bytes data]
* TLSv1.3 (IN), TLS Unknown, Certificate Status (22):
{ [1 bytes data]
* TLSv1.3 (IN), TLS handshake, Unknown (8):
{ [6 bytes data]
* TLSv1.3 (IN), TLS Unknown, Certificate Status (22):
{ [1 bytes data]
* TLSv1.3 (IN), TLS handshake, Certificate (11):
{ [4026 bytes data]
* TLSv1.3 (IN), TLS Unknown, Certificate Status (22):
{ [1 bytes data]
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
{ [264 bytes data]
* TLSv1.3 (IN), TLS Unknown, Certificate Status (22):
{ [1 bytes data]
* TLSv1.3 (IN), TLS handshake, Finished (20):
{ [36 bytes data]
* TLSv1.3 (OUT), TLS change cipher, Client hello (1):
} [1 bytes data]
* TLSv1.3 (OUT), TLS Unknown, Certificate Status (22):
} [1 bytes data]
* TLSv1.3 (OUT), TLS handshake, Finished (20):
} [36 bytes data]
* SSL connection using TLSv1.3 / TLS_AES_128_GCM_SHA256
* ALPN, server did not agree to a protocol
* Server certificate:
*  subject: CN=nextstrain.org
*  start date: May 12 04:49:43 2022 GMT
*  expire date: Aug 10 04:49:42 2022 GMT
*  subjectAltName: host "nextstrain.org" matched cert's "nextstrain.org"
*  issuer: C=US; O=Let's Encrypt; CN=R3
*  SSL certificate verify ok.
} [5 bytes data]
* TLSv1.3 (OUT), TLS Unknown, Unknown (23):
} [1 bytes data]
> PUT /groups/blab-private/zika HTTP/1.1
> Host: nextstrain.org
> User-Agent: curl/7.58.0
> Accept: */*
> Transfer-Encoding: chunked
> Content-Type: application/vnd.nextstrain.dataset.main+json
> Expect: 100-continue
> 
{ [5 bytes data]
* TLSv1.3 (IN), TLS Unknown, Certificate Status (22):
{ [1 bytes data]
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
{ [130 bytes data]
* TLSv1.3 (IN), TLS Unknown, Unknown (23):
{ [1 bytes data]
< HTTP/1.1 100 Continue
} [5 bytes data]
* TLSv1.3 (OUT), TLS Unknown, Unknown (23):
} [1 bytes data]
* TLSv1.3 (OUT), TLS Unknown, Unknown (23):
} [1 bytes data]
* TLSv1.3 (OUT), TLS Unknown, Unknown (23):
} [1 bytes data]
* TLSv1.3 (OUT), TLS Unknown, Unknown (23):
} [1 bytes data]
* TLSv1.3 (OUT), TLS Unknown, Unknown (23):
} [1 bytes data]
* TLSv1.3 (OUT), TLS Unknown, Unknown (23):
} [1 bytes data]
* TLSv1.3 (OUT), TLS Unknown, Unknown (23):
} [1 bytes data]
* TLSv1.3 (IN), TLS Unknown, Unknown (23):
{ [1 bytes data]
< HTTP/1.1 503 Service Unavailable
< Connection: keep-alive
< Server: Cowboy
< Date: Wed, 22 Jun 2022 23:46:39 GMT
< Content-Length: 506
< Content-Type: text/html; charset=utf-8
< Cache-Control: no-cache, no-store
* HTTP error before end of send, stop sending
< 
{ [506 bytes data]
* Closing connection 0
} [5 bytes data]
* TLSv1.3 (OUT), TLS Unknown, Unknown (21):
} [1 bytes data]
* TLSv1.3 (OUT), TLS alert, Client hello (1):
} [2 bytes data]
<!DOCTYPE html>
	<html>
	  <head>
		<meta name="viewport" content="width=device-width, initial-scale=1">
		<meta charset="utf-8">
		<title>Application Error</title>
		<style media="screen">
		  html,body,iframe {
			margin: 0;
			padding: 0;
		  }
		  html,body {
			height: 100%;
			overflow: hidden;
		  }
		  iframe {
			width: 100%;
			height: 100%;
			border: 0;
		  }
		</style>
	  </head>
	  <body>
		<iframe src="//www.herokucdn.com/error-pages/application-error.html"></iframe>
	  </body>
	</html>
```

</details>

### Understanding
I believe this is because we close the connection from Heroku's routing layer while it's still sending us data, which it interprets as a protocol error on our part or a network error. Notably, curl handles this fine, as the above local example demonstrates, c.f. the message "HTTP error before end of send, stop sending".

It's possible our immediate 100 Continue response (Node's default behaviour) is confusing Heroku because it thinks we agreed to accept the request data when really we're just saying "Mmmhmm, go on…" while we do our work before determining the 401/403 response.

Notably, [MDN says](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Expect) that 100 Continue should be used when:

> if the information from the request header is sufficient to cause immediate success

which is certainly not the case here. Not clear what accepted practice is, however.

### Possible solutions

Try each of these in order in a review app or other dev context on Heroku.

- [x] Enable Heroku's "experimental" [pass thru support](https://devcenter.heroku.com/articles/http-routing#enabling-100-continue-support) of `Expect: 100-continue` request headers and `100 Continue` responses to take Heroku's routing layer more out of the way of the client and our server here.
- [x] Send `Connection: close` on error, to better notify Heroku's routing layer that we're going to hang up and not read request data.
- [x] Send `Connection: close` + [half-close the TCP socket](https://nodejs.org/docs/latest-v14.x/api/net.html#net_socket_end_data_encoding_callback) ourselves on error (`res.socket.end()`), to avoid the "TCP reset problem" and notify at a lower level that we're really going to hang up now. (But beware, doing this behind Express' back may have other unintended consequences.)
- [x] Always read the whole request body even if we're just going to throw it away before returning an error.

### Additional context
Potentially related to https://github.com/nextstrain/cli/issues/189 and why we're seeing a broken connection instead of an HTTP error as expected (which would get handled).

tsibley:
# Description

Resolves #564.

# Deploy

- [ ] Merge, monitor deploy to canary, and verify behaviour on next.nextstrain.org
- [ ] `heroku labs:enable http-end-to-end-continue -a nextstrain-canary`
- [ ] Verify behaviour again on next.nextstrain.org
- [ ] Promote canary to prod
- [ ] `heroku labs:enable http-end-to-end-continue -a nextstrain-server`
- [ ] Verify behaviour on nextstrain.org
- [ ] Resolve https://github.com/nextstrain/cli/issues/189

### DIFF
--- a/server.js
+++ b/server.js
@@ -30,7 +30,32 @@ const app = require("./src/app");
 
 app.set("port", port);
 
-const server = http.createServer(app)
+const server = http
+  .createServer({
+    ServerResponse: class ServerResponse extends http.ServerResponse {
+      /**
+       * Sets res.wroteContinue so later handlers can know if res.writeContinue()
+       * was called already.  The standard parent class tracks this already, but
+       * only in an internal property.
+       */
+      writeContinue() {
+        super.writeContinue();
+        this.wroteContinue = true;
+      }
+    },
+  })
+  .on("request", app)
+  .on("checkContinue", (req, res) => {
+    /* When this event fires, the normal "request" event that calls app is not
+     * fired.
+     *
+     * Endpoints that receive request bodies (PUT, POST, etc) will check the
+     * req.expectsContinue property to see if they should call
+     * res.writeContinue() at an appropriate point.
+     */
+    req.expectsContinue = true;
+    return app(req, res);
+  })
   .on("listening", () => {
     console.log("  -------------------------------------------------------------------------");
     console.log(nextstrainAbout);
--- a/src/app.js
+++ b/src/app.js
@@ -413,14 +413,15 @@ app.useAsync(async (err, req, res, next) => {
   }
 
   /* Read the entire request body (discarding it) if the request might have a
-   * body.  This ensures that the request is finished being sent before we
-   * return a (error) response, which some clients require (such as Heroku's
-   * routing/proxy layer and the Python "requests" package, but notably not
-   * curl).
+   * body and wasn't made with Expect: 100-continue, or if it was and we wrote
+   * a 100 Continue response but then ended up here.  This ensures that the
+   * request is finished being sent before we return a (error) response, which
+   * some clients require (such as Heroku's routing/proxy layer and the Python
+   * "requests" package, but notably not curl).
    */
   const mayHaveBody = !["GET", "HEAD", "DELETE", "OPTIONS"].includes(req.method);
 
-  if (mayHaveBody) {
+  if (mayHaveBody && (!req.expectsContinue || res.wroteContinue)) {
     const reqFinished = streamFinished(req);
     req.unpipe();
     req.resume();
--- a/src/app.js
+++ b/src/app.js
@@ -7,6 +7,14 @@ const utils = require("./utils");
 const cors = require('cors');
 const {addAsync} = require("@awaitjs/express");
 const {Forbidden, NotFound, Unauthorized} = require("http-errors");
+const stream = require("stream");
+const {promisify} = require("util");
+
+/* XXX TODO: Replace promisify() with require("stream/promises") once we
+ * upgrade to Node 15+.
+ *   -trs, 5 Nov 2021
+ */
+const streamFinished = promisify(stream.finished);
 
 const PRODUCTION = process.env.NODE_ENV === "production";
 const CANARY_ORIGIN = process.env.CANARY_ORIGIN;
@@ -402,6 +410,21 @@ app.useAsync(async (err, req, res, next) => {
   if (res.headersSent) {
     utils.verbose("Headers already sent; using Express' default error handler");
     return next(err);
+  }
+
+  /* Read the entire request body (discarding it) if the request might have a
+   * body.  This ensures that the request is finished being sent before we
+   * return a (error) response, which some clients require (such as Heroku's
+   * routing/proxy layer and the Python "requests" package, but notably not
+   * curl).
+   */
+  const mayHaveBody = !["GET", "HEAD", "DELETE", "OPTIONS"].includes(req.method);
+
+  if (mayHaveBody) {
+    const reqFinished = streamFinished(req);
+    req.unpipe();
+    req.resume();
+    await reqFinished;
   }
 
   res.vary("Accept");
--- a/src/endpoints/sources.js
+++ b/src/endpoints/sources.js
@@ -352,6 +352,11 @@ function receiveSubresource(subresourceExtractor) {
 
     authz.assertAuthorized(req.user, authz.actions.Write, subresource.resource);
 
+    /* Notify the client they can continue sending the request body now that
+     * we're past authz.
+     */
+    if (req.expectsContinue) res.writeContinue();
+
     /* Proxy the data through us:
      *
      *    client (browser, CLI, etc) ⟷ us (nextstrain.org) ⟷ upstream source


server.js
```diff
@@ -30,7 +30,32 @@ const app = require("./src/app");
...
+    /* When this event fires, the normal "request" event that calls app is not
+     * fired.
+     *
+     * Endpoints that receive request bodies (PUT, POST, etc) will check the
+     * req.expectsContinue property to see if they should call
+     * res.writeContinue() at an appropriate point.
+     */
+    req.expectsContinue = true;
+    return app(req, res);
+  })
```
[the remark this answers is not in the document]
victorlin: Ok, I realized while reading the [`http.createServer()` docs](https://nodejs.org/api/http.html#httpcreateserveroptions-requestlistener) that `app` is a `requestListener <Function>`, and confirmed by [this SO answer](https://stackoverflow.com/a/26641372). This makes more sense now.

src/endpoints/sources.js
```diff
@@ -352,6 +352,11 @@ function receiveSubresource(subresourceExtractor) {
 
     authz.assertAuthorized(req.user, authz.actions.Write, subresource.resource);
 
+    /* Notify the client they can continue sending the request body now that
+     * we're past authz.
+     */
+    if (req.expectsContinue) res.writeContinue();
```
victorlin, line 358: Is `100-Continue` only used for `PUT` requests in this app? If so, any chance it'll be used elsewhere?

src/app.js
```diff
@@ -404,6 +412,22 @@ app.useAsync(async (err, req, res, next) => {
...
+   * request is finished being sent before we return a (error) response, which
+   * some clients require (such as Heroku's routing/proxy layer and the Python
+   * "requests" package, but notably not curl).
+   */
+  const mayHaveBody = !["GET", "HEAD", "DELETE", "OPTIONS"].includes(req.method);
+
+  if (mayHaveBody && (!req.expectsContinue || res.wroteContinue)) {
+    const reqFinished = streamFinished(req);
+    req.unpipe();
+    req.resume();
```
victorlin, line 427: I am looking through [Stream docs](https://nodejs.org/api/stream.html) since this is new to me. What's the purpose of these two lines?

#### `req.unpipe()`

[doc](https://nodejs.org/api/stream.html#readableunpipedestination):

> If the destination is not specified, then all pipes are detached

Where were any attached?

#### `req.resume()`

[doc](https://nodejs.org/api/stream.html#readableresume):

> can be used to fully consume the data from a stream without actually processing any of that data

This one makes sense to me and is described by the block comment.

victorlin (commented):
Looks good – would've been an approval pre-merge. Just adding some questions and remarks from my learning here.

src/endpoints/sources.js
tsibley: It'd be used for any body-bearing methods, like PUT, POST, or PATCH, but this endpoint is the only one we currently have.

src/app.js
tsibley: Ah, since this is the app's global error handler, we don't really "know" what may or may not have previously called `req.pipe(…)`. It seemed safest to `.unpipe()` before resuming the stream so that we didn't resume to previous destinations.

server.js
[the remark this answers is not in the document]
tsibley: Indeed, it's made of functions/callables. :-) This is a common pattern across languages/frameworks (WSGI, ASGI, PSGI, etc).

Issue #76, opened 2020-01-10
groups: Stop hardcoded assumption that group name matches source name exactly


kairstenfay:
See @tsibley's [comments](https://github.com/nextstrain/nextstrain.org/pull/74#discussion_r364941068) on a [recent groups-related PR](https://github.com/nextstrain/nextstrain.org/pull/74) for context.

jameshadfield:
@tsibley @kairstenfay could you elaborate on this here please?

We currently have a setup where nextstrain group `x` is backed by S3 bucket `nextstrain-x` and, if private, then authenticated users belong to the cognito group `x`. Are you suggesting moving away from this?

tsibley:
@jameshadfield I'm suggesting that Nextstrain Group `x` being backed by Source `x` which is backed by S3 bucket `nextstrain-x` and Cognito group `x` is a temporary state in the implementation that we should bake into as little code as possible.

As I just mentioned in review comments on #82, we're missing a first class concept/object/record in the codebase for a Nextstrain Group. Sources are a distinct concept, and I want to avoid conflating them with Groups. For example, Groups have a Source (just one for now), but not all Sources are Groups. In the not too distant future, I think we need to introduce a proper Group concept, which knows which Source it is associated with (or, more likely, can dynamically instantiate that Source) and also knows other "group metadata" things, like what Cognito group it is associated with. The Group concept may still use our existing naming conventions, but it will be the sole place that is hardcoded and the rest of the system has those details abstracted away. The rest of the system will deal with Source objects (and the Group objects to-come) which have a consistent API. The design principle here is that components of the system should have clear ownership of data and actions related to them.

PR #570, opened 2022-07-11
Add RESTful API endpoints for Groups customizations

### LINKED ISSUE #547: Add RESTful API endpoints for Groups customizations
The goal is to allow for Groups using the new multi-tenant backend storage to self-manage logo and overview customizations previously managed via the special-cased files `group-logo.png` and `group-overview.md`.

## Proposal

As a starting proposal, consider two new routes (exact paths up for discussion) × three methods (GET, PUT, DELETE) for six new endpoints:

```
/groups/:groupName/logo

    GET                           → 200 OK (Content-Type: image/png)
    PUT (Content-Type: image/png) → 204 No Content
    DELETE                        → 204 No Content

/groups/:groupName/overview

    GET                               → 200 OK (Content-Type: text/markdown)
    PUT (Content-Type: text/markdown) → 204 No Content
    DELETE                            → 204 No Content
```

These are just the "happy paths". 401/403/404 should also be returned as appropriate.

## Implementation considerations

- ~If using the proposed routes, then they should come before more generic routes for datasets. Or, it might be preferable to change the routes instead to avoid conflict with user-provided paths, e.g. to `/groups/:groupName/-/logo` or similar.~ Settled on using the `/groups/:groupName/settings` namespace going forwards: https://github.com/nextstrain/nextstrain.org/pull/581#discussion_r954148909

- I don't think we need [custom media types](https://docs.nextstrain.org/projects/nextstrain-dot-org/en/latest/api-restful.html#media-types) here but this could be discussed.

- Ok to accept just `image/png` for now. We can expand this in the future to more `image/*` types.

- The storage of this data can remain objects like `group-overview.md`, but we should ensure it can't conflict with other group data, like narratives. For example, `/groups/…/narratives/group-overview` shouldn't be able to touch the content uploaded for the group overview customization.

  ~This can likely be accomplished several ways. The best might be a layout change to the backend storage so that datasets and narratives live under a further prefix of the per-group prefix, e.g. `s3://nextstrain-groups/blab/datasets/zika.json` instead of `s3://nextstrain-groups/blab/zika.json`, so that the top-level is reserved for internal control data. I have a local prototype of this if we want to pursue that.~ Done in #563 and #562.

victorlin:

Adds RESTful API endpoints for downloading/uploading/removing group overviews and logos.

### Related issue(s)

- Closes #547

### Save for later

- 34f02d57e29d99d2725fa7bb2dc20a2a8ee49866 from https://github.com/nextstrain/nextstrain.org/pull/570#discussion_r960106390

### Testing

Automated:

- [x] ~Added test for `GET /groups/test/overview`~ removed, see https://github.com/nextstrain/nextstrain.org/pull/570#discussion_r958789759

Manual:

<details>

<summary>

Local testing from a member of `test-private/owners`. [Maybe automate?](https://github.com/nextstrain/nextstrain.org/pull/570#discussion_r923922221)

</summary>

Token setup

```sh
nextstrain login
id_token_line=$(grep id_token ~/.nextstrain/secrets)
token=${id_token_line#"id_token = "}
```

GET private overview works

```sh
curl http://localhost:5000/groups/test-private/settings/overview \
    --header "Authorization: Bearer $token" \
    --header 'Accept: text/markdown' \
    --compressed \
        | tee overview.md
```

DELETE private overview works

```sh
curl http://localhost:5000/groups/test-private/settings/overview \
    --header "Authorization: Bearer $token" \
    --request "DELETE"
```

PUT private overview works

```sh
curl http://localhost:5000/groups/test-private/settings/overview \
    --header "Authorization: Bearer $token" \
    --header 'Content-Type: text/markdown' \
    --upload-file overview.md
```

GET private logo works

```sh
curl http://localhost:5000/groups/test-private/settings/logo \
    --header "Authorization: Bearer $token" \
    --header 'Accept: image/png' \
    --compressed > logo.png
```

DELETE private logo works

```sh
curl http://localhost:5000/groups/test-private/settings/logo \
    --header "Authorization: Bearer $token" \
    --request "DELETE"
```

PUT private logo works

```sh
curl http://localhost:5000/groups/test-private/settings/logo \
    --header "Authorization: Bearer $token" \
    --header 'Content-Type: image/png' \
    --upload-file logo.png
```

**Test these since the underlying code was refactored**

GET private narrative still works

```sh
curl http://localhost:5000/groups/test-private/narratives/foo \
    --header "Authorization: Bearer $token" \
    --header 'Accept: text/markdown' \
    --compressed \
        > narrative.md
```

DELETE private narrative still works

```sh
curl http://localhost:5000/groups/test-private/narratives/victorlin-test \
    --header "Authorization: Bearer $token" \
    --request "DELETE"
```

PUT private narrative still works

```sh
curl http://localhost:5000/groups/test-private/narratives/victorlin-test \
    --header "Authorization: Bearer $token" \
    --header 'Content-Type: text/vnd.nextstrain.narrative+markdown' \
    --upload-file narrative.md
```

GET private dataset still works

```sh
curl http://localhost:5000/groups/test-private/victorlin-test \
    --header "Authorization: Bearer $token" \
    --header 'Accept: application/json' \
    --compressed \
        > dataset.json
```

DELETE private dataset still works

```sh
curl http://localhost:5000/groups/test-private/victorlin-test \
    --header "Authorization: Bearer $token" \
    --request "DELETE"
```

PUT private dataset still works

```sh
curl http://localhost:5000/groups/test-private/victorlin-test \
    --header "Authorization: Bearer $token" \
    --header 'Content-Type: application/vnd.nextstrain.dataset.main+json' \
    --upload-file dataset.json
```

</details>

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -43,6 +43,15 @@ const {
 } = endpoints.sources;
 
 const {
+  getGroupLogo,
+  putGroupLogo,
+  deleteGroupLogo,
+  getGroupOverview,
+  putGroupOverview,
+  deleteGroupOverview,
+} = endpoints.groups;
+
+const {
   CoreSource,
   CoreStagingSource,
   CommunitySource,
@@ -304,6 +313,18 @@ app.routeAsync("/groups/:groupName")
 
 app.use("/groups/:groupName/settings",
   endpoints.groups.setGroup(req => req.params.groupName));
+
+app.routeAsync("/groups/:groupName/settings/logo")
+  .getAsync(getGroupLogo)
+  .putAsync(putGroupLogo)
+  .deleteAsync(deleteGroupLogo)
+;
+
+app.routeAsync("/groups/:groupName/settings/overview")
+  .getAsync(getGroupOverview)
+  .putAsync(putGroupOverview)
+  .deleteAsync(deleteGroupOverview)
+;
 
 app.route("/groups/:groupName/settings/*")
   .all(() => { throw new NotFound(); });
--- a/src/app.js
+++ b/src/app.js
@@ -302,6 +302,12 @@ app.routeAsync("/groups/:groupName")
   .getAsync(endpoints.static.sendGatsbyEntrypoint)
 ;
 
+app.use("/groups/:groupName/settings",
+  endpoints.groups.setGroup(req => req.params.groupName));
+
+app.route("/groups/:groupName/settings/*")
+  .all(() => { throw new NotFound(); });
+
 // Avoid matching "narratives" as a dataset name.
 app.routeAsync("/groups/:groupName/narratives")
   .getAsync((req, res) => res.redirect(`/groups/${esc(req.params.groupName)}`));
--- a/src/endpoints/groups.js
+++ b/src/endpoints/groups.js
@@ -1,5 +1,7 @@
 import * as authz from "../authz/index.js";
 import { Group } from "../groups.js";
+import {contentTypesProvided, contentTypesConsumed} from "../negotiate.js";
+import {deleteByUrls, proxyFromUpstream, proxyToUpstream} from "../upstream.js";
 
 
 const setGroup = (nameExtractor) => (req, res, next) => {
@@ -12,6 +14,147 @@ const setGroup = (nameExtractor) => (req, res, next) => {
 };
 
 
+/* Group customizations
+ */
+
+
+/* Group logo
+ */
+
+
+/* GET
+ */
+const getGroupLogo = contentTypesProvided([
+  ["image/png", sendGroupLogo],
+]);
+
+
+/* PUT
+ */
+const putGroupLogo = contentTypesConsumed([
+  ["image/png", receiveGroupLogo],
+]);
+
+
+/* DELETE
+ */
+const deleteGroupLogo = async (req, res) => {
+  authz.assertAuthorized(req.user, authz.actions.Write, req.context.group);
+
+  const method = "DELETE";
+  const url = await req.context.group.source.urlFor("group-logo.png", method);
+  await deleteByUrls([url]);
+
+  return res.status(204).end();
+};
+
+
+/* Group overview
+ */
+
+
+/* GET
+ */
+const getGroupOverview = contentTypesProvided([
+  ["text/markdown", sendGroupOverview],
+  ["text/plain", sendGroupOverview],
+]);
+
+
+/* PUT
+ */
+const putGroupOverview = contentTypesConsumed([
+  ["text/markdown", receiveGroupOverview],
+]);
+
+
+/* DELETE
+ */
+const deleteGroupOverview = async (req, res) => {
+  authz.assertAuthorized(req.user, authz.actions.Write, req.context.group);
+
+  const method = "DELETE";
+  const url = await req.context.group.source.urlFor("group-overview.md", method);
+  await deleteByUrls([url]);
+
+  return res.status(204).end();
+};
+
+
+/**
+ * An Express endpoint that sends a group overview determined by the request.
+ *
+ * @param {express.request} req - Express-style request instance
+ * @param {express.response} res - Express-style response instance
+ * @returns {expressEndpointAsync}
+ */
+async function sendGroupOverview(req, res) {
+  authz.assertAuthorized(req.user, authz.actions.Read, req.context.group);
+
+  return await proxyFromUpstream(req, res,
+    await req.context.group.source.urlFor("group-overview.md"),
+    "text/markdown"
+  );
+}
+
+
+/**
+ * An Express endpoint that receives a group overview determined by the request.
+ *
+ * @param {express.request} req - Express-style request instance
+ * @param {express.response} res - Express-style response instance
+ * @returns {expressEndpointAsync}
+ */
+async function receiveGroupOverview(req, res) {
+  authz.assertAuthorized(req.user, authz.actions.Write, req.context.group);
+
+  return await proxyToUpstream(req, res,
+    async (method, headers) => await req.context.group.source.urlFor("group-overview.md", method, headers),
+    "text/markdown"
+  );
+}
+
+
+/**
+ * An Express endpoint that sends a group logo determined by the request.
+ *
+ * @param {express.request} req - Express-style request instance
+ * @param {express.response} res - Express-style response instance
+ * @returns {expressEndpointAsync}
+ */
+async function sendGroupLogo(req, res) {
+  authz.assertAuthorized(req.user, authz.actions.Read, req.context.group);
+
+  return await proxyFromUpstream(req, res,
+    await req.context.group.source.urlFor("group-logo.png"),
+    "image/png"
+  );
+}
+
+
+/**
+ * An Express endpoint that receives a group logo determined by the request.
+ *
+ * @param {express.request} req - Express-style request instance
+ * @param {express.response} res - Express-style response instance
+ * @returns {expressEndpointAsync}
+ */
+async function receiveGroupLogo(req, res) {
+  authz.assertAuthorized(req.user, authz.actions.Write, req.context.group);
+
+  return await proxyToUpstream(req, res,
+    async (method, headers) => await req.context.group.source.urlFor("group-logo.png", method, headers),
+    "image/png"
+  );
+}
+
+
 export {
   setGroup,
+  getGroupLogo,
+  putGroupLogo,
+  deleteGroupLogo,
+  getGroupOverview,
+  putGroupOverview,
+  deleteGroupOverview,
 };
--- a/src/endpoints/groups.js
+++ b/src/endpoints/groups.js
@@ -1,0 +1,17 @@
+import * as authz from "../authz/index.js";
+import { Group } from "../groups.js";
+
+
+const setGroup = (nameExtractor) => (req, res, next) => {
+  const group = new Group(nameExtractor(req));
+
+  authz.assertAuthorized(req.user, authz.actions.Read, group);
+
+  req.context.group = group;
+  return next();
+};
+
+
+export {
+  setGroup,
+};
--- a/src/endpoints/sources.js
+++ b/src/endpoints/sources.js
@@ -4,25 +4,12 @@
  *   no-multi-spaces: Alignment makes for easier reading
  */
 
-import { parse as parseContentType } from 'content-type';
+import { NotFound } from '../httpErrors.js';
 
-import negotiateMediaType from 'negotiator/lib/mediaType.js';
-import stream from 'stream';
-import { promisify } from 'util';
-import zlib from 'zlib';
-import readStream from 'raw-body';
-import { InternalServerError, NotFound, UnsupportedMediaType } from '../httpErrors.js';
 import * as authz from '../authz/index.js';
 import { contentTypesProvided, contentTypesConsumed } from '../negotiate.js';
-import { fetch, Request } from '../fetch.js';
 import { sendAuspiceEntrypoint } from './static.js';
-
-
-/* XXX TODO: Replace promisify() with require("stream/promises") once we
- * upgrade to Node 15+.
- *   -trs, 5 Nov 2021
- */
-const pipeline = promisify(stream.pipeline);
+import { deleteByUrls, proxyFromUpstream, proxyToUpstream } from "../upstream.js";
 
 
 /**
@@ -189,13 +176,7 @@ const deleteResource = resourceExtractor => async (req, res) => {
 
   const method = "DELETE";
   const upstreamUrls = await Promise.all(resource.subresources().map(s => s.url(method)));
-  const upstreamResponses = await Promise.all(upstreamUrls.map(url => fetch(url, {method})));
-
-  const goodStatuses = new Set([204, 404]);
-
-  if (!upstreamResponses.every(r => goodStatuses.has(r.status))) {
-    throw new InternalServerError("failed to delete all subresources");
-  }
+  await deleteByUrls(upstreamUrls);
 
   return res.status(204).end();
 };
@@ -305,39 +286,10 @@ function sendSubresource(subresourceExtractor) {
 
     authz.assertAuthorized(req.user, authz.actions.Read, subresource.resource);
 
-    const subresourceUrl = await subresource.url();
-
-    /* Proxy the data through us:
-     *
-     *    client (browser, CLI, etc) ⟷ us (nextstrain.org) ⟷ upstream source
-     */
-    return await proxyResponseBodyFromUpstream(req, res, new Request(subresourceUrl, {
-      headers: {
-        Accept: subresource.accept,
-
-        /* Normalize Accept-Encoding to gzip or nothing, as realistically we
-         * only expect upstreams to support gzip and normalizing reduces the
-         * cache burden of varying on many equivalent but effectively the same
-         * client Accept-Encoding values.
-         */
-        "Accept-Encoding": req.acceptsEncodings("gzip")
-          ? "gzip"
-          : "identity",
-      },
-
-      /* Use "no-cache" mode to always revalidate with the upstream but avoid
-       * re-transferring the content if we have a cached copy that matches
-       * upstream.
-       */
-      cache: "no-cache",
-
-      /* Disable automatic body decompression.  Instead, we pass through the
-       * Accept-Encoding of the client to the upstream and return the
-       * upstream Content-Encoding and body directly.  This avoids needlessly
-       * decompressing and recompressing ourselves.
-       */
-      compress: false,
-    }));
+    return await proxyFromUpstream(req, res,
+      await subresource.url(),
+      subresource.accept
+    );
   };
 }
 
@@ -351,7 +303,6 @@ function sendSubresource(subresourceExtractor) {
  */
 function receiveSubresource(subresourceExtractor) {
   return async (req, res) => {
-    const method = "PUT";
     const subresource = subresourceExtractor(req);
 
     authz.assertAuthorized(req.user, authz.actions.Write, subresource.resource);
@@ -361,262 +312,11 @@ function receiveSubresource(subresourceExtractor) {
      */
     if (req.expectsContinue) res.writeContinue();
 
-    /* Proxy the data through us:
-     *
-     *    client (browser, CLI, etc) ⟷ us (nextstrain.org) ⟷ upstream source
-     */
-    // eslint-disable-next-line prefer-const
-    let headers = {
-      "Content-Type": subresource.mediaType,
-      ...copyHeaders(req, ["Content-Encoding", "Content-Length"]),
-
-      /* XXX TODO: Consider setting Cache-Control rather than relying on
-       * ambiguous defaults.  Potentially impacts:
-       *
-       *   - Our own fetch() caching, including in sendSubresource() above
-       *   - Our Charon endpoints, if upstream headers are sent to the browser?
-       *   - CloudFront caching (not sure about its default behaviour)
-       *   - Browsers, if fetched directly, such as by redirection
-       *
-       * I think a cautious initial value would be to set "private" or "public"
-       * depending on the Source and then always set "must-revalidate,
-       * proxy-revalidate, max-age=0".  This would allow caches (ours,
-       * browsers, CloudFront?) to store the data but always check upstream
-       * with conditional requests.
-       *   -trs, 7 Dec 2021
-       */
-    };
-
-    // Body of the request as a Node stream
-    let body = req;
-
-    // Compress on the fly to gzip if it's not already gzip compressed.
-    if (!headers["Content-Encoding"]) {
-      delete headers["Content-Length"]; // won't be valid after compression
-      headers["Content-Encoding"] = "gzip";
-      body = body.pipe(zlib.createGzip());
-    }
-
-    if (headers["Content-Encoding"] !== "gzip") {
-      throw new UnsupportedMediaType("unsupported Content-Encoding; only gzip is supported");
-    }
-
-    /* Our upstreams for PUTs are all S3, and S3 requires a Content-Length
-     * header (i.e. doesn't accept streaming PUTs).  If we don't have a
-     * Content-Length from the request (i.e. the request is a streaming PUT or
-     * we're doing on-the-fly compression), then we have to buffer the entire
-     * body into memory so we can calculate length for S3.  When passed a
-     * buffer instead of a stream, fetch() will calculate Content-Length for us
-     * before sending the request.
-     *
-     * An alternative to buffering the whole body is to use S3's multipart
-     * upload API, but the minimum part size is 5MB so some buffering would be
-     * required anyway.  Multipart uploads would add inherent complexity at
-     * runtime and also design time, as we'd have to rework our data model.
-     *
-     * In a review of all the (compressed) core and group datasets (nearly
-     * 11k), over 99% are less than 5MB and none are more than 15MB.  Given
-     * that we'd only be able to use multipart uploads for less than 1% of
-     * datasets and even the largest datasets would fit comfortably in memory,
-     * it doesn't seem worth implementing.
-     *
-     * Allow buffering up to 20MB of data after gzip compression (guaranteed by
-     * Content-Encoding handling above).  Requests that exceed this will get a
-     * 413 error (thrown by readStream()), and if this becomes an issue we can
-     * consider bumping the limit.  Clients also have the option of
-     * pre-compressing the data and including a Content-Length themselves so we
-     * don't have to buffer it, in which case we don't limit request sizes.
-     *   -trs, 21 Jan 2022
-     */
-    if (!headers["Content-Length"]) {
-      body = await readStream(body, { limit: 20_000_000 /* 20MB */ });
-    }
-
-    const subresourceUrl = await subresource.url(method, {
-      "Content-Type": headers["Content-Type"],
-      "Content-Encoding": headers["Content-Encoding"],
-    });
-
-    const upstreamRes = await fetch(subresourceUrl, {method, body, headers});
-
-    switch (upstreamRes.status) {
-      case 200:
-      case 204:
-        break;
-
-      default:
-        throw new InternalServerError(`upstream said: ${upstreamRes.status} ${upstreamRes.statusText}`);
-    }
-
-    return res.status(204).end();
+    return proxyToUpstream(req, res,
+      async (method, headers) => await subresource.url(method, headers),
+      subresource.mediaType
+    );
   };
-}
-
-
-/**
- * Fetch from an upstream server and stream the response body back through as
- * our own response body.
- *
- * Only 200 and 304 responses are passed through.  403 and 404 statuses are
- * raised as 404s.  Everything else raises a 500.
- *
- * If our upstream request included an `Accept` header, makes sure the upstream
- * response satisfies it.
- *
- * @param {express.request} req - Express-style request instance
- * @param {express.response} res - Express-style response instance
- * @param {Response} upstreamReq - Request instance (WHATWG fetch()-style) for
- *    the upstream resource
- */
-async function proxyResponseBodyFromUpstream(req, res, upstreamReq) {
-  const upstreamRes = await fetch(upstreamReq);
-
-  switch (upstreamRes.status) {
-    case 200:
-    case 304:
-      break;
-
-    case 403:
-    case 404:
-      throw new NotFound();
-
-    default:
-      throw new InternalServerError(`upstream said: ${upstreamRes.status} ${upstreamRes.statusText}`);
-  }
-
-  /* Check that the upstream returned something acceptable to us.  This ensures
-   * we honor the content negotiation for our own response.
-   */
-  const accept      = upstreamReq.headers.get("Accept");
-  const contentType = upstreamRes.headers.get("Content-Type");
-
-  if (!acceptableContentType(accept, contentType)) {
-    throw new InternalServerError(`source response has unacceptable Content-Type: ${contentType}`);
-  }
-
-  const forwardedUpstreamResHeaders = [
-    /* It's ok to forward the ETag back without weakening it (W/…) even if
-     * upstreamReq.compress is enabled, because though local decoding of the
-     * Content-Encoding renders a strong ETag incorrect, this incorrectness
-     * (and potentially buggy caching) would only be visible to the client if
-     * they also made direct requests to our upstream using the same ETag.
-     * That is not a situation we need to worry about.  As long the local
-     * decompression behaviour we use is always the same and not different for
-     * the same client at different times, then the same ETag won't be sent for
-     * two different response versions.
-     *   -trs, 7 Dec 2021
-     */
-    "ETag",
-    "Last-Modified",
-    "Content-Length",
-
-    /* Forward Content-Encoding if local decompression was disabled so the
-     * client can decode the body itself.
-     */
-    ...(!upstreamReq.compress ? ["Content-Encoding"] : []),
-  ];
-
-  res.set(copyHeaders(upstreamRes.headers, forwardedUpstreamResHeaders));
-
-  /* Check if the request conditions (e.g. If-None-Match) are satisfied (e.g.
-   * against our response ETag) and short-circuit with a 304 if we can!
-   */
-  if (req.fresh) {
-    // Remove headers about the body, which we aren't sending now.
-    res.removeHeader("Content-Type");
-    res.removeHeader("Content-Length");
-    res.removeHeader("Content-Encoding");
-    return res.status(304).end();
-  }
-
-  /* Most of our upstreams send unconditional Content-Encoding: gzip responses,
-   * which is fine for browsers because all browsers support gzip, but not all
-   * API clients may by default (e.g. curl doesn't without the --compressed
-   * option).  Play nice by decompressing the response if necessary.  This
-   * could be a global middleware, but punting on that for now as it's a bit
-   * more fiddly.  (My ideal would be if our existing "compression()"
-   * middleware also handled this case, so maybe an upstream patch is in
-   * order.)
-   *   -trs, 9 Dec 2021
-   */
-  if (res.get("Content-Encoding") === "gzip" && !req.acceptsEncodings("gzip")) {
-    res.removeHeader("Content-Encoding");
-    res.removeHeader("Content-Length");
-
-    // Weaken any strong ETag since we're now modifying response body
-    const etag = res.get("ETag");
-    if (etag && !etag.startsWith("W/")) res.set("ETag", `W/${etag}`);
-
-    await pipeline(upstreamRes.body, zlib.createGunzip(), res);
-  } else {
-    try {
-      await pipeline(upstreamRes.body, res);
-    } catch (err) {
-      /* Nothing we can do about the client closing the connection on us before
-       * we were "ready"; see rationale in commit message introducing this
-       * handling.
-       *   -trs, 15 July 2022
-       */
-      if (err?.code !== 'ERR_STREAM_PREMATURE_CLOSE') throw err;
-    }
-  }
-
-  return res.end();
-}
-
-
-/**
- * Checks if the `Content-Type` of a response satisfies the `Accept` of a
- * request.
- *
- * If `accept` is falsey (i.e. the request accepts anything), returns `true`.
- *
- * If `contentType` is falsey or invalid (i.e. we don't know what type the
- * response data is), returns `false`.
- *
- * @param {string} accept - `Accept` header value from a request
- * @param {string} contentType - `Content-Type` header value from the related response
- * @returns {boolean}
- */
-function acceptableContentType(accept, contentType) {
-  if (!accept) return true;
-  if (!contentType) return false;
-
-  let mediaType;
-
-  try {
-    mediaType = parseContentType(contentType).type;
-  } catch (err) {
-    // parseContentType() throws a TypeError if contentType is bogus
-    if (!(err instanceof TypeError)) {
-      throw err;
-    }
-  }
-
-  if (!mediaType) return false;
-
-  return negotiateMediaType(accept, [mediaType]).length > 0;
-}
-
-
-/**
- * Copy a given list of headers from a source object into a new object,
- * omitting any without a value.
- *
- * @param {express.request|express.response|fetch.Headers} headerSource -
- *   Express request/response object or WHATWG Fetch API Headers object from
- *   which to copy headers.
- *
- * @param {String[]} headerNames - Array of header names to copy from the source object
- *
- * @returns {Object}
- */
-function copyHeaders(headerSource, headerNames) {
-  return Object.fromEntries(
-    headerNames
-      .map(name => [name, headerSource.get(name)])
-      .filter(([name, value]) => value != null && value !== "") // eslint-disable-line no-unused-vars
-  );
 }
 
 
--- a/src/groups.js
+++ b/src/groups.js
@@ -1,4 +1,5 @@
 import { strict as assert } from 'assert';
+import * as authz from "./authz/index.js";
 import { NotFound } from './httpErrors.js';
 
 /* eslint-disable-next-line import/first, import/newline-after-import */
@@ -82,6 +83,35 @@ class Group {
       ["editors", `${this.name}/editors`],
       ["owners", `${this.name}/owners`],
     ]);
+  }
+
+
+  /**
+   * Policy for this Group itself.  Separate from the policy for the group's
+   * Source, which is the container for the group's datasets and narratives.
+   */
+  get authzPolicy() {
+    const {Read, Write} = authz.actions;
+    const {Type} = authz.tags;
+
+    const viewers = this.membershipRoles.get("viewers");
+    const editors = this.membershipRoles.get("editors");
+    const owners = this.membershipRoles.get("owners");
+
+    return [
+      /* All membership roles in a Nextstrain Group can see information about
+       * the group itself (e.g. members, role names, etc.), but only owners can
+       * update it.  Note that the datasets and narratives within the group are
+       * authz'd separately on the GroupSource.
+       */
+      {tag: Type.Group, role: viewers, allow: [Read]},
+      {tag: Type.Group, role: editors, allow: [Read]},
+      {tag: Type.Group, role: owners, allow: [Read, Write]},
+    ];
+  }
+
+  get authzTags() {
+    return new Set([authz.tags.Type.Group]);
   }
 }
 
--- a/src/groups.js
+++ b/src/groups.js
@@ -1,6 +1,7 @@
 import { strict as assert } from 'assert';
 import * as authz from "./authz/index.js";
 import { NotFound } from './httpErrors.js';
+import { GroupSource } from "./sources/index.js";
 
 /* eslint-disable-next-line import/first, import/newline-after-import */
 import { readFile } from 'fs/promises';
@@ -85,6 +86,12 @@ class Group {
     ]);
   }
 
+  /**
+   * Source for this Group.
+   */
+  get source() {
+    return new GroupSource(this);
+  }
 
   /**
    * Policy for this Group itself.  Separate from the policy for the group's
--- a/docs/authz.rst
+++ b/docs/authz.rst
@@ -26,7 +26,7 @@ Policies
 
 There is no single policy for all of nextstrain.org but different policies for
 different parts of the site.  Currently, policies are defined for and attached
-to each :term:`Source`.
+to each :term:`Source` and :term:`Group`.
 
 The design of the system allows for policies to be easily stacked or combined
 (e.g. concatenate all the rules), so if necessary we could introduce a global
@@ -126,8 +126,7 @@ The main enforcement function used to guard access-controlled code is::
     
 It throws an ``AuthzDenied`` exception if the *user* is **not** allowed to
 perform the *action* on the *object* as determined by the policy covering the
-object (i.e.  from the object's ``Source`` currently).  Otherwise, it returns
-nothing.
+object (e.g. from the object's ``Source``).  Otherwise, it returns nothing.
 
 It is the responsibility of the enforcement function to determine the policy in
 force for the given object.
--- a/package-lock.json
+++ b/package-lock.json
@@ -466,18 +466,23 @@
       }
     },
     "@aws-sdk/abort-controller": {
-      "version": "3.52.0",
-      "resolved": "https://registry.npmjs.org/@aws-sdk/abort-controller/-/abort-controller-3.52.0.tgz",
-      "integrity": "sha512-Z+4uVtgwbKSChruh6R/WIrGb5uvvXi/d6EQ7zC6hyghtn9EGQc+WJ3BVB4bIUshwMunlgjA3nDiPb5V3t5zv8Q==",
+      "version": "3.127.0",
+      "resolved": "https://registry.npmjs.org/@aws-sdk/abort-controller/-/abort-controller-3.127.0.tgz",
+      "integrity": "sha512-G77FLYcl9egUoD3ZmR6TX94NMqBMeT53hBGrEE3uVUJV1CwfGKfaF007mPpRZnIB3avnJBQGEK6MrwlCfv2qAw==",
       "requires": {
-        "@aws-sdk/types": "3.52.0",
-        "tslib": "^2.3.0"
+        "@aws-sdk/types": "3.127.0",
+        "tslib": "^2.3.1"
       },
       "dependencies": {
+        "@aws-sdk/types": {
+          "version": "3.127.0",
+          "resolved": "https://registry.npmjs.org/@aws-sdk/types/-/types-3.127.0.tgz",
+          "integrity": "sha512-e0wtx2IkOl7rwfKfLH5pPTzQ+d45V7b1WrjeL0WDI8kOu6w+sXmhNxI6uM2kf0k4NiTLN84lW290AEWupey9Og=="
+        },
         "tslib": {
-          "version": "2.3.1",
-          "resolved": "https://registry.npmjs.org/tslib/-/tslib-2.3.1.tgz",
-          "integrity": "sha512-77EbyPPpMz+FRFRuAFlWMtmgUWGe9UOG2Z25NqCwiIjRhOf5iKGuzSe5P2w1laq+FkRy4p+PCuVkJSGkzTEKVw=="
+          "version": "2.4.0",
+          "resolved": "https://registry.npmjs.org/tslib/-/tslib-2.4.0.tgz",
+          "integrity": "sha512-d6xOpEDfsi2CZVlPQzGeux8XMwLT9hssAsaPYExaQMuYskwb+x1x7J371tWlbBdWHroy99KnVB6qIkUbs5X3UQ=="
         }
       }
     },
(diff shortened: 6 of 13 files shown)


src/sources/groups.js
```diff
@@ -255,6 +258,13 @@ class GroupNarrative extends Narrative {
 }
 
 
+class GroupOverview extends Overview {
+  get baseName() {
+    return "group-overview";
+  }
+}
```
victorlin, line 265: https://github.com/nextstrain/nextstrain.org/labels/request%20for%20comments

I initially started with 0010a642285089dbc3befc79a82b1d405dd6c639...4f5452ce84d4a5e32714f7a617997894d20204d4 which gets the URL by `GroupSource.urlFor("group-overview.md")` and uses the group's [`authz.tags.Type.Source`](https://github.com/nextstrain/nextstrain.org/blob/da26d59de198471b8948532ec3e52f2c2d780456/src/authz/tags.js#L17). Then I switched to 46990fdee31be7137ecb3f451fb61abbd23ff0c4 which object-ifys `group-overview.md`, comparable to datasets/narratives. These are 2 very different approaches. Is the second one overkill?

src/sources/models.js
```diff
@@ -302,6 +303,47 @@ class NarrativeSubresource extends Subresource {
 }
 
 
+// TODO: rename to GroupOverview if this will only be used by groups? If so:
+// 1. Change the name of sources.groups.GroupOverview
+// 2. Change the name of authz.tags.Type.Overview
+class Overview extends Resource {
```
victorlin, line 309: https://github.com/nextstrain/nextstrain.org/labels/request%20for%20comments

src/endpoints/sources.js
```diff
@@ -254,6 +254,21 @@ const getNarrative = contentTypesProvided([
 ]);
 
 
+const sendGroupOverviewSubresource = type =>
+  // TODO: link req.context.groupOverview to req.context.source.overview()?
```
victorlin, line 258: https://github.com/nextstrain/nextstrain.org/labels/request%20for%20comments

victorlin (commented):
@tsibley this is still in early stages, just focused on `GET /groups/:groupName/overview`. I have some https://github.com/nextstrain/nextstrain.org/labels/request%20for%20comments to make sure I'm doing this right!

src/sources/groups.js
tsibley: Ah, yeah. I think the second is probably overkill since this is a single, system-defined file as opposed to arbitrary user-defined sets of files.

tsibley: One consideration would be if we need write ACLs on the overview to be separate from write ACLs on the source itself, but I don't think we do. As described [in the group roles docs I wrote](https://docs.nextstrain.org/en/latest/learn/groups/index.html#roles), I think editing the overview should be purview of the owners role.

src/sources/models.js
tsibley: This is currently specific to groups with no plans/vision for extending it, so my preference is to keep it tightly scoped.

src/sources/groups.js
victorlin: Ok, I'll revert back then. But I did learn much more when making all the classes to get the second working, so that exercise didn't go to waste!

src/endpoints/sources.js
```diff
@@ -292,6 +319,27 @@ function pathParts(path = "") {
...
+ *
+ * @param {sourceExtractor} sourceExtractor - Function to provide the Source instance from the request
+ * @returns {expressEndpointAsync}
+ */
+function sendGroupOverview(sourceExtractor) {
+  return async (req, res) => {
+    const groupSource = sourceExtractor(req);
+
+    authz.assertAuthorized(req.user, authz.actions.Read, groupSource);
+
```
victorlin, line 334: I thought it would be good to add tests in authz.test.js like:

https://github.com/nextstrain/nextstrain.org/blob/0afb7b4235fd2b8575e51a25e951022bec77bd78/test/authz.test.js#L98-L102

but realized those tests only work on the dataset/narrative/group objects. This line asserts authorized access on `groupSource`, which is tested already but not with the intended effect of covering path differences (e.g. `/groups/test-private` vs. `groups/test-private/overview`).

[requests.test.js](https://github.com/nextstrain/nextstrain.org/blob/0afb7b4235fd2b8575e51a25e951022bec77bd78/test/requests.test.js) so far is testing requests without auth. It would be good to have some functional tests for requests involving auth but I'm not sure how to best implement. See PR description for manual testing which I'd like to automate.

victorlin (commented):
@tsibley progress here is just GET/PUT overview. Would you be able to review before I extend this approach to more endpoints?

src/endpoints/sources.js
tsibley: Nod. The authz.test.js file tests the authz system/framework itself is functioning as expected, but not that authz has been applied correctly to our request handlers. That would be, as you suggest, the job of requests.test.js (or a subsidiary).

Agreed it would be good to test authz of the actual handlers, e.g. using equivalents of what you're manually testing! I punted on this at the time because I didn't want to (purely due to dev time) extend the authn system to support test/dev-only fake authn… but that is ultimately the way to implement this.

src/endpoints/sources.js
```diff
@@ -400,100 +400,115 @@ async function sendAny(req, res, url, accept) {
...
-      throw new UnsupportedMediaType("unsupported Content-Encoding; only gzip is supported");
-    }
+/**
+ * Proxy the data through us:
+ *
+ *    client (browser, CLI, etc) ⟷ us (nextstrain.org) ⟷ upstream source
+ *
+ * @param {express.request} req - Express-style request instance
+ * @param {express.response} res - Express-style response instance
+ * @param {urlCallback} urlCallback - Callback to retrieve resource URL
```
tsibley, line 459: To fit with the rest of the codebase/system here, I'd call this an `upstreamUrlExtractor` (or maybe just `urlExtractor`, but I like the explicitness of what's upstream and what's not).

src/endpoints/sources.js
```diff
@@ -400,100 +400,115 @@ async function sendAny(req, res, url, accept) {
...
+ * @param {string} contentType - `Content-Type` header value to send
+ * @returns {expressEndpointAsync}
+ */
+async function receiveAny(req, res, urlCallback, contentType) {
```
tsibley, line 463: Similar to my reservations about `sendAny`'s name, I think this isn't as descriptive a name as it could be. Perhaps `receiveUrl` or `receiveToUpstream` or `proxyToUpstream`?

src/endpoints/sources.js
```diff
@@ -400,100 +400,115 @@ async function sendAny(req, res, url, accept) {
...
-       * with conditional requests.
-       *   -trs, 7 Dec 2021
-       */
-    };
+    const urlCallback = async (method, headers) => await subresource.url(method, headers);
 
-    // Body of the request as a Node stream
-    let body = req;
+    return receiveAny(req, res,
+      urlCallback,
```
tsibley, line 445: _Minor, non-blocking_

The need for `urlCallback` here and it's relationship to `receiveAny` had me briefly scratching my head first. I think it would read clearer (at least to me! ;-) to inline the function into the argument list:

```js
return receiveAny(req, res,
  async (method, headers) => await subresource.url(method, headers),
  subresource.mediaType
);
```

or

```js
return receiveAny(req, res,
  subresource.url.bind(subresource),
  subresource.mediaType
);
```

src/endpoints/sources.js
```diff
@@ -391,6 +403,29 @@ async function sendAny(req, res, url, accept) {
 }
 
 
+/**
+ * Generate an Express endpoint that receives a group overview
+ * determined by the request.
+ *
+ * @param {sourceExtractor} sourceExtractor - Function to provide the Source instance from the request
```
tsibley, line 410: Ditto about not needing `sourceExtractor` here either.

src/endpoints/sources.js
```diff
@@ -285,6 +285,18 @@ const getGroupOverview = contentTypesProvided([
...
+/* PUT
+ */
+const putGroupOverviewMarkdown = contentTypesProvided([
+  ["text/markdown", receiveGroupOverview(req => req.context.source)],
+]);
+
+
+const putGroupOverview = contentTypesProvided([
+  ["text/markdown", putGroupOverviewMarkdown],
+]);
```
tsibley, line 297: Ditto about this division of handlers.

src/endpoints/sources.js
```diff
@@ -391,6 +403,29 @@ async function sendAny(req, res, url, accept) {
...
+ * determined by the request.
+ *
+ * @param {sourceExtractor} sourceExtractor - Function to provide the Source instance from the request
+ * @returns {expressEndpointAsync}
+ */
+function receiveGroupOverview(sourceExtractor) {
+  return async (req, res) => {
+    const groupSource = sourceExtractor(req);
+
+    authz.assertAuthorized(req.user, authz.actions.Read, groupSource);
```
tsibley, line 417: `authz.actions.Write`, not read!

src/endpoints/sources.js
```diff
@@ -391,6 +403,29 @@ async function sendAny(req, res, url, accept) {
...
+
+    const urlCallback = async (method, headers) => await groupSource.urlFor("group-overview.md", method, headers);
+
+    return await receiveAny(req, res,
+      urlCallback,
+      "text/markdown" // TODO: put this in a class?
+    );
```
tsibley, line 424: Ditto about inlining `urlCallback` and not worrying about abstracting the media type.

tsibley (changes requested):
Overall I think this is the right direction!

Left lots of minor notes and comments for your discretion, plus a few actual requested changes.

tsibley:
BTW, in your testing notes above, the PUT of the overview markdown uses Accept instead of Content-Type.

src/endpoints/sources.js
victorlin: @tsibley

> extend the authn system to support test/dev-only fake authn

I want to do this, but not sure where to start or what's appropriate. I've looked over the relevant pages in [the dev docs](https://docs.nextstrain.org/projects/nextstrain-dot-org/en/latest/index.html). A few questions:

1. Does this call for a new strategy for `passport.authenticate()`, something like [passport-mock-strategy](https://github.com/cszatmary/passport-mock-strategy)?
2. If these tests are added in requests.test.js, do we need to send requests with header `Authorization: Bearer <token>`, like [how `Accept: <mediatype>` is currently sent](https://github.com/nextstrain/nextstrain.org/blob/98e6c6c186edc0f867f81cfb30da7e7c88a76652/test/requests.test.js#L305-L311)? Or would a mock strategy from (1) eliminate the need for this?
3. Would it be appropriate to create "test users" in the `nextstrain.org` Cognito User Pool under `test`/`test-private` groups, with varying role membership? If so, what's the proper way to authenticate as these users?

tsibley: 1. I don't think a new Passport strategy is necessary, because…
2. …we can modify what JWTs [our Bearer auth accepts](https://github.com/nextstrain/nextstrain.org/blob/98e6c6c186edc0f867f81cfb30da7e7c88a76652/src/authn/index.js#L133-L146) in dev/test mode and then pass something acceptable in the test requests. Very important to make sure the dev/test code paths can't be enabled in production.
3. I don't think this will be necessary now (we can just pretend users exist), although it may be in the future. In either case, we would authenticate as them by the tests offering a fake JWT, not by interacting with Cognito.

FWIW, while automated tests are good here, I don't think we need to block implementation or deploy on having tests sorted out.

src/sources/groups.js
tsibley: I'm thinking about authz again and how it relates to groups as part of implementing the API for group membership management. In that work, I need to control the ability to modify the group itself (e.g. add/remove members) separately from the ability to modify the _contents_ of the group (e.g. the group's Source, with its datasets and narratives). I'm (so far) opting to add `authz.tags.Type.Group` and run authz checks for the membership management endpoints against the `Group` instead of the `GroupSource`. This had me thinking about our brief discussion here.

I realized that how I'm extending the authz system is applicable to modifying the group overview and logo, and perhaps is what those endpoints here should be authz'd using.

In re-reading my commentary here, I also realized I made a mistake. I said:

> One consideration would be if we need write ACLs on the overview to be separate from write ACLs on the source itself, but I don't think we do. As described [in the group roles docs I wrote](https://docs.nextstrain.org/en/latest/learn/groups/index.html#roles), I think editing the overview should be purview of the owners role.

But the ACLs on the GroupSource itself allow not just owners the Write action, but also editors. This is contrary to what I think should be the case!

src/endpoints/sources.js
```diff
@@ -305,40 +305,53 @@ function sendSubresource(subresourceExtractor) {
 
     authz.assertAuthorized(req.user, authz.actions.Read, subresource.resource);
 
-    const subresourceUrl = await subresource.url();
+    return await sendAny(req, res,
```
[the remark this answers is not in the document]
victorlin: Renamed to `proxyFromUpstream` in 5e91e82fff4cd673d424674958f428a99a207299 + 8e8fad78f9500b3338ddf27b5b62d38755226b23.

src/endpoints/sources.js
```diff
@@ -270,6 +270,21 @@ const putNarrative = contentTypesConsumed([
...
+/* GET
+ */
+const getGroupOverviewMarkdown = contentTypesProvided([
+  ["text/markdown", sendGroupOverview(req => req.context.source)],
+]);
+
+
+const getGroupOverview = contentTypesProvided([
+  ["text/markdown", getGroupOverviewMarkdown],
+]);
```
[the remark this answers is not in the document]
victorlin: Yeah, I largely followed the existing patterns for narrative endpoints since I imagined they'd be the most similar to group overview given Markdown type. 

So is the suggestion here to remove the `getGroupOverviewMarkdown` intermediate? I'm confused because
your other suggestions https://github.com/nextstrain/nextstrain.org/pull/570#discussion_r924794105 and https://github.com/nextstrain/nextstrain.org/pull/570#discussion_r924795051 seem to rely on this intermediate.

src/endpoints/sources.js
```diff
@@ -292,6 +307,27 @@ function pathParts(path = "") {
...
 
 
+/**
+ * Generate an Express endpoint that sends a group overview
+ * determined by the request.
+ *
+ * @param {sourceExtractor} sourceExtractor - Function to provide the Source instance from the request
+ * @returns {expressEndpointAsync}
+ */
+function sendGroupOverview(sourceExtractor) {
```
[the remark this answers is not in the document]
victorlin: Oh, didn't notice `sourceExtractor` is currently used for `setSource` to populate `req.context.source`. I see it's not needed here. Removed: 24ef89db41c36207fc18cca6ec84363f3e942f5a

src/endpoints/sources.js
```diff
@@ -292,6 +307,27 @@ function pathParts(path = "") {
...
+  return async (req, res) => {
+    const groupSource = sourceExtractor(req);
+
+    authz.assertAuthorized(req.user, authz.actions.Read, groupSource);
+
+    return await sendAny(req, res,
+      await groupSource.urlFor("group-overview.md"),
+      "text/markdown" // TODO: put this in a class?
```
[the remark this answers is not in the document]
victorlin: Removed TODO: f4755d41ccfb263932b04f6695eec9b053b8422b

src/endpoints/sources.js
victorlin: Renamed to `upstreamUrlExtractor` and defined the type which I forgot to do earlier: a7e98021804bd257bbffaf1eb64dc4de5444ccba

src/endpoints/sources.js
victorlin: Renamed to `proxyToUpstream` in 50c4c1eeaf8ec8dd6c605a55624165dd48293e42 + 054f05b21089def72659626d866410f40642b706.

src/endpoints/sources.js
victorlin: Inlined: f233f81f5cbd18de7ba3b95c2688dd0d89e26646.

I'm not familiar with `bind()`, gonna try to understand that one tomorrow.

src/endpoints/sources.js
victorlin: Removed: 84e61b76ce5b59ee5e9fab51f42bfc19733bf4c6

src/endpoints/sources.js
victorlin: Oops, fixed in 9168b758615c8163486b6c8a6cee2e9415116b32.

src/endpoints/sources.js
victorlin: Inlined: 21a91cfd89eafae234ec860224638f94259416fb

Removed TODO: ed4ce9bb50750c03c5d3f1ffb801c13908795652

victorlin:
> BTW, in your testing notes above, the PUT of the overview markdown uses Accept instead of Content-Type.

Good catch, but also this means it worked without the `Content-Type` header when it shouldn't. Will fix this.

src/endpoints/sources.js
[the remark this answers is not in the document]
tsibley: Ah, suggestion is either to a) remove `getGroupOverviewMarkdown` or b) keep it but make the other two linked changes. So correct, not both. :-) I think I have a slight preference for (a), but it's not a big deal either way.

tsibley:
Ah, good point. Looking closer, that's because the PUT endpoints are using `contentTypesProvided` instead of `contentTypesConsumed`.

src/endpoints/sources.js
```diff
@@ -351,100 +431,113 @@ function sendSubresource(subresourceExtractor) {
...
+ *
+ *    client (browser, CLI, etc) ⟷ us (nextstrain.org) ⟷ upstream source
+ *
+ * @param {express.request} req - Express-style request instance
+ * @param {express.response} res - Express-style response instance
+ * @param {upstreamUrlExtractor} upstreamUrlExtractor - Callback to retrieve resource URL
+ * @param {string} contentType - `Content-Type` header value to send
+ * @returns {expressEndpointAsync}
+ */
+async function proxyToUpstream(req, res, urlCallback, contentType) {
```
tsibley, line 457: `urlCallback` → `upstreamUrlExtractor` in the function signature too.

victorlin: 🤦 fixed: f171d88f85d1b75fea3b0826be0d55ba7ee2273b

src/endpoints/sources.js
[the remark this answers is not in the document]
victorlin: Ok, removed `getGroupOverviewMarkdown`: 627831b36b9e35e7542b3d3f0ceee15270ec685b

src/endpoints/sources.js
```diff
@@ -270,6 +270,21 @@ const putNarrative = contentTypesConsumed([
 ]);
 
 
+/* Group customizations
+ */
+
+/* GET
+ */
+const getGroupOverviewMarkdown = contentTypesProvided([
+  ["text/markdown", sendGroupOverview(req => req.context.source)],
```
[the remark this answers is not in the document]
victorlin: This isn't needed after 627831b36b9e35e7542b3d3f0ceee15270ec685b

src/endpoints/sources.js
```diff
@@ -270,6 +270,21 @@ const putNarrative = contentTypesConsumed([
...
+]);
+
+
+const getGroupOverview = contentTypesProvided([
+  ["text/markdown", getGroupOverviewMarkdown],
```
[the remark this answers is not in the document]
victorlin: Added `text/plain` in 627831b36b9e35e7542b3d3f0ceee15270ec685b.

src/endpoints/sources.js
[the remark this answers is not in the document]
victorlin: Forgot to remove parameter from the function call earlier: 9d14fe47cabcf156b2770d11fac535c1d6181b3f

src/endpoints/sources.js
victorlin: also 899a5e9cc73358a48969fefa0c38eaa0174b79d6

src/endpoints/sources.js
victorlin: Done: 04311615a7d85cba1f0c07bb32162e62e5b0999f

Also fixed to use `contentTypesConsumed`: 8aa733ca29f41f5eab8b2edbf8a060d966dbd937

src/sources/groups.js
victorlin: I see, yeah the current implementation allows editors to edit overview when it should be just owners.

> add `authz.tags.Type.Group` and run authz checks for the membership management endpoints against the `Group` instead of the `GroupSource`

The difference between `Group` vs. `GroupSource` isn't entirely clear to me. Is the `Group` also available in `req.context` like the `GroupSource` currently is through `req.context.source`?

https://github.com/nextstrain/nextstrain.org/blob/25bccef5b004c3e4965c8ba6c2af3c9d0cf0fa15/src/endpoints/sources.js#L319-L323

src/endpoints/sources.js
```diff
@@ -305,39 +343,72 @@ function sendSubresource(subresourceExtractor) {
...
+ * Generate an Express endpoint that receives a group overview
+ * determined by the request.
+ *
+ * @returns {expressEndpointAsync}
+ */
+function receiveGroupOverview() {
+  return async (req, res) => {
+    const groupSource = req.context.source;
+
+    authz.assertAuthorized(req.user, authz.actions.Write, groupSource);
```
victorlin, line 406: Hmm, after 8aa733ca29f41f5eab8b2edbf8a060d966dbd937 seems like this doesn't work anymore:

```sh
curl http://localhost:5000/groups/test-private/overview \
    --header "Authorization: Bearer $token" \
    --header 'Content-Type: text/markdown' \
    --upload-file overview.md
```

I get a 403 when logged in as my user, a member of `test-private/owners`. Gonna take a deeper look at this next week, much easier now with [a nice VSCode debugging setup](https://gist.github.com/victorlin/e9e1177aef9cdd0d14ddbc10c54434da).

src/endpoints/sources.js
```diff
@@ -270,6 +270,24 @@ const putNarrative = contentTypesConsumed([
...
+  ["text/markdown", sendGroupOverview()],
+  ["text/plain", sendGroupOverview()],
+]);
+
+
+/* PUT
+ */
+const putGroupOverview = contentTypesConsumed([
+  ["text/markdown", receiveGroupOverview()],
+]);
```
victorlin, line 288: I'm also curious why this "worked" before 8aa733ca29f41f5eab8b2edbf8a060d966dbd937, when using `contentTypesProvided` instead of `contentTypesConsumed`.

src/endpoints/sources.js
tsibley: Did your token expire?

victorlin: No, I'm able to `GET` the file right before `PUT`. If I pass an invalid token (I assume this is equivalent to an expired token) I get `401 Unauthorized` which makes sense, but this is `403 Forbidden`.

victorlin: This was fixed by 5fce14b7cbac318456f293cece839847797eab7e.

src/sources/groups.js
victorlin: > the ACLs on the GroupSource itself allow not just owners the Write action, but also editors.

Is this true? Based on the snippet below, I don't think any role currently has Write access to the GroupSource itself:

https://github.com/nextstrain/nextstrain.org/blob/0975bc0650f13c8b0a7abd57f9d8c7e5cd0e5684/src/sources/groups.js#L279-L284

From your statement I'd expect it to be:

```js
{tag: Type.Source, role: viewers, allow: [Read]}, 
{tag: Type.Source, role: editors, allow: [Read, Write]}, 
{tag: Type.Source, role: owners,  allow: [Read, Write]}, 
```

src/endpoints/sources.js
victorlin: Oh duh, this "worked" before 9168b758615c8163486b6c8a6cee2e9415116b32.

src/endpoints/sources.js
victorlin: Now I understand how `bind()` works here, but choosing to keep the first option. I think it's more readable mostly because the parser on my text editor does not provide any context on `subresource.url`, so I have to manually search for [the `url` property under `class Subresource`](https://github.com/nextstrain/nextstrain.org/blob/64a338e92f9a7e3d413503455d523578f75cef60/src/sources/models.js#L190) to understand that it is a function which takes `method` and `headers` (same for [`urlFor` under `class GroupSource`](https://github.com/nextstrain/nextstrain.org/blob/64a338e92f9a7e3d413503455d523578f75cef60/src/sources/groups.js#L72)). Having the explicit call makes it easier to tell what's going on.

src/endpoints/sources.js
```diff
@@ -292,6 +310,26 @@ function pathParts(path = "") {
...
 
 
+/**
+ * Generate an Express endpoint that sends a group overview
+ * determined by the request.
+ *
+ * @returns {expressEndpointAsync}
+ */
+function sendGroupOverview() {
+  return async (req, res) => {
```
tsibley, line 320: Now that this doesn't take any parameters, there's no need for the indirection of returning a function. The `sendGroupOverview` symbol can have the `async (req, res)` signature directly instead. Ditto with `receiveGroupOverview`.

victorlin: Done: a28c3c97475cac2c209e4d02cf8cab678dc6717a, 6f3867d88476b6d7c191faf0b2f54d24d98d16a9

src/sources/groups.js
tsibley: > The difference between `Group` vs. `GroupSource` isn't entirely clear to me.

My mental model distinguishes between the concepts of groups and sources like so:

Sources are abstract containers for datasets and narratives and provide the interface for shared code to work across the various concrete ways we store/access datasets and narratives (core, staging, community, "fetch", groups, etc). Sources are a concept mostly internal to the codebase, i.e. not exposed to users.

Groups have members, roles, a visibility, overview/description, a logo, etc. and also contain datasets and narratives. So every group has a source, but not all sources have a group. Groups are an externally-visible concept, i.e. exposed to users.

See also the [glossary](https://docs.nextstrain.org/projects/nextstrain-dot-org/en/latest/glossary.html#term-Source) and [previous commentary](https://github.com/nextstrain/nextstrain.org/issues/76#issuecomment-574303470).

This said, there are likely some places in the codebase where these two concepts are currently a bit blurred together. However, I think we should be further distinguishing them over time. (And this is what I'm doing in my work on implementing the API for group membership management.)

> Is the `Group` also available in `req.context` like the `GroupSource` currently is through `req.context.source`?

Not currently, as it's not needed, but it could be.

> Is this true? Based on the snippet below, I don't think any role currently has Write access to the GroupSource itself:

You're right! :man_facepalming: I confused myself on my earlier re-read and thought I'd made a mistake I hadn't, thus making a new mistake instead. :upside_down_face:

victorlin: Ok! Sounds like a new `authz.tags.Type.Group` will be ideal here instead of 2adabdb426743f2c4b254ceb2cbc7bcbe1e9ec6b. If you already have work in progress for that, I don't mind waiting until that's merged to use it here.

src/endpoints/sources.js
```diff
@@ -305,40 +342,72 @@ function sendSubresource(subresourceExtractor) {
...
+ *    client (browser, CLI, etc) ⟷ us (nextstrain.org) ⟷ upstream source
+ *
+ * @param {express.request} req - Express-style request instance
+ * @param {express.response} res - Express-style response instance
+ * @param {String} url - Resource URL
+ * @param {String} accept - HTTP Accept header value
+ * @returns {expressEndpointAsync}
+ */
+async function proxyFromUpstream(req, res, url, accept) {
+  return await proxyResponseBodyFromUpstream(req, res, new Request(url, {
```
victorlin, line 365: `@returns {expressEndpointAsync}` isn't right here since it's effectively:

https://github.com/nextstrain/nextstrain.org/blob/2adabdb426743f2c4b254ceb2cbc7bcbe1e9ec6b/src/endpoints/sources.js#L632

victorlin: Fixed: ecc6916b0db78d4066aad838ed8c6ceed75b72d0, aaaf67cb9335ead33ee44f690a90d63f6c706259

victorlin:
@tsibley I added DELETE which was simpler: 1542f8fb56cc3472b72eb2d457cef6393532ff34. Could you review before I use these approaches for `/groups/:groupName/logo`?

src/sources/groups.js
tsibley: Ok. See https://github.com/nextstrain/nextstrain.org/pull/581/commits/b54b35ac4309b9a0d8c0f792b8fb8508d0793c72.

tsibley:
@victorlin Functionally, that DELETE implementation looks good by inspection to me! Organizationally, I'm not sure it makes sense to have `proxyDeleteUpstream` as-is. I'd probably either a) inline it into `deleteGroupOverview` (like the similar inline in `deleteResource`) or b) refactor `deleteResource`'s multi-URL implementation into a `delete(urls: String[])` function and use it in both `deleteResource` and `deleteGroupOverview`. Happy to sketch out either or both tomorrow if that'd be clearer.

One thing I note is that there's scant little "proxying" happening, particularly since no client request headers are sent upstream (`req` isn't even passed into `proxyDeleteUpstream`) and no upstream response headers/content make it back to the client (`res` is passed in, but only used to send a static 204). The proxy moniker would make more sense to me if intended to support conditional deletes, e.g. with [`If-Match`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Match) or [`If-Unmodified-Since`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/If-Unmodified-Since) headers, but S3 doesn't support those so we can't.

src/sources/groups.js
victorlin: Planning to rebase onto b54b35ac4309b9a0d8c0f792b8fb8508d0793c72 and move all the changes in `src/endpoints/sources.js` to a new `src/endpoints.groups.js`, to be consistent with #581.

src/endpoints/sources.js
```diff
@@ -270,6 +270,56 @@ const putNarrative = contentTypesConsumed([
...
+};
+
+/**
+ * Send a DELETE request to the upstream URL.
+ *
+ * @param {express.response} res - Express-style response instance
+ * @param {upstreamUrlExtractor} upstreamUrlExtractor - Callback to retrieve resource URL
+ */
+// TODO: use this function in deleteResource to avoid code duplication?
+async function proxyDeleteUpstream(res, upstreamUrlExtractor) {
```
victorlin, line 309: Replying to https://github.com/nextstrain/nextstrain.org/pull/570#issuecomment-1225010406 in a "thread".

> Organizationally, I'm not sure it makes sense to have `proxyDeleteUpstream` as-is. I'd probably either a) inline it into `deleteGroupOverview` (like the similar inline in `deleteResource`) or b) refactor `deleteResource`'s multi-URL implementation into a `delete(urls: String[])` function and use it in both `deleteResource` and `deleteGroupOverview`. Happy to sketch out either or both tomorrow if that'd be clearer.

That makes sense. I did a sketch of both: (a) f5acfc7c4b9e1b3aa75a5192dff2521b2932da52 (b) 13ff607117fef230cf0eb9eb60cdd3418beb16d7

Here's some pros and cons to both as-is:

- f5acfc7c4b9e1b3aa75a5192dff2521b2932da52
    - Pro: has a custom error message specifying what failed to be deleted.
    - Con: duplicate logic around `fetch` and checking the response status.
- 13ff607117fef230cf0eb9eb60cdd3418beb16d7
    - Pro: less code duplication.
    - Con: error message is generic: `At least one delete request failed.`

Maybe less code duplication is preferred, otherwise it would be duplicated 3 times (`deleteResource`, `deleteGroupOverview`, `deleteGroupLogo`) and potentially more in the future.

To mitigate the con of generic error messages, we could add a parameter `deleteByUrls(urls, errorMessage)` but not sure if that's too extra.

tsibley: I like the refactored version in 13ff607. I don't think we need a second `errorMessage` parameter. Instead, we could extend the `deleteByUrls` function to produce a more specific error message with the URL(s) that failed and their statuses. Also note that, IIRC, the error message will be for our eyes only during troubleshooting/debugging; it won't be shown to the end user, who will just see a generic "server error" message of some sort. So any error message improvements we make here are for our own sake, and we could punt on the errors now at the expensive of our future selves. ;-)

src/sources/groups.js
tsibley: Maybe base off 29449eb instead? Contains work I'd forgotten to push up yesterday.

tsibley: You can see an example of how it's used in 9431522e9c099c0b970a25f8b179e7da887e81f9.

src/endpoints/groups.js
```diff
@@ -0,0 +1,100 @@
...
+]);
+
+
+/* DELETE
+ */
+const deleteGroupOverview = async (req, res) => {
+  authz.assertAuthorized(req.user, authz.actions.Write, req.context.group);
+
+  // TODO: somehow have a req.context.group.urlFor?
+  const groupSource = req.context.source;
```
victorlin, line 47: Conceptually, the Group overview should be retrieved from the `Group` rather than the `GroupSource`. However, with the current functionality, there isn't a way to retrieve the S3 object URL from the `Group` itself. It just happens to work with `GroupSource.urlFor`, even though I assume the intended use case is for datasets/narratives. `Group.urlFor` is achievable though, see 3908f012d5136d3e01ba211c7f8f681257948c9a as an example.

Or, for a bigger refactor, should the method of storage (S3 here) be its own entity that's shared between both the `Group` and `GroupSource`?

src/sources/groups.js
victorlin: Done. Added some new thoughts after the rebase: https://github.com/nextstrain/nextstrain.org/pull/570#pullrequestreview-1087578047

src/upstreamRequestHandlers.js
```diff
@@ -0,0 +1,353 @@
+const {parse: parseContentType} = require("content-type");
```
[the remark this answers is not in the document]
tsibley: I might call it just `src/upstream.js`. But also fine as is, even if not the catchiest.

tsibley (commented):
Aside from the discussion around `req.context.source` and `GroupSource` in the other thread, this latest iteration looks like a good direction to me.

src/endpoints/groups.js
tsibley: Consider this, for example: https://gist.github.com/tsibley/a42514eca2a44a73e5bf7eb33cfe3235

test/requests.test.js
```diff
@@ -242,6 +242,40 @@ describe("narratives", () => {
   }
 });
 
+describe("groupOverview", () => {
+
+  describe("groups", () => {
+    testPaths([
+      { case: "overview", path: "/groups/test/settings/overview" },
+    ]);
+  });
```
victorlin, line 251: These tests now fail after rebasing to include d36f8441e0fd7d9d288cd92aa898caeb356b8a9d and use the `authz.tags.Type.Group` on the overview endpoint, without conditionally adding `authz.tags.Visibility.Public` here:

https://github.com/nextstrain/nextstrain.org/blob/d36f8441e0fd7d9d288cd92aa898caeb356b8a9d/src/groups.js#L112-L114

`Visibility.Public` seems intended for datasets/narratives, so this is probably fine. I'll remove these tests given manual tests cover the same thing with authz.

src/upstreamRequestHandlers.js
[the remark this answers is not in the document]
victorlin: Renamed to `src/upstream.js`, that seems more consistent with the other filenames.

test/requests.test.js
victorlin, line 251: These tests now fail after rebasing to include d36f8441e0fd7d9d288cd92aa898caeb356b8a9d and use the `authz.tags.Type.Group` on the group overview endpoint, without conditionally adding `authz.tags.Visibility.Public` here:

https://github.com/nextstrain/nextstrain.org/blob/d36f8441e0fd7d9d288cd92aa898caeb356b8a9d/src/groups.js#L112-L114

`Visibility.Public` seems intended for datasets/narratives, so this is probably fine. I'll remove these tests given manual tests cover the same thing with authz.

victorlin: Nevermind. After inspecting failing tests, I think adding `Visibility.Public` to the Group's authz policy is needed if we want to `setGroup()` on the base endpoint `/groups/:groupName`:

https://github.com/nextstrain/nextstrain.org/blob/69ae64418f0a2e5ab3da03e364de0382ca3a9603/src/app.js#L286-L287

since that still calls `assertAuthorized()` on the `Group`:

https://github.com/nextstrain/nextstrain.org/blob/69ae64418f0a2e5ab3da03e364de0382ca3a9603/src/endpoints/groups.js#L8-L11

src/groups.js
```diff
@@ -81,6 +91,58 @@ class Group {
...
+    const publicPolicy = this.authzPublicPolicy();
+
+    return basePolicy.concat(publicPolicy);
+  }
+
+  /**
+   * A public policy for the group if appropriate.
+   * To be used in authzPolicy and the group's Source.
+   */
+  authzPublicPolicy() {
```
victorlin, line 128: I added this in 371efb9c77e4fc37f4644ae879a1a63a3a877434 but might be appropriate to add alongside d36f8441e0fd7d9d288cd92aa898caeb356b8a9d in #581

test/requests.test.js
tsibley: > since that still calls `assertAuthorized()` on the `Group`:

Ah, [it doesn't](https://gist.github.com/tsibley/a42514eca2a44a73e5bf7eb33cfe3235#file-nopaste-diff-L105-L115) in [the patch I pasted](https://github.com/nextstrain/nextstrain.org/pull/570#discussion_r957687615). (Though that patch was off-the-cuff so may not be fully baked in other ways.)

src/groups.js
tsibley: Hmm. My first reaction is that we don't want to mark the Group with the Public authz tag because we want to condition members-only stuff like members and member roles on Read, even for public groups.

Also, it's likely ok for the management endpoints here for the overview/logo to be non-public even if the group's overview and logo are otherwise publicly viewable. Or we can change where authz is gated in the routing stack (e.g. like I noted in the comment I made a few minutes ago).

So I think there's a few ways to get from here to there, but all these intersecting bits may be easier to talk out in person/real time (though unfortunately it looks like I'll be out with a sick kiddo tomorrow).

victorlin: Ok, I just pushed 055efa733fdf52ea237b11f43198345afb242831 which removes authorization check on `/groups/:groupName`, and also dropped the test a1dec33b035e594e974a64a896fbe1534900eb9d since that will fail on authorization now (when I added it, it worked by using the public read access on `GroupSource` but we've moved to using `Group` as discussed in https://github.com/nextstrain/nextstrain.org/pull/570#discussion_r953081820).

test/requests.test.js
victorlin: I see, somehow I missed that while splitting your patch up. I added that one line change in 055efa733fdf52ea237b11f43198345afb242831 with my understanding of its purpose in the commit message.

src/endpoints/groups.js
```diff
@@ -12,6 +15,86 @@ const setGroup = (nameExtractor) => (req, res, next) => {
...
+
+  const method = "DELETE";
+  const url = await groupSource.urlFor("group-overview.md", method);
```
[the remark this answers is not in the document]
victorlin: This doesn't work, I think because it's ES6 syntax but this project uses CommonJS.

This works though, noting that the name `method` is irrelevant to the interpreter (`foo="DELETE"` works just fine):

```js
urlFor("group-overview.md", method="DELETE");
```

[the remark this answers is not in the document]
victorlin: Oh nevermind, `method="DELETE"` fails on eslint's `no-undef`. My understanding is that `method="DELETE"` is an assignment that returns the value being assigned, based off of this [SO post](https://stackoverflow.com/q/16027247). It only "works" because the value is returned.

I don't think we want this here, so I'll just revert to the original code.

src/endpoints/groups.js
victorlin: Ok, this has been addressed by 82e3909318e6244a9c13dddc335721fdb198e721.

---

Re: your patch, I applied parts of it in 39d4db2fb6a5e7be49edd239427609a4d4296329...055efa733fdf52ea237b11f43198345afb242831. However, I'm a bit unsure about adding the Group `authz.actions.Read` assertion on `/groups/:groupName/settings`. Would that also suggest removing the existing read assertions on specific endpoints such as this?

https://github.com/nextstrain/nextstrain.org/blob/1de0bc91faf8eb21c4be7f2234b20b085950649a/src/endpoints/groups.js#L92-L98

I prefer keeping the calls to `assertAuthorized` closer to the invocation like how it is currently.

I also prefer having the calls to `assertAuthorized` closer to each other like how both `sendGroupOverview` and `receiveGroupOverview` have calls. I'd be more OK with the suggestion if read was the only option under `/groups/:groupName/settings`.

src/endpoints/sources.js
victorlin: Ok, I'll punt on it for now.

Maybe something like 34f02d57e29d99d2725fa7bb2dc20a2a8ee49866 can work if we want failed URLs and statuses.

src/endpoints/groups.js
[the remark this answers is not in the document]
tsibley: So sorry, my suggestion misled you because I made a minor mistake! I didn't mean to suggest using named parameters. I thought the original call was `urlFor(…, {method})` and so suggested `urlFor(…, {method: "DELETE"})`. I should have suggested `urlFor(…, "DELETE")`.

src/endpoints/groups.js
tsibley: > Would that also suggest removing the existing read assertions on specific endpoints such as this?

No. It wasn't my intention to suggest that.

> I prefer keeping the calls to `assertAuthorized` closer to the invocation like how it is currently. … I also prefer having the calls to `assertAuthorized` closer to each other…

Yes, agreed.

src/endpoints/groups.js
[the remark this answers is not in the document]
victorlin: Ah ok, I see where you're coming from now. If the original call was `urlFor(…, {method})` then I would be fully on board with the original suggestion. However, I think I'll opt to keep it as-is, since:

1. Passing it directly as `urlFor(…, "DELETE")` loses the small bit of context that `"DELETE"` is an HTTP method. Even if that can be obvious with everything else around it, I prefer having the extra bit of context. Especially since my IDE (I assume others too) can't determine the function signature for `groupSource.urlFor`.
2. Consistency - `const method = "DELETE"` is also used elsewhere in the file.

tsibley:
@victorlin Sorry this fell off my radar for a bit. I submitted some comments above (some of which were sitting pending for a week without me realizing they hadn't been submitted :-/). I want to get this across the line soon—it's been dragging out thru no fault of your own—so I'll prioritize more conversation around it. I also need to swap context back in for how it interacts with #581 and your comments there.

src/app.js
```diff
@@ -300,19 +308,38 @@ app.routeAsync("/groups/:groupName")
   .getAsync(endpoints.static.sendGatsbyEntrypoint)
 ;
```
tsibley, line 309: I'm still not sure the authz layers are quite right in this latest pass. For example, this route isn't protected any more and it should be.

victorlin:
Force-pushing rebase onto latest changes from #583.

src/endpoints/sources.js
```diff
@@ -351,275 +303,18 @@ function sendSubresource(subresourceExtractor) {
...
   return async (req, res) => {
-    const method = "PUT";
     const subresource = subresourceExtractor(req);
 
     authz.assertAuthorized(req.user, authz.actions.Write, subresource.resource);
 
-    /* Notify the client they can continue sending the request body now that
-     * we're past authz.
-     */
-    if (req.expectsContinue) res.writeContinue();
```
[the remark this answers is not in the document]
victorlin: Good catch! Fixed in 932b01673b67e18dfa1b491806cd801d36d5eec7.

src/sources/groups.js
```diff
@@ -15,50 +14,55 @@ const MULTI_TENANT_BUCKET = "nextstrain-groups";
 
 
 class GroupSource extends Source {
-  constructor(groupOrName) {
+  constructor(name) {
     super();
+    this.name = name;
+  }
```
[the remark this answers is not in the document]
victorlin: I briefly mentioned this in the commit message of 82e3909318e6244a9c13dddc335721fdb198e721 but it's not obvious from the change itself. This is all because of circular dependencies. If both are to be set in their constructors, we can have either `Group.source` or `GroupSource.group`, but not both.

For example, errors will arise if you try to set `this.group` in the constructor here.

In CJS, this was a bit cryptic:

```
nextstrain.org/src/sources/groups.js:20
    this.group = groupOrName instanceof Group
                             ^

TypeError: Right-hand side of 'instanceof' is not an object
```

With ESM, it's more obvious:

```
RangeError: Maximum call stack size exceeded
    at normalizeGroupName (file://.../nextstrain.org/src/groups.js:214:15)
    at new Group (file://.../nextstrain.org/src/groups.js:43:43)
    at new GroupSource (file://.../nextstrain.org/src/sources/groups.js:23:9)
    at new Group (file://.../nextstrain.org/src/groups.js:58:19)
    at new GroupSource (file://.../nextstrain.org/src/sources/groups.js:23:9)
    at new Group (file://.../nextstrain.org/src/groups.js:58:19)
    at new GroupSource (file://.../nextstrain.org/src/sources/groups.js:23:9)
    at new Group (file://.../nextstrain.org/src/groups.js:58:19)
    at new GroupSource (file://.../nextstrain.org/src/sources/groups.js:23:9)
    at new Group (file://.../nextstrain.org/src/groups.js:58:19)
    at new GroupSource (file://.../nextstrain.org/src/sources/groups.js:23:9)
    at new Group (file://.../nextstrain.org/src/groups.js:58:19)
    at new GroupSource (file://.../nextstrain.org/src/sources/groups.js:23:9)
    at new Group (file://.../nextstrain.org/src/groups.js:58:19)
    at new GroupSource (file://.../nextstrain.org/src/sources/groups.js:23:9)
    at new Group (file://.../nextstrain.org/src/groups.js:58:19)
npm ERR! code ELIFECYCLE
```

src/groups.js
```diff
@@ -1,5 +1,9 @@
 import { strict as assert } from 'assert';
+import authzActions from "./authz/actions.js";
+import authzTags from "./authz/tags.js";
```
victorlin, line 3: @tsibley

I removed the comment here in the rebase onto ESM:

https://github.com/nextstrain/nextstrain.org/blob/d36f8441e0fd7d9d288cd92aa898caeb356b8a9d/src/authz/index.js#L135-L144

However, circular imports are still an issue. After moving the import back to top-level in 6401dda32e25de051d7b77323325fdc2439aef24, tests still fail due to circular imports. This one is easy to fix though, since a more selective import isn't circular anymore: c2fe5d84e21a659143ae498af3bd5f3b2a2a4499

Does this seem reasonable to you?

src/groups.js
```diff
@@ -48,6 +50,12 @@ class Group {
     this.name = groupRecord.name;
     assertValidGroupName(this.name);
 
+    /**
+     * Source for this Group.
+     * @type {GroupSource}
+     */
+    this.source = new GroupSource(this.name);
+
```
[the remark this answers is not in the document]
victorlin: Duplicate of https://github.com/nextstrain/nextstrain.org/pull/570#discussion_r965317329

src/sources/groups.js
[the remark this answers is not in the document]
victorlin: I refactored the ESM to have a top-level import which differs from the non-global `require` in `GroupSource.group` (didn't want to deal with non-global async `await import()`). This means circular dependency is an issue again, at least in the tests. Instead of trying to hack around it, I think there's an inherent issue with trying to cross-reference the two classes within each other. When I started on 1258b213335f4f1dd623f0b627b74c4de17f675b, my goal was the one-way referencing described in the title which would avoid circular dependency, but it's not trivial (and maybe not desirable) to remove `GroupSource.group`.

An alternative would be to revert to the original one-way referencing of GroupSource -> Group by dropping 1258b213335f4f1dd623f0b627b74c4de17f675b in favor of something like 9123008895a012a76323039266e93522291e72ee.

src/sources/groups.js
```diff
@@ -15,50 +14,55 @@ const MULTI_TENANT_BUCKET = "nextstrain-groups";
...
 
-    this.group = groupOrName instanceof Group
-      ? groupOrName
-      : new Group(groupOrName);
+  /**
+   * Group of this GroupSource.
+   */
+  group() {
+    const {Group} = require("../groups"); // eslint-disable-line global-require
+    return new Group(this.name);
```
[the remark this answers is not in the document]
victorlin: This is an easy change, added in 7d3d2541ee752183b9c8892cbb3f23dec887c91e. But see the alternative proposed in https://github.com/nextstrain/nextstrain.org/pull/570#discussion_r975908086

src/groups.js
```diff
@@ -4,6 +4,8 @@ const {NotFound} = require("http-errors");
 const authz = require("./authz");
 
 const GROUPS_DATA = require("../data/groups.json");
+// TODO: figure out why this doesn't work without /groups
+const {GroupSource} = require("./sources/groups");
```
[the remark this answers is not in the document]
victorlin: The error with CJS was cryptic:

```
nextstrain.org/src/groups.js:57
    this.source = new GroupSource(this.name);
                  ^

TypeError: GroupSource is not a constructor
```

But with ESM now it's more obvious:

```
nextstrain.org/src/groups.js:59
    this.source = new GroupSource(this.name);
                      ^

ReferenceError: Cannot access 'GroupSource' before initialization
```

So, this has something to do with the circular dependency issue and would be resolved by the alternative in https://github.com/nextstrain/nextstrain.org/pull/570#discussion_r975908086

src/sources/groups.js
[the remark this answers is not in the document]
tsibley: > I think there's an inherent issue with trying to cross-reference the two classes within each other.

Hmm. There should be a way to do this. It's not an uncommon pattern for classes, though it's perhaps easier if the classes aren't in separate files.

> If both are to be set in their constructors, we can have either `Group.source` or `GroupSource.group`, but not both.

I don't think we need to set both in their constructors. `Group.source` can be a computed property. `GroupSource.group` should probably be a singleton, but could be lazily set.

src/groups.js
tsibley: I suppose? I think it's unfortunate that we have to, e.g., directly use `authzTags` and `authz/tags.js` instead of delegating via `authz.tags` since restructurings will require updating the callsites, but it's a minor thing in the big picture.

src/sources/groups.js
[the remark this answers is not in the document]
victorlin: Oh!!! Thanks for making me think twice about this. Actually, these circular dependencies are [totally appropriate with ES6](http://calculist.org/blog/2012/06/29/static-module-resolution/#cyclic-dependencies). The actual app runs fine. The issue is with an older version of Jest not behaving well. Upgrading Jest fixes things: 6ef1ec2d45574f19d1bc4c9bd35c3f94e5bdbf31

src/groups.js
victorlin: Nevermind, this workaround isn't needed after upgrading Jest. See https://github.com/nextstrain/nextstrain.org/pull/570#discussion_r977050514

victorlin:
@tsibley see 9a1902b97e2fab6e33e51118bc3fcc9bb37adb33...833ea30fa2b5b97c931a9b8d94a4560b255f34ae which is the new version of e98a3ea1e573e693fe51fee7488fd8cf2a0c7a80...39d4db2fb6a5e7be49edd239427609a4d4296329 (commits that were shared between this PR + #581). You may want to rebase your PR onto 833ea30fa2b5b97c931a9b8d94a4560b255f34ae and drop the old commits.

Notably, 6ef1ec2d45574f19d1bc4c9bd35c3f94e5bdbf31 has been added so that tests will pass with the circular dependency used in bf7eecf1fad01692122217db004a039f0631af10. Context: https://github.com/nextstrain/nextstrain.org/pull/570#discussion_r975893884

src/sources/groups.js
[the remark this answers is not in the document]
victorlin: I modified and reworded/split 1258b213335f4f1dd623f0b627b74c4de17f675b into 833ea30...0af95cc. So this is a pretty minimal change now – adding the computed property `Group.source`.

src/sources/groups.js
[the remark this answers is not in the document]
victorlin: This change is no longer present in c3275458fc8bddc667dabe4163820b7ca56e280a.

src/sources/groups.js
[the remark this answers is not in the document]
tsibley: Awesome!

src/endpoints/groups.js
```diff
@@ -43,7 +43,7 @@ const deleteGroupLogo = async (req, res) => {
   authz.assertAuthorized(req.user, authz.actions.Write, req.context.group);
 
   const method = "DELETE";
-  const url = await req.context.group.source.urlFor("group-logo.png", method);
+  const url = await req.context.group.source.urlFor(req.context.group.source.logoFile, method);
```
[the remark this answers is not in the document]
victorlin: > Part of that weirdness comes from the `x.y.z.foo(x.y.z.bar)`

We could use new methods `urlForLogo()`/`urlForOverview()` to avoid this, but that probably goes deeper into the "no man's land".

> using `urlFor()` + a well-known string

I'd be more ok with the "well-known string" if it was used locally in just one file. My main motivation is to avoid duplicate hard-coding of this value across different files. What if we make it a `const` exported from this file?

> we could just as easily be calling this object in S3 `group-logo` with no extension

Is this common practice? I've not had to deal with supporting various image formats in storage via API before, but it kinda makes sense. Should this also be applied to `group-overview`, in case we decide to allow other text formats later on?

src/groups.js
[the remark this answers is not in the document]
victorlin: Not needed anymore with c3275458fc8bddc667dabe4163820b7ca56e280a.

src/app.js
victorlin: Good catch here, that wasn't intentional. I dropped 0af95cc316c56f19f94b553ff72110f7b4532273 which should fix things since:

1. Without it, the authz assertion from `setSource` is retained.
2. Having the Group in the request's context is only necessary for the new `settings` endpoints, and that's already set from 833ea30fa2b5b97c931a9b8d94a4560b255f34ae.

src/endpoints/groups.js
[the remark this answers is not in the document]
victorlin: Ok, in that case I plan to drop 352603ae53c37e7f93c2c4c022999addfe9f35f5 and continue with `group-logo.png` since that's used elsewhere in the code already. To support more image types in the future, it'd be good to store in S3 without the extension, but that requires more careful thought for proper transition. Splitting off as a separate issue: #594

src/endpoints/sources.js
```diff
@@ -351,275 +303,18 @@ function sendSubresource(subresourceExtractor) {
...
-     *   -trs, 21 Jan 2022
-     */
-    if (!headers["Content-Length"]) {
-      body = await readStream(body, { limit: 20_000_000 /* 20MB */ });
-    }
-
-    const subresourceUrl = await subresource.url(method, {
-      "Content-Type": headers["Content-Type"],
-      "Content-Encoding": headers["Content-Encoding"],
-    });
```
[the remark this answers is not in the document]
victorlin: Oh, I probably made this change while adding `upstreamUrlExtractor` into the mix. Wasn't intentional so I've reverted (a18ffa306fcce7508a26a050e027efb630e04dd0), but I also wasn't aware of the reasoning for the inline object. Thanks for clarifying!

tsibley:
Does 7184cd7 still apply?

src/app.js
victorlin: > Does 7184cd7295ac006cfb6f5fa981c1d4b6146ea2f4 still apply?

Oh right it doesn't after the change above. Also dropped.

src/endpoints/groups.js
victorlin: This should be resolved with c3275458fc8bddc667dabe4163820b7ca56e280a.

tsibley (approved):
LGTM! :sparkles: Sorry for the several dead ends and fits and starts review that turned this into a bit of a slog. :pray:

victorlin:
@tsibley no worries! Thanks for your helpful comments here, learned a lot during the process!

PR #571, opened 2022-07-12
Update gitignore

victorlin:

Removing some seemingly unused entries.

### Related issue(s)

_N/A_

### Testing

_N/A_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,14 +1,15 @@
 # environment variables
-environment*
+/environment*
 
 # auspice build
-auspice-client/index.html
-auspice-client/dist/
+/auspice-client/index.html
+/auspice-client/dist/
 
 # misc #
 .DS_Store
-node_modules/
-npm-debug.log
+/node_modules/
+/static-site/node_modules/
+/npm-debug.log
 
 # local sessions
 /sessions/
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@ auspice-client/dist/
 .DS_Store
 node_modules/
 npm-debug.log
-scratch
 
 # local sessions
 /sessions/


tsibley (changes requested):
I'm personally on board with the second commit (5a3e88c10d54871e44504a42aa2a358c181b1171)… but I know many lab folks keep private `environment.sh` and similar files that they manually `source` in their shell sessions when their secrets are needed in the environment. I believe that's the motivation behind this ignore pattern, which for better or worse you'll find as boilerplate in many of the Nextstrain (and blab) repos.

+1 in general for tidying the ignore patterns, though.

I do also wish more of the existing gitignore patterns were anchored to the root (e.g. `/environment*`), so that they don't accidentally match files deeper in the source tree (e.g. `src/environment.js`). But that's not yet something I've had the fortitude to fix systematically.

victorlin:
Thanks for providing that context. Makes more sense why `environment*` is there now, though I would also rather remove it and push individuals to use custom files like `~/.gitignore_global`.

I applied your suggestion here: 1dce0e10750bbd2caee534fcb94510807c28f177

.gitignore
```diff
@@ -1,15 +1,14 @@
...
+/auspice-client/index.html
+/auspice-client/dist/
 
 # misc #
-.DS_Store
-node_modules/
-npm-debug.log
-scratch
+/.DS_Store
+/node_modules/
```
tsibley, line 10: `.DS_Store` wants to be unanchored since it's per-directory (IIRC, present in any directory browsed by the Finder on macOS, or maybe browsed + custom view attrs).

`/node_modules/` will want to expand to two lines to also cover `/static-site/node_modules/`.

tsibley (changes requested):

.gitignore
victorlin: Done: 68a2765ac37c129f526c58ceb951a5cd6aaf091a

tsibley (approved):

PR #557, opened 2022-06-16
charon: Use new "sources" handlers for getDataset, getNarrative, and getSourceInfo requests

tsibley:
### Description of proposed changes
Several related changes, all detailed in commit messages, but the top-line change is a consolidation of code paths between our Charon endpoints' handlers and the new "sources" handlers powering the RESTful API endpoints.

Resolves #150.

### Testing
- [x] CI passes locally
- [x] CI passes on GitHub (wtf)
- [x] Poking around locally doesn't find any issues

PR #573, opened 2022-07-15
Fix mixed up 503s and 404s for what should be successful responses

tsibley:
Fixes sporadic 503s (from Heroku's routing layer) and 404s (from us) on requests that should have been 304s. First noticed on /charon/getDataset requests after the deploy of #557, but present for longer on other endpoints.

See commit messages (and the linked notes + Slack thread within) for all the details.

### Testing
- [x] Lots of local testing and debugging, see [notes](https://github.com/tsibley/blab-standup/blob/master/2022-07-14.md#readme) for an inkling of it.
- [x] Tests pass locally
- [x] CI passes

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3,11 +3,6 @@
   "requires": true,
   "lockfileVersion": 1,
   "dependencies": {
-    "@awaitjs/express": {
-      "version": "0.6.3",
-      "resolved": "https://registry.npmjs.org/@awaitjs/express/-/express-0.6.3.tgz",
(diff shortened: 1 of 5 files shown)


tsibley:
I'll plan to merge this once CI passes so that it deploys to the canary. This will both a) let it get some shakedown time on the canary over the weekend and early next week before promoting to prod and b) immediately make the canary less "precarious" in that a promotion to prod will no longer include known-bad code (just normal, unknown-bad code :joy:).

tsibley:
RTD failure is unrelated, c.f. https://github.com/nextstrain/ncov/issues/982.

joverlee521 (commented):
Thanks for the detailed debugging notes! It was great to follow along, especially with such a hidden bug!

PR #581, opened 2022-08-23
Add RESTful API for managing Groups members

tsibley:
Adds RESTful API endpoints for listing/adding/removing existing users to group membership roles.

Things to address later:

- [ ] Currently no way to invite new users. Need to sort out the desired interface/mechanism for this. [Discussion](https://github.com/nextstrain/nextstrain.org/pull/581#discussion_r956509505).

- [ ] Automated tests for these endpoints/functions now that we have a shiny new testing environment (#598, #656, #657)

- [x] Currently this branch is missing a good, safe solution for dev and testing of these API endpoints.  I really don't want to grant all dev/test instances (local, GitHub Actions, Heroku review apps, etc) permissions to modify all groups, but AWS IAM policies don't support a finer resource specificity than the user pool, so we can't use the same approach we do with S3 buckets and prefix-scoping.

  Given these conflicting constraints, a separate dev-only user pool seems most appropriate, but that's a bit more initial work to setup and more ongoing work to manage/keep relevant over time.  I'm a little loathe to address that now or as a prerequisite for this work, but I also don't like not having a dev/testing story here.

   We're rapidly reaching (or are at?) the point where I really want separate dev vs. prod AWS accounts controlled by Terraform or CloudFormation something.

### Testing
- [x] Manual testing of endpoints
- [ ] ~Add new tests~
- [x] Local tests pass
- [x] CI passes

### After merge

- [ ] Deploy `env/production` Terraform config

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -348,6 +348,21 @@ app.routeAsync("/groups/:groupName/settings/overview")
   .optionsAsync(optionsGroup)
 ;
 
+app.routeAsync("/groups/:groupName/settings/members")
+  .getAsync(endpoints.groups.listMembers);
+
+app.routeAsync("/groups/:groupName/settings/roles")
+  .getAsync(endpoints.groups.listRoles);
+
+app.routeAsync("/groups/:groupName/settings/roles/:roleName/members")
+  .getAsync(endpoints.groups.listRoleMembers);
+
+app.routeAsync("/groups/:groupName/settings/roles/:roleName/members/:username")
+  .getAsync(endpoints.groups.getRoleMember)
+  .putAsync(endpoints.groups.putRoleMember)
+  .deleteAsync(endpoints.groups.deleteRoleMember)
+;
+
 app.route("/groups/:groupName/settings/*")
   .all(() => { throw new NotFound(); });
 
--- a/src/cognito.js
+++ b/src/cognito.js
@@ -1,0 +1,85 @@
+/**
+ * Cognito user pool (IdP) management.
+ *
+ * @module cognito
+ */
+/* eslint-disable no-await-in-loop */
+import {
+  CognitoIdentityProviderClient,
+  AdminAddUserToGroupCommand,
+  AdminRemoveUserFromGroupCommand,
+  paginateListUsersInGroup,
+  UserNotFoundException,
+} from "@aws-sdk/client-cognito-identity-provider";
+
+import { COGNITO_USER_POOL_ID } from "./config.js";
+import { NotFound } from "./httpErrors.js";
+
+
+const REGION = COGNITO_USER_POOL_ID.split("_")[0];
+
+const cognito = new CognitoIdentityProviderClient({ region: REGION });
+
+
+/**
+ * Retrieve AWS Cognito users in a Cognito group.
+ *
+ * @param {string} name - Name of the AWS Cognito group
+ * @yields {object} user, see <https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-cognito-identity-provider/interfaces/usertype.html>
+ */
+export async function* listUsersInGroup(name) {
+  const paginator = paginateListUsersInGroup({client: cognito}, {
+    UserPoolId: COGNITO_USER_POOL_ID,
+    GroupName: name,
+  });
+
+  for await (const page of paginator) {
+    yield* page.Users;
+  }
+}
+
+
+/**
+ * Add an AWS Cognito user to a Cognito group.
+ *
+ * @param {string} username
+ * @param {string} group - Name of the AWS Cognito group
+ * @throws {NotFound} if username is unknown
+ */
+export async function addUserToGroup(username, group) {
+  try {
+    await cognito.send(new AdminAddUserToGroupCommand({
+      UserPoolId: COGNITO_USER_POOL_ID,
+      Username: username,
+      GroupName: group,
+    }));
+  } catch (err) {
+    if (err instanceof UserNotFoundException) {
+      throw new NotFound(`unknown user: ${username}`);
+    }
+    throw err;
+  }
+}
+
+
+/**
+ * Remove an AWS Cognito user from a Cognito group.
+ *
+ * @param {string} username
+ * @param {string} group - Name of the AWS Cognito group
+ * @throws {NotFound} if username is unknown
+ */
+export async function removeUserFromGroup(username, group) {
+  try {
+    await cognito.send(new AdminRemoveUserFromGroupCommand({
+      UserPoolId: COGNITO_USER_POOL_ID,
+      Username: username,
+      GroupName: group,
+    }));
+  } catch (err) {
+    if (err instanceof UserNotFoundException) {
+      throw new NotFound(`unknown user: ${username}`);
+    }
+    throw err;
+  }
+}
--- a/src/endpoints/groups.js
+++ b/src/endpoints/groups.js
@@ -1,7 +1,9 @@
 import * as authz from "../authz/index.js";
+import { NotFound } from "../httpErrors.js";
 import { Group } from "../groups.js";
 import {contentTypesProvided, contentTypesConsumed} from "../negotiate.js";
 import {deleteByUrls, proxyFromUpstream, proxyToUpstream} from "../upstream.js";
+import { slurp } from "../utils/iterators.js";
 import * as options from "./options.js";
 
 
@@ -152,13 +154,94 @@ async function receiveGroupLogo(req, res) {
 }
 
 
+/* Members and roles
+ */
+const listMembers = async (req, res) => {
+  const group = req.context.group;
+
+  authz.assertAuthorized(req.user, authz.actions.Read, group);
+
+  return res.json(await group.members());
+};
+
+
+const listRoles = (req, res) => {
+  const group = req.context.group;
+
+  authz.assertAuthorized(req.user, authz.actions.Read, group);
+
+  const roles = [...group.membershipRoles.keys()];
+  return res.json(roles.map(name => ({name})));
+};
+
+
+const listRoleMembers = async (req, res) => {
+  const group = req.context.group;
+  const {roleName} = req.params;
+
+  authz.assertAuthorized(req.user, authz.actions.Read, group);
+
+  return res.json(await slurp(group.membersWithRole(roleName)));
+};
+
+
+const getRoleMember = async (req, res) => {
+  const group = req.context.group;
+  const {roleName, username} = req.params;
+
+  authz.assertAuthorized(req.user, authz.actions.Read, group);
+
+  for await (const member of group.membersWithRole(roleName)) {
+    if (member.username === username) {
+      return res.status(204).end();
+    }
+  }
+
+  throw new NotFound(`user ${username} does not have role ${roleName} in group ${group.name}`);
+};
+
+
+const putRoleMember = async (req, res) => {
+  const group = req.context.group;
+  const {roleName, username} = req.params;
+
+  authz.assertAuthorized(req.user, authz.actions.Write, group);
+
+  await group.grantRole(roleName, username);
+
+  return res.status(204).end();
+};
+
+
+const deleteRoleMember = async (req, res) => {
+  const group = req.context.group;
+  const {roleName, username} = req.params;
+
+  authz.assertAuthorized(req.user, authz.actions.Write, group);
+
+  await group.revokeRole(roleName, username);
+
+  return res.status(204).end();
+};
+
+
 export {
   setGroup,
   optionsGroup,
+
   getGroupLogo,
   putGroupLogo,
   deleteGroupLogo,
+
   getGroupOverview,
   putGroupOverview,
   deleteGroupOverview,
+
+  listMembers,
+  listRoles,
+  listRoleMembers,
+
+  getRoleMember,
+  putRoleMember,
+  deleteRoleMember,
 };
--- a/aws/iam/main.tf
+++ b/aws/iam/main.tf
@@ -37,7 +37,11 @@ resource "aws_iam_policy" "server" {
   name        = local.server_policy_name
   description = local.server_policy_description
 
-  policy = file("${path.module}/policy/${local.server_policy_name}.json")
+  policy = templatefile(
+    "${path.module}/policy/${local.server_policy_name}.tftpl.json", {
+      COGNITO_USER_POOL_ID = var.COGNITO_USER_POOL_ID
+    }
+  )
 }
 
 moved {
--- a/aws/iam/policy/NextstrainDotOrgServerInstance-testing.tftpl.json
+++ b/aws/iam/policy/NextstrainDotOrgServerInstance-testing.tftpl.json
@@ -18,7 +18,16 @@
       ],
       "Resource": [
         "arn:aws:s3:::nextstrain-groups"
-      ]
+      ],
+      "Condition": {
+        "StringLike": {
+          "s3:prefix": [
+            "blab/*",
+            "test/*",
+            "test-private/*"
+          ]
+        }
+      }
     },
     {
       "Sid": "MultitenantBucketObjectActions",
@@ -30,7 +39,21 @@
         "s3:DeleteObject"
       ],
       "Resource": [
-        "arn:aws:s3:::nextstrain-groups/*"
+        "arn:aws:s3:::nextstrain-groups/blab/*",
+        "arn:aws:s3:::nextstrain-groups/test/*",
+        "arn:aws:s3:::nextstrain-groups/test-private/*"
+      ]
+    },
+    {
+      "Sid": "CognitoUserPoolActions",
+      "Effect": "Allow",
+      "Action": [
+        "cognito-idp:AdminAddUserToGroup",
+        "cognito-idp:AdminRemoveUserFromGroup",
+        "cognito-idp:ListUsersInGroup"
+      ],
+      "Resource": [
+        "arn:aws:cognito-idp:us-east-1:827581582529:userpool/${COGNITO_USER_POOL_ID}"
       ]
     }
   ]
--- a/aws/iam/policy/NextstrainDotOrgServerInstance.tftpl.json
+++ b/aws/iam/policy/NextstrainDotOrgServerInstance.tftpl.json
@@ -18,16 +18,7 @@
       ],
       "Resource": [
         "arn:aws:s3:::nextstrain-groups"
-      ],
-      "Condition": {
-        "StringLike": {
-          "s3:prefix": [
-            "blab/*",
-            "test/*",
-            "test-private/*"
-          ]
-        }
-      }
+      ]
     },
     {
       "Sid": "MultitenantBucketObjectActions",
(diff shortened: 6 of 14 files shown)


src/app.js
```diff
@@ -300,6 +300,24 @@ app.routeAsync("/groups/:groupName")
   .getAsync(endpoints.static.sendGatsbyEntrypoint)
 ;
 
+app.routeAsync("/groups/:groupName/members")
+  .getAsync(endpoints.groups.listMembers);
+
+app.routeAsync("/groups/:groupName/roles")
+  .getAsync(endpoints.groups.listRoles);
```
victorlin, line 307: This has me thinking about your remark in #547:

> If using the proposed routes, then they should come before more generic routes for datasets. Or, it might be preferable to change the routes instead to avoid conflict with user-provided paths, e.g. to `/groups/:groupName/-/logo` or similar.

With this PR + the proposed endpoints in that issue, we would have to document that `members`, `roles`, `overview`, and `logo` can't be dataset names, and there might come more down the road in the same realm of groups management.

I'd be in favor of adding an intermediate in the path so there would be only one restricted dataset name for all groups management endpoints. We'd just have to settle on the name.

tsibley: Nod. This sort of namespacing makes sense to me. I forwent it here following your lead in #570 and because I figured we had to document a few restrictions anyhow so it wouldn't be too hard to document a short list. But happy to namespace instead and make it easy to keep that list short. Regardless of what we do for the management endpoints, though, we also need to document that your dataset can't start with `narratives/`. ;-) 

Without namespacing, the list of invalid dataset names would be:

  - `narratives/**`
  - `members/**` (soon, [#581][])
  - `roles/**` (soon, [#581][])
  - `overview` (soon, [#570][])
  - `logo` (soon, [#570][])

[#570]: https://github.com/nextstrain/nextstrain.org/pull/570
[#581]: https://github.com/nextstrain/nextstrain.org/pull/581

With namespacing, it'd be just two items.

For the namespace itself, I'm fond of short characters unlikely to be in anyone's name, like:

```
/groups/:groupName/-/**
/groups/:groupName/_/**
/groups/:groupName/=/**
```

and there's precedent set by lots of other people's URLs for things like that, but something more mundane and self-descriptive might be better, e.g.:

```
/groups/:groupName/manage/**
```

victorlin: I haven't noticed short characters used by other people's URLs, though I've probably come across it before.

I'm a bit more inclined to the self-descriptive option since we'll need to code this into Nextstrain CLI. Short characters might be slightly easier on users but might not provide enough context to developers.

If the concern is clashing with user dataset names, another idea would be to combine the two (e.g. `/groups/:groupName/_manage/**`). Not sure if there's any precedent for that.

The word "manage" seems to imply some sort of edit access, so it might be confusing to users who can only read a list of members at `/groups/:groupName/manage/members` but not actually "manage" it. How about something more generic like "settings"?

Also, noting that dataset names can't be `narratives` either, and presumably we'd do the same for the new namespace here:

https://github.com/nextstrain/nextstrain.org/blob/e98a3ea1e573e693fe51fee7488fd8cf2a0c7a80/src/app.js#L303-L305

tsibley: Fair point about "manage". I guess I was thinking in terms of "management API", which doesn't imply a level of edit access to me.

"settings" would be fine by me and follows the convention of using a plural noun for a collection.

victorlin: Shall we go with "settings" here and in #570, or is this something worth discussing with a larger group since it's a decision that will stick?

For "settings" specifically, I'll shoot down my previous idea of `_settings` since:

1. I imagine users wouldn't want to name a dataset `settings` anyways, though I can't speak for them.
2. `narratives` is already an existing namespace without the `_` prefix, and it'd be good to keep these namespace names consistent.

tsibley: Agreed re: not going with `_settings`. If we _were_ going to go that route, I think I'd prefer `.X` (e.g. `.settings`) to keep with the filesystem-based convention for "hidden" or "private" or "system" things (whereas prefixing with `_` is a code-based convention for roughly the same).

I'd be comfortable making the call to use `settings` also happy to solicit input from a larger group by posting to #nextstrain-dev for a quick poll/thoughts.

victorlin: [Posted to #nextstrain-dev](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1661452428629519), but we can start working with it in our PRs now.

tsibley:
Code is ready to review as-is as it's in what I'd consider to be a deployable state (though still missing some important functionality). I appreciate feedback/thoughts from folks on the two items "to address" listed at the top. I haven't had a chance to think on them any more yet, but will do so tomorrow or next week as well.

src/endpoints/index.js
```diff
@@ -1,12 +1,14 @@
...
 const cli = require("./cli");
+const groups = require("./groups");
 const sources = require("./sources");
 const static_ = require("./static");
 const users = require("./users");
 
 module.exports = {
   charon,
   cli,
+  groups,
```
victorlin, line 11: Changes in this file should come with 77edeb6c7024c91de4f2d89b291a43b1814ec103, not f2b1eaffad45ba93bf54fe8d360edf74b3719892, otherwise this doesn't work:

https://github.com/nextstrain/nextstrain.org/blob/77edeb6c7024c91de4f2d89b291a43b1814ec103/src/app.js#L303-L304

victorlin (changes requested):

src/app.js
```diff
@@ -300,6 +300,27 @@ app.routeAsync("/groups/:groupName")
   .getAsync(endpoints.static.sendGatsbyEntrypoint)
 ;
 
+app.use("/groups/:groupName/settings",
+  endpoints.groups.setGroup(req => req.params.groupName));
+
+app.routeAsync("/groups/:groupName/settings/members")
+  .getAsync(endpoints.groups.listMembers);
```
victorlin, line 307: > Currently no way to invite new users. Need to sort out the desired interface/mechanism for this.

Is this something you want to start implementing in this PR with an endpoint like `POST /groups/:groupName/settings/members/add`? I assume it would behave similarly to some functions in the existing [`scripts/provision-group`](https://github.com/nextstrain/nextstrain.org/blob/e98a3ea1e573e693fe51fee7488fd8cf2a0c7a80/scripts/provision-group).

---

For naming reference, I went on a tangent and compiled a list of GitHub's endpoints for adding a new member to an organization. Their URLs are all over the place, so probably not a good source of inspiration:

- `https://github.com/orgs/<org>/people`: members list
- `https://github.com/orgs/<org>/outside-collaborators`: outside collaborators
- `https://github.com/orgs/<org>/pending_collaborators`: pending collaborators
- `https://github.com/orgs/<org>/invitations/member_adder_add`: clicking the **Invite** button sent a request to this URL
- `https://github.com/orgs/<org>/invitations/<username>/edit`: direct link that works in browser
- ...

src/endpoints/index.js
tsibley: Whoops. Thanks!

src/cognito.js
```diff
@@ -0,0 +1,91 @@
+/**
```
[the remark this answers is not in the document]
tsibley: The user pool config is the initial work (not much). The setup and upkeep of individual users/groups, whatever ones we'd need/want in dev and testing, is the ongoing work. It's not onerous, just annoying.

src/app.js
tsibley: Something like that. Not sure. Behaviour would be similar to internal provisioning, yep. I don't think I want to gate this PR on it, though, as I don't think it's strictly necessary to do so. We still have work to do for UI for these endpoints anyway (CLI and/or web).

Some questions around invitations:

- Are we implementing a true invite → accept/decline invite → membership added flow from the start? Or just letting folks add anyone without an accept/decline step? Latter is much easier (esp. when we don't have our own database), but former is what we'll eventually want (although it may be quite a while till it's actually necessary to prevent abuse).

- How do we customize emails from Cognito when a new person is invited for a group? We probably have to send our own separately if we're going to include details of the membership transaction.

- Inviting is a little awkward with Cognito alone because it means the the invitee doesn't get a say in their username. Either we let the person inviting pick it (ha, nope!) or we force it internally (e.g. to match email). The latter is probably workable.

and there are surely more.

tsibley: So we almost certainly need to do the inviting new users part mostly outside of Cognito, and only touch Cognito when the user accepts (at which point they get a say in their user details and we add them to the right group+role).

joverlee521: Do we want to open up the user pool for self sign-up with admin confirmation? We can add an endpoint for the owner to just confirm new users and add them to the appropriate roles. 

Then the user -> group communication can happen outside of our domain.

tsibley: Ooh, that sounds like a useful piece of the puzzle. I'd plain forgot about the self sign-up options! I'm not sure we'd even need the admin confirmation bit. Would need to think thru all the flows a bit more.

Earlier in the day, I'd idly noodled a bit on issuing stateless invitations (where all the state is encoded in a token we send the invited user, who hands it back to us (or not) when accepting/declining (or ignoring)). This ~works if invitations have a TTL and we're ok holding a small amount of state for a short time (TTL+1) when an invitation is rescinded. But what we can do with that approach is necessarily less than if we hold state on outstanding invitations, so it's not entirely appealing.

joverlee521: I suggested the admin confirmation as a way for group owners to "vouch" for new users, but I guess adding them to their group has the same effect. 

How much do we care about random people signing up for accounts? They won't have access to things unless they are added to a group, but do we have to worry about filling up Cognito quotas?

src/app.js
```diff
@@ -300,6 +300,27 @@ app.routeAsync("/groups/:groupName")
...
+  .getAsync(endpoints.groups.listRoleMembers);
+
+app.routeAsync("/groups/:groupName/settings/roles/:roleName/members/:username")
+  .getAsync(endpoints.groups.getRoleMember)
+  .putAsync(endpoints.groups.putRoleMember)
+  .deleteAsync(endpoints.groups.deleteRoleMember)
+;
+
+app.route("/groups/:groupName/settings/*")
+  .all(() => { throw new NotFound(); });
```
victorlin, line 322: This makes sense on the CLI level, but the UI renders a "dataset not found" page.

Suggestion: 3c2d31a6d059981a734d67d269af6fd22cf91cf1

src/app.js
```diff
@@ -300,6 +300,27 @@ app.routeAsync("/groups/:groupName")
...
+
 // Avoid matching "narratives" as a dataset name.
 app.routeAsync("/groups/:groupName/narratives")
   .getAsync((req, res) => res.redirect(`/groups/${esc(req.params.groupName)}`));
```
victorlin, line 326: Suggestion: add this redirect for `/groups/:groupName/settings` as well.

victorlin (dismissed):

src/app.js
tsibley: I don't think we care too much about random signups unaffiliated with a group. It's pointless, but I don't think we need to prevent it if it's easier to do a group invite flow without a confirmation step. We need not be concerned about user quotas in Cognito: the limit is 40 million. There are also [operational quotas](https://docs.aws.amazon.com/cognito/latest/developerguide/limits.html) that apply to user signups, logins, etc. which would be impacted by additional not-necessary users, but I think we're unlikely to hit them either.

src/app.js
tsibley: Ugh, really dislike our Gatsy setup. Your suggested fix is appropriate, though, thanks!

tsibley:
Reconverted to back to draft. I'm going to:

- [ ] Rebase this on top of @victorlin's work in #547 and wait to mark this ready for review until that's finalized (soon I think!).
- [ ] Bite the bullet and make the dev/test Cognito user pool so we can properly test this. I think it'll be best in the long run too, so I might as well do it now.
- [ ] Think thru the story for new user invitations / user signup and try to write something up for those flows… but not condition this PR on it in order to keep things moving forward.
- [ ] Address @victorlin's two small review comments above

src/app.js
tsibley: Applied as 8e99844.

src/app.js
tsibley: Done as 7cb5663.

tsibley:
As part of figuring out the story here for dev/test, I'm going to (finally) get a Terraform config set up for us. I'll start by modeling a subset of our current AWS resources (the most relevant ones), then importing live state for them into Terraform, and then from there have a better basis for a prod/dev+test split.

tsibley:
Ok, I wrote a base Terraform configuration in #598. Next week I'll rebase this branch on top of that and then write the bits for a dev/prod split (at least in Cognito, but maybe larger? not sure yet).

src/cognito.js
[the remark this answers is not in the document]
tsibley: Well, several months later I got back to this and #656 and #657 have done the bulk of the work to carve out a testing environment for Cognito. I've rebased this branch onto that work. Now I just have to write some tests for this branch!

tsibley:
Taking this out of draft because I'd like to merge it without blocking any longer on comprehensive automated tests for the new endpoints. The Terraform and testing environment work I've done (e.g. [see this comment](https://github.com/nextstrain/nextstrain.org/pull/581#discussion_r1107899703)) allows for safe manual testing and development. Automated tests would still be great, and I'd started at one point down that hole, but quickly realized that some crucial infrastructural pieces were still missing, including:

1. obtaining authn tokens (e.g. logging in) in tests (or mechanisms to bypass authn in tests :grimacing:)
2. test fixtures for groups and users, and how to deal with concurrency (e.g. multiple instances of the tests running at the same time)

At that point I started to shimmy back up out of the rabbit hole. These missing pieces are totally doable and we should absolutely have those things some day, but I can't (and don't want to) keep implementing development infrastructure at the expense of feature development.

The desire to merge this comes up again now because I've been using this PR branch as the base branch for [my prototyping/WIP](https://github.com/nextstrain/nextstrain.org/compare/trs/groups-members-api...trs/wip/generalized-auth) for CDC to generalize authn to support any OIDC/OAuth2 IdP. That's because this branch has changes that would need to be accounted for by that work, and I'd rather account for that up front rather than later, given that we do still want this branch's endpoints so we can make the UI and other stuff.

aws/iam/policy/NextstrainDotOrgServerInstanceDev.tftpl.json
victorlin, line 1: I cherry-picked ff5599774ae22ad38338b73ec7d2890c23fe1f94 and applied these changes via Terraform while testing.

src/app.js
```diff
@@ -347,6 +347,21 @@ app.routeAsync("/groups/:groupName/settings/overview")
...
+app.routeAsync("/groups/:groupName/settings/roles")
+  .getAsync(endpoints.groups.listRoles);
+
+app.routeAsync("/groups/:groupName/settings/roles/:roleName/members")
+  .getAsync(endpoints.groups.listRoleMembers);
+
+app.routeAsync("/groups/:groupName/settings/roles/:roleName/members/:username")
+  .getAsync(endpoints.groups.getRoleMember)
+  .putAsync(endpoints.groups.putRoleMember)
+  .deleteAsync(endpoints.groups.deleteRoleMember)
```
victorlin, line 362: I noticed that after every PUT/DELETE, I would get `UnauthorizedError`s from [here](https://github.com/nextstrain/nextstrain.org/blob/ad853a2445d6fd726ada036b73cc05608cf00011/src/endpoints/groups.js#L13) until I restarted the server. I haven't looked deeper, so not sure if it's a problem.

src/endpoints/groups.js
```diff
@@ -152,13 +154,94 @@ async function receiveGroupLogo(req, res) {
...
+
+  throw new NotFound(`user ${username} does not have role ${roleName} in group ${group.name}`);
+};
+
+
+const putRoleMember = async (req, res) => {
+  const group = req.context.group;
+  const {roleName, username} = req.params;
+
+  authz.assertAuthorized(req.user, authz.actions.Write, group);
```
victorlin, line 208: Is it intentional that both `editors` and `owners` of a group can add/remove members? I'm not sure what the distinction is, but I'd assume that `editors` are more for adding/removing resources and that admin-level settings can only be changed by `owners`.

victorlin (commented):
<details>

<summary>Testing commands</summary>

```sh
# Get a token
export NEXTSTRAIN_DOT_ORG=http://localhost:5000
export NEXTSTRAIN_COGNITO_USER_POOL_ID="$(jq -r .COGNITO_USER_POOL_ID ../nextstrain.org/env/testing/config.json)"
export NEXTSTRAIN_COGNITO_CLI_CLIENT_ID="$(jq -r .COGNITO_CLI_CLIENT_ID ../nextstrain.org/env/testing/config.json)"
nextstrain login

# Listing members works
curl http://localhost:5000/groups/test/settings/members \
    --header @<(nextstrain authorization) | jq

# Listing roles works
curl http://localhost:5000/groups/test/settings/roles \
    --header @<(nextstrain authorization) | jq

# Listing role members works
curl http://localhost:5000/groups/test/settings/roles/editors/members \
    --header @<(nextstrain authorization) | jq

# Add myself as an editor
curl -X PUT http://localhost:5000/groups/test/settings/roles/editors/members/victorlin-owner \
    --header @<(nextstrain authorization)

# Remove myself as an owner
curl -X DELETE http://localhost:5000/groups/test/settings/roles/owners/members/victorlin-owner \
    --header @<(nextstrain authorization)

# Add myself as an owner
curl -X PUT http://localhost:5000/groups/test/settings/roles/owners/members/victorlin-owner \
    --header @<(nextstrain authorization)
```

<details>

tsibley:
@victorlin Note that for easier curl usage, you can use the `nextstrain authorization` command like so:

    curl ... --header @<(nextstrain authorization)

(This also has the advantage of not leaving the token in shell variables and not making it machine-visible via process arguments.)

src/app.js
tsibley: Ahhh. This is expected in this case because you're operating on your own user. Changing the role of a user X marks as stale any token X was issued before the change. X must renew/refresh their token to continue making authenticated requests. This happens automatically and transparently in a web context (e.g. if client-side JS was making the API requests) because the server manages the tokens, but here you're managing them. Nextstrain CLI [detects this condition and prompts the user](https://github.com/nextstrain/cli/blob/4797a5fee9631e034be76377e243b7d46d220bfc/nextstrain/cli/remote/nextstrain_dot_org.py#L715-L736) to `nextstrain login --renew` and then retry, but it could also make the renewal automatic and transparent in the future.

A server restart only fixed it by happenstance because in local dev you don't have Redis to persist the "stale before" bits, so an in-process memory store is used. This is one bit of janky test/dev set up that would be nice to improve at some point so parity with production is better. We could totally be using Redis locally in dev more routinely; I only use it now when I want to specifically test the Redis bits. But not that high priority to address it yet.

src/endpoints/groups.js
tsibley: No, and editors are not allowed to modify members. Your assumption of the role split is correct, but you were being mislead by local dev server restarts, per https://github.com/nextstrain/nextstrain.org/pull/581#discussion_r1351059469.

victorlin: Oh! I just tested again and, as a member of `editors`, I can't add or remove users from roles. This makes sense now.

src/app.js
victorlin: Got it! Thanks for the detailed explanation.

tsibley:
Rebased to resolve conflict introduced by #720. Will wait for tests to pass in CI before merging.

PR #582, opened 2022-08-25
Add endpoint for the Nextstrain CLI standalone archive installers

tsibley:
Currently these resolve through redirects to GitHub where the installers
are served directly out of the CLI's repo.¹  This will make maintenance
and updates easier, and the indirection of the nextstrain.org endpoints
allows us to swap things around behind the scenes as necessary later.

Planned to be used by forthcoming updates to a) our installation docs²
and b) the self-upgrade instructions of the CLI itself (when already
running a standalone install).

¹ https://github.com/nextstrain/cli/pull/217
² https://github.com/nextstrain/docs.nextstrain.org/pull/123

### Testing
- [x] Manually tested redirects work
- [ ] CI passes

### DIFF
--- a/src/endpoints/cli.js
+++ b/src/endpoints/cli.js
@@ -41,6 +41,23 @@ const download = async (req, res) => {
 };
 
 
+const installer = (req, res) => {
+  const os = req.params.os;
+  switch (os) {
+    case "linux":
+    case "mac":
+      return res.redirect("https://github.com/nextstrain/cli/raw/master/bin/standalone-installer-unix");
+
+    case "windows":
+      return res.redirect("https://github.com/nextstrain/cli/raw/master/bin/standalone-installer-windows");
+
+    default:
+      throw new NotFound(`No installer for OS: ${os}`);
+  }
+};
+
+
 module.exports = {
   download,
+  installer,
 };
--- a/src/app.js
+++ b/src/app.js
@@ -342,6 +342,9 @@ app.route(["/users", "/users/:name"])
 app.routeAsync("/cli/download/:version/:assetSuffix")
   .getAsync(endpoints.cli.download);
 
+app.routeAsync("/cli/installer/:os")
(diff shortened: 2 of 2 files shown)


src/endpoints/cli.js
```diff
@@ -41,6 +41,23 @@ const download = async (req, res) => {
...
 
+const installer = (req, res) => {
+  const os = req.params.os;
+  switch (os) {
+    case "linux":
+    case "mac":
+      return res.redirect("https://github.com/nextstrain/cli/raw/master/bin/standalone-installer-unix");
+
+    case "windows":
+      return res.redirect("https://github.com/nextstrain/cli/raw/master/bin/standalone-installer-windows");
```
victorlin, line 52: _(non-blocking)_

Future-proof in case the default branch ever changes:

```suggestion
    case "linux":
    case "mac":
      return res.redirect("https://github.com/nextstrain/cli/raw/HEAD/bin/standalone-installer-unix");

    case "windows":
      return res.redirect("https://github.com/nextstrain/cli/raw/HEAD/bin/standalone-installer-windows");
```

victorlin (approved):

src/endpoints/cli.js
tsibley: Ah, good call. As a cute trick it seems even `@` is supported, but I'll spell it out `HEAD`. :-)

tsibley: Updated with repush.

victorlin: Also, `-` sometimes works too (e.g. https://github.com/nextstrain/nextstrain.org/blob/-/README.md) and doesn't redirect to an exact commit.

Source: https://stackoverflow.com/a/65158264

tsibley: Makes sense to me, as `-` can be used with `git` commands to mean `the last ref I had checked out`, e.g. `git checkout -`, which is similar in spirit to `cd -`.

PR #583, opened 2022-09-02
Convert top-level project from CJS to ESM

victorlin:

Convert the top-level project from CommonJS modules (CJS) to EcmaScript modules (ESM). Note that this doesn't apply the project in `static-site/`.

### Related issue(s)

- Unmerged: https://github.com/nextstrain/nextstrain.org/blob/92cfbdf347fc1b81707b073629e5b1cb26d5dbb6/src/authz/index.js#L135-L144
- https://github.com/nextstrain/nextstrain.org/pull/570#discussion_r958831774

### Tasks

- [x] Get build to succeed
- [x] Get Jest to run properly
- [x] address ignored ESLint rules
- [x] get ESLint to work with top-level `await`
- [ ] Figure out what to do with old ESLint related packages such as `eslint-config-airbnb`: https://github.com/nextstrain/nextstrain.org/pull/583#discussion_r974530079

### Testing

- [x] Checks pass


src/app.js
```diff
@@ -4,6 +4,7 @@ import sslRedirect from 'heroku-ssl-redirect';
 import nakedRedirect from 'express-naked-redirect';
 import express from 'express';
 import compression from 'compression';
+import {createProxyMiddleware} from "http-proxy-middleware";
```
tsibley, line 7: Moving these to the top-level will also necessitate moving them from `devDependencies` to `dependencies` in `package.json`, I think.

package.json
```diff
@@ -15,7 +15,7 @@
     "build": "./build.sh",
     "lint": "eslint --ext .js,.jsx .",
     "set-up": "npm run build",
-    "server": "node server.js",
+    "server": "node --experimental-modules --es-module-specifier-resolution=node server.js",
```
tsibley, line 18: Hmm, as someone who routinely runs `node server.js` directly, I'd love if there was a way to enable these options/features another way. Are `package.json` options doable?

src/app.js
```diff
@@ -160,7 +160,7 @@ app.routeAsync("/charon/getSourceInfo")
 app.routeAsync("/charon/*")
   .all((req) => {
     utils.warn(`(${req.method}) ${req.url} has not been handled / has no handler`);
-    throw new NotFound();
+    throw new httpErrors.httpErrors.NotFound();
```
tsibley, line 163: Doubling of property lookup here and a few other places.

Also, these exception classes are used so extensively in the codebase and are a fundamental part of the unit of computation (responses) that I would prefer shorter symbols than `httpErrors.NotFound()`. That could mean boilerplate like:

```js
import httpErrors from 'http-errors';
const {Forbidden, NotFound, Unauthorized} = httpErrors;
```

or wrapping that up into our own little module to ease the boilerplate.

.eslintrc
```diff
@@ -42,6 +42,10 @@ rules:
   react/prop-types: off # possibly reinstate
   react/sort-comp: off # possibly reinstate
   jsx-a11y/no-static-element-interactions: off
+  import/extensions: off
+  import/first: off
+  import/newline-after-import: off
+  import/no-extraneous-dependencies: off
```
tsibley, line 48: Do we have a plan for turning them back on? Or are we ok with indefinitely disabling them?

tsibley (commented):
\o/ would love to see this! I [very briefly started down this path](https://gist.github.com/tsibley/e044784453e68f851a779c62a8fded10) a year ago, with the intent of having a transition period with a mix of CJS and ESM in the codebase, but I never saw it thru. Moving to all ESM like this does would be nicer anyway! I left a few notes that came to mind as I skimmed thru the commits, but I realize this PR isn't marked for review yet, so don't feel obligated to attend to them now.

src/app.js
victorlin: Whoops, good catch on my bulk-editing error! I'll look into adding a wrapper module since this import is used in quite a few places.

.eslintrc
victorlin: I'm going to try fixing and removing them in this PR - will see if that gets too tedious.

package.json
victorlin: I think it either has to be passed as flags directly or in the `NODE_OPTIONS` env var. Some quick internet searching along the lines of "set node options in package.json" makes me think that there isn't a good solution here if you still want to use `node server.js`. See https://github.com/nodejs/node/issues/37409 for some related discussion.

tsibley: Ah, hmm. I see.

Do we really need `--experimental-modules`? It's [marked as deprecated](https://nodejs.org/docs/latest-v14.x/api/cli.html#cli_experimental_modules) and seems to be a historical thing replaced now by `"type": "module"` in `package.json`.

If we wanted to avoid needing `--es-module-specifier-resolution=node`, we'd just have to update the imports to use `.js` and `…/index.js`? This doesn't seem so burdensome to me and arguably makes the mapping of code to files more explicit/clear.

Relatedly, `Procfile` will also need updating if we do stick with required extra `node` options.

victorlin: Good points, I removed both and added f1f1f972b14f38012ef6c32650b369eed445a6f5 to resolve this.

src/app.js
victorlin: Done: 93d5b248c4bf0f421aec4c8b254530d7e18da79a

src/groups.js
```diff
@@ -1,7 +1,6 @@
-const assert = require("assert").strict;
-const {NotFound} = require("http-errors");
-
-const GROUPS_DATA = require("../data/groups.json");
+import { strict as assert } from 'assert';
+import httpErrors from 'http-errors';
+import GROUPS_DATA from '../data/groups.json';
```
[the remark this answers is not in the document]
victorlin: Yes, added: 326c4b492be59237e4820464fb6c36954c5b3029

.eslintrc
victorlin: Done, wasn't too bad. Relevant commits: 54dfb22d0f4927c6b3b8adb506df5aa9266989ee...cb853ed56c90507f2e3360e37c2c7dce9c68ff77

src/app.js
victorlin: I updated to keep the imports at their existing levels, so that's no longer necessary: 7cc84e85f8a8c557fe9c96ff787bceae774dd2c5

src/app.js
```diff
@@ -397,17 +397,20 @@ app.route("/dist/*")
...
-  app.use(createProxyMiddleware(app.locals.gatsbyDevUrl));
+    // WebSocket endpoint
+    app.get("/socket.io/", createProxyMiddleware(app.locals.gatsbyDevUrl, {ws: true}));
+
+    /* This is the end of the line in gatsbyDevUrl mode, as the proxy middleware
+    * doesn't fallthrough on 404 and even if it did the Gatsby dev server
+    * returns a 200 Ok for missing pages.
+    */
+    app.use(createProxyMiddleware(app.locals.gatsbyDevUrl));
+  })();
```
tsibley, line 414: Hmm. I think this is problematic because it's now non-deterministic when the `app.get(…)` and `app.use(…)` handlers get registered, e.g. they may not register until after things later in this file. This matters because registration order is routing order.

src/user.js
```diff
@@ -78,7 +78,7 @@ async function markUserStaleBeforeNow(username) {
 }
 
 
-module.exports = {
+export {
   userStaleBefore,
   markUserStaleBeforeNow,
 };
```
victorlin, line 84: I'd also be happy to refactor these end-of-file exports to be "inline" exports, but that can be opinionated. Won't do it unless I get a thumbs up from someone else here.

src/app.js
victorlin: Oh that's a valid point. How about something like fd145bf357625af4ae0df0883829f6efa7fc43ef?

Are there any other implications of the CJS->ESM switch from synchronous to asynchronous module loading that we should be aware of?

tsibley: Not your fault, but that's fairly ugly IMO. Why can't we use [top-level await](https://nodejs.org/api/esm.html#top-level-await) since these are now ES modules?

victorlin: Oh actually you're right, it runs fine without these hacks. This is just an ESLint error, so I'll look into getting ESLint to behave properly.

src/user.js
victorlin: I looked for an ESLint rule to prefer inline exports over end-of-file exports, but could not find one. I only found [import/exports-last](https://github.com/import-js/eslint-plugin-import/blob/06cc47c4e60246a4caa120b2d8cc0a98e07e894d/docs/rules/exports-last.md) which prefers end-of-file over inline.

Now I'm thinking that maybe it's better to keep as-is and enforce `import/exports-last` for future consistency.

src/app.js
tsibley: :pray:

src/user.js
tsibley: Hmm. I wouldn't take ESLint as gospel here, although I know that opinion itself is blasphemy in some circles. I don't care too much about the export styles we use (maybe I don't know enough about vagaries of either to care), but I'd have a slight preference to not enforce one or the other. For example, it can be nice to use inline exports as you can see what's exported and what's not when reading the file without jumping to the bottom.

src/app.js
```diff
@@ -396,11 +398,10 @@ app.route("/dist/*")
  * instead of served straight from disk.
  */
 if (app.locals.gatsbyDevUrl) {
-  // eslint-disable-next-line global-require, import/no-extraneous-dependencies
-  const {createProxyMiddleware} = require("http-proxy-middleware");
+  const {createProxyMiddleware} = await import("http-proxy-middleware");
```
victorlin, line 401: After some ESLint upgrading/tinkering, there's still [this error](https://github.com/nextstrain/nextstrain.org/runs/8239083567?check_suite_focus=true#step:7:11):

```
  401:34  error  Parsing error: Can not use keyword 'await' outside an async function (401:34)
```

Supposedly, top-level await should work with ESLint 8...

- https://github.com/eslint/eslint/issues/14632#issuecomment-908795546
- https://github.com/eslint/eslint/discussions/13934
- https://eslint.org/docs/latest/user-guide/configuring/language-options

Will investigate more another day.

src/app.js
victorlin: Continuing ESLint endeavors here: https://github.com/nextstrain/nextstrain.org/pull/583#discussion_r965396359

src/app.js
victorlin: After playing in a [test repo](https://github.com/victorlin/nodejs-test-apps), I deduced that the Babel parser used in this project also has an affect on supporting top-level await. Notably, it can work with earlier versions of ESLint as long as the Babel parser is at least v7.15.0 based on [the changelog entry](https://github.com/babel/babel/blob/3057a3b2f7782993a293aeb0ef629b153ac4422f/CHANGELOG.md#v7150-2021-08-04).

The reason this error persisted is because 4fd47d425a77b27ddf45a82047152baf87b58c27 upgraded the Babel parser but did not update Babel itself. After doing that in 177b0aefa2c327b0c28f677776ed3b787cb99421, ESLint passes but there are now build issues with Auspice's dependency pinning not supporting the latest version of Babel. Need to dig deeper to fix that, but at least this is making sense now.

package.json
```diff
@@ -61,19 +63,19 @@
     "yaml-front-matter": "^4.0.0"
   },
   "devDependencies": {
-    "babel-core": "^6.26.0",
-    "babel-eslint": "^8.2.3",
+    "@babel/core": "^7.19.0",
```
victorlin, line 66: Interesting that 177b0aefa2c327b0c28f677776ed3b787cb99421 which merely updates the Babel core dependency in the top-level project causes `gatsby build` to fail in `static-site`:

```
failed Building production JavaScript and CSS bundles - 63.137s
error Generating JavaScript bundles failed

Can't resolve 'regenerator-runtime/runtime' in '/home/runner/work/nextstrain.org/nextstrain.org/auspice-client/customisations'

Perhaps you need to install the package 'regenerator-runtime/runtime'?
```

src/app.js
victorlin: Moving the tangent issue to https://github.com/nextstrain/nextstrain.org/pull/583#discussion_r966353409

package.json
victorlin: I don't understand why, but 4a4f1839d9d02f4c5d85c11b5a7309b6a605d0a8 fixed this.

Maybe relevant:

- https://github.com/gatsbyjs/gatsby/issues/23676
- https://github.com/gatsbyjs/gatsby/issues/1905
- https://stackoverflow.com/a/68091315

src/user.js
victorlin: I personally find smaller files with inline exports easier to understand. But I don't have a strong preference or will to make a large refactor here, so I'll keep things as-is without adding an `import/exports-last` rule.

package.json
tsibley: > I don't understand why, but



> Interesting that [177b0ae](https://github.com/nextstrain/nextstrain.org/commit/177b0aefa2c327b0c28f677776ed3b787cb99421) which merely updates the Babel core dependency in the top-level project causes `gatsby build` to fail in `static-site`:

Did you have a populated `static-site/node_modules/` when this happened? If not, Node would have found the one in the repo root and picked up the new Babel version.

Or maybe it's because, when resolving imports, Node will, I believe, keep walking upwards looking for other `node_modules/` if the first one it finds doesn't have the module being imported. So if something in `static-site/node_modules/` was doing a conditional import or similar, the two deps could get tangled up.

tsibley: image

victorlin: > Did you have a populated `static-site/node_modules/` when this happened?

Yes, this was reproduced on [the CI run](https://github.com/nextstrain/nextstrain.org/runs/8256127348?check_suite_focus=true) which runs `npm ci` + `build.sh`, which is effectively:

```
npm ci
cd static-site
npm ci
gatsby build
```

Also noting that the path in the error is `nextstrain.org/auspice-client/customisations`. `auspice-client` is only used in one file in `static-site/`. Removing that reference allows the build to succeed: c9ba5f9abdd55dc06822441c01906389b0c0fb38 (not in PR)

Maybe similar to what you described, because there is no `node_modules` under that path, it then walks up to `nextstrain.org/node_modules/` when really `nextstrain.org/static-site/node_modules` would be more appropriate.

victorlin: So maybe the answer to "why did 4a4f1839d9d02f4c5d85c11b5a7309b6a605d0a8 fix this?" is multiple parts:

1. `static-site/src/sections/sars-cov-2-page.jsx` relies on the top-level `node_modules`
2. 177b0aefa2c327b0c28f677776ed3b787cb99421 effectively removed `regenerator-runtime` from the top-level `node_modules`
3. Using the older version of Gatsby to create the page from (1) relies on `regenerator-runtime`. The newer version of Gatsby does not have this dependency.

package.json
```diff
@@ -17,7 +17,8 @@
     "set-up": "npm run build",
     "server": "node server.js",
     "start": "npm run server",
-    "test": "jest --roots ./test/",
+    "//test": "echo 'NODE_OPTIONS is required until this issue is resolved: https://github.com/facebook/jest/issues/9430 '",
+    "test": "NODE_OPTIONS='--experimental-vm-modules' jest --roots ./test/",
```
tsibley, line 21: _non-blocking_

An ugly hack. :-) Just noting here that this env var will be needed when running individual test files too, e.g.:

    NODE_OPTIONS='--experimental-vm-modules' npx jest test/auspice_client_requests.test.js

tsibley (approved):
Lots of fiddly work. Thank you for doing this! :pray:

package.json
jameshadfield: More notes (or partial rememberings) for posterity -- node's "ability" to walk up is why the folder structure when building auspice with custom extensions really matters, and why we put the JS in a subdirectory (`auspice-client/customisations`).

P.S. How have we persevered with Gatsby for this long?!?

tsibley: > P.S. How have we persevered with Gatsby for this long?!?

I think I'm going to rip it out once we start doing more UI work… :see_no_evil:

package.json
```diff
@@ -70,7 +70,6 @@
     "babel-plugin-lodash": "^3.2.11",
     "benchmark": "^2.1.4",
     "eslint": "^8.23.0",
-    "eslint-config-airbnb": "^19.0.4",
```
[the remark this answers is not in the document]
victorlin: [A lot](https://github.com/airbnb/javascript/tree/eslint-config-airbnb-v15.1.0/packages/eslint-config-airbnb-base/rules)... I don't know if it's worth sifting through and picking out what we think is valuable. I realized that the motivation for dropping this (updating it, which results in new linting errors) isn't actually necessary, so I force-pushed without the update.

The downside of keeping it is that we use pretty outdated versions not written for the newer version of ESLint, as mentioned in `npm install` output:

```
npm WARN eslint-config-airbnb@15.1.0 requires a peer of eslint@^3.19.0 || ^4.3.0 but none is installed. You must install peer dependencies yourself.
npm WARN eslint-config-airbnb-base@11.3.2 requires a peer of eslint@^3.19.0 || ^4.5.0 but none is installed. You must install peer dependencies yourself.
npm WARN eslint-plugin-jsx-a11y@5.1.1 requires a peer of eslint@^2.10.2 || ^3 || ^4 but none is installed. You must install peer dependencies yourself.
```

test/auspice_client_requests.test.js
```diff
@@ -1,6 +1,6 @@
 import {jest} from '@jest/globals';
-import urlsToTest from './auspice_client_requests.json';
 import fetch from 'node-fetch';
+import urlsToTest from './auspice_client_requests.json';
```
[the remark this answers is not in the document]
victorlin: I wanted to say this is something with `--experimental-vm-modules` having the same effect as `--experimental-json-modules`, but...

1. Using `--experimental-vm-modules` for running the project doesn't work.

    ```diff
    diff --git a/package.json b/package.json
    index cc22156..b546c95 100644
    --- a/package.json
    +++ b/package.json
    @@ -15,7 +15,7 @@
         "build": "./build.sh",
         "lint": "eslint --ext .js,.jsx .",
         "set-up": "npm run build",
    -    "server": "node server.js",
    +    "server": "node --experimental-vm-modules server.js",
         "start": "npm run server",
         "//test": "echo 'NODE_OPTIONS is required until this issue is resolved: https://github.com/facebook/jest/issues/9430 '",
         "test": "NODE_OPTIONS='--experimental-vm-modules' jest --roots ./test/",
    diff --git a/src/groups.js b/src/groups.js
    index 1d5ca23..81c58b8 100644
    --- a/src/groups.js
    +++ b/src/groups.js
    @@ -2,8 +2,7 @@ import { strict as assert } from 'assert';
     import { NotFound } from './httpErrors.js';
     
     /* eslint-disable-next-line import/first, import/newline-after-import */
    -import { readFile } from 'fs/promises';
    -const GROUPS_DATA = JSON.parse(await readFile(new URL('../data/groups.json', import.meta.url)));
    +import GROUPS_DATA from '../data/groups.json';
     
     const PRODUCTION = process.env.NODE_ENV === "production";
    ```

    ```
    $ npm run server
    ...
    TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".json"
    ...
    ```

2. "JSON/WASM module" support is listed separately from "Running the module in the correct context",  which I think is what `--experimental-vm-modules` addresses.

So I don't have a good answer here 😕

src/groups.js
```diff
@@ -1,6 +1,9 @@
 import { strict as assert } from 'assert';
 import { NotFound } from './httpErrors.js';
-import GROUPS_DATA from '../data/groups.json';
+
+import { createRequire } from 'module';
+const require = createRequire(import.meta.url);
+const GROUPS_DATA = require('../data/groups.json');
```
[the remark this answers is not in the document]
victorlin: Makes sense, updated in 4511c261756323ca896c08784b759722db0116fd.

package.json
victorlin: Good point, I updated the README in 781a0237afd1bbd6a069917c903a2ebd23864b90.

package.json
victorlin: Commit message updated in 383adab98fe1375d772125f2f7117458a4aaf860.

package.json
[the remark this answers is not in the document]
jameshadfield: It predates my joining the group too - it was part of the first auspice prototype and I lifted it wholesale when we built nextstrain.org (at least, that's my memory). 

I haven't reviewed the rules in a very long time however I'm a big fan of the linting we've been using (some things are overly pedantic, but in general I think it's been extremely helpful for the codebase).

test/auspice_client_requests.test.js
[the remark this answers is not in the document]
tsibley: Hmm. Jest might be doing funny business with loaders. (Separately from anything here, I wish we weren't using Jest.)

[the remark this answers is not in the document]
tsibley: Yeah, `jest` is setting up global loaders or something for `.json`. If I run:

```js
import urlsToTest from './auspice_client_requests.json';
console.log(Object.keys(urlsToTest));
```

as

```
NODE_OPTIONS=--experimental-vm-modules node test/foo.test.js
```

I get

```
TypeError [ERR_IMPORT_ASSERTION_TYPE_MISSING]: Module "file:///home/tom/nextstrain/nextstrain.org/test/auspice_client_requests.json" needs an import assertion of type "json"
    at new NodeError (node:internal/errors:372:5)
    at validateAssertions (node:internal/modules/esm/assert:82:15)
    at defaultLoad (node:internal/modules/esm/load:24:3)
    at ESMLoader.load (node:internal/modules/esm/loader:407:26)
    at ESMLoader.moduleProvider (node:internal/modules/esm/loader:326:22)
    at new ModuleJob (node:internal/modules/esm/module_job:66:26)
    at ESMLoader.#createModuleJob (node:internal/modules/esm/loader:345:17)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:304:34)
    at async ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:82:21)
    at async Promise.all (index 0) {
  code: 'ERR_IMPORT_ASSERTION_TYPE_MISSING'
```

but the import works when run under:

```
NODE_OPTIONS=--experimental-vm-modules npx jest --run-tests-by-path test/foo.test.js
```

tsibley (approved):
Repush LGTM!

tsibley:
Hmm. Ran into this just now, which used to work fine before:

```
$ aws-profile blab/trs ./scripts/provision-group KU-song --members ~/Downloads/ku-song.yaml 
internal/process/esm_loader.js:74
    internalBinding('errors').triggerUncaughtException(
                              ^

TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension "" for /home/tom/nextstrain/nextstrain.org/scripts/provision-group
    at new NodeError (internal/errors.js:322:7)
    at Loader.defaultGetFormat [as _getFormat] (internal/modules/esm/get_format.js:71:15)
    at Loader.getFormat (internal/modules/esm/loader.js:105:42)
    at Loader.getModuleJob (internal/modules/esm/loader.js:243:31)
    at async Loader.import (internal/modules/esm/loader.js:177:17)
    at async Object.loadESM (internal/process/esm_loader.js:68:5)
    at async handleMainPromise (internal/modules/run_main.js:59:12) {
  code: 'ERR_UNKNOWN_FILE_EXTENSION'
}

$ head -n1 ./scripts/provision-group 
#!/usr/bin/env node
```

PR #585, opened 2022-09-15
auspice 2.38.0

jameshadfield:


victorlin:
@jameshadfield with #572 merged, these manual PRs shouldn't be needed anymore!

jameshadfield:
Ahhh sweet! I saw that for auspice.us but missed it for nextstrain.org

PR #587, opened 2022-09-19
Convert package-lock.json back to a v1 lockfile

victorlin:

In 5e47c94d5e04606a1ec9c2403158ba2a021f6981, package-lock.json was updated using NPM >=7, which converted the file to a v2 lockfile.

The top-level project is using NPM 6.14 (Node 14) as specified in package.json and used by Heroku. While the newer lockfile version is backwards-compatible, it makes more sense to have a lockfile generated by the NPM version being used in production.

To avoid unnecessary version changes that would arise from regenerating the file solely based on package.json, this was done by restoring the last v1 lockfile, clearing cache, then installing:

    git checkout 5e47c94d5e04606a1ec9c2403158ba2a021f6981^ -- package-lock.json
    rm -rf node_modules/
    npm cache clear --force
    npm install
    git add package-lock.json

### Related issue(s)

_N/A_

### Testing

- [ ] Checks pass


tsibley (approved):
+1 in general (assuming CI passes), but (thru no fault of your own!) I find it hard to review this since there's no easily comparable state in the same package lock format. What we have is:

1. format v1 prior to Auspice upgrade
2. format v2 after Auspice upgrade
3. format v1 after Auspice upgrade

Nevertheless, briefly skimming the (huge!) diff here (i.e. between 2 and 3 above), I do see that it seems to be more than a purely syntactic transformation. For example, I see bumps in resolved versions, e.g.

```diff
     "@babel/core": {
-      "version": "7.11.6",
-      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.11.6.tgz",
-      "integrity": "sha512-Wpcv03AGnmkgm6uS6k8iwhIwTrcP0m17TL1n1sy7qD0qelDu4XNeW0dN0mHfa+Gei211yDaLoEe/VlbXQzM4Bg==",
+      "version": "7.19.0",
+      "resolved": "https://registry.npmjs.org/@babel/core/-/core-7.19.0.tgz",
+      "integrity": "sha512-reM4+U7B9ss148rh2n1Qs9ASS+w94irYXga7c2jaQv9RVzpS7Mv1a9rnYYwuDa45G+DkORt9g6An2k/V4d9LbQ==",
```

I guess the common wisdom in the JS community is to just Trust The Process…

victorlin:
[7.11.6](https://www.npmjs.com/package/@babel/core/v/7.11.6) -> [7.19.0](https://www.npmjs.com/package/@babel/core/v/7.19.0) is a pretty big jump given the 2 year difference. Maybe each incremental change of package-lock.json only bumps the bare minimum, while regenerating the file pulls the latest possible version given all the constraints. Skilling through package-lock.json's history, I don't see any regeneration so that would make sense.

Well, tests are failing now so I'll fix that before merging. I'm guessing all these upgrades also included some new ESLint rules. Will see if I can avoid the upgrades.

jameshadfield:
Using dependabot to upgrade auspice should help here. But the underlying issue is that an on-the-surface simple change to a node project's dependency can induce an enormous set of changes to the dependency tree.

victorlin:
Ok, now this PR should effectively be what would have happened to package-lock.json if 5e47c94d5e04606a1ec9c2403158ba2a021f6981 was created using NPM 6.14. See package-lock.json here for the diff with v2 conversion "canceled out": e945a423f6e871399d32618a9fcc9089d0e47152...9ca679fd31cba56707be0b59b38053526ff2579d

jameshadfield:
> However, these changes are still pretty surprising because [there were no dependency changes in the new version of Auspice](https://github.com/nextstrain/auspice/compare/v2.37.3...v2.38.0#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519).

As far as I understand, if you run `npm install` it'll bump up dependencies as much as it can, as allowed by the pinning in `package.json`. For instance, if you define `package^1.2.3` and `npm install` today you might get `1.3.0`, next week the same `npm install` command might get you `1.4.2`. Pinning dependencies tightly doesn't help here as the dependencies of `package` may not be tightly pinned, so you'll still get version bumps. The idea behind semver seems sound, but in practice some degree of testing is always required. 

(It's common to rant about dependencies & packaging, but every language seems to have their own set of problems. Node has bitten me a lot over the years, but I've also used Node the most.)

victorlin:
Ah, this got merged with #583 since I had rebased that branch onto this change. Feel free to comment here if you want to see anything else changed related to the lockfile.

victorlin:
> For instance, if you define `package^1.2.3` and `npm install` today you might get `1.3.0`, next week the same `npm install` command might get you `1.4.2`

I'll just assume that's the reasoning for the large diff in e945a423f6e871399d32618a9fcc9089d0e47152...9ca679fd31cba56707be0b59b38053526ff2579d, though I don't feel like digging deeper to find out. CI passed and I didn't notice anything odd when running locally, so new versions should be ok here.

PR #595, opened 2022-09-27
Use Node.js version 18

victorlin:

Although the current version (14) is still in active LTS status, the latest LTS version comes with performance improvements [[1]]. The performance improvements from version 16 should also apply [[2]].

Also update the npm version to a minimum of 8.6.0 since it corresponds with Node.js 18.0.0 per https://nodejs.org/en/download/releases/.

[1]: https://nodejs.org/en/blog/release/v18.0.0/
[2]: https://nodejs.org/en/blog/release/v16.0.0/

### Related issue(s)

_N/A_

### Tasks

- [ ] Merge before this PR: https://github.com/nextstrain/auspice/pull/1560
- [ ] Check `npm install` and build outputs for any dependency compatibility issues

### Testing

- [ ] Checks pass
- [ ] Manually test preview build

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,8 +8,8 @@
   "repository": "github:nextstrain/nextstrain.org",
   "homepage": "https://nextstrain.org",
   "engines": {
-    "node": "^14",
-    "npm": "^6.14"
+    "node": "^18",
+    "npm": "^8.6.0"
   },
   "scripts": {
     "build": "./build.sh",
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: '14'
+          node-version: '18'
       - run: node --version
       - run: npm ci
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -8,8 +8,8 @@
   "repository": "github:nextstrain/nextstrain.org",
   "homepage": "https://nextstrain.org",
   "engines": {
-    "node": "^14",
-    "npm": "^6.14"
+    "node": "^18",
```
victorlin, line 11: [This error from `npm run build`](https://github.com/nextstrain/nextstrain.org/actions/runs/3137943731/jobs/5096707413#step:6:143):

```
error UNHANDLED REJECTION error:0308010C:digital envelope routines::unsupported
```

was also observed in https://github.com/nextstrain/auspice/pull/1520. Webpack, an Auspice dependency, [is the culprit](https://github.com/webpack/webpack/issues/14532). This will be fixed once that PR is merged, released, and the Auspice version is bumped for this project.

victorlin:
This was premature and incomplete. Closing in favor of #635.

PR #596, opened 2022-09-27
Remove `set-up` → `build` npm command alias

victorlin:

This alias has been around since 84a45a8bb9b1a8c967b33ab6964ab24c793084ad and should be unused at this point.

### Related issue(s)

_N/A_

### Testing

- [ ] Checks pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,7 +14,6 @@
   "scripts": {
     "build": "./build.sh",
     "lint": "eslint --ext .js,.jsx .",
-    "set-up": "npm run build",
     "server": "node server.js",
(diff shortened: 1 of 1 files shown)


victorlin:
@jameshadfield looks like you added `set-up` long ago in 3b077f0c566be1c25d1f5df86ab61156f48cc6af. If you don't use it anymore, I think we can just remove it.

jameshadfield (approved):
Yeah I think "set-up" was what we originally used so when "build" came along we aliased it. Happy for this PR to be merged ✅

PR #599, opened 2022-10-04
CI: Build Auspice client

victorlin:

It can be useful to test that a newer Auspice version builds properly in the context of this project.

### Related issue(s)

Testing scenario brought up by @jameshadfield in https://github.com/nextstrain/auspice/pull/1520#discussion_r985331370.

### Testing

- [ ] Checks pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,6 +33,17 @@ jobs:
           name: logs
           path: test/server.log
 
+  build-auspice-client:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '14'
+      - run: node --version
+      - run: npm ci
+      - run: node_modules/.bin/auspice build --verbose --extend auspice-client/customisations/config.json
+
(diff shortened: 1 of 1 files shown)


jameshadfield:
This should be covered by our normal CI (`npm run build` will build the auspice client), right?

The scenario I brought up is where we want to try to build a specific (new / development) version of auspice within the nextstrain.org repo. This requires something along the lines of `npm install nextstrain/auspice#commit` (you can do it locally with `npm link` but I find it more robust to use a github commit). I think the way to achieve this is for an auspice action to make a testing PR within nextstrain.org (i.e. https://github.com/nextstrain/auspice/pull/1565). Perhaps I'm missing something here.

victorlin:
> This should be covered by our normal CI (`npm run build` will build the auspice client), right?

@jameshadfield oh you're right! This PR isn't needed then. Yes, https://github.com/nextstrain/auspice/pull/1565 should be sufficient.

tsibley:
I'd still +1 this though: [b057b77](https://github.com/nextstrain/nextstrain.org/pull/599/commits/b057b77d285ab254a03f614e0beaeb875b79372c)

victorlin:
@tsibley oh I forgot about that! Done with #601

PR #605, opened 2022-10-06
Add comments to dependabot.yml

victorlin:

Adds some useful links.

### Related issue(s)

_N/A_

### Testing

_N/A_

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,7 +1,11 @@
+# Dependabot configuration file
+# Reference: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
 version: 2
 updates:
   - package-ecosystem: "npm"
     directory: "/"
+    # If you want to run this manually, go to https://github.com/nextstrain/nextstrain.org/network/updates
+    # and click "Last checked …" then "Check for updates".
     schedule:
       interval: "daily"
     allow:


.github/dependabot.yml
```diff
@@ -1,7 +1,11 @@
+# Dependabot configuration file
+# Reference: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
 version: 2
 updates:
   - package-ecosystem: "npm"
     directory: "/"
+    # If you want to run this manually, go to https://github.com/nextstrain/nextstrain.org/network/updates
+    # and click "Last checked …" then "Check for updates".
     schedule:
       interval: "daily"
```
jameshadfield, line 10: Can we make this monthly? 🙏

victorlin: @jameshadfield I'm assuming you want the security updates to be monthly, but [that isn't possible](https://github.com/nextstrain/auspice/issues/1462#issuecomment-1270750330). This is just for the Auspice version checks, which should be more useful on a daily interval.

jameshadfield: That's exactly what I wanted. Oh well :( thanks for looking into it!

> This is just for the Auspice version check

Where doe the config specify only auspice? (Asking for my own understanding.)

victorlin: The last line! Not so obvious from the PR changes view.

https://github.com/nextstrain/nextstrain.org/blob/c2f3d84a398dc9cb9de55bf632f63c52d5f47826/.github/dependabot.yml#L5-L12

jameshadfield (approved):

PR #615, opened 2022-10-21
Remind us to promote the canary to production

tsibley:
…if we're tardy about it.

### Changes outside PR

- Generated a [`read-protected` scoped](https://devcenter.heroku.com/articles/oauth#scopes) Heroku API token for our Heroku bot user
- Added the token as a `HEROKU_TOKEN_READ_PROTECTED` GitHub org-level secret for Actions
- Created a ["Nextstrain Bot" Slack app](https://api.slack.com/apps/A047F02PTE2) and ["installed" it into our Slack org](https://bedfordlab.slack.com/apps/A047F02PTE2-nextstrain-bot) in order to generate [a suitable bot token](https://api.slack.com/apps/A047F02PTE2/oauth?).
- Added the bot token as a `SLACK_TOKEN` GitHub org-level secret for Actions

### Testing

- [x] Checks pass
- [x] Manually invoke to test ([workflow runs](https://github.com/nextstrain/nextstrain.org/actions/workflows/remind-to-promote.yml), [Slack message from last run](https://bedfordlab.slack.com/archives/C02LTGS6NLQ/p1666391445716149))

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "http-proxy-middleware": "^1.3.1",
     "jest": "^27.5.1",
     "jest-extended": "^1.1.0",
+    "luxon": "^3.0.4",
     "request": "^2.88.2",
     "start-server-and-test": "^1.11.4"
   },
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,5 +99,6 @@ jobs:
           curl "$(jq -r .output_stream_url build.json)" \
             --fail --silent --show-error --location
 
-      - name: Logout of Heroku
+      - if: always()
+        name: Logout of Heroku
         run: sed -i -e '/^machine api\.heroku\.com/d' ~/.netrc
--- a/.github/workflows/remind-to-promote.yml
+++ b/.github/workflows/remind-to-promote.yml
@@ -1,0 +1,59 @@
+name: remind to promote
+
+on:
+  # Run every hour (a little before the top of the hour), but note that
+  # scripts/remind-to-promote may choose not to check at certain times.
+  schedule:
+    - cron: '48 * * * *'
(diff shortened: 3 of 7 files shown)


package.json
```diff
@@ -47,6 +47,7 @@
     "jose": "npm:jose-node-cjs-runtime@^3.11.3",
     "js-yaml": "^4.0.0",
     "lodash.partition": "^4.6.0",
+    "luxon": "^3.0.4",
```
victorlin, line 50: Can this be in `devDependencies` since scripts/remind-to-promote.js is not actually used by the app?

tsibley: Ah, yes.

tsibley: Fixed in repush.

victorlin: Lint is now failing:

```
> eslint --ext .js,.jsx .


/home/runner/work/nextstrain.org/nextstrain.org/scripts/remind-to-promote.js
Error:   3:1  error  'luxon' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
```

This is tricky because we have multiple projects in this repo and ESLint is using the same rules for all projects. Yet, some rules such as `import/no-extraneous-dependencies` are project-specific.

One solution would be to limit ESLint to just `src/`, but it's nice to have some form of linting on other JS files.

Another solution, which retains global linting, would be to create a new `scripts/package.json`. This rule would then search there based on https://github.com/import-js/eslint-plugin-import/issues/1174.

tsibley: Oy. Fixed with [34df2bc](https://github.com/nextstrain/nextstrain.org/pull/615/commits/34df2bc3a9ece6551de451b200c848aeb1d305c9).

victorlin (approved):

PR #618, opened 2022-10-28
Feat/narratives test

jameshadfield:
This sets up a review app to test the narratives debugger (https://github.com/nextstrain/auspice/pull/1590).

It is a draft as it shouldn't be merged in it's current state!

Note that the changes to nextstrain.org are minimal and amount to routing a single URL pathname towards auspice.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "@aws-sdk/client-iam": "^3.53.0",
     "@aws-sdk/client-s3": "^3.53.1",
     "argparse": "^1.0.10",
-    "auspice": "2.40.1",
+    "auspice": "2.41.0",
     "aws-sdk": "^2.908.0",
     "chalk": "^2.4.1",
     "compression": "^1.7.3",
--- a/src/app.js
+++ b/src/app.js
@@ -418,6 +418,11 @@ app.route("/schemas/*")
 app.route("/dist/*")
   .all(endpoints.static.auspiceAssets, (req, res, next) => next(new NotFound()));
 
+/* Auspice has a special /edit/narratives route -
+ * It is not backed by a dataset, and only exists for GET requests
+ */
+app.routeAsync("/edit/narratives")
+  .getAsync(endpoints.static.sendAuspiceEntrypoint);
 
 /* Gatsby static HTML pages and other assets.
  *
--- a/package-lock.json
+++ b/package-lock.json
@@ -5252,9 +5252,9 @@
           }
         },
         "@babel/generator": {
-          "version": "7.20.3",
-          "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.20.3.tgz",
-          "integrity": "sha512-Wl5ilw2UD1+ZYprHVprxHZJCFeBWlzZYOovE4SDYLZnqCOD11j+0QzNeEWKLLTWM7nixrZEh7vNIyb76MyJg3A==",
+          "version": "7.20.4",
+          "resolved": "https://registry.npmjs.org/@babel/generator/-/generator-7.20.4.tgz",
+          "integrity": "sha512-luCf7yk/cm7yab6CAW1aiFnmEfBJplb/JojV56MYEK7ziWfGmFlTfmL9Ehwfy4gFhbjBfWO1wj7/TuSbVNEEtA==",
           "requires": {
             "@babel/types": "^7.20.2",
             "@jridgewell/gen-mapping": "^0.3.2",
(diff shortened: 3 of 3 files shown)


package.json
```diff
@@ -27,7 +27,7 @@
     "@aws-sdk/client-iam": "^3.53.0",
     "@aws-sdk/client-s3": "^3.53.1",
     "argparse": "^1.0.10",
-    "auspice": "2.40.0",
+    "auspice": "github:nextstrain/auspice#aa150e3",
```
jameshadfield, line 30: @victorlin do you know why this is failing? It uses the same approach as the automatic PRs made from the auspice repo (https://github.com/nextstrain/auspice/blob/master/.github/workflows/make_pr_for_downstream_repo.yaml#L59)

```
Run npm ci
npm ERR! Error while executing:
npm ERR! /usr/bin/git ls-remote -h -t ssh://git@github.com/nextstrain/auspice.git
npm ERR! 
npm ERR! Warning: Permanently added the ECDSA host key for IP address '140.82.114.4' to the list of known hosts.
npm ERR! git@github.com: Permission denied (publickey).
npm ERR! fatal: Could not read from remote repository.
npm ERR! 
npm ERR! Please make sure you have the correct access rights
npm ERR! and the repository exists.
npm ERR! 
npm ERR! exited with error code: 128
```

jameshadfield: It also seems to have worked in the past (e.g. https://github.com/nextstrain/nextstrain.org/pull/533), so I wonder if this is a GitHub action change or we need to add SSH credentials?

jameshadfield: The heroku review app also failed, with some more helpful logs:

```
       npm ERR! Error while executing:
       npm ERR! /usr/bin/git ls-remote -h -t ssh://git@github.com/nextstrain/auspice.git
       npm ERR! 
       npm ERR! Host key verification failed.
       npm ERR! fatal: Could not read from remote repository.
       npm ERR! 
       npm ERR! Please make sure you have the correct access rights
       npm ERR! and the repository exists.
       npm ERR! 
       npm ERR! exited with error code: 128
```

victorlin: @joverlee521 that sounds right. @jameshadfield it looks like you used a newer version of NPM during `npm install` which upgraded the lockfile version. I'll regenerate with npm 6.14.15 which should fix it.

jameshadfield: Aha - another gotcha. Thanks both!

I knew I was using a newer version of node/npm than we want for this repo as we don't want to get the nasy lockfile-version-changes, but didn't realise it would have flow on effects such as this!

victorlin: Not your fault, it's too easy for this to happen! Created a separate issue: #620

jameshadfield:
@tsibley could I ask you to review d145a5fc6b41aecb5f0e957ee105e8a006be6963? (The actual pathname will probably change, so that one day it can expand beyond "just" a debugger.)

src/app.js
```diff
@@ -418,6 +418,15 @@ app.route("/schemas/*")
...
 
+/* debug-narrative is a special page within auspice
+ * (currently experimental). It is not backed by a dataset,
+ * and only exists for GET requests for a HTML page
+ */
+app.route("/debug-narrative")
+  .get(async (req, res) => {
+    if (!req.accepts("text/html")) throw new NotAcceptable();
+    return await endpoints.static.sendAuspiceEntrypoint(req, res);
+  });
```
tsibley, line 429: To safely use async handlers, we have to use `routeAsync`, `getAsync`, etc. These are safe to use with non-async handlers too, so it's acceptable to always use them if you're unsure.

Furthermore, I would likely simplify this to:

```suggestion
app.routeAsync("/debug-narrative")
  .getAsync(endpoints.static.sendAuspiceEntrypoint);
```

It's ok for routes that don't support multiple content types to not bother with content negotiation, especially routes not intended for programmatic usage.

tsibley (changes requested):

src/app.js
```diff
@@ -20,7 +20,7 @@ const CANARY_ORIGIN = process.env.CANARY_ORIGIN;
 /* eslint-disable import/first */
 import * as utils from './utils/index.js';
 import { addAsync } from './async.js';
-import { Forbidden, NotFound, Unauthorized } from './httpErrors.js';
+import { Forbidden, NotFound, Unauthorized, NotAcceptable } from './httpErrors.js';
```
tsibley, line 23: Remember to drop this if you cherry-pick the suggestion in https://github.com/nextstrain/nextstrain.org/pull/618/files#r1012106232.

src/app.js
jameshadfield: Thanks!

If we don't do this, then were the `getAsync` to fail (reject) then it's unhandled because the middleware chain has already completed, whereas if we use `app.routeAsync` then the middleware will await the `getAsync` and so we can catch errors there?

tsibley: Yeah, errors can get lost/subvert the typical handling.

src/app.js
```diff
@@ -418,6 +418,11 @@ app.route("/schemas/*")
 app.route("/dist/*")
   .all(endpoints.static.auspiceAssets, (req, res, next) => next(new NotFound()));
 
+/* debug-narrative is a special page within auspice.
```
jameshadfield, line 421: typo - commit needs a fixup

PR #619, opened 2022-10-28
scripts/collect-search-results: Update `fileToUrl`

joverlee521:

Replace all instances of underscore in the filename to create the URL. Never noticed this issue before, but now the underscores in the URLs lead to a BadRequest error since we explicitly forbid underscores in the resource path in #611.

### Testing

Tested locally for both SARS-CoV-2 and seasonal flu. 

- [x] Checks pass


joverlee521:
Merging since the only users of the script are the [update-search workflow](https://github.com/nextstrain/nextstrain.org/blob/master/.github/workflows/update-search.yml) and myself.

tsibley:
Thanks for catching this!

jameshadfield:
Thanks Jover!

PR #630, opened 2022-11-16
migrate-group: Exclude S3 object keys containing "/" from syncing

tsibley:
…even if they match *.json or *.md.  Instead, report them as unsynced files.

This correctly handles an observed case of an object named "archive/zika.json" that was dutifully copied over, though it would be inaccessible.

Related-to: <https://github.com/nextstrain/private/issues/66>

### Testing

- [x] Manually ran (in dry run mode!) against original group where this was observed
- [x] Checks pass (but are mostly meaningless)

### DIFF
--- a/scripts/migrate-group.js
+++ b/scripts/migrate-group.js
@@ -103,6 +103,7 @@ async function syncData({dryRun = true, group}) {
     filters: [
       "--exclude=*",
       "--include=*.json",
+      "--exclude=*/*",
     ]
   });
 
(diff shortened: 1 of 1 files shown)


jameshadfield:
I forgot to comment here but this seems sensible and the right thing to do 👍 

I don't think we're deleting the old buckets immediately, but may be worth an email to those who had files that are now no longer stored on the new groups interface.

tsibley:
> I don't think we're deleting the old buckets immediately, but may be worth an email to those who had files that are now no longer stored on the new groups interface.

Yep, I've been noting this in my emails to owners when telling them their group's migration is complete. (Datasets and narratives with slashes in their name aren't the only such cases.)

PR #631, opened 2022-11-17
Analytics

jameshadfield:
Because of the way plausible domains work (they're a tag, not actually domain-restricted) the review app here _should_ result in dev.nextstrain.org activity on plausible. Let's see.


jameshadfield:
This works 👍 (make sure to enable the plausible request on uBlockOrigin if you're using it).

I'll go update the domain (to nextstrain.org) & release a compatable version of auspice before updating this PR

tsibley:
This is pending https://github.com/nextstrain/auspice/pull/1597.

tsibley:
I'll pick this up.

tsibley:
One thing worth noting is that the analytics in Plausible will also include traffic to next.nextstrain.org, dev.nextstrain.org (eventually), our PR review apps, etc. as the Plausible domain (nextstrain.org) is statically compiled into Auspice and Gatsby rather than a dynamic runtime config. We could parameterize the Auspice and Gatsby compile-time configs at Heroku build time, but not sure that's worth it at the moment. Anyhow, something to note for now and maybe do something about later.

PR #580, opened 2022-08-15
SARS-CoV-2 Forecasts

joverlee521:
### Description of proposed changes
Adds a Gatsby page for SARS-CoV-2 forecasts that describes the forecast models and displays the static images of plots generated by the [forecasts-ncov pipeline](https://github.com/nextstrain/forecasts-ncov) (not yet implemented). 

The display of static images is expected to be temporary as we will develop interactive visualizations for modeling results. 

### TODOs
- [x] replace placeholder texts (@trvrb)
- [x] replace placeholder images

PR #641, opened 2022-12-21
Use stand-alone forecasts viz app

jameshadfield:
The viz app development has happened in https://github.com/nextstrain/forecasts-ncov/pull/8.

I tried many steps to avoid having another app in our build steps, but ultimately this was the cleanest! Specifically

* our version of gatsby is not compatible with d3 (specifcially the d3 libraries which use DOM selections). There are a few random GitHub issue comments indicating a potential solution, but after spending an afternoon on it I decided this wasn't worth it. Note that (a) we're on Gatsby v2 still and (b) we want to move off Gatsby.

* I tried using rollup to produce a bundle which Gatsby could import. Gatsby (perhaps as expected) tries to transpile this which we don't want. We could imagine having a specific version of the app which renders (in an IIFE) into a specifc element, and importing this in a <script> tag, but I didn't explore this.

* I tried cloning / copying the viz source into nextstrain.org and importing certain static-site components (e.g. header, footer) but this is not possible in create-react app.

* I tried the above, but copying the (static-site) files (header, footer) however the dependencies go deep and this felt like a bad road to go down!

Finally I settled on this (temporary) solution. See `./build.sh` for how to build.


build.sh
```diff
@@ -46,4 +49,15 @@ build-auspice() {
     cd ..
 }
 
+build-forecasts() {
+    git clone git@github.com:nextstrain/forecasts-ncov.git
```
huddlej, line 53: Do we want this line to look like the following instead?

```
git clone https://github.com/nextstrain/forecasts-ncov.git
```

[CI fails to clone this repo due to a public key error](https://github.com/nextstrain/nextstrain.org/actions/runs/3746695720/jobs/6362237409#step:6:395) and I get the same error from local command line.

.gitignore
```diff
@@ -27,3 +27,6 @@
 
 # docs build
 /docs/build/
+
+# forecasts-ncov is a directory (git repo) introduced at build time
+forecasts-ncov/
```
tsibley, line 32: I'd rather not introduce whole new chunks of code at build time that are unrelated to the code in this repo, as it adds lots of unnecessary complexity during deployment which can be a rich source of bugs. For example, it allows changes in the forecasts-ncov repo to break our build (or deployed app) on Heroku when things worked fine locally or in CI. It also means that redeploys without any code changes on our end will "auto-upgrade" the forecasts-ncov viz app. Some of those are addressable by fetching a static ref instead or trying to be extra careful, but it seems better to me to make them non-issues altogether.

Instead, we can vendor the forecasts-ncov repo and then control in this repo when we take changes from it. This provides a much more controlled and reproducible build and deployment process. I've had good experiences over the years with `git subrepo` (a third-party tool) to do this, and there's now also `git subtree` in core git to accomplish similar ends.

tsibley (changes requested):

src/app.js
```diff
@@ -126,6 +126,18 @@ app.use((req, res, next) => {
...
+/**
+ * The forecasts-viz app is standalone until such time as we remove gatsby,
+ * then it should be part of the app which replaces the static-site.
+ */
+app.routeAsync("/sars-cov-2/forecasts/interactive")
+  .getAsync(endpoints.static.sendForecastsVizPage("index.html"));
+;
+// the app has one (hardcoded) script import at /static/js/main.<hash>.js)
+app.route("/static/js/*")
+  .all(endpoints.static.forecastsVizAssets, (req, res, next) => next(new NotFound()));
```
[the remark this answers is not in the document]
jameshadfield: I'll take a look if create-react-app can configure this prefix

.gitignore
jameshadfield: We can vendor it for sure. While this is a temporary solution, I understand we want it to be reproducible while it lasts. I see the medium term direction as something like this:
 
1. Turn the viz app into published, versioned importable react components (and probably this no longer lives in forecasts-ncov)
2. Remove Gatsby, and whatever we use to replace it then has forecasting pages which import those react components

src/app.js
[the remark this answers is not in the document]
jameshadfield: Precarious is the right word -- I expected this to be fixed up as I moved from a [wip] to a finalised PR 😄 

The viz app now requests static assets from `/forecasting-viz-assets/*` which makes routing more straightforward and less prone to conflict, and I've moved the routes down to the end of `app.js` (well, as far as I can whilst being above the 'Everything else gets 404ed.' route 😉 )

build.sh
jameshadfield: Thanks John! Implemented this change and both CI + heroku can now build.

(I'm going to change to something like `git subtree`, but that's separate.)

.gitignore
jameshadfield: This was a good example to test out `git subtree` which I'd read about but not used - thanks!

jameshadfield:
P.S. [Latest CI failure seems transient](https://github.com/nextstrain/nextstrain.org/actions/runs/3753746737/jobs/6377289786)

jameshadfield:
Rebased onto the updated #580, and ammended to use the latest forecasts-ncov main branch. I've addressed the review concerns but if there are still desired changes we can make them in #580.

PR #648, opened 2023-01-11
SC2 Forecasts Improvement(s)

joverlee521:
Nitpick improvement(s) to the SC2 forecasts page.


jameshadfield (approved):
LGTM. I'll write up some thoughts on the direction this should go so we can work out the next steps -- a single PNG per panel is just the first/easiest implementation but certainly not the best.

joverlee521:
Closing since we are now using different sized images for different resolutions in #660

PR #649, opened 2023-01-23
Revert "SARS-CoV-2 Forecasts"

joverlee521:
Reverts nextstrain/nextstrain.org#580

During the forecasting-automation meeting on 2023/01/23, we decided to rollback the sc2/forecasts pages for now. 
Reverting so that we don' t have any unused code hanging around.


joverlee521:
Merging without review since this is purely a revert and the test app built successfully.

PR #654, opened 2023-02-08
sources: Bypass CloudFront for core and staging source manifests

tsibley:
This usage was missed by "sources: Bypass CloudFront for core and staging source S3 buckets" (4365b26).  The rationale for that commit applies here as well.

Related-to: <https://github.com/nextstrain/nextstrain.org/pull/552>

### Testing

- [x] Manifest loads locally
- [x] Checks pass


tsibley:
Merging and will deploy. I think this is low chance of negative impact.

PR #660, opened 2023-02-27
Restore SARS-CoV-2 Forecasts

joverlee521:

Restore SARS-CoV-2 Forecasts page + interactive app. 

### Testing

- [x] Checks pass


joverlee521:
Text on the main forecasts page (`/sars-cov-2/forecasts`) should be edited within the static section of the nextstrain.org repo (`/static-site/src/sections/sars-cov-2-forecasts-page.jsx`). 

Text on the interactive viz page (`/sars-cov-2/forecasts/interactive`) should be edited within the viz app of the forecasts-ncov repo (`viz/src/Panels.js`).

joverlee521:
In offline chat with @jameshadfield, we decided the easiest path forward is to only include with the static page/static images and remove the interactive viz app for now. This will give us more time to improve the viz app itself and to decide on how to move forward with replacing Gatsby to allow D3 components within nextstrain.org. It is also simpler to edit the static text just within the nextstrain.org repo and not have to work with forecasts-ncov repo as a git subtree. 

This PR has been updated with cherry-picked commits from #580 that only add/update the static web page.

static-site/src/sections/sars-cov-2-forecasts-page.jsx
```diff
@@ -40,6 +40,29 @@ const introContents = [
   },
 ];
 
+
+const nextstrainCladesSrcBase = "https://nextstrain-data.s3.amazonaws.com/files/workflows/forecasts-ncov/gisaid/trial/viz-library-mk2/nextstrain_clades/global";
```
joverlee521, line 44: Need to update this URL once the latest forecasts figures have been uploaded to S3

static-site/src/sections/sars-cov-2-forecasts-page.jsx
```diff
@@ -0,0 +1,203 @@
...
+          {text}
+        </splashStyles.FocusParagraph>
+        <MediumSpacer />
+        <FullWidthPicture>
+          {images[props.cladeType]?.srcSets
+            ? images[props.cladeType].srcSets
+                .map((source) => <source key={source.srcset} media={source.media} srcSet={source.srcSet}/>)
+            : null
+          }
+          <img
```
jameshadfield, line 193: On iPhone, the screen width is 375px (ish) but the `_small.png` is 504px and it overflows. Setting `max-width: 100%;` on the `` element scales this down (has to be on the `` from my testing).

Otherwise the responsive images look great!

joverlee521: Thanks for the tip! I updated the style in the force-push.

static-site/src/sections/sars-cov-2-forecasts-page.jsx
joverlee521: Updated in force-push.

jameshadfield (commented):
The images + banner look good @joverlee521. We still need abstracts / titles ammended.

huddlej (commented):
I like the new disclaimer and lead text at the top of the page! It's much clearer now that this is a work in progress and it's nice to have the links out to the GitHub repos that produce the plots.

I wish there was some text explaining why the countries that are shown were selected. Following the link to the forecasts-ncov README doesn't give me any details about this. Could that information be included in either or both of the main page and README?

I also found myself wishing there was a line at y=1 in the growth advantage plots, since that is the reference point relative to BA.2.

I noticed that there is a lot of white space between columns in both sections of plots. What if the overall figures (representing all three columns) were narrower or the individual panels were wider? I see why one might prefer to maintain the equal aspect ratio, so narrowing the outer figure might be better.

Finally, both the page title and the URL emphasize "forecasts", but the current figures do not include forecasts. Is there a broader name we could use to describe these types pages that live outside of the standard phylogenetic view? Something like "/dynamics/" or "/frequencies/" could capture the contents without ruling out forecasts in the future.

marlinfiggins:
Lovely work! I do like the new disclaimer and the lead text. I think there's still use for a technical details page but that can wait for now.

I also agree with John and Richard that current title is a bit confusing. I would prefer switching it to dynamics as John suggests. This also highlights that the point is not restricted purely to frequency forecasts and that we're also interested in the growth advantages as well.

Also, seconding that we should have sort of line at GA=1, so that it is clear that the reference point for BA.2. growth advantage is clear.

### Things I think should be changed now
- Changing title to dynamics
- Adding line for GA = 1

### Things that should occur over next couple of weeks
- Moving back to interactive viz
- Toggle to show underlying frequencies (perhaps with Binomial Sampling CIs)
- Toggle to logit-scale
- Adding in proper forecasts and a toggle here to show them
- Adding in DirMultinomial option to allow for over-dispersion in sequence count
- Technical details page
- (Pie in the sky) Adding in toggle for posterior predictive checks

joverlee521:
Thanks for reviewing @huddlej and @marlinfiggins! 

> I wish there was some text explaining why the countries that are shown were selected. Following the link to the forecasts-ncov README doesn't give me any details about this. Could that information be included in either or both of the main page and README?

I've added a [Data Prep section](https://github.com/nextstrain/forecasts-ncov#data-prep-configurations) to the forecasts-ncov README. I'm hesitant to put this in the webpage in case data prep configs change over time. 

> I noticed that there is a lot of white space between columns in both sections of plots. What if the overall figures (representing all three columns) were narrower or the individual panels were wider? I see why one might prefer to maintain the equal aspect ratio, so narrowing the outer figure might be better.

I believe @jameshadfield will be adjusting the white space for the larger screensizes in the [viz library](https://github.com/nextstrain/forecasts-viz). (see [Slack thread](https://bedfordlab.slack.com/archives/C036L7V615F/p1680308807072109?thread_ts=1680229857.935619&cid=C036L7V615F))

> I also agree with John and Richard that current title is a bit confusing. I would prefer switching it to dynamics as John suggests. 

I've renamed things from "forecasts" to "dynamics" in 9b1d7295d2bb363d471231b42d675e07046a1955.

> Also, seconding that we should have sort of line at GA=1, so that it is clear that the reference point for BA.2. growth advantage is clear.

This will also have to be changed in the [viz library](https://github.com/nextstrain/forecasts-viz) used to generate the static images. Once updated in the viz library and we update forecasts-ncov to use the latest tarball, then the underlying static images will include the threshold line without any further changes here.

PR #667, opened 2023-03-22
Add workflow to auto-merge Auspice bump PRs

victorlin:

The auto-merge is reasonable because the target branch is deployed to the canary site (next.nextstrain.org), not the production site. Auto-merge allows us to test the latest Auspice release on the canary site without manually merging the automatically created version bump PRs.

### Changes outside of this PR

- [x] Regenerated `[org secret] GH_TOKEN_NEXTSTRAIN_BOT_REPO` and renamed as `[org Actions+Dependabot secret] GH_TOKEN_NEXTSTRAIN_BOT_REPO`
- [x] Created a new organization Dependabot secret on named `GH_TOKEN_NEXTSTRAIN_BOT_REPO` with the regenerated value
- [x] Updated the value of the existing organization Actions secret with the same name

### Related issue(s)

Prompted by [Slack discussion](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1679447507855999).

### Testing

- [x] Auto-merge was successful through manual testing on my personal fork repo ([PR](https://github.com/victorlin/nextstrain.org/pull/4), [run](https://github.com/victorlin/nextstrain.org/actions/runs/4513731600/jobs/7948936949)). Dependabot will wait for CI to pass then merge.
- [ ] Post-merge: new job is skipped on non-dependabot PRs (can't test this until the workflow exists in `master`)

### DIFF
--- a/.github/workflows/dependabot-automation.yml
+++ b/.github/workflows/dependabot-automation.yml
@@ -1,0 +1,27 @@
+name: Dependabot automation
+
+on: pull_request
+
+permissions:
+  pull-requests: write
+  contents: write
+
+jobs:
+  merge-auspice-bump:
+    if: github.event.pull_request.user.login == 'dependabot[bot]'
+    runs-on: ubuntu-latest
+    steps:
+      - id: metadata
+        uses: dependabot/fetch-metadata@v1
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      # Run if Auspice is an updated dependency and there are no major version bumps
+      # to allow for manual testing before merging.
+      - if: |
+          contains(fromJSON(steps.metadata.outputs.updated-dependencies-json).*.dependencyName, 'auspice')
+          && steps.metadata.outputs.update-type != 'version-update:semver-major'
+        # Dependabot should ensure CI passes before merging.
+        run: gh pr comment "${{ github.event.pull_request.html_url }}" --body "@dependabot merge"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN_NEXTSTRAIN_BOT_REPO }}


.github/workflows/dependabot-automation.yml
```diff
@@ -0,0 +1,21 @@
+name: Dependabot automation
+
+on: pull_request
```
tsibley, line 3: This will run as soon as the PR is opened, yeah? Maybe let's wait until CI passes for it so we don't auto-merge broken stuff?

victorlin: > This will run as soon as the PR is opened, yeah?

Yeah good point, I overlooked this while testing.

> Maybe let's wait until CI passes for it so we don't auto-merge broken stuff?

I thought of 2 ways to do this:

1. Use GitHub's [auto-merge](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request) feature. This requires enabling the following features:
    1. **Require status checks to pass before merging** under the repository's `master` branch protection rule
    2. **Allow auto-merge** under the repository's general settings
2. Trigger on `workflow_run` instead. This is what I went with in 13fbd76ed86d5a29bb9f2eb95bb195aae8549a10. It looks slightly hacky since `dependabot/fetch-metadata` [can no longer be used](https://github.com/dependabot/fetch-metadata/blob/aea2135c95039f05c64436f1d14638c300e10b2b/src/dependabot/verified_commits.ts#L16-L17), but avoids the implications of (1).

victorlin: A slight downside to (2): it is not obvious which workflow run is doing the auto-merging whereas previously the `pull_request` trigger allows the workflow run to show on the commit.

`pull_request` trigger:

image

`workflow_run` trigger:

image

victorlin: Oh smart, why didn't I think of that... would've saved me a morning of tinkering with GitHub Actions. I'll try it out on the fork and update this PR.

victorlin: Done: 91355967558fc8101c2c8a98b7cb64893ed35da0

Tested over at https://github.com/victorlin/nextstrain.org/pull/4.

.github/workflows/dependabot-automation.yml
```diff
@@ -0,0 +1,22 @@
...
+jobs:
+  merge-auspice-bump:
+    if: github.event.pull_request.user.login == 'dependabot[bot]'
+    runs-on: ubuntu-latest
+    steps:
+      - id: metadata
+        uses: dependabot/fetch-metadata@v1
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+      - if: contains(steps.metadata.outputs.dependency-names, 'auspice')
```
tsibley, line 18: This output is [described as](https://github.com/dependabot/fetch-metadata):

> A comma-separated list of the package names updated by the PR.

which means that `contains()` will be a naive match and may produce false positives.

How about instead using `steps.metadata.outputs.updated-dependencies-json` and looking for an entry with `dependencyName == 'auspice'`?

Relatedly, we might want to limit auto-merges to cases where `updateType != 'version-update:semver-major'` (or similar)? (This is a bit far forward looking, but still…)

tsibley: It'd be something like:

```yaml
      - if: contains(fromJSON(steps.metadata.outputs.updated-dependencies-json).*.dependencyName, 'auspice')
```

victorlin: Good suggestions! I'll try them out on my fork repo.

victorlin: Tested, works well:

- https://github.com/victorlin/nextstrain.org/pull/5
- https://github.com/victorlin/nextstrain.org/pull/6

.github/workflows/dependabot-automation.yml
```diff
@@ -0,0 +1,27 @@
...
+      - id: metadata
+        uses: dependabot/fetch-metadata@v1
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      # Run if Auspice is an updated dependency and it is not a major version bump
+      # to allow for manual testing before merging.
+      - if: |
+          contains(fromJSON(steps.metadata.outputs.updated-dependencies-json).*.dependencyName, 'auspice')
+          && steps.metadata.outputs.update-type != 'version-update:semver-major'
```
tsibley, line 23: _non-blocking, informational only_

This top-level `update-type` will be the aggregate (sort of a `max()`) of everything updated, which may include deps other than Auspice, right? But I guess in practice, for now, only Auspice will appear because of our current dependabot config?

The `auspice` entry in `updated-dependencies-json` also has its own specific `update-type`, but it looks like it's not possible in GitHub's expression syntax to extract that entry from the array, so we can't address its `update-type` in a condition. (We could extract it elsewhere, in another job step, but that seems unnecessary for now I guess.)

tsibley (approved):

.github/workflows/dependabot-automation.yml
victorlin: Oh I didn't realize that, thanks. I updated the comment in [force-push](https://github.com/nextstrain/nextstrain.org/compare/670b3b0df07d9792437e50fe67696475f4a2033b..2bb5028d6e2ae143354192a9771c93aacdc34093) to be more accurate.

victorlin:
[Failing test](https://github.com/nextstrain/nextstrain.org/actions/runs/4577824490/jobs/8083704853) seems transient – it [passed](https://github.com/nextstrain/nextstrain.org/actions/runs/4577836018/jobs/8083731957) the merged commit.

joverlee521:
Just made an Auspice release for the first time since this workflow was implemented! It totally worked as expected 🎉 

The only hiccup I ran into was the [deploy job](https://github.com/nextstrain/nextstrain.org/actions/runs/4986204078/attempts/1) did not have access to our Heroku secrets because the GH Action was triggered by dependabot. I've added the Heroku secrets to the [Dependabot secrets](https://github.com/nextstrain/nextstrain.org/settings/secrets/dependabot) and the [latest run](https://github.com/nextstrain/nextstrain.org/actions/runs/4986204078) is working as expected.

PR #672, opened 2023-04-04
scripts: collect all S3 objects when truncated

### LINKED ISSUE #670: `collect-dataset` and `collect-search-results` returning empty datasets
### Current Behavior
Scripts `collect-dataset` and `collect-search-results`, which collect dataset info from the `nextstrain-data` AWS S3 bucket return empty arrays for datasets. (See [example](https://github.com/nextstrain/nextstrain.org/actions/runs/4599878907/jobs/8125808754#step:8:16) in our scheduled GH Actions.)

This is resulting in an empty dataset list at https://nextstrain.org/influenza/ and empty search list at https://nextstrain.org/search/sars-cov-2. 

I noticed the empty results when I was running `collect-search-results` manually for seasonal flu data.

joverlee521:
Collect all S3 objects by using the `ContinuationToken` param and fetching all content until `IsTruncated` is no longer true.

### Related issue(s)
Fixes #670.

### Testing
Scripts worked locally with:
```
./scripts/collect-datasets --keyword flu 
./scripts/collect-datasets --keyword staging

./scripts/collect-search-results --pathogen sars-cov-2
./scripts/collect-search-results --pathogen seasonal-flu
```

This now pulls >6000 datasets from staging and all of the dated datasets for ncov. 
Should we add additional filters to limit the number of datasets? Maybe limit objects by their `LastModified` date?

### DIFF
--- a/scripts/collect-search-results.js
+++ b/scripts/collect-search-results.js
@@ -1,9 +1,9 @@
 #!/usr/bin/env node
 import argparse from 'argparse';
-import AWS from 'aws-sdk';
 import fetch from 'node-fetch';
 import fs from 'fs';
 import pLimit from 'p-limit';
+import S3 from '@aws-sdk/client-s3';
 
 const bucket = "nextstrain-data";
 
@@ -98,22 +98,12 @@ async function collectFromBucket({BUCKET, fileToUrl, inclusionTest}) {
 }
 
 async function listAllObjects({BUCKET}) {
-  const S3 = new AWS.S3();
-  const s3ListObjectParams = {
-    Bucket: BUCKET,
-    ContinuationToken: null
-  };
+  const client = new S3.S3Client();
   const s3Objects = [];
 
   try {
-    let isTruncated = true;
-    while (isTruncated) {
-      // eslint-disable-next-line no-await-in-loop
-      const s3Response = await S3.listObjectsV2(s3ListObjectParams).promise();
-      s3Objects.push(...s3Response.Contents);
-
-      s3ListObjectParams.ContinuationToken = s3Response.NextContinuationToken;
-      isTruncated = s3Response.IsTruncated;
+    for await (const page of S3.paginateListObjectsV2({client}, {Bucket: BUCKET})) {
+      s3Objects.push(...page.Contents);
     }
   } catch (err) {
     console.log("Error listing objects via the S3 API -- were credentials correctly set?");
--- a/scripts/collect-search-results.js
+++ b/scripts/collect-search-results.js
@@ -66,18 +66,9 @@ async function main({args}) {
 // -------------------------------------------------------------------------------- //
 
 async function collectFromBucket({BUCKET, fileToUrl, inclusionTest}) {
-  const S3 = new AWS.S3();
   console.log(`Collecting datasets from ${BUCKET} to link sample names -> datasets...`);
-  let s3Objects;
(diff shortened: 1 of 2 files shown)


tsibley:
Hmm, can we use a built in paginator here instead (i.e. from the AWS SDK)?

scripts/collect-search-results.js
```diff
@@ -34,7 +34,11 @@ async function main({args}) {
         BUCKET: `nextstrain-data`,
         fileToUrl: (filename) => `https://nextstrain.org/${filename.replace('.json', '').replace(/_/g, '/')}`,
         inclusionTest: (fn) => {
-          return (fn.startsWith("ncov_") && !fn.endsWith("_gisaid.json") && !fn.endsWith("_zh.json"));
+          return (fn.startsWith("ncov_")
+            && !fn.endsWith("_gisaid.json")
+            && !fn.endsWith("_zh.json")
+            // Exclude all dated JSONs to avoid OOM error
+            && !fn.match(new RegExp(/_[0-9]{4}-[0-9]{2}-[0-9]{2}\.json/)));
```
tsibley, line 41: Would be good to anchor this pattern to the end of the name, e.g.

```suggestion
            && !fn.match(new RegExp(/_[0-9]{4}-[0-9]{2}-[0-9]{2}\.json$/)));
```

also, FYI, you can use regex literal syntax:

```suggestion
            && !fn.match(/_[0-9]{4}-[0-9]{2}-[0-9]{2}\.json$/));
```

tsibley (approved):
Looks fine to me from a code perspective, although…

> Avoid OOM error by excluding dated SARS-CoV-2 JSONs.
> This will limit the search page results to the latest datasets.

Is that an acceptable outcome here? ISTR part of the point of the https://nextstrain.org/search/sars-cov-2 page was to cover all those dated builds? Could someone with more background context—maybe @jameshadfield?—chime in here?

scripts/collect-search-results.js
joverlee521: Ah, thanks! Updated in force push.

joverlee521:
> Is that an acceptable outcome here? ISTR part of the point of the https://nextstrain.org/search/sars-cov-2 page was to cover all those dated builds? 

Good question. This brings up the broader question of whether we want to continue maintaining the /search pages. The /search/sars-cov-2 page was _very_ slow when I was testing locally with only the latest build search results. Should we invest time into speeding up the `Select` component?

tsibley:
I think the consensus is we want functionality like the search pages (and to extend that functionality well beyond where it is today). I see the current implementation, though, as an initial "MVP", a stepping stone that was (is?) useful, but isn't the foundation for more extensive, longer-term search functions.

jameshadfield:
It's on my to-do list to remind me what all our various "S3 crawling" scripts are actually doing, and which functionality is still relevant. SARS-CoV-2 strain search is almost certainly not needed anymore - this was created in a time with (multiple?) orders of magnitude fewer sequences and so there was a decent chance your strain of interest was in a recent build (and this page told you which one). We now subsample out ~99.9% of strains, and more importantly Nextclade exists to analyse your sequences of interest. I'd suggest stopping the script running & adding some text to the page to indicate it's no longer supported.

I still want us to aim for generalised dataset collection with search / filtering, but I don't think this script is a stepping stone to it.

PR #673, opened 2023-04-06
Move Auspice dep from the server code to the auspice-client customizations

tsibley:
See commit messages.

### Related issue(s)

https://github.com/nextstrain/auspice/pull/1661

### Testing

- [x] Build works locally
- [x] Checks pass

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -22,8 +22,8 @@ main() {
             build-auspice;;
         all)
             echo "Running the nextstrain.org build script"
-            build-static
             build-auspice
+            build-static    # Depends on Auspice being built
             echo "Build complete. Next step: \"npm run server\"";;
         *)
             echo "Unknown build step \"$step\"" >&2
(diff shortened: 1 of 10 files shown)


build.sh
victorlin, line 1: Based on the CI failure, it looks like this file needs updating to run `build-auspice` before `build-static`.

victorlin (approved):
Glad to see this! Looks good pending the CI fix.

build.sh
tsibley: Ah, indeed. Fixed! Thanks for the diagnosis. It worked for me locally, but I guess that's because I already had `auspice-client/node_modules/`.

tsibley:
So my [first fix](https://github.com/nextstrain/nextstrain.org/compare/36e511c5031714273787eb9f6a8e86d7e1d749f4..6dc47dc8a8db86d751b4fc9c433d37606dd354da) worked in CI… but failed locally when I tested it. That's because I tested it by doing:

```
rm -rf node_modules/ auspice-client/node_modules/ static-site/node_modules/
./build.sh
```

which failed when the Gatsby build couldn't find the `iso-639-1` package. That's because it was still specified in the top-level deps instead of the auspice-client deps, and I'd removed the former. It worked in CI because `npm ci` was run before `./build.sh`.

I [fixed it again](https://github.com/nextstrain/nextstrain.org/compare/6dc47dc8a8db86d751b4fc9c433d37606dd354da..3205a85c0d0a52a84cf1eee326213790024d1443) by moving that dep into auspice-client, which works locally with the above test.

As long as CI passes now, I'll merge.

PR #676, opened 2023-04-11
Split groups data between production and testing environments

tsibley:
These environments do not share the resources behind group—they have separate Cognito user pools and a shared but access-partitioned S3 bucket—so it doesn't make that much sense to bundle both into the same data file anymore.  Pragmatically, this also avoids a bunch of noisy errors at local dev server start and periodically thereafter.

### Testing

- [x] `npm run test:ci` passes locally
- [x] `npm run lint` passes locally
- [x] Checks pass

### DIFF
--- a/docs/infrastructure.md
+++ b/docs/infrastructure.md
@@ -65,6 +65,9 @@ Deploys to the production app are performed by manually [promoting](https://devc
   - `CONFIG_FILE` is the path to a JSON file defining the defaults for the required variables below.
     If not provided, the checked-in files `env/production/config.json` and `env/testing/config.json` are used (depending on the value of `NODE_ENV`).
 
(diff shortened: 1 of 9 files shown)


victorlin (approved):
Looks good!

- [x] I loaded the groups page locally with `nextstrain.org-testing` credentials and saw the groups in testing/groups.json

---

Small thing: I'm assuming the first line in the description of 20a0a63d0becfc6f4ab04a93418299162085a001 should start with

> These environments do not share the **same** resources behind group**s**

tsibley:
> Small thing: I'm assuming the first line in the description of [20a0a63](https://github.com/nextstrain/nextstrain.org/commit/20a0a63d0becfc6f4ab04a93418299162085a001) should start with
> 
> > These environments do not share the **same** resources behind group**s**

Ah, thanks for the copy edit. I changed it to:

image

PR #678, opened 2023-04-16
Move manifest.json files into source code

jameshadfield:
The manifest files are a continual point of confusion, and part of the
reason may be their location on S3. Bringing them into source code has
benefits for visibility/understanding, ease of tracking changes,
and should help with their eventual removal.

Historically, these were part of Auspice's source code and were extracted
into their own files in order to allow other groups to more easily use
Auspice. As nextstrain.org is not intended for external use there is no
problem bringing them back into source code.

Manifests taken 2023-04-17 at ~11h00 (NZ time) using
```
curl --compressed "https://nextstrain-data.s3.amazonaws.com/manifest_guest.json" --output data/manifest_core.json
curl --compressed "https://nextstrain-staging.s3.amazonaws.com/manifest_guest.json" --output data/manifest_staging.json
```

(I'll separate out the first commit as it's own PR if this PR isn't merged for some reason.)

### DIFF
--- a/data/manifest_core.json
+++ b/data/manifest_core.json
@@ -1,0 +1,396 @@
+{
+    "user": "guest",
+    "pathogen": {
+      "dengue": {
+        "serotype": {
+          "denv1": "",
+          "denv2": "",
+          "denv3": "",
+          "denv4": "",
+          "all": "",
+          "default": "denv1"
+        }
+      },
+      "ebola": "",
+      "enterovirus": {
+        "lineage": {
+          "d68": {
+            "segment": {
+              "genome": "",
+              "vp1": "",
+              "default": "genome"
+            }
+          },
+          "default": "d68"
+        }
+      },
+      "flu": {
+        "category": {
+          "avian": {
+            "subtype": {
+              "h5n1": {
+                "segment": {
+                  "ha": "",
+                  "mp": "",
+                  "na": "",
+                  "ns": "",
+                  "np": "",
+                  "pa": "",
+                  "pb1": "",
+                  "pb2": "",
+                  "default": "ha"
+                }
+              },
+              "h5nx": {
+                "segment": {
+                  "ha": "",
+                  "mp": "",
+                  "na": "",
+                  "ns": "",
+                  "np": "",
+                  "pa": "",
+                  "pb1": "",
+                  "pb2": "",
+                  "default": "ha"
+                }
+              },
+              "h7n9": {
+                "segment": {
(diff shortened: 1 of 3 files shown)


src/utils/index.js
```diff
@@ -17,7 +17,18 @@ const getGitHash = () => {
...
 const verbose = (msg, ...rest) => {
+  /**
+   * During server start-up there is a race condition where calls to `utils.verbose()`
+   * can occur before arg parsing, which means global.verbose has not been set.
+   * We store these & print them immediately following our "server loaded" message.
+   */
+  if (global.verbose===undefined) {
+    earlyVerboseStatements.push([msg, ...rest]);
+    return;
+  }
```
tsibley, line 31: _non-blocking_

This seems very hacky to me. For example, the verbose messages will still be lost if there's a crash between the verbose call and the call to `_logEarlyVerboseStatements`.

I'd rather see us move from an option to an environment variable, e.g. `NEXTSTRAIN_DEBUG=1` or similar. Env vars are commonly used to control debugging logs for this reason, i.e. that they're actually global at process start (among other reasons).

Maybe as part of a larger [logging improvement](https://github.com/nextstrain/nextstrain.org/issues/511) effort…

tsibley (approved):
+1 for this move!

After merge and deploy, would you also be sure to

```
aws s3 rm s3://nextstrain-data/manifest_guest.json
aws s3 rm s3://nextstrain-staging/manifest_guest.json
```

?

src/utils/index.js
tsibley: This also shouldn't be needed… it used to work. I suspect the change to use ES modules affected global handling in a non-obvious way.

trvrb:
The ability to have Heroku review apps to test changes here seems very valuable. Currently, the necessity to edit live file places a scary bar on modifying manifest.

src/utils/index.js
tsibley: I propose a different fix that restores how it used to work: https://github.com/nextstrain/nextstrain.org/pull/679

jameshadfield: Yeah, +1 for the larger logging improvement.

> This also shouldn't be needed… it used to work. I suspect the change to use ES modules affected global handling in a non-obvious way.

The manifest `utils.verbose()` calls used to work because they were inside a promise (and thus delayed). I'm not sure if the logging from 6dff7d23abf6f478d6ce2345c81c568992f8e86f ever worked, but it may well have done!

PR #683, opened 2023-04-19
Use same manifest for core + staging

jameshadfield:
As discussed in slack [1]:

> Previously, a big part of the reason we had a manifest_guest.json for
staging, was to be able to test updates to manifest without worrying about breaking the server too badly.

This has some negatives if there are datasets missing on the staging bucket, but the positives greatly outweigh these (and staging is for internal use only, really).

[1] https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1681784685980769

### Testing

- [x] test review app for staging + core URLs


trvrb:
Thanks for attending to this @jameshadfield. Everything looks good from my perspective.

This does surface issue with a few datasets present in manifest and absent from staging.nextstrain.org, but really don't think this is any worse than the current inverse situation with many datasets present at staging.nextstrain.org, but absent from the manifest.

trvrb (approved):

tsibley (approved):
Seems good to me.

PR #687, opened 2023-04-27
Use "href" template literals for safe-by-construction URL strings

tsibley:
Avoids having to write out encodeURIComponent(), which is quite a mouthful!

### Testing

- [x] Checks pass

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -28,6 +28,7 @@ import { replacer as jsonReplacer } from './json.js';
 import * as middleware from './middleware.js';
 import * as redirects from './redirects.js';
 import * as sources from './sources/index.js';
+import { uri } from './templateLiterals.js';
 
 const {
   setSource,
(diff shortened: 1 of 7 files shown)


tsibley:
In [discussion in Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1682624875485889), @joverlee521 suggested `uri` instead of `href`. I'd be fine with this name too.

tsibley:
`uri`'s probably better, even if `href` feels cuter to me.

joverlee521 (approved):
I recommend adding a link to some tagged template docs (e.g. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates). 

I definitely had no idea how the function worked the first time I read through it.

tsibley:
@joverlee521 Fair enough! I added comments and [repushed](https://github.com/nextstrain/nextstrain.org/compare/ac052ed19a347e0c8e1d551e756db8d0f416da7b..63da24bd5a2ad648b1f3ad396e5f20c1604c93e3).

PR #688, opened 2023-05-06
Update ESLint rules

victorlin:

Replace [Airbnb base config](https://github.com/airbnb/javascript/tree/eslint-config-airbnb-v18.2.0/packages/eslint-config-airbnb-base/rules) with other "recommended" base configs, and make other adjustments around that change.

### Related issue(s)

This is the equivalent of https://github.com/nextstrain/auspice/pull/1665 for this repo.

### Testing

- [x] Checks pass, except for unrelated failing docs build


jameshadfield:
Nice to have consistency between the projects! There are other JS projects (e.g. [forecasts-viz](github.com/nextstrain/forecasts-viz)) so I wonder about the pros/cons of making this it's own repo & then the .eslintrc files would just look like:
```
extends:
  - nextstrain
```

.eslintrc.yaml
```diff
@@ -23,7 +23,7 @@ rules:
   prefer-const: ["error", {"destructuring": "all"}]
 
   # Code style rules
-  # arrow-parens: ["error", "always"]
+  arrow-parens: ["error", "always"]
```
tsibley, line 26: Hmm. I'm not a fan of this rule and think it adds noise to functional composition, which is widely used in this codebase. (I'm probably responsible for most of the places this commit adds parens.)

I think `=>` is already an unambiguous marker of a function that the parens aren't needed.

src/authn/index.js
```diff
@@ -153,7 +153,7 @@ function setup(app) {
   // requests to Cognito when we need to load back a user profile from their
   // session cookie.
   passport.serializeUser((user, done) => {
-    let serializedUser = {...user};
+    const serializedUser = {...user};
```
tsibley, line 156: > but I don't see any rationale.

The rationale is unstated, but I used `let` because the variables are modified later. The `let` signals to the reader that modification will happen.

The reason `const` works here is because it only prevents reassignment, not modification. So it's not constant in an intuitive nor useful sense when assigned an object or array (vs. a string/integer literal).

src/app.js
```diff
@@ -205,7 +205,7 @@ const coreBuildRoutes = coreBuildPaths.map(path => [
   `${path}:*`, // Tangletrees at top-level, e.g. /a:/a/b
 ]);
 
-app.use([coreBuildRoutes, "/narratives/*"], setSource(req => new CoreSource()));
+app.use([coreBuildRoutes, "/narratives/*"], setSource(_req => new CoreSource()));
```
tsibley: Changes like this make the linting rule seem a bit silly to me. It's plainly obvious to any reader that `req` isn't used without needing the `_` prefix. We're essentially adding the `_` prefix to the name for the computer alone, as a replacement for adding `// eslint-disable-line no-unused-vars` instead.

src/sources/models.js
```diff
@@ -92,7 +91,7 @@ class Source {
   async baseUrl() {
     throw new Error("async baseUrl() must be implemented by subclasses");
   }
-  async urlFor(path, method = 'GET', headers = {}) { // eslint-disable-line no-unused-vars
+  async urlFor(path, _method = 'GET', _headers = {}) {
```
tsibley, line 94: Another downside of the `_` prefixing instead of the `no-unused-vars` disabling is that if `_method` _does_ start being used in this body, ESLint doesn't care if you rename it to remove the underscore or not. But we can detect unnecessary disablings.

.eslintrc.yaml
victorlin: Would you be in favor of enforcing consistency with other options (i.e. b2c466bdda38f5fdf89286207112b4e1b323b81e instead of 82e6755e5a18c475fdcd3fd0697905f389af63b3), or just dropping this rule entirely?

src/authn/index.js
victorlin: Your proposal is to keep the original code:

```js
let serializedUser = {...user}; // eslint-disable-line prefer-const
```

How about using `const` to be "syntactically correct" along with a comment that more clearly signals modification to the reader? Example:

```js
// Note that this variable will be modified later.
const serializedUser = {...user};
```

src/app.js
victorlin: I'll re-add the existing per-line exceptions but keep the rule enabled. Does that sound good? To me, this rule is still useful for catching things such as unused imports and other unintended unused variables.

.eslintrc.yaml
tsibley: I'd tend towards dropping the rule, as it's stylistic, and I don't think consistency for consistency's sake adds much here.

src/authn/index.js
tsibley: Hmm, I don't see how there's anything more "syntactically correct" about `const` than `let` here. It's a question of style since `const` isn't about actual immutability (and we're not dealing with memory layout implications, like in other languages).

As far as adding a comment goes, I think it's a little bit silly to have a comment saying "Hey, this thing we're calling a constant will, in fact, be modified below."  :upside_down_face: Seems less convoluted to just make it `let`.

src/app.js
tsibley: Sure. I guess I really wish we had a better rule. (We could, but would have to write it, it seems.)

.eslintrc.yaml
victorlin: Replaced 82e6755e5a18c475fdcd3fd0697905f389af63b3 with bb530526987dbe774c220842f41812f970a2aacd.

src/authn/index.js
victorlin: I didn't read up on differences and assumed there were some benefits to using `const`. Looking more, it seems like there isn't.

Replaced aed061754cc0c368c9affb9424f2a1df858a2da9 with f658bd7d43e10b78705c16a1a2b6e677b550c2a9.

src/sources/models.js
victorlin: Got rid of `_`-prefixing in 8a3d795b1c63af3d073e28f2b085c5120a6f3ab1.

tsibley (approved):

jameshadfield (approved):

src/authn/index.js
jameshadfield: [Interested in thoughts -- I'm not trying to re-add this rule to this repo]

For booleans / numbers / strings diligently using `let/const` is useful (for me) when reading and writing code.
For objects (and all their derivatives), framing `let/const` in terms of immutability does make using them diligently seem somewhat pointless. However I've come to read it as something else, namely "this reference won't change"; which _is_ useful to me when reading code. It has particular relevance for things like React which use reference-based equality checks. My 2 cents 😄

tsibley: > However I've come to read it as something else, namely "this reference won't change"

I mean, that's literally what it means! so that's a reasonable reading of it. :-) I just think it's a not very useful choice of meaning and so prefer to either a) ignore the let/const distinction or b) assign `const` a meaning that I think is more useful (if not enforced by the language).

I'm not sure I understand how it's useful in relation to React. React begs for immutability but that's not what it gets with `const`. So if I see React code using a `const` variable and I read `const` as "this reference won't change" (it's literal meaning), I still have to think "well, is anything mutating the members of this object directly instead of passing React a modified copy?". When using `const` instead as "this variable won't change" (my preferred approach (b) above), then I know I don't have to think about what React will see because I wouldn't have marked it `const` if I was mutating it.

PR #698, opened 2023-06-14
Upgrade to Node.js 18

victorlin:

Node.js 16 will reach EOL on 2023-09-11¹. Node.js 18 is the next LTS version, with projected EOL on 2025-04-30.

¹ https://github.com/nodejs/Release/blob/6a881a11363eb601f17b2d46e3f1664db5ce2a3b/README.md

### Testing

- [ ] Checks pass


victorlin:
The build is [failing](https://github.com/nextstrain/nextstrain.org/actions/runs/5261612895/jobs/9509922913#step:6:196) due to Webpack <5 being used in `static-site/` with the same error that prompted https://github.com/nextstrain/auspice/pull/1520 (addressed while adding Node.js 18 support to Auspice):

```
Error:error:0308010C:digital envelope routines::unsupported
```

Based on the [lockfile](https://github.com/nextstrain/nextstrain.org/blob/f7a854a4a4e6779658e6ba9165fd20bda51ebaa4/static-site/package-lock.json#L68162), Webpack 4.46.0 is installed as a dependency of Gatsby. The solution is to use Webpack ≥5, and that is only possible by either upgrading Gatsby to the [next major version](https://www.gatsbyjs.com/docs/reference/release-notes/v3.0/#webpack-5) or removing it.

In other words, upgrading nextstrain.org to use Node.js 18 may be dependent on resolving https://github.com/nextstrain/private/issues/88.

jameshadfield:
Replaced by #810 (which jumps v18 and moves straight to v20)

PR #700, opened 2023-07-13
List datasets + files from core + staging sources

jameshadfield:
This is a pretty major PR and most detail is in commit messages + comments in the code. A high level overview of the functionality introduced:

* For our core & staging sources (`s3://nextstrain-data` and `s3://nextstrain-staging`, respectively) we create daily inventories on S3, which are stored in the private bucket `s3://nextstrain-inventories`. These are essentially free to generate (<$1/year/bucket).
* Our server pulls these inventories down daily and maintains a set of datasets and files, including past versions of each where available.
* A new API is introduced to access this data in a cross-source fashion, including filtering.
* The `/pathogens` page shows the core datasets using the UI [I recently prototyped](github.com/nextstrain/workflow-asset-summary/). A new `/pathogens/inputs` page shows the core files. The `/staging` page (previously not working) also now uses this new UI.

As is the case with feature pushes of this scope there is no clear end point. There are a huge number of potential improvements we can make, so for this PR please indicate if you consider something blocking vs an improvement we can tackle in subsequent PRs 🙏 

From my point of view, the following needs to be done before merge (in the same vein as how I approached the prototype, I'm trying to share work at an early stage):

* Client-side error handling if the API calls 404 / time out / contain no data.
* Amend the production IAM policy to have access to the new bucket (review apps use the testing one which I’ve already modified)
* [Could be done after merge] Implement a S3 lifecycle to expire inventory objects after 30 days (?)

I'll make some notes via in-line comments about feature pushes I don't think are blocking here, but which come up as a result of this work, so that discussions can be threaded.

### Testing

It should 🤞 all work via review apps. To test locally, you can avoid the S3 API calls by creating a (git-ignored) `./devData` folder and adding a manifest+inventory for each bucket with the following filenames:
```console
./devData/core.manifest.json          ./devData/core.inventory.csv.gz
./devData/staging.manifest.json       ./devData/staging.inventory.csv.gz
```
(You can pick any day's inventory, that's not so important for dev purposes.) Then run the server with a `LOCAL_INVENTORY=true` environment variable.

P.S.  _manifest JSON_ here refers to [the S3 inventory manifest](https://docs.aws.amazon.com/AmazonS3/latest/userguide/storage-inventory-location.html#storage-inventory-location-manifest) and is completely unrelated to our existing usage of manifest JSONs. However this work should eventually allow us to remove those manifests. So that's nice.


static-site/src/components/CardsV2/card-component.jsx
```diff
@@ -0,0 +1,205 @@
...
+      data-tooltip-content="Click to view associated pathogen page"
+      href={`https://nextstrain.org${pages[p1]}`}  target="_blank" rel="noreferrer"
+    >
+      <MdCoronavirus size={"1.2em"}/>
+      {pages[p1]}
+    </a>
+  )
+}
+
+function Details({data, view}) {
```
jameshadfield, line 132: [Potentially blocking] The ability to show more detailed information about versions below a card has been disabled here - partly as it requires a bit of polish, and partly as it used d3 to manipulate the DOM and d3 + gatsby = 😡.

Because we group datesamped datasets with the non-datestamped versions (i.e. `.../6m/2023-01-01` is shown as a version of `.../6m`, rather than its own dataset), there is currently no way to actually access this datestamped dataset. These were exposed in the (now removed) table, so this is arguably a downgrade. Although that table was incomplete, so...

static-site/src/components/CardsV2/helpers.js
```diff
@@ -0,0 +1,322 @@
...
+ * available filtering options. These options are dependent on the
+ * current subset of cards displayed -- e.g. if you've previously
+ * filtered by "flu", then the updated set of available filtering
+ * options should only consider keywords in flu-like datasets.
+ */
+export function useFilterOptions(data) {
+  const [state, setState] = useState([]);
+  const counts = {};
+  useEffect(() => {
+    console.log("----- useEffect // useFilterOptions")
```
jameshadfield, line 44: Logs will be removed before merge, but they may be helpful for people during review

src/sources/core.js
```diff
@@ -29,15 +35,52 @@ class CoreSource extends Source {
...
+    this._allResources.set(
+      'dataset',
+      Array.from(datasets).map(([, objects]) => new CollectedResources(this, objects))
+    );
+    this._allResources.set(
+      'file',
+      Array.from(files).map(([, objects]) => new CollectedResources(this, objects))
+    );
+    // TODO XXX narratives
+  }
```
jameshadfield, line 76: There's an inherent linking between (some) datasets and (some) intermediate files. It would be nice to surface this - e.g. when looking at the card for zika you could see the available intermediate files as well. I ummed and ahhed about trying to build this into this version but ultimately left it as a future todo. Beyond the obvious "how do we programmatically know which datasets are connected to which files", the desired UI needs to be worked out before we know how best to approach it. Note that such a linking is cross-source, so if it's done ahead of time on the server, we'll have to perform a per-linked-resource authz filtering (e.g. the intermediate files for a dataset might be stored in a source you don't have permission to view).

src/sources/core.js
```diff
@@ -29,15 +35,52 @@ class CoreSource extends Source {
...
+    }
+    /* Comment (to drop before merge)
+    This conditional could be replaced by overwriting `collectResources` in CoreStagingSource.
+    That would require most of the logic in this function to be pulled out into a utility function,
+    which may not be a bad idea */
+    const CollectedResources = this.name==='core' ?
+      CoreCollectedResources : CoreStagingCollectedResources;
+    const datasets = new Map(), files = new Map();
+    s3objects.forEach((object) => {
+      const [name, resourceType] = CollectedResources.objectName(object);
```
jameshadfield, line 59: Grouping of resources is done by those which have the same _name_, so there are decisions to be made around what name to give things. For instance, I think it's clear that we should group datestamped-datasets with their non-datestamped selves. We should probably also group `.zst` and `.gz` files together (although here I don't). Would be great to have others' thoughts.

tsibley:
Mostly a note to self. Things to make sure to review here based on the Blab Nextstrain meeting just now:

- [ ] Lifecycle policies for `s3://nextstrain-inventories`
- [ ] Terraform config for `s3://nextstrain-inventories`
- [ ] Terraform config for IAM policies
- [ ] Lifecycle policies for other S3 buckets with trial/test/dev stuff… what are we missing?

src/sources/core.js
```diff
@@ -128,6 +175,78 @@ class CoreDataset extends Dataset {
...
+  lastUpdated(version=this._versions[0]) {
+    return version.LastModifiedDate;
+  }
+  
+  filter(options) {
+    /**
+     * We can perform authorization here by either calling 
+     * authz.authorized(user, authz.actions.Read, object),
+     * See note above for more details.
+     */
```
jameshadfield, line 193: @tsibley threading the authz conversation here. For our other APIs, the authz check is done on a single resource (`Resource` subclass instance). Here we have a _collection_ of resources, which may include resources with different filenames etc.

We can easily perform an authorization check on the `CollectedResources` instance, but to stay properly aligned with the current approach the check should potentially be on each individual resource within the collection.

tsibley (commented):
I read thru this linearly, leaving comments on specifics I saw as I went. I mostly skipped over the UI bits as I wanted to focus on the bigger picture. The major takeaways for me:

- I don't think the internal design/data model here is the right one, as I elaborated elsewhere.
- I don't think we should be collapsing /x/y/z and /x/y/z/YYYY-MM-DD, as I elaborated on elsewhere. I don't _think_ this would be a loss of any actual versions, and it makes the conceptual integration with the codebase a lot lot simpler.
- I think the collection/indexing should be happening in a separate process, external from the web server, producing an artifact (a giant JSON blob, a SQLite database, etc) that's consumed by the web server.
- A lot of the code seems to me to suffer from accidental complexity that's not been managed away/minimized. As accidental complexity builds up it drags down a codebase and makes things harder to understand/modify/maintain. I think of it a bit like barnacles or other biofouling on a ship. I realize some of this may be attributable to being a WIP/prototype, but I think it'll be important to address beyond that point rather than forget about.

Those are all really frank, but don't get me wrong, I think we absolutely want the functionality you've prototyped here! I think it will be an important set of features and foundational to future work, enough that we'll really want it on solid footings.

This might be a good focus (like we've been discussing) for the team to collaborate on? Or at least you and I? At least I'd be interested in elaborating in code instead of just commentary. And maybe you'd be interested in further refining the UI-side in parallel?

jameshadfield:
Replaced by #803 and #719

PR #712, opened 2023-09-25
Add mpox as the default monkeypox URL prefix

jameshadfield:
All 3 current /mpox datasets (/mpox/lineage-B.1, /mpox/clade-IIb, /mpox/all-clades) are added to the manifest (and thus Auspice's dropdowns), with the last one set as the default. The card on the main splash page is switched to /mpox.

Context (slack): <https://bedfordlab.slack.com/archives/C03G8HQEV18/p1695508115055009>

The original /monkeypox/* URLs are left unchanged however we can remove these and add redirects instead if we wish to. Steps to do so:
* Removing `monkeypox` from the (committed) manifest JSON will remove it from the auspice sidebar dropdowns and stop the ability for default URL redirects such as 'monkeypox' → 'monkeypox/hmpxv1'
* Removing `/monkeypox` from `app.js` will stop nextstrain.org/monkeypox URLs working at all (they'll 404)
* Redirects to go from /monkeypox URLs to /mpox ones should be added in `./src/redirects.js`. All the dataset names have changed (i.e. it's not just `/monkeypox/mpox/`) so we should add explicit redirects for all the old URLs.

### DIFF
--- a/data/manifest_core.json
+++ b/data/manifest_core.json
@@ -201,6 +201,14 @@
           "default": "hmpxv1"
         }
       },
+      "mpox": {
+        "resolution": {
+          "all-clades": "",
+          "clade-IIb": "",
+          "lineage-B.1": "",
+          "default": "clade-IIb"
+        }
+      },
       "mumps": {
         "resolution": {
           "na": "",
--- a/data/manifest_core.json
+++ b/data/manifest_core.json
@@ -194,13 +194,6 @@
         }
       },
       "measles": "",
-      "monkeypox": {
-        "resolution": {
-          "hmpxv1": "",
-          "mpxv": "",
-          "default": "hmpxv1"
-        }
-      },
       "mpox": {
         "resolution": {
           "all-clades": "",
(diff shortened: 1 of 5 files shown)


corneliusroemer (approved):

corneliusroemer:
Looks good, you chose to default to `all-clade` which is fine I guess, though most mpox these days is `clade-IIb` so we might also want to choose that. But this decision not blocking and easily changed later.

jameshadfield:
> most mpox these days is clade-IIb so we might also want to choose that

Force-pushed to change the default to clade-IIb

corneliusroemer:
Test work! Let's merge :) 

But shall we also rename the tile from `monkeypox` to `mpox`? 

image

https://nextstrain-s-manifest-m-n73mf6.herokuapp.com/mpox/all-clades

jameshadfield:
Redirects added, as per @corneliusroemer's suggestion.

/monkeypox/mpxv → /mpox/all-clades
/monkeypox/hmpxv1 → /mpox/clade-IIb
/monkeypox/hmpxv1/big → /mpox/lineage-B.1

While the monkeypox option no longer exists in the auspice sidebar, valid URLs will still work, e.g. /monkeypox/hmpxv1/2022-09-04

PR #716, opened 2023-09-28
Remove the nonsense that is utils.getGitHash()

tsibley:
This function is used in one place, to provide a "version" that's not necessary.  Since the implementation is nonsense to begin with, just remove it.

## Checklist

- [x] Checks pass

### DIFF
--- a/server.js
+++ b/server.js
@@ -2,7 +2,6 @@ import argparse from 'argparse';
 import http from 'http';
 import * as utils from './src/utils/index.js';
 
-const version = utils.getGitHash();
 const nextstrainAbout = `
   Nextstrain is an open-source project to harness the scientific and public
   health potential of pathogen genome data.
@@ -11,9 +10,8 @@ const nextstrainAbout = `
   See https://github.com/nextstrain/nextstrain.org for more.
 `;
 const parser = new argparse.ArgumentParser({
-  version,
   addHelp: true,
-  description: `Nextstrain.org server version ${version}`,
+  description: `Nextstrain.org server`,
   epilog: nextstrainAbout
 });
 parser.addArgument('--verbose', {action: "storeTrue", help: "verbose server logging"});
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -1,21 +1,7 @@
-import fs from 'fs';
 import chalk from 'chalk';
 import fetch from 'node-fetch';
 import { NotFound } from '../httpErrors.js';
 
-
-const getGitHash = () => {
-  /* https://stackoverflow.com/questions/34518389/get-hash-of-most-recent-git-commit-in-node */
-  try {
-    const rev = fs.readFileSync('.git/HEAD').toString();
-    if (rev.indexOf(':') === -1) {
-      return rev;
-    }
-    return fs.readFileSync('.git/').toString() + rev.substring(5).replace(/\n/, '');
-  } catch (err) {
-    return "unknown";
-  }
-};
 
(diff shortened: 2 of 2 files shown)


victorlin (approved):

jameshadfield (changes requested):
What do you mean by "nonsense"? The current implementation is broken, but I wouldn't refer to the concept as nonsensical at all. Reporting the current SHA is useful in many cases, all the more important since we don't version the server code.

I'd rather we fix the code and surface it more widely by displaying it in the server startup message. A fix is below, but we may choose to use a library instead.

```diff
diff --git a/server.js b/server.js
index 1a2c5e89..dfcb10d5 100644
--- a/server.js
+++ b/server.js
@@ -59,6 +59,7 @@ const server = http
     console.log("  -------------------------------------------------------------------------");
     console.log(nextstrainAbout);
     console.log(`  Server listening on port ${server.address().port}`);
+    console.log(`  Server version (git SHA): ${version}`);
     console.log(`  Accessible at https://nextstrain.org or http://localhost:${server.address().port}`);
     console.log(`  Server is running in ${app.locals.production ? 'production' : 'development'} mode`);
     console.log("\n  -------------------------------------------------------------------------\n\n");
diff --git a/src/utils/index.js b/src/utils/index.js
index 926062e8..8bc0bb71 100644
--- a/src/utils/index.js
+++ b/src/utils/index.js
@@ -7,11 +7,13 @@ import { NotFound } from '../httpErrors.js';
 const getGitHash = () => {
   /* https://stackoverflow.com/questions/34518389/get-hash-of-most-recent-git-commit-in-node */
   try {
-    const rev = fs.readFileSync('.git/HEAD').toString();
+    let rev = fs.readFileSync('.git/HEAD').toString();
     if (rev.indexOf(':') === -1) {
       return rev;
+    } else {
+      rev = rev.replace(/.+:\s+/, '').replace("\n", '');
     }
-    return fs.readFileSync('.git/').toString() + rev.substring(5).replace(/\n/, '');
+    return fs.readFileSync(`.git/${rev}`).toString().slice(0, 8)
   } catch (err) {
     return "unknown";
   }

```

tsibley:
Agreed the concept is not nonsense; that's why I said, "Since the _implementation_ is nonsense" in the commit message.

While reporting the current SHA can be useful in many cases, I don't see how it's actually useful here. It's been broken since it was added in November 2018 and in 5 years no one's missed it. In local development, you can see what revision you're using easily. On Heroku dynos, the app slugs don't contain `.git/` so it'd always be "unknown" anyway. So that's why I removed it instead of fixing it.

Fixing the code is doable, of course, but the correct approach is not to read `.git/HEAD` and resolve its contents ourselves. I realize the quick patch you suggest is for demonstration purposes, but it would break on a bunch of cases because of incorrect assumptions it makes. One example is in a "detached HEAD" state (e.g. you've checked out a specific commit, perhaps for testing), but there are others. Using `git rev-parse HEAD` or similar (e.g. `git describe --always --dirty`) would be a better fix. But again, it doesn't seem worth bothering.

tsibley:
That said, if you'd rather fix it one way or another, please go ahead and we can close this. No problem with that.

tsibley:
I incorporated more of my rationale from https://github.com/nextstrain/nextstrain.org/pull/716#issuecomment-1740068559 into the commit message and merged.

If we want to add back git revision reporting in the future, we can, that's fine, and there's pointers to the right way to do so in the commit message now.

Issue #726, opened 2023-10-10
GitHub Actions deploy job succeeds even when Heroku build fails


tsibley:
Consider this GitHub Actions [deploy job from the morning](https://github.com/nextstrain/nextstrain.org/actions/runs/6472687000/job/17574035909). It kicked off [this build on Heroku](https://dashboard.heroku.com/apps/nextstrain-canary/activity/builds/964110bf-d487-4f9d-96b6-f2100f79c8be). That build failed, as you can see in Heroku logs and also the [GitHub job logs](https://github.com/nextstrain/nextstrain.org/actions/runs/6472687000/job/17574035909#step:5:201), but the GitHub job's conclusion was success.

It should check the build status after the build completes (successfully or not) and assert success.

tsibley:
I'm re-running the deploy job in that workflow to see if this was transient or not...

tsibley:
[That worked](https://github.com/nextstrain/nextstrain.org/actions/runs/6472687000/job/17583712545#step:5:285), so the errors themselves were transient (or at least don't occur consistently).

That said, that's still separate from this issue: making sure our CI deploy status reflects Heroku accurately.

PR #727, opened 2023-10-11
dev: Check the Heroku build status during our GitHub Actions deploy job

tsibley:
We'd been assuming builds on Heroku were always successful, which meant the GitHub Actions status could (and did) mask build issues on Heroku.

Resolves <https://github.com/nextstrain/nextstrain.org/issues/726>.

## Checklist

- [x] Tested the snippets manually; testing the workflow is not possible without merge first
- [x] Checks pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,6 +99,22 @@ jobs:
           curl "$(jq -r .output_stream_url build.json)" \
             --fail --silent --show-error --location
 
+      - name: Check Heroku build
+        run: |
+          id="$(jq -r .id build.json)"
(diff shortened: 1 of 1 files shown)


joverlee521 (approved):
Changes look by inspection!

victorlin (approved):

tsibley:
Worked in the success condition: https://github.com/nextstrain/nextstrain.org/actions/runs/6486469947/job/17614896676#step:6:1

We'll see I guess in the future if it works in the failure condition. I don't think we need to intentionally fail a deploy to test it (but we could).

PR #731, opened 2023-10-13
authn: Generalize support to ~any OIDC/OAuth2 IdP, not just AWS Cognito

tsibley:
OIDC is OpenID Connect 1.0, which is an identity/authentication protocol layered on top of OAuth 2.0's authorization protocol.  AWS Cognito implements OIDC/OAuth2 but our authn code hardcoded some assumptions about Cognito specifically.  Undo that and parameterize and generalize the code to work (in theory) with other OIDC identity providers (IdPs). In practice, some additional changes may be necessary for specific other IdPs, but as-is I can get this generalized authn code to work against a test Azure AD IdP.

Outside of authn, there are still some other bits of the codebase which require Cognito.  Those will be addressed in subsequent work.

This work is motivated by CDC AMD's efforts to host a copy of nextstrain.org internally in order to avail themselves of Groups internally.

Related-to: <https://github.com/nextstrain/private/issues/94>

## Checklist

- [x] Checks pass

### DIFF
--- a/env/testing/config.json
+++ b/env/testing/config.json
@@ -1,6 +1,9 @@
 {
-  "COGNITO_BASE_URL": "https://nextstrain-testing.auth.us-east-1.amazoncognito.com",
-  "COGNITO_CLIENT_ID": "6qiojrhr8tibt0f6hphnm1osp1",
-  "COGNITO_CLI_CLIENT_ID": "9opa27o74f4jsq8g4a34e1mqr",
-  "COGNITO_USER_POOL_ID": "us-east-1_zqpCrjM7I"
+  "OIDC_IDP_URL": "https://cognito-idp.us-east-1.amazonaws.com/us-east-1_zqpCrjM7I",
+  "OAUTH2_CLIENT_ID": "6qiojrhr8tibt0f6hphnm1osp1",
+  "OAUTH2_CLI_CLIENT_ID": "9opa27o74f4jsq8g4a34e1mqr",
+  "OAUTH2_LOGOUT_URL": "https://nextstrain-testing.auth.us-east-1.amazoncognito.com/logout",
+  "COGNITO_USER_POOL_ID": "us-east-1_zqpCrjM7I",
+  "OIDC_USERNAME_CLAIM": "cognito:username",
+  "OIDC_GROUPS_CLAIM": "cognito:groups"
 }
--- a/src/authn/index.js
+++ b/src/authn/index.js
@@ -19,7 +19,7 @@ import { JOSEError, JWTClaimValidationFailed, JWTExpired } from 'jose/util/error
 import partition from 'lodash.partition';
 import BearerStrategy from './bearer.js';
 import { getTokens, setTokens, deleteTokens } from './session.js';
-import { PRODUCTION, OIDC_ISSUER_URL, OIDC_JWKS_URL, OAUTH2_AUTHORIZATION_URL, OAUTH2_TOKEN_URL, OAUTH2_LOGOUT_URL, OAUTH2_SCOPES_SUPPORTED, OIDC_USERNAME_CLAIM, OIDC_GROUPS_CLAIM, OAUTH2_CLIENT_ID, OAUTH2_CLIENT_SECRET, OAUTH2_CLI_CLIENT_ID } from '../config.js';
+import { PRODUCTION, OIDC_ISSUER_URL, OIDC_JWKS_URL, OAUTH2_AUTHORIZATION_URL, OAUTH2_TOKEN_URL, OAUTH2_LOGOUT_URL, OAUTH2_SCOPES_SUPPORTED, OIDC_USERNAME_CLAIM, OIDC_GROUPS_CLAIM, OIDC_IAT_BACKDATED_BY, OAUTH2_CLIENT_ID, OAUTH2_CLIENT_SECRET, OAUTH2_CLI_CLIENT_ID } from '../config.js';
 import { AuthnRefreshTokenInvalid, AuthnTokenTooOld } from '../exceptions.js';
 import { fetch } from '../fetch.js';
 import { copyCookie } from '../middleware.js';
@@ -716,8 +716,8 @@ async function verifyIdToken(idToken, client = OAUTH2_CLIENT_ID) {
     if (typeof claims.iat !== "number") {
       throw new JWTClaimValidationFailed(`"iat" claim must be a number`, "iat", "invalid");
     }
-    if (claims.iat < staleBefore) {
-      throw new AuthnTokenTooOld(`"iat" claim less than user's staleBefore: ${claims.iat} < ${staleBefore}`);
+    if (claims.iat + OIDC_IAT_BACKDATED_BY < staleBefore) {
+      throw new AuthnTokenTooOld(`"iat" claim (plus any backdating) less than user's staleBefore: ${claims.iat} + ${OIDC_IAT_BACKDATED_BY} < ${staleBefore}`);
     }
   }
 
--- a/src/authn/index.js
+++ b/src/authn/index.js
@@ -19,7 +19,7 @@ import { JOSEError, JWTClaimValidationFailed, JWTExpired } from 'jose/util/error
 import partition from 'lodash.partition';
 import BearerStrategy from './bearer.js';
 import { getTokens, setTokens, deleteTokens } from './session.js';
-import { PRODUCTION, COGNITO_USER_POOL_ID, COGNITO_BASE_URL, COGNITO_CLIENT_ID, COGNITO_CLI_CLIENT_ID } from '../config.js';
+import { PRODUCTION, OIDC_ISSUER_URL, OIDC_JWKS_URL, OAUTH2_AUTHORIZATION_URL, OAUTH2_TOKEN_URL, OAUTH2_LOGOUT_URL, OAUTH2_SCOPES_SUPPORTED, OIDC_USERNAME_CLAIM, OIDC_GROUPS_CLAIM, OAUTH2_CLIENT_ID, OAUTH2_CLIENT_SECRET, OAUTH2_CLI_CLIENT_ID } from '../config.js';
 import { AuthnRefreshTokenInvalid, AuthnTokenTooOld } from '../exceptions.js';
 import { fetch } from '../fetch.js';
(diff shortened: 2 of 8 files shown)


tsibley:
@victorlin Tagged you for review specifically since you volunteered to be involved in this work.

tsibley:
@victorlin Some notes on testing.

I've tested this against our testing env (e.g. using AWS Cognito) and also against [an Azure AD](https://portal.azure.com/?feature.msaljs=true#view/Microsoft_AAD_IAM/ActiveDirectoryMenuBlade/~/Overview) I set up for the purpose (somewhat matching what CDC is using).

For Azure AD, I used a JSON file like so:

```json
{ 
  "OIDC_IDP_URL": "https://login.microsoftonline.com/0ce9e8dc-e009-4cb4-8512-7989bd6906a8/v2.0", 
  "OAUTH2_CLIENT_ID": "c3d6647f-dccc-4a85-a2bb-fb8fbc7524a9", 
  "OAUTH2_CLIENT_SECRET": "SECRET FROM AZURE AD APP REGISTRATION", 
  "OAUTH2_CLI_CLIENT_ID": "REQUIRED; NOT YET SORTED OUT; WILL BE PART OF FUTURE NEXTSTRAIN CLI CONFIGURATION TOO", 
  "OIDC_USERNAME_CLAIM": "preferred_username", 
  "OIDC_GROUPS_CLAIM": "groups", 
  "COGNITO_USER_POOL_ID": "REQUIRED; NOT YET SORTED OUT", 
} 
```

and then set `CONFIG_FILE=path/to/that/file.json` when running `node server.js`. The `OAUTH2_CLIENT_SECRET` needs to be replaced by the actual client secret.

@victorlin You can login to the Azure portal as azure-admin@nextstrain.org (creds in 1Password). From there, you can [generate another client secret](https://portal.azure.com/?feature.msaljs=true#view/Microsoft_AAD_RegisteredApps/ApplicationMenuBlade/~/Credentials/appId/c3d6647f-dccc-4a85-a2bb-fb8fbc7524a9/isMSAApp~/false) to use (or I can share the one I generated) under "App registration", find the app/client, click the secrets link, add a new one. Once you have the nextstrain.org server running locally, you can try logging in with test-user@nextstraintesting.onmicrosoft.com (password in 1Password).

I can also walk you thru all of this.

I should verify our testing env still works with this PR, as there've been some changes since my last check.

src/config.js
```diff
@@ -70,6 +74,23 @@ const fromEnvOrConfig = (name, default_) => {
 };
 
 
+/**
+ * Deserialize a value that might be JSON, passing it thru if it isn't.
+ *
+ * @param {any} x - Value which might be JSON
+ */
+function maybeJSON(x) {
```
victorlin, line 82: _(minor)_

Is this for `OIDC_IAT_BACKDATED_BY`? If so I think it should be in 9644d8c9d946b554bdaa11370e491a4eb71030fb.

src/authn/index.js
victorlin, line 1: Since `OAUTH2_CLI_CLIENT_ID` is not sorted out yet, I assume the only things that can be tested with AAD are: `/login`, `/logout`, and token renewal on the server side, and once logged in, `/whoami`, `/groups/test-private/*` on the client side. Is that correct?

src/authn/index.js
```diff
@@ -91,10 +102,12 @@ function setup(app) {
...
       {
-        authorizationURL: `${COGNITO_BASE_URL}/oauth2/authorize`,
-        tokenURL: `${COGNITO_BASE_URL}/oauth2/token`,
-        clientID: COGNITO_CLIENT_ID,
+        authorizationURL: OAUTH2_AUTHORIZATION_URL,
+        tokenURL: OAUTH2_TOKEN_URL,
+        clientID: OAUTH2_CLIENT_ID,
         callbackURL: "/logged-in",
+        scope: OAUTH2_SCOPES,
+        clientSecret: OAUTH2_CLIENT_SECRET,
```
victorlin, line 110: My understanding is that `OAUTH2_CLIENT_SECRET` is necessary for AAD and not Cognito because the `NextstrainDotOrgServerInstance-testing` IAM user has access to the user pool. And this Passport call is able to handle both cases. Is that correct?

src/config.js
victorlin, line 1: *(commenting on an arbitrary file to start a thread)*

Coming from a recent background in Azure AD (now Entra ID but I'll still call it AAD for now), this is all pretty interesting to me. I'm curious, what docs did you reference for this?

env/testing/config.json
```diff
@@ -1,6 +1,9 @@
 {
-  "COGNITO_BASE_URL": "https://nextstrain-testing.auth.us-east-1.amazoncognito.com",
-  "COGNITO_CLIENT_ID": "6qiojrhr8tibt0f6hphnm1osp1",
-  "COGNITO_CLI_CLIENT_ID": "9opa27o74f4jsq8g4a34e1mqr",
-  "COGNITO_USER_POOL_ID": "us-east-1_zqpCrjM7I"
+  "OIDC_IDP_URL": "https://cognito-idp.us-east-1.amazonaws.com/us-east-1_zqpCrjM7I",
```
victorlin, line 2: It's interesting how in AWS, there is a clear distinction between IAM users/groups and Cognito users/groups, while in Azure with the current setup, it's all under AAD. I wonder if this has implications for organizations that manages Azure resources with AAD users.

Did you look into something like Azure AD B2C? I haven't used it personally, but it [looks to be](https://learn.microsoft.com/en-us/azure/active-directory-b2c/supported-azure-ad-features) quite different from AAD, made clearer by name with the recent rebranding.

I'm asking these questions, but I don't think we should ponder too much because this seems to be working well for CDC's use case.

victorlin (commented):
- [x] Tested using AWS Cognito with the testing user pool. I used the same commands as in https://github.com/nextstrain/nextstrain.org/pull/581#pullrequestreview-1656344763 with `s/COGNITO_CLI_CLIENT_ID/OAUTH2_CLI_CLIENT_ID/`.
- [x] Tested using AAD with the existing app registration (via a new client secret) and a user that I created for myself and added to `test-private/owners`. I tested the `/login` and `/logout` routes but not token renewal.

notes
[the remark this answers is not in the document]
tsibley: Whoops, yes indeed.

src/config.js
tsibley: Ah, no, it's for `OIDC_CONFIGURATION` which is introduced in the same commit. But yes, it's also used subsequently by `OIDC_CONFIGURATION`.

src/authn/index.js
tsibley: Correct. The issue more precisely is not `OAUTH2_CLI_CLIENT_ID` itself—it's straightforward to create the OIDC client in Azure AD and get an id—but that Nextstrain CLI doesn't support any IdP other than ours right now.

src/authn/index.js
tsibley: The Passport call is able to handle both the presence and absence of the client secret. A client secret is an optional part of client configuration in the spec. Azure AD requires one for the type of auth flow we're using and Cognito does not, though it supports them. For nextstrain.org, we could change to using a client secret with Cognito (and probably should). For Nextstrain CLI, as a public client, we can't secure a client secret and so wouldn't use one (true with Azure AD too).

src/config.js
tsibley: I had some prior experience with OIDC and OAuth2, so that helped. But I still extensively referenced both the OIDC and OAuth2 specs as well as to a lesser extent Microsoft's Azure AD documentation, and then also did some experiments/exploration against an actual Azure AD instance.

Links to Microsoft's documentation are mostly in #nextstrain-cdc (haven't dug them out), where as the specs are linked to in comments in this PR and include:

- https://openid.net/specs/openid-connect-core-1_0.html
- https://openid.net/specs/openid-connect-discovery-1_0.html
- https://openid.net/specs/openid-connect-rpinitiated-1_0.html
- https://datatracker.ietf.org/doc/html/rfc6749

src/config.js
victorlin: Oh! I missed that as being optional since I was focused on changes to the config files and only saw `OIDC_IAT_BACKDATED_BY` as a non-string value. Makes sense now.

src/authn/index.js
victorlin: Ok, I sort of get it now. [This](https://stackoverflow.com/a/47156625) StackExchange answer was also helpful. I think I'll need to do more reading on OIDC/OAuth2 to fully grasp these concepts.

tsibley: Ok. Also happy to walk thru it in person!

env/testing/config.json
tsibley: I didn't look into Azure AD B2C, but looking now, yes, it's much more akin to Cognito. e.g. conceptually we could swap out our usage of Cognito for Azure AD B2C, but we _wouldn't_ swap Cognito for Azure AD/Entra ID.

In CDC's case, they want to use their existing user directory in Azure AD and since they're also hosting the application themselves, it's simplest to authenticate against Azure AD directly. While they could set up an Azure AD B2C instance just for this app and then invite specific users from their Azure AD into their Azure AD B2C, it would just be more layers of indirection for not any functional gain. The gain would be more administrative/organizational, e.g. if for some reason they didn't want to directly authn against Azure AD or if they wanted a harder adminstrative split.

tsibley: Maybe to sum that up: There's a lot of ways to do user directories for authn purposes, but no one right way; it's highly dependent on the needs/context/where the users already are.

src/authn/index.js
victorlin, line 1: _(non-blocking)_

This file still has many references to Cognito in comments and variable names even though it is generalized for more than that. I'd think to replace most references to "Cognito" with "authentication service", and "Cognito group" with "authentication group" or something along those lines.

victorlin (approved):

src/authn/index.js
tsibley: Yeah... there's also other bits elsewhere (code, comments, documentation) which assume Cognito. For example, the documentation in `docs/infrastructure.md` is about our specific setup, not the general infra required. My thinking was not to excise Cognito entirely, but to just make it possible to swap it out. I think it's reasonable for the codebase to generally assume Cognito for now. We can see how this goes over time and do the work to more fully excise it later.

PR #736, opened 2023-10-23
authn: Update the refresh token when renewing tokens, if applicable

tsibley:
Reading the OAuth 2.0 spec again¹, I noted that:

> The authorization server MAY issue a new refresh token, in which
> case the client MUST discard the old refresh token and replace it
> with the new refresh token.  The authorization server MAY revoke the
> old refresh token after issuing a new refresh token to the client.

I had assumed the refresh token itself was never renewed as I never observed AWS Cognito doing so in practice, but the spec is clear here and we want to support a wider variety of IdPs.  Renewable refresh tokens would actually be welcome in Cognito as then we could limit their lifetime to be much closer to the max session age we set.  Or perhaps Cognito does renew them, but only as necessary and we've never happened to notice this issue!

¹ <https://datatracker.ietf.org/doc/html/rfc6749#section-6>

## Checklist

- [x] Tested no updated refresh token renewal path with Cognito
- [x] Tested updated refresh token renewal path with Cognito + MITMed response including a new (but bogus) refresh token.
- [x] Checks pass

### DIFF
--- a/src/authn/index.js
+++ b/src/authn/index.js
@@ -215,7 +215,7 @@ function setup(app) {
            */
           if (err instanceof JWTExpired || err instanceof AuthnTokenTooOld) {
             debug(`Renewing tokens for ${user.username} (session ${req.session.id.substr(0, 7)}…)`);
-            ({idToken, accessToken} = await renewTokens(refreshToken));
+            ({idToken, accessToken, refreshToken} = await renewTokens(refreshToken));
 
             const updatedUser = await userFromTokens({idToken, accessToken, refreshToken});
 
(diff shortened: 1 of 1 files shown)


src/authn/index.js
```diff
@@ -788,11 +795,12 @@ async function renewTokens(refreshToken) {
       throw new Error(`failed to renew tokens: ${response.status} ${response.statusText}: ${body}`);
   }
 
-  const {id_token: idToken, access_token: accessToken} = JSON.parse(body);
+  const {id_token: idToken, access_token: accessToken, refresh_token: refreshToken_} = JSON.parse(body);
 
   return {
     idToken,
     accessToken,
+    refreshToken: refreshToken_ ?? refreshToken,
```
victorlin, line 803: _(non-blocking)_

Suggestion: Use a more meaningful variable name, e.g.

```suggestion
    refreshToken: refreshTokenFromIdP ?? refreshToken,
```

victorlin (approved):

src/authn/index.js
tsibley: Ah, ha, I briefly dithered on names for that before choosing the trailing `_` form (as `x_` is fairly conventional for "`x`, but modified"). Seemed ok to me as declaration and usage were so close together, but yeah, relies on connotations rather than being descriptive on its own.

I've changed it to `refreshTokenUpdate`.

PR #739, opened 2023-10-26
Relax Redis requirement

tsibley:
See commit messages.

Related-to: <https://github.com/nextstrain/private/issues/94>

## Checklist

- [x] Checks pass


src/redis.js
victorlin, line 1: Suggestion: Update this comment to not infer a development environment.

https://github.com/nextstrain/nextstrain.org/blob/2f861771a1bfa60b2af5d94193705da136810dd4/src/authn/index.js#L295-L300

If I'm understanding correctly, the usage of a local file store is fine in production as long as there is no horizontal scaling.

victorlin, line 1: My testing plan:

- [ ] Run [group membership API testing commands](https://github.com/nextstrain/nextstrain.org/pull/581#pullrequestreview-1656344763) in dev mode with env/testing/config.json
- [ ] Run [group membership API testing commands](https://github.com/nextstrain/nextstrain.org/pull/581#pullrequestreview-1656344763) in dev mode with [Azure AD config](https://github.com/nextstrain/nextstrain.org/pull/731#issuecomment-1760825272)

tsibley: Ah, good point! Your understanding is correct. I'll update the comment as you suggest.

During dev yesterday, I did look into swapping out the `FileStore` / `RedisStore` combo for a single store implementation hooked up to the new kv interface. That is, effectively changing out the `FileStore` for storing in SQLite and making the setup of session storage easier and more consistent with how we use Redis elsewhere. The [session store API requirements](https://github.com/expressjs/session#session-store-implementation) seemed minimal so I was hopeful something existed. Found [other desire for the same thing](https://github.com/jaredwray/keyv/issues/637), but no implementations… and decided I could save our own implementation for another time (if ever).

victorlin: I don't think I'll get to this today, so feel free to run it yourself first.

tsibley: [Updated](https://github.com/nextstrain/nextstrain.org/compare/2f861771a1bfa60b2af5d94193705da136810dd4..64f7d92b68b1648579c097db1eb501adcada5126).

victorlin: Tested and works!

victorlin (approved):

src/config.js
```diff
@@ -307,6 +307,30 @@ export const SESSION_SECRET = fromEnvOrConfig("SESSION_SECRET", PRODUCTION ? und
...
+ * how we configure review apps.  Currently they're configured via the Heroku
+ * Dashboard¹.  I think we need to switch to the app.json file² instead, as
+ * described in the review app documentation.³
+ *   -trs, 12 Oct 2023
+ *
+ * ¹ <https://dashboard.heroku.com/pipelines/38f67fc7-d93c-40c6-a182-501da2f89d9d/settings>
+ * ² <https://devcenter.heroku.com/articles/app-json-schema>
+ * ³ <https://devcenter.heroku.com/articles/github-integration-review-apps>
+ */
+export const REDIS_REQUIRED = fromEnvOrConfig("REDIS_REQUIRED", PRODUCTION && !REVIEW_APP);
```
tsibley, line 331: Welp, this doesn't work at all when setting `REDIS_REQUIRED=false` in the environment or `"REDIS_REQUIRED": false` in the config because of the use of `||` in `fromEnvOrConfig()`. orz

I tested the functionality and it worked! but it only _appeared_ to work because I was testing against a previous version of the code which provided a default of `null` instead of `PRODUCTION && !REVIEW_APP`. In my testing, the production config set `REDIS_REQUIRED=true`, which worked, and the testing config set it to `false`, which worked, but the actual value was resolved to `null` (the default). This had the same effect as `false` and so masked the problematic use of `||`. When I changed the default, the behaviour changed, and I didn't re-test. Should have written proper tests, but I didn't because then we need an actual `REDIS_URL` for the connection. We could provide one in CI I suppose…

tsibley: https://github.com/nextstrain/nextstrain.org/pull/746

PR #745, opened 2023-11-01
Check redirects

victorlin:

Went through `src/redirects.js` and addressed things that I found were out of date.

## Related issue(s)

N/A

## Checklist

- [x] Checks pass
- [x] `/narratives` redirect works properly on review app ([link](https://nextstrain-s-victorlin--rhf9ry.herokuapp.com/narratives))

### DIFF
--- a/src/redirects.js
+++ b/src/redirects.js
@@ -74,10 +74,6 @@ const setup = (app) => {
       return next('route');
     });
 
-  /* We redirect /ncov/zh -> /ncov/global/zh here rather than the manifest file as the URL structure currently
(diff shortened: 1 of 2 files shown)


tsibley (approved):
LGTM. One odd thing I noticed though (presumably because of Gatsby or some other JS scroll routing) is that `/narratives` redirects to `/#narratives` but on that page the URL is then updated in-situ to remove the `#narratives` anchor. This leaves the user at a place they don't have a link back to.

victorlin:
@tsibley good observation. Seems like it only happens on valid IDs (`/#not-an-anchor` will stay there), unless a valid anchor is visited first. Odd, but I don't think we use these enough on nextstrain.org to justify digging deeper.

PR #746, opened 2023-11-02
config: Support falsy numeric and boolean values…

tsibley:
…instead of considering them missing.

The values considered missing are now more clear:

  1. undefined (i.e. not present)
  2. null (i.e. present but explicitly null)
  3. "" (i.e. present but empty, particularly for env vars)

This allows falsy values¹ like the number 0 and boolean false to be
configurable values², as expected to be for OIDC_IAT_BACKDATED_BY and
REDIS_REQUIRED.

For example, setting REDIS_REQUIRED=false is now functional as intended
by "Allow and document how to run safely in production mode without
Redis" (64f7d92b).  And setting OIDC_IAT_BACKDATED_BY=0 in the
environment now overrides any value from the config file instead of
being ignored.

Related-to: <https://github.com/nextstrain/nextstrain.org/pull/739>

¹ <https://developer.mozilla.org/en-US/docs/Glossary/Falsy>

² Other falsy values like NaN and 0n are also now considered
  non-missing, but there's no way to express those in our current config
  sources, which permit only strings or JSON values.  We may find
  ourselves wanting to treat NaN as missing too if config sources
  change.

## Checklist

- [x] Checks pass

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -82,7 +82,7 @@ const configFile = CONFIG_FILE
  * @param {function} [options.fromConfig] - conversion function to apply to values obtained from the config file; defaults to the identity function
  * @throws {Error} if no value is found and default is undefined
  */
-const fromEnvOrConfig = (name, default_, {fromEnv = maybeJSON, fromConfig = x => x} = {}) => {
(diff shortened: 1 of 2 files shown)


jameshadfield (approved):
Works for setting an env variable to false, which will be immediately useful for my work in nextstrain.org

PR #747, opened 2023-11-03
GitHub action to automate resource index rebuilding

jameshadfield:
GitHub action to rebuild the resource collection index. ~Action has not been tested yet.~ [Action ran successfully](https://github.com/nextstrain/nextstrain.org/actions/runs/6777395697) and the uploaded index was tested locally.


.github/workflows/resource-indexer.yml
```diff
@@ -0,0 +1,40 @@
+name: rebuild the resource collection index
```
jameshadfield, line 1: Could someone remind me the best way to test this? (Once it is at the stage where I believe it'll work, of course.) Do I make a commit in master with an empty `.github/workflows/resource-indexer.yml`?

jameshadfield: cc @joverlee521 @victorlin @tsibley

joverlee521: [This is still the workaround](https://github.com/nextstrain/docs.nextstrain.org/commit/aa6f75e103b83676345e3a0f7ac60291aa99e4da) I use, not sure if others have discovered new methods.

victorlin: Another workaround is to push a commit that adds the `pull_request` trigger. I just tested it in https://github.com/victorlin/github-test/pull/7.

Note that once it's added, it'll be run on every subsequent push to the PR, and should be dropped from the PR before merging if it's only for testing purposes. If you only want it to run once, you can push a commit to add it then push another one to revert it. Then before merging, just drop those two commits in a rebase.

jameshadfield: Ha! Wonderfully creative solutions. Thanks both :)

.github/workflows/resource-indexer.yml
```diff
@@ -0,0 +1,39 @@
...
+
+on:
+  # Run at ~4am UTC time which is (± an hour) 4am UK, 5am Switzerland, midnight
+  # US east coast, 9pm US west coast so that for most users (and most
+  # developers) the index regenerates overnight
+  schedule:
+    - cron: '* 4 * * *'
```
tsibley, line 8: Noticed this as I was quickly looking: this crontab spec is "every minute of the 04 hour". You want `0 4 * * *` (or similar, replacing 0 with 0–59).

.github/workflows/resource-indexer.yml
```diff
@@ -0,0 +1,39 @@
...
+          aws-region: us-east-1
+          role-to-assume: arn:aws:iam::827581582529:role/GitHubActionsRoleResourceIndexer
+          credential-lifetime: 3600 # seconds, or 1 hour, which is currently the default and should be plenty
+      - name: Rebuild the index
+        run: |
+          node resourceIndexer/main.js \
+            --gzip --output resources.json.gz
+      - name: Make a backup of the existing (current) index so that we can restore if needed
+        run: |
+          aws s3 cp s3://nextstrain-inventories/resources.json.gz s3://nextstrain-inventories/resources.json.gz.previous
```
tsibley, line 36: It'd be very easy for two bad runs to not leave us with a backup to restore from. I'd think to do this more robustly by enabling object versioning on the bucket and setting a bucket lifecycle applied to this key prefix which permanently deletes past versions after a fixed time.

docs/resource-collection.rst
```diff
@@ -67,8 +71,13 @@ Index creation (Inventory access and index upload)
...
 
-*This section will be updated once the
-index creation is automated.*
+The GitHub action assumes necessary AWS permissions via the IAM role
+`GitHubActionsRoleResourceIndexer
+<https://us-east-1.console.aws.amazon.com/iamv2/home?region=us-east-1#/roles/details/GitHubActionsRoleResourceIndexer>`__
+which is obtained using OIDC. This role uses permissions from the IAM policy
+`NextstrainResourceIndexer
+<https://us-east-1.console.aws.amazon.com/iamv2/home?region=us-east-1#/policies/details/arn%3Aaws%3Aiam%3A%3A827581582529%3Apolicy%2FNextstrainResourceIndexer>`__
+to list & read the S3 inventories, as well as upload the new index.
```
tsibley, line 80: Maybe these are in Terraform on the `james/resources` base branch and I'm not up-to-date there yet, but these sorts of new resources are the kind of thing I think would be great to have in Terraform from their introduction.

.github/workflows/resource-indexer.yml
```diff
@@ -0,0 +1,39 @@
+name: rebuild the resource collection index
```
tsibley, line 1: Minor, but I'd name this something slightly less unwieldy in the GitHub Actions web UI.  Maybe just "resource indexer" or "index resources"?

docs/resource-collection.rst
jameshadfield: Added as to-dos in #748

.github/workflows/resource-indexer.yml
jameshadfield: Thanks - fixed up 👍

.github/workflows/resource-indexer.yml
jameshadfield: Thanks - versioning enabled on the bucket, lifecycles adjusted accordingly, backup step removed, and docs updated

.github/workflows/resource-indexer.yml
jameshadfield: Sure - done

tsibley (approved):

.github/workflows/resource-indexer.yml
tsibley, line 1: Minor thing, but nice to make the filename match the workflow name (and this is how all the others are), so I suggest this become `index-resources.yml`.

jameshadfield:
The commits in this PR have been combined and shifted to #719. The changes since this PR are:
* Action filename renamed as suggested
* The added documentation slightly changed after editing
* The indexer is now run with additional arguments `--resourceTypes dataset --collections core` as that's the only data that the server will currently use for the functionality introduced in #719. These arguments will be removed as we expand the functionality of nextstrain.org to allow further information in the index to be used.

PR #749, opened 2023-11-07
[bugfix] default params for signedUrl()

jameshadfield:
The default parameter values for this function would only be used in the case where no argument was provided, which makes no sense as a bucket and key will always be needed. These changes preserve the intended defaults of headers={} and method='GET'. (Reading the previous code it seems the intention was to allow a default method of 'GET', despite the docstring indicating it was required.)

- [ ] Checks pass

### DIFF
--- a/src/s3.js
+++ b/src/s3.js
@@ -42,15 +42,15 @@ export async function* listObjects({bucket, prefix}) {
  * Generate a signed URL for an S3 object.
  *
  * @param {object} params
- * @param {string} params.method - HEAD, GET, PUT, DELETE
  * @param {string} params.bucket - name of bucket
  * @param {string} params.key - object key
+ * @param {string} [params.method] - HEAD, GET, PUT, DELETE. Default: GET
  * @param {object} [params.headers] - Content-Type and/or Content-Encoding headers for PUT
  * @param {number} [params.expiresIn] - seconds until the URL expires
  * @param {number} [params.issuedAt] - absolute time in seconds since the Unix epoch at which the signed URL should be considered issued at, i.e. when the countdown for expiresIn starts
  * @returns {string} signed URL
  */
-export async function signedUrl({method, bucket, key, headers, expiresIn, issuedAt} = {method: "GET", headers: {}}) {
+export async function signedUrl({method="GET", bucket, key, headers={}, expiresIn, issuedAt}) {
   const normalizedHeaders = normalizeHeaders(headers);
 
   const commands = new Map([


src/s3.js
```diff
@@ -42,16 +42,16 @@ export async function* listObjects({bucket, prefix}) {
...
  * @param {string} params.key - object key
+ * @param {string} [params.method] - HEAD, GET, PUT, DELETE. Default: GET
  * @param {object} [params.headers] - Content-Type and/or Content-Encoding headers for PUT
  * @param {number} [params.expiresIn] - seconds until the URL expires
  * @param {number} [params.issuedAt] - absolute time in seconds since the Unix epoch at which the signed URL should be considered issued at, i.e. when the countdown for expiresIn starts
  * @returns {string} signed URL
  */
-export async function signedUrl({method, bucket, key, headers, expiresIn, issuedAt} = {method: "GET", headers: {}}) {
-  const normalizedHeaders = normalizeHeaders(headers);
+export async function signedUrl({method, bucket, key, headers, expiresIn, issuedAt}) {
```
tsibley, line 53: Ahh, thanks for the catch! I get this backwards sometimes. I meant to write:

```diff
-export async function signedUrl({method, bucket, key, headers, expiresIn, issuedAt} = {method: "GET", headers: {}}) {
+export async function signedUrl({method = "GET", bucket, key, headers = {}, expiresIn, issuedAt} = {}) {
```

which makes the `headers || {}` and `method || 'GET'` in the function body unnecessary. This is my preferred way to support optional args in a destructuring, as it naturally documents them in the function signature.

jameshadfield:
Updated to use [the requested syntax](https://github.com/nextstrain/nextstrain.org/pull/749/files#r1385254018) and tested locally. Will merge once CI passes.

src/s3.js
```diff
@@ -42,15 +42,15 @@ export async function* listObjects({bucket, prefix}) {
...
  * @param {string} params.bucket - name of bucket
  * @param {string} params.key - object key
+ * @param {string} [params.method] - HEAD, GET, PUT, DELETE. Default: GET
  * @param {object} [params.headers] - Content-Type and/or Content-Encoding headers for PUT
  * @param {number} [params.expiresIn] - seconds until the URL expires
  * @param {number} [params.issuedAt] - absolute time in seconds since the Unix epoch at which the signed URL should be considered issued at, i.e. when the countdown for expiresIn starts
  * @returns {string} signed URL
  */
-export async function signedUrl({method, bucket, key, headers, expiresIn, issuedAt} = {method: "GET", headers: {}}) {
+export async function signedUrl({method="GET", bucket, key, headers={}, expiresIn, issuedAt}) {
```
tsibley, line 53: FWIW, this dropped the trailing `= {}` that I'd intentionally suggested in `{method="GET", …} = {}`. Without it, a call to `signedUrl()` generates an error like:

    Uncaught TypeError: Cannot read properties of undefined (reading 'method')

While calling `signedUrl()` is always an error, omitting the default `= {}` seems a weirder way to throw the error compared to `signedUrl({})`:

    Uncaught Error: No value provided for input HTTP label: Bucket.

PR #756, opened 2023-11-20
authn: Stop request processing on failed Bearer authn attempts

tsibley:
Fixes regression where after a failed Bearer authn attempt the request processing continued on (still unauthenticated).  Failed attempts now once again produce an error response (typically 400 or 401) with the appropriate challenge (if any) in the WWW-Authenticate header.  In particular, this restores responses like:

    HTTP/1.1 401 Unauthorized
    WWW-Authenticate: error="invalid_token" …

when a stale token is sent in the request.  Nextstrain CLI relies on these responses to form a more helpful error message to the user.

The regression was introduced by "authn: Switch to custom authenticator callbacks" (aca4d07f) because the Passport "strategy" we use here stops performing the default behaviour we were relying on once a callback is provided.  Instead, it expects the callback to provide all the behaviour.

Note that this regression was merely a usability issue, not a security issue, as the rest of the request processing remained unauthenticated.

Good tests of the interaction of Bearer authn with our "stale before" functionality would have caught this regression, but we continue to not have suitable testing scaffolding for tests involving authentication and so adding them (at the time or now) is not trivial.

## Checklist

- [x] Checks pass

### DIFF
--- a/src/authn/index.js
+++ b/src/authn/index.js
@@ -22,6 +22,7 @@ import { getTokens, setTokens, deleteTokens } from './session.js';
 import { PRODUCTION, OIDC_ISSUER_URL, OIDC_JWKS_URL, OAUTH2_AUTHORIZATION_URL, OAUTH2_TOKEN_URL, OAUTH2_LOGOUT_URL, OAUTH2_SCOPES_SUPPORTED, OIDC_USERNAME_CLAIM, OIDC_GROUPS_CLAIM, OIDC_IAT_BACKDATED_BY, OAUTH2_CLIENT_ID, OAUTH2_CLIENT_SECRET, OAUTH2_CLI_CLIENT_ID, SESSION_COOKIE_DOMAIN, SESSION_SECRET, SESSION_MAX_AGE } from '../config.js';
 import { AuthnRefreshTokenInvalid, AuthnTokenTooOld } from '../exceptions.js';
 import { fetch } from '../fetch.js';
+import { Unauthorized, HttpErrors } from '../httpErrors.js';
 import { copyCookie } from '../middleware.js';
 import { REDIS } from '../redis.js';
 import { userStaleBefore } from '../user.js';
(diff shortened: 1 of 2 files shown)


src/authn/index.js
```diff
@@ -437,8 +438,27 @@ function setup(app) {
...
+     */
+
+    // error: something went wrong internally; can't be remediated by client
     if (err) return next(err);
+
+    // failure: authn attempted but failed; may be remediated by client
+    if (!user && (challenge || status)) {
+      if (challenge) {
+        res.set("WWW-Authenticate", challenge);
+      }
```
victorlin, line 453: From `passport.authenticate`'s [docstring](https://github.com/jaredhanson/passport/blob/fc0fdc804fb5dbf99c510b6b43fa05dded9f4f48/lib/middleware/authenticate.js#L38C20-L38C20):

> An optional `info` argument will be passed, containing additional details provided by the strategy's verify callback - this could be information about a successful authentication or a challenge message for a failed authentication.

It looks like `challenge` is assuming a special case of `info`. Would the `if (challenge)` condition ever fail? If not, would it be better to verify that it's actually a challenge message (e.g. based on your example, `challenge.startsWith("error")`)?

tsibley: Nod. It is indeed assuming a special case of `info`, gated by the `!user` condition. It's not always truthy, however: if the request's `Authorization` header when split on a space contains fewer than or more than 2 parts.

https://github.com/nextstrain/nextstrain.org/blob/1557c827003de7bfcbee0e07f9f70687c7270b1f/src/authn/bearer.js#L101-L113

The `this.fail(400)` call there results in a callback invocation like:

    // err, user, challenge, status
    callback(null, false, undefined, 400);

Calls to `this.fail(…)` are collected and only invoke the callback if all strategies fail:

https://github.com/jaredhanson/passport/blob/cfdbd4a762b51e339ebfea931d65bccbbde53282/lib/middleware/authenticate.js#L107-L116

tsibley:
Repushed to fix the commit message since I realized I omitted part of the WWW-Authenticate header in the example.

image

victorlin (approved):

src/authn/index.js
victorlin: Thanks for the explanation!

PR #761, opened 2023-12-01
Use inline exports

victorlin:

This a follow-up to "Use cjs-to-es6 to help conversion" (1963fcd7), which left exports at the bottom of the file. That was necessary for CJS, but now ESM syntax can be leveraged for more concise code.

I searched for instances of "export {" and manually applied the conversion where possible. I skipped src/async.js since that's a vendored copy of an external script.

## Related issue(s)

Proposed in https://github.com/nextstrain/nextstrain.org/pull/735#discussion_r1401216668.

## Checklist

- [x] Checks pass
- [x] Review app: clicking around a few pages works
- [x] Locally (because groups doesn't work on review app): Login, logout, accessing groups works

### DIFF
--- a/src/endpoints/sources.js
+++ b/src/endpoints/sources.js
@@ -15,7 +15,7 @@ import { deleteByUrls, proxyFromUpstream, proxyToUpstream } from "../upstream.js
  *   instance from the request
  * @returns {expressMiddleware}
  */
-const setSource = (sourceExtractor) => (req, res, next) => {
+export const setSource = (sourceExtractor) => (req, res, next) => {
   const source = sourceExtractor(req);
 
   authz.assertAuthorized(req.user, authz.actions.Read, source);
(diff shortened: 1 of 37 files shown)


src/endpoints/sources.js
```diff
@@ -391,26 +391,3 @@ function receiveSubresource(subresourceExtractor) {
...
-  setNarrative,
-  ifNarrativeExists,
-  getNarrative,
-  putNarrative,
-  deleteNarrative,
-  optionsNarrative,
-
-  sendDatasetSubresource,
-  sendNarrativeSubresource,
-};
```
tsibley, line 416: Ahh, I think it's a bit of a misstep to switch from a clear export list like this to scattered per-symbol exports. To me, some files make more sense with per-symbol exports and some make more sense with a single export block at the bottom.

My suggestion in https://github.com/nextstrain/nextstrain.org/pull/735#discussion_r1401216668 was not intended to mean "use per-symbol exports everywhere", but I see now that you proposed that in response and I thumbs upped it without realizing.

victorlin: Thanks for following up – I misunderstood your initial comment and took it further than your intended scope of "import + export". New PR which should align better: #834

PR #762, opened 2023-12-07
Add lassa build to manifest

victorlin:

The lassa repo¹ has a link to nextstrain.org/lassa that doesn't work. This should fix it.

¹ <https://github.com/nextstrain/lassa>

## Checklist

- [x] `/lassa` works on preview build: https://nextstrain-s-victorlin--gc6piw.herokuapp.com/lassa
- [x] Checks pass
- [ ] Post-merge: create a PR for adding a lassa card to the pathogens page

### DIFF
--- a/data/manifest_core.json
+++ b/data/manifest_core.json
@@ -193,6 +193,13 @@
           "default": "seasonal"
         }
       },
+      "lassa": {
+        "resolution": {
+          "l": "",
+          "s": "",
+          "default": "s"
+        }
+      },
       "measles": "",
       "mpox": {
         "resolution": {


jameshadfield:
:+1:

Lassa is already a `coreBuildPath` ([here](https://github.com/nextstrain/nextstrain.org/blob/master/src/routing/core.js#L28)) which is why the /lassa URL gets sent to auspice. Without the corresponding entry in the manifest you need the full route (e.g. https://nextstrain.org/lassa/s). So this PR will essentially allow https://nextstrain.org/lassa to redirect (to /lassa/s) as well as having lassa in the auspice dropdown (because the manifest is used for the `getAvailable` API)

PR #769, opened 2023-12-19
whoami: List public group memberships…

victorlin:
… in addition to private group memberships.

If anything, I think it's even more valuable to have the list of public group memberships, as private group memberships can be easily found by logging in, going to the Groups index page, scrolling to the bottom, and looking for groups with a padlock icon.

## Related issue(s)

N/A, came across this while testing Groups stuff.

## Checklist

- [x] Checks pass

### DIFF
--- a/static-site/src/pages/whoami.jsx
+++ b/static-site/src/pages/whoami.jsx
@@ -12,9 +12,21 @@ const UserPage = () => {
     <Fragment>
       You&apos;re logged in as <strong>{user.username}</strong>.
       <SubText>
-        You have access to the following private Nextstrain groups, which each
+        You have access to the following Nextstrain groups, which each
         contain a collection of datasets and/or narratives:
       </SubText>
+
+      Public:
+
+      <UserGroupsList>
+        {visibleGroups.filter((group) => !group.private).map((group) => (
+          <li>
+            <a href={`/groups/${group.name}`}>{group.name}</a>
+          </li>
+        ))}
+      </UserGroupsList>
+
+      Private:
 
       <UserGroupsList>
         {visibleGroups.filter((group) => group.private).map((group) => (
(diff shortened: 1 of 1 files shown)


j23414 (approved):
I like the idea of listing both public and private groups. LGTM

static-site/src/pages/whoami.jsx
```diff
@@ -12,17 +12,30 @@ const UserPage = () => {
     <Fragment>
       You&apos;re logged in as <strong>{user.username}</strong>.
       <SubText>
-        You have access to the following private Nextstrain groups, which each
+        You have access to the following Nextstrain groups, which each
```
joverlee521, line 15: +1 for listing user membership of public groups. 

Slight devil's advocate of whether this will confuse users because technically they can access _all_ public groups via nextstrain.org.

static-site/src/pages/whoami.jsx
```diff
@@ -12,17 +12,30 @@ const UserPage = () => {
...
 
+      Public:
+
+      <UserGroupsList>
+        {visibleGroups.filter((group) => !group.private).map((group) => (
+          <li>
+            <a href={`/groups/${group.name}`}>{group.name}</a>
+          </li>
+        ))}
+      </UserGroupsList>
```
tsibley, line 27: This lists all public groups, i.e. the list will be the same for any logged in user. Was the intent to list just the public groups that the user is a member of?

victorlin: 🤦 should have checked canary after merging... I'll open a fix before we promote to nextstrain.org.

tsibley: No worries. Thanks!

victorlin: Reverted in #779 and re-opened as #780.

static-site/src/pages/whoami.jsx
victorlin: Updated wording in #780:

> You are a member of the following Nextstrain groups, which each …

PR #771, opened 2024-01-02
Update manifest with dengue gene datasets

j23414:

Per user request, update dengue builds to include E genes as well as whole genome.

## Related issue(s)

* https://github.com/nextstrain/dengue/issues/17 

## Checklist

- [ ] Checks pass

### DIFF
--- a/data/manifest_core.json
+++ b/data/manifest_core.json
@@ -3,12 +3,42 @@
     "pathogen": {
       "dengue": {
         "serotype": {
-          "denv1": "",
-          "denv2": "",
-          "denv3": "",
-          "denv4": "",
-          "all": "",
-          "default": "denv1"
+          "denv1": {
+            "segment": {
+              "genome": "",
+              "E": "",
+              "default": "genome"
+            }
+          },
+          "denv2": {
+            "segment": {
+              "genome": "",
+              "E": "",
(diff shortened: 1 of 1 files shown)


j23414:
Merging, since Dataset dropdown options look correct:

* https://nextstrain-s-dengue-man-th1myu.herokuapp.com/dengue/all/E

jameshadfield:
From reading through the changes here, URLs such as "https://nextstrain.org/dengue/denv1" will now be redirected to "https://nextstrain.org/dengue/denv1/genome". On the live site (which doesn't yet have the changes in this PR) both URLs are valid and return the same dataset, so I presume this is deliberate and part of a migration of URLs to the new format?

j23414:
Thanks for checking in! This is intentional; I've uploaded files such as `dengue_denv1_genome.json` with the intention of removing the original `dengue_denv1.json` files within the next week or month.

Feel free to suggest alternative approaches to display both "genome" and "E" in the Dataset dropdown menu, or on the site. It's quite possible I've missed some preferred route for this or am misunderstanding something.

jameshadfield:
This is the best way - auspice doesn't handle mixed levels like  `.../A` + `.../A/B` well. Just wanted to make sure this path was intentional!

PR #772, opened 2024-01-02
Add script to update GitHub token on Heroku apps

victorlin:

Using a script allows for minimal time between GitHub token regeneration and config var updates, especially when the same token is used across multiple apps. Also useful to somewhat automate the process given GitHub requires token regeneration at least once every year.

Note that it'll still take some time between manually regenerating the token on the GitHub UI, pasting it as an argument to this script, then running the script. In the case of this GitHub token, it's not currently being used at a level where this matters critically.

In the future if this becomes more critical, the downtime could be eliminated by generating a new token while keeping the old one valid.

## Related issue(s)

N/A

## Checklist

- [x] Checks pass
- [x] Used after the latest token regeneration
- [ ] Post-merge: Update [internal wiki page](https://nextstrain.atlassian.net/wiki/spaces/NEXTSTRAIN/pages/170688571/GitHub+nextstrain-bot+usage)

### DIFF
--- a/scripts/update-github-token
+++ b/scripts/update-github-token
@@ -1,0 +1,59 @@
+#!/bin/bash
+# Updates nextstrain.org Heroku apps with a new GitHub token.
+#
+set -euo pipefail
+
+main() {
+    if [ "$#" -ne 1 ]; then
+      echo "Usage: $0 <GITHUB_TOKEN>"
+      exit 1
+    fi
+
+    check-deps
+
+    check-heroku-access
+
+    update-github-token "$1"
+}
+
+check-deps () {
+    echo "This script depends on heroku and jq. Checking that they are available..."
+
+    heroku --help > /dev/null
+    jq --help > /dev/null
+
+    echo "Done."
+}
+
+check-heroku-access() {
+    echo "Checking that you are logged in. If you see an error, please log in to Heroku CLI."
+
+    heroku whoami >/dev/null
+
+    echo "You are logged in. Checking access..."
+
+    is_nextstrain_admin=$(heroku teams --json | jq '.[] | select(.name == "nextstrain" and .role == "admin") | length == 1')
+    if [[ -z "$is_nextstrain_admin" ]]; then
+      echo "ERROR: Insufficient access with Heroku CLI. Please log out then log back in to Heroku CLI as a member of the Nextstrain team."
+      exit 1
+    fi
+
+    echo "You have access."
+}
+
+update-github-token() {
+    new_token="$1"
+
+    echo "Updating tokens..."
+
+    for app in "nextstrain-dev" "nextstrain-canary" "nextstrain-server"
+    do
+        old_token=$(heroku config:get GITHUB_TOKEN --app "$app")
+        echo "[$app] Replacing old token $old_token with $new_token"
+        heroku config:set GITHUB_TOKEN="$new_token" --app "$app"
+    done
+
+    echo "Done."
+}
+
+main "$@"


scripts/update-github-token
victorlin, line 1: Current output examples:

```
$ ./scripts/update-github-token XXXXXXXXXXXXXXXX
Checking that you are logged in. If you see an error, please log in to Heroku CLI.
You are logged in. Checking access...
You have access.
Updating tokens...
[nextstrain-dev] Replacing old token XXXXXXXXXXXXXXXX with XXXXXXXXXXXXXXXX
Setting GITHUB_TOKEN and restarting ⬢ nextstrain-dev... done, v170
GITHUB_TOKEN: XXXXXXXXXXXXXXXX
[nextstrain-canary] Replacing old token XXXXXXXXXXXXXXXX with XXXXXXXXXXXXXXXX
Setting GITHUB_TOKEN and restarting ⬢ nextstrain-canary... done, v241
GITHUB_TOKEN: XXXXXXXXXXXXXXXX
[nextstrain-server] Replacing old token XXXXXXXXXXXXXXXX with XXXXXXXXXXXXXXXX
Setting GITHUB_TOKEN and restarting ⬢ nextstrain-server... done, v896
GITHUB_TOKEN: XXXXXXXXXXXXXXXX
Done.
```

Errors:

```
$ ./scripts/update-github-token XXXX Y
Usage: ./scripts/update-github-token <GITHUB_TOKEN>
```

```
$ ./scripts/update-github-token XXXXXXXXXXXXXXX
This script depends on heroku and jq. Checking that they are available...
/opt/homebrew/Cellar/heroku/7.60.1/lib/client/bin/heroku: line 16: /opt/homebrew/Cellar/heroku/7.60.1/lib/client/bin/../8.7.1-3f5e369/bin/heroku: No such file or directory
```

```
$ ./scripts/update-github-token XXXXXXXXXXXXXXX
This script depends on heroku and jq. Checking that they are available...
Done.
Checking that you are logged in. If you see an error, please log in to Heroku CLI.
 ›   Error: not logged in
```

```
$ ./scripts/update-github-token XXXXXXXXXXXXXXX
This script depends on heroku and jq. Checking that they are available...
Done.
Checking that you are logged in. If you see an error, please log in to Heroku CLI.
You are logged in. Checking access...
ERROR: Insufficient access with Heroku CLI. Please log out then log back in to Heroku CLI as a member of the Nextstrain team.
```

scripts/update-github-token
```diff
@@ -0,0 +1,59 @@
...
+update-github-token() {
+    new_token="$1"
+
+    echo "Updating tokens..."
+
+    for app in "nextstrain-dev" "nextstrain-canary" "nextstrain-server"
+    do
+        old_token=$(heroku config:get GITHUB_TOKEN --app "$app")
+        echo "[$app] Replacing old token $old_token with $new_token"
+        heroku config:set GITHUB_TOKEN="$new_token" --app "$app"
```
tsibley, line 53: It'd also be better to not have the token in plain text in the output and in the process args. Doing the latter is a bit more tedious, but can be done with `EDITOR=… heroku config:edit GITHUB_TOKEN` where `…` is the semi-tedious part.

scripts/update-github-token
tsibley: It would be better to not have the token passed in the process args, esp. where it'll end up in (usually long-lived) shell history. Secrets are better passed by stdin or a filename or an env var.

tsibley (approved):
A script is fine… but I'd think to extend our Terraform config to cover this if we want to start automating management of this task. I know there's a Heroku provider for Terraform which would let us set config vars on Heroku apps, and we can provide the token to Terraform using an input variable marked as sensitive. This is a common pattern, and using Terraform will will provide a measure of self-documentation as well.

jameshadfield:
How long does updating the tokens manually take? Reading our [(internal) wiki](https://nextstrain.atlassian.net/wiki/spaces/NEXTSTRAIN/pages/170688571/GitHub+nextstrain-bot+usage) I'd guess 15 minutes to obtain a new token and update the 3 apps. So let's say 30min. And we have to do it once per year. [In my opinion it's not worth automating this.](https://xkcd.com/1205/) (And this script still requires us to manually obtain the token from GitHub if I'm understanding it correctly.)

victorlin:
@jameshadfield you're right, it doesn't take much time to do this manually. I wanted to do it via Heroku CLI, and once I had the commands, I figured it could be documented for future selves via a script.

If time expenditure is the concern, unfortunately I've already spent the time on this and can't take it back, but I can break it down. Based on my activity tracker, I spent ~1.5 hours on this task from seeing GitHub's reminder email to creating this PR. In that time, ~1 extra hour went into creating this PR versus had I gone in and looked for how to update the config vars and updated them directly. In this extra time, I learned more about how Heroku CLI works, how to author bash scripts, and put some thought into the secret rotation process - none of which was necessary, but something inside of me that day wanted to do it. This isn't important in the bigger picture of Nextstrain development and not something we would prioritize as a group, so consider it part of the 50% time that I'm using for unprioritized effort.

victorlin:
@jameshadfield and I chatted about this briefly on a call while we were discussing other things. There was a misunderstanding - the time expenditure described above is not a concern.

I think it's ok for there to be multiple approaches to handling this process. I made this to save myself from clicking around on the Heroku Dashboard. But also fine to make the same changes using Heroku Dashboard. Maybe in a year things will change and we won't even have to do this process again. But if we do, I would likely reach for such a script.

The process to do this using Heroku Dashboard wasn't written anywhere, so I just added them [here](https://nextstrain.atlassian.net/wiki/pages/diffpagesbyversion.action?pageId=170688571&selectedPageVersions=14&selectedPageVersions=15). That should clarify the scope of this script. I'd only continue work here if others on the team would also use the script, so let me know if you would! Otherwise it's been a while since I've touched this and we don't need this any time soon, so I'll close for now.

PR #773, opened 2024-01-03
Use anchor for resource links

victorlin:

The previous "links" relied on onClick which is bad for accessibility. If div is still desired, the link role would be appropriate.¹ Or, use an anchor element instead, as done in this commit.

¹ <https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/link_role>

## Related issue(s)

N/A

## Checklist

- [ ] Checks pass
- [ ] Make the same change in Auspice

### DIFF
--- a/auspice-client/customisations/splash.js
+++ b/auspice-client/customisations/splash.js
@@ -45,11 +45,11 @@ const Splash = ({available, browserDimensions, dispatch, errorMessage, changePag
         }
         <div style={{marginTop: "30px"}}/>
         {pageInfo.showDatasets ?
-          <ListAvailable type="datasets" data={available.datasets} width={browserDimensions.width} dispatch={dispatch} changePage={changePage}/> :
+          <ListAvailable type="datasets" data={available.datasets} width={browserDimensions.width}/> :
           null
         }
         {pageInfo.showNarratives ?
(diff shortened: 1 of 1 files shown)


auspice-client/customisations/splash.js
```diff
@@ -114,17 +114,19 @@ function ErrorMessage({errorMessage}) {
...
 
+const ListItem = styled.li`
+  margin: 5px 0;
+`
+
 /* lifted from auspice's default splash page */
-function formatDataset(requestPath, dispatch, changePage) {
+function formatDataset(requestPath) {
+
+  // FIXME: Do we actually need dispatch and changePage here? I've removed them for now.
```
victorlin, line 124: This is my main concern - it's not clear to me what `dispatch` and `changePage` were used for. Seems to work fine without it.

jameshadfield: The behaviour will be a little different -- dispatching stays within the SPA, whereas links will go to a new URL, fetch the auspice HTML entrypoint, fetch the auspice JS (cache hit), reparse the JS and so on. In practice, the Auspice splash page within nextstrain.org is used rarely - perhaps only for community data?

P.S. I've wanted proper link behaviour for a while now, especially in the default Auspice splash page.

PR #754, opened 2023-11-16
static-site: Remove sharp-related dependencies

victorlin:
## Description of proposed changes

These don't seem to be necessary, and make it harder to set up a dev environment on Apple silicon macs.

## Related issue(s)

- Resolves #597

## Checklist

- [ ] Checks pass
- [ ] Check that the removal of these gatsby plugins are acceptable
    - `gatsby-remark-images` ([thread](https://github.com/nextstrain/nextstrain.org/pull/754#discussion_r1448116570))

PR #778, opened 2024-01-09
static-site: Configure and use ESLint

### LINKED ISSUE #775: eslint disabled for static-site
### Current Behavior

Currently linting is (seemingly) disabled for the static-site contents. Running with `--debug` indicates it is inheriting the `static-site/` ignore pattern from the parent directory. Removing the line

https://github.com/nextstrain/nextstrain.org/blob/c3dd7262f8cff0a5093567629cc4d55d55668ad6/.eslintrc.yaml#L17

allows `npx eslint --ext .js,.jsx  .` to work as expected within the static-site directory and (but?) also flags up linting errors in the static-site when run from the root directory.

The static site has a `.eslintrc.yaml` which indicates that our intention is to lint it (but using our own rules):

https://github.com/nextstrain/nextstrain.org/blob/c3dd7262f8cff0a5093567629cc4d55d55668ad6/static-site/.eslintrc.yaml#L1-L2

@victorlin I think you're best placed here at the moment -- is this expected behaviour? I would think we want to be linting the static-site. It currently has around 30k (yes 30,000) linting errors. 

### Eslint versions

This is an aside, but I'll mention it her so others don't have to investigate.

From the root directory we have eslint v8.23.1 (via `npm ci`), however within the static site we have version 6.8.0 (also via `npm ci`). The static-site's eslint is via:

`gatsby` (has a dependency of) `eslint-config-react-app` (has a peer dep of) `@typescript-eslint/eslint-plugin` (has a peer dependency of) `eslint "^5.0.0 || ^6.0.0"`

This doesn't matter, as running `../node_modules/.bin/eslint --ext .js,.jsx   .` from the static-site gives the same output as below (but with the reported eslint version being v8).

### Expected behavior
Static site JS is linted.

### How to reproduce
Steps to reproduce the current behavior:
1. `cd static-site`
2. `npm ci`
3. `npx eslint --ext .js,.jsx   .` 
4. Error:
```
Oops! Something went wrong! :(

ESLint: 6.8.0.

You are linting ".", but all of the files matching the glob pattern "." are ignored.

If you don't want to lint these files, remove the pattern "." from the list of arguments passed to ESLint.

If you do want to lint these files, try the following solutions:

* Check your .eslintignore file, or the eslintIgnore property in package.json, to ensure that the files are not configured to be ignored.
* Explicitly list the files from this glob that you'd like to lint on the command-line, rather than providing a glob as an argument.
```

victorlin:

This project was previously unlinted. Add the necessary configuration including rule exceptions to make the linter run without error on the code as-is.

## Related issue(s)

Fixes #775

## Checklist

- [x] Checks pass
- [x] Check violations of recommended rules ([thread](https://github.com/nextstrain/nextstrain.org/pull/778#discussion_r1446713406))


static-site/.eslintrc.yaml
```diff
@@ -1,2 +1,28 @@
...
+ignorePatterns:
+  - public/
+
+rules:
+  react/jsx-key: off
+  react/jsx-no-target-blank: off
+  react/no-unescaped-entities: off
+  react/no-unknown-property: off
+  react/prop-types: off
+  react-hooks/exhaustive-deps: off
```
jameshadfield, line 21: I'd suggest enabling this rule -- missing deps is such as common problem with hooks

static-site/package.json
```diff
@@ -12,6 +12,7 @@
     "test": "gatsby build",
     "build:pp": "gatsby build --prefix-paths",
     "clean": "rm -rf public",
+    "lint": "eslint --ext .js,.jsx .",
```
jameshadfield, line 15: This rule works for me now 🎉  

Static-site files are not linting in VS-Code for me (the linting of server-code works as expected). Perhaps this isn't possible as it'd require vs-code to be running multiple eslint versions? If it's working for you please ignore the following which must be specific to my setup & which we can debug elsewhere. 

Here's the eslint output (from within vs-code):
```
[Info  - 12:16:47] ESLint server is starting.
[Info  - 12:16:47] ESLint server running in node v18.15.0
[Info  - 12:16:47] ESLint server is running.
Uncaught exception received.
Error: spawn /Applications/Visual Studio Code.app/Contents/Frameworks/Code Helper (Plugin).app/Contents/MacOS/Code Helper (Plugin) ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:283:19)
    at onErrorNT (node:internal/child_process:476:16)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
[Info  - 12:16:47] ESLint library loaded from: /Users/naboo/github/nextstrain/nextstrain.org/node_modules/eslint/lib/api.js
[Info  - 12:16:47] ESLint library loaded from: /Users/naboo/github/nextstrain/nextstrain.org/static-site/node_modules/eslint/lib/api.js
```

Note: I had to test with this PR in combination with #754 so my package-lock isn't identical to this PR's.

victorlin: No problem, I think fine to debug in this thread. I am using VSCode extension [ESLint](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) v2.4.2 and this is my output (I had to open a JS file in both static-site and server code to get the last 2 lines to show):

```
[Info  - 3:30:18 PM] ESLint server is starting.
[Info  - 3:30:19 PM] ESLint server running in node v18.15.0
[Info  - 3:30:19 PM] ESLint server is running.
[Info  - 3:30:19 PM] ESLint library loaded from: /Users/vlin/Dropbox/repos/nextstrain/nextstrain.org/static-site/node_modules/eslint/lib/api.js
[Info  - 3:30:19 PM] ESLint library loaded from: /Users/vlin/Dropbox/repos/nextstrain/nextstrain.org/node_modules/eslint/lib/api.js
```

What version of the extension do you have?

jameshadfield: I also have v2.4.2. Apart from the uncaught exception I'm experiencing the output looks the same. Does the linting actually work for you in the open JS files? As in, if you introduce an error do you get red-squiggly underlines & eslint on-hover infobox?

victorlin: That's weird. Linting on static-site works fine for me. For example, if I remove the bypass `react/no-unknown-property: off` and open `static-site/src/components/Datasets/dataset-map.jsx`, I see:

static-site VSCode ESLint error

Linting on server code also works fine. When adding a unused variable to server.js, I see:

server VSCode ESLint error

jameshadfield: Thanks for confirming! I'll do a bit more digging when I get the chance.

static-site/.eslintrc.yaml
victorlin: This is mostly done. There's one which I haven't dug deep enough to know how to best address: e17c5dca9b4d69699b1a92e2408d4f70f05b5295

@joverlee521 you might have a better sense of this part of the code.

joverlee521: Hmm, I think it's fine to disable the rule for hooks that are only expected to be run once. 

(Also, this should be raising a similar warning for `getOverview` within `edit-overview-form.jsx`)

victorlin: The expectation of running once is what this is rule addresses - in theory it should run every time props are updated, not just initial load (see https://github.com/facebook/react/issues/14920#issuecomment-470945051).

In practice, I ran this locally and couldn't find any cases where the props would change after initial load. If there was a button to go directly from editing one group's settings to another, maybe this would start to break, but currently it seems fine from a bit of testing.

I'll add a comment to disable with this reasoning.

victorlin: > Also, this should be raising a similar warning for `getOverview` within `edit-overview-form.jsx`

That's odd, I see the warning locally but not in CI logs. Will look into it.

jameshadfield: From reading the code the idiomatic way to do this is to memoise the function. I don't think you can use `useMemo` for an async function, but you can achieve the same effect with `useEffect`. In this case it's simpler (and arguably clearer) to disable the linting & add a comment 👍

victorlin: Comments added in 1c111308942e2793c40172b8bf9d61ecfc7786bb.

static-site/package.json
jameshadfield: I did some digging and think I understand what's going on, although it'd be interesting to know if you can replicate.

The code I was adding to test whether linting was working doesn't seem to be covered by the current `static-site/.eslintrc.yaml`. I was testing via adding things like:
```js
const unused = 1; // expect no-unused-vars error
const bbb = {key: un_used}; // expect 'un_used is not defined no-undef'
```
For me to get these errors in the static site I had to add `eslint:recommended` as an extension. I think this should be added to our config, but I'm interested to know if you get these kinds of errors flagged up as well? I'm guessing the answer is no because there are a few of these errors currently in the static-site code.

Adding that extension flags up 91 linting errors in the current code. Adding `browser: true` to `env` reduces this to 42. Most of them are `error  'require' is not defined` and `error  'exports' is not defined` as gatsby code uses this style. However there are also legitimate errors flagged up.

P.S. Without adding `eslint:recommended`, I _am_ getting linting working correctly (vs-code and cli) for specific JSX/hooks errors - for instance if I remove the `react/prop-types: off` rule then I get those errors flagged up. This was previously not working as expected, so I'll continue to monitor this - perhaps it was another vscode extension (I disabled them all & restarted to debug this), perhaps it was transient. I don't know.

victorlin: Oh right, I forgot about `eslint:recommended`. We should be using that just as we do in the server code. I'll add it.

victorlin: Added `eslint:recommended` in 875ff4b...6ed471e.

.github/workflows/ci.yml
```diff
@@ -6,6 +6,17 @@ on:
...
+  static-site-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '16'
+      - run: cd static-site/
+      - run: npm ci
+      - run: npm run lint
```
victorlin: 🤦 This wasn't working as expected because `cd` only applies to the current step. Need to use `working-directory`.

jameshadfield: Good catch! Relevant [docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun):

> Each run keyword represents a new process and shell in the runner environment. 

Looks like you can also set a working directory for all steps in the job (although in this case one line more than your solution!)

```yaml
  static-site-lint:
    runs-on: ubuntu-latest
    defaults:
      run:
        working-directory: static-site
```

static-site/package.json
```diff
@@ -12,6 +12,7 @@
     "test": "gatsby build",
     "build:pp": "gatsby build --prefix-paths",
     "clean": "rm -rf public",
+    "lint": "DEBUG=eslint:cli-engine eslint --ext .js,.jsx .",
```
jameshadfield, line 15: Did you mean to leave this DEBUG variable set? I quite like the output it gives, so I'm happy for it to stay.

jameshadfield (approved):
LGTM

.github/workflows/ci.yml
victorlin: Oh, I like the `defaults` approach better. Even though it's more lines, it makes more sense because the scope of the directory is on the job level, not step level.

static-site/package.json
victorlin: Yes, I added it when investigating the discrepancy between local and GitHub runner ESLint runs. It's not overly verbose, so I don't see any harm in keeping it around.

.github/workflows/ci.yml
victorlin: Updated in [force-push](https://github.com/nextstrain/nextstrain.org/compare/6083dc569735afc877bb85cffbd85a1b6578c406..225accdf503a35290769b5e0caef57704b39e236).

PR #780, opened 2024-01-10
whoami: List public group memberships (take 2)

victorlin:

See commit messages.

## Related issue(s)

- Re-attempt at #769

## Checklist

- [x] Checks pass
- [x] Local testing: public testing env group memberships are listed
- [ ] Post-merge: public production env group memberships are listed on canary site


src/endpoints/users.js
```diff
@@ -22,6 +22,24 @@ const visibleGroups = (user) => ALL_GROUPS
...
+ *                  `private` -> {bool} (group is private)
+ */
+const groupMemberships = (user) => ALL_GROUPS
+  .map(g => [g.name, g.source])
+  // Check roles (formatted as '<groupName>/<role>') for group membership.
+  .filter(([name ,]) => [...user.authzRoles].some(role => role.split('/')[0] === name))
+  .map(([name, source]) => ({
+    name,
+    private: !authz.authorized(null, authz.actions.Read, source),
+  }));
```
tsibley, line 40: Code outside of `src/authn/` (and `src/groups.js`) generally shouldn't be splitting (or joining) on `/` to form authz roles. That implementation detail should remain encapsulated. There's also a list of the user's groups (by name) already on the user object, so filtering down in this way from `ALL_GROUPS` isn't necessary.

```suggestion
const groupMemberships = (user) => user?.groups?
  .map(name => new Group(name))
  .map(group => ({
    name: group.name,
    private: !group.isPublic,
  }))
  ??
  []
;
```

The nullish coalescing operators used make it so the caller can simply do `groupMemberships(req.user)` even when `req.user === null`.

Longer term, it probably makes sense to convert `req.user.groups` to Group objects in the authn code rather than here. There can also be some additional error/edge-case handling (e.g. group names reported by Cognito that the server doesn't know about).

tsibley (changes requested):

src/endpoints/users.js
victorlin: Good suggestion. Switched to use `user.groups` in 3a4dcc79a8978bda401726d43c7caf7b72dc9c5a. I took it a bit further by returning the list of Groups to use `name` and `isPublic` directly instead of through an intermediate object.

src/endpoints/users.js
```diff
@@ -30,6 +43,7 @@ export const getWhoami = contentTypesProvided([
     res.json({
       user: req.user || null,
       visibleGroups: visibleGroups(req.user),
+      groupMemberships: groupMemberships(req.user),
```
tsibley, line 46: _non-blocking_

I'm hesitant to start exposing all the basic properties of the Group class this way, e.g.:

```json
{
  "name": "test",
  "isPublic": true,
  "membershipRoles": {
    "viewers": "test/viewers",
    "editors": "test/editors",
    "owners": "test/owners"
  }
}
```

It seems fine at the moment (although `membershipRoles` is really internal and shouldn't be needed by the frontend; any need is a misuse). But I'm worried we won't realize we're automatically and publicly serializing the Group class some time down the road and we'll leak something we shouldn't.

tsibley (commented):
This needs rebasing for the Gatsby → Next.js switch.

src/endpoints/users.js
victorlin: Good point. Addressed this with bd882f9503422fb7545e86ba72fbe8696947f988

tsibley (approved):

src/endpoints/users.js
```diff
@@ -22,6 +22,24 @@ const visibleGroups = (user) => ALL_GROUPS
...
+/**
+ * Returns an array of Nextstrain groups that a given *user* is a member of. The
+ * order of groups returned matches the order in the `groups.json` data file.
+ *
+ * @param {Object | undefined} user. `undefined` represents a non-logged-in user
+ * @returns {Array} Each element represents a group with a a subset of properties from
+ *                  the Group class.
+ */
+const groupMemberships = (user) => user?.groups
+  ?.map(name => new Group(name))
```
tsibley, line 34: This ended up breaking `/whoami` for me (noticed in production just now) because my Cognito user is a member of Cognito groups for which there are no corresponding Nextstrain groups. This is a historical artifact and likely only affects me (and maybe one or two other devs), but it'd still be better to be robust in the face of unexpected input.

I noticed the issue in the web, where `/whoami` 404s for me when JSON is requested, but reproduced it below with `curl` for an example:

```console
$ nextstrain whoami
# Logged into https://nextstrain.org as…
username: trs
email: tsibley@fredhutch.org
groups: 
  - !flags/canary
  - blab-private/editors
  - blab/editors
  - seattleflu/editors
  - test/editors
  - trs-test/owners

$ curl -s https://nextstrain.org/whoami -H 'Accept: application/json' -H @<(nextstrain authorization)
{"error":"unknown group: test"}
```

tsibley: I'd [previously said](https://github.com/nextstrain/nextstrain.org/pull/780#discussion_r1448077922):

> Longer term, it probably makes sense to convert `req.user.groups` to Group objects in the authn code rather than here. There can also be some additional error/edge-case handling (e.g. group names reported by Cognito that the server doesn't know about).

and I guess I shouldn't have said "Longer term"!

victorlin: Ahh, fixing in #839

Issue #707, opened 2023-08-24
Add Cache-Control header to Charon API endpoints


tsibley:
Current responses are caching-eligible but don't define acceptable cache behaviour via Cache-Control. This leaves the decision up to browsers' heuristics and can cause understandable user confusion if they upload a new dataset but don't see that reflected in their browser because it's got a fresh cache entry for the `/charon/getDataset` request. This happened recently to Katie.

We should tell browsers what fresh means to us using `Cache-Control`.

tsibley:
I think our best option is likely:

```
Cache-Control: no-cache
```

which permits caching by browsers (like now) but instructs them to always check for updates (revalidate) before using the cached response (unlike now). This doesn't save a request/response cycle, but it does save response transfer time when a cached response can be used. It also forbids using the cached response when the browser is offline, though I'm not sure we care about that use case (and not sure other bits of nextstrain.org support it, like the Auspice client).

Alternatively, we could use something like:

```
Cache-Control: max-age=0, stale-while-revalidate=300
```

to permit using a cached response for up to 5 minutes (300s) while the browser tries to revalidate it. The wouldn't save a request/response cycle, but it would move it out of the foreground and into the background. It does mean users might have to reload several times before they saw an update to the cached copy. This supports offline use cases, but again, that may not matter.

tsibley:
Also, Auspice could, either instead of or additionally to the above, make the _requests_ to `/charon/getDataset` with `Cache-Control: no-cache`.

jameshadfield:
`Cache-Control: no-cache` for charon responses seems good to me

corneliusroemer:
Bumping this as I independently thought about it not realizing there's already is an issue!
https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1705507737089689

PR #782, opened 2024-01-17
upstream: Add a Cache-Control header to resource responses

tsibley:
This permits private caching (e.g. by browsers), like now, but instructs them to always check for updates (revalidate) before using the cached response (unlike now).  Prior to this, browsers would use their own heuristics to figure out when to revalidate which could lead to staleness and user confusion.

This new caching policy doesn't save a request/response cycle, but it does save response transfer time when a cached response can be used. It also forbids using the cached response when the browser is offline, though I'm not sure we care about that use case (and not sure other bits of nextstrain.org support it, like the Auspice client).

This affects the RESTful API and Charon API endpoints for datasets and narratives, as well as the endpoints for a group's overview and logo.

Resolves <https://github.com/nextstrain/nextstrain.org/issues/707>.

## Checklist

- [x] Checked caching behaviour locally; seems as expected
- [x] Checks pass

### DIFF
--- a/src/upstream.js
+++ b/src/upstream.js
@@ -238,6 +238,12 @@ async function proxyResponseBodyFromUpstream(req, res, upstreamReq) {
 
   res.set(copyHeaders(upstreamRes.headers, forwardedUpstreamResHeaders));
 
+  /* Allow private (e.g. browser) caches to store this response, but require
+   * them to revalidate it every time before use.  They'll make conditional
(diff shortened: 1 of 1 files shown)


jameshadfield (approved):
Looks good by inspection. This won't be set for v1 datasets (`sendV1Dataset`) but they shouldn't be being created anymore so cache validation seems unnecessary.

PR #783, opened 2024-01-17
Surface previous dataset versions across URL redirects

### LINKED ISSUE #777: [resource index] use core manifest to update datasets whose URLs are redirected
This issue originates from [this slack thread](https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1704395193748589)

The current index reflects the datasets on (the core) s3 bucket. Some of the dataset URLs produced there are inaccessble as the server redirects based on data in the [manifest](https://github.com/nextstrain/nextstrain.org/blob/master/data/manifest_core.json). One example is `dengue/denv1` which now redirects to `dengue/denv1/genome` (via [this recent PR](https://github.com/nextstrain/nextstrain.org/pull/771)). Using the abridged index shown below, a request for `/dengue/denv1@2023-12-25` should ideally return version "A", however the server uses the manifest to resolve `/dengue/denv1` to `/dengue/denv1/genome` and then, because `2023-12-25` is older than the earliest `/dengue/denv1/genome` dataset we throw a `NotFoundError: Version descriptor 2023-12-25 predates available versions` error. 

To improve this the indexer should use the manifest data to remove `dengue/denv1` resources and include them as part of the `dengue/denv1/genome` resource. This should allow the above requests to "just work" - the server will send a redirect to `/dengue/denv1/genome@2023-12-25`, that URL will be a 200 and the underlying dataset returned will be version "A".  But I haven't tested this.

Abridged version of the current index:

```
{
  "dengue/denv1": {
    "versions": [
      ...,
      {
        "date": "2023-12-25",
        "fileUrls": {
          "main": "https://nextstrain-data.s3.amazonaws.com/dengue_denv1.json?versionId=A"
        }
      }
    ]
  },
  "dengue/denv1/genome": {
    "versions": [
      {
        "date": "2024-01-04",
        "fileUrls": {
          "main": "https://nextstrain-data.s3.amazonaws.com/dengue_denv1_genome.json?versionId=B"
        }
      },
      ...
    ]
  }
}
```

jameshadfield:
Closes #777. See that issue or the commit messages for full details.

### Example URLs which are enabled via this work:

The earliest nCoV dataset we have is from 2020-01-19, and was back then accessed via `/ncov`. Now `/ncov@2020-01-19` redirects to `/ncov/gisaid/global/6m@2020-01-19` and loads that dataset. 

We can access the `/dengue/denv1` datasets prior to Jan 2024 via either `/dengue/denv1@YYYY-MM-DD` or `/dengue/denv1/genome@YYYY-MM-DD`, with the former redirecting to the latter.

### Testing

Best done locally, via something like
```bash
mkdir -p devData
node resourceIndexer/main.js --output devData/index.json --save-inventories \
  --indent --resourceTypes dataset --collections core # needs AWS creds to access s3://nextstrain-inventories
RESOURCE_INDEX="devData/index.json" node server.js --verbose
```

The heroku review app will not correctly use this functionality as the live index has not been updated with the changes in  this PR. If you want a non-local way to test this let me know and I'll make it happen. (You can test a subset of the functionality, e.g. [/dengue/denv1@2024-01-03](https://nextstrain-s-james-reso-uw9ge9.herokuapp.com/dengue/denv1@2024-01-03) works as expected.)

### DIFF
--- a/src/endpoints/sources.js
+++ b/src/endpoints/sources.js
@@ -43,7 +43,8 @@ export const setDataset = (pathExtractor) => (req, res, next) => {
 
 /**
  * Generate Express middleware that redirects to the canonical path for the
- * current {@link Dataset} if it is not fully resolved.
+ * current {@link Dataset} if it is not fully resolved. Any provided version
+ * descriptor is included in the redirect.
  *
  * @param {pathBuilder} pathBuilder - Function to build a fully-specified path
  * @returns {expressMiddleware}
@@ -54,9 +55,11 @@ export const canonicalizeDataset = (pathBuilder) => (req, res, next) => {
 
   if (dataset === resolvedDataset) return next();
 
+  const version = dataset.versionDescriptor ? `@${dataset.versionDescriptor}` : '';
+
   const canonicalPath = pathBuilder.length >= 2
-    ? pathBuilder(req, resolvedDataset.pathParts.join("/"))
-    : pathBuilder(resolvedDataset.pathParts.join("/"));
+    ? pathBuilder(req, resolvedDataset.pathParts.join("/") + version)
+    : pathBuilder(resolvedDataset.pathParts.join("/") + version);
 
   /* 307 Temporary Redirect preserves request method, unlike 302 Found, which
    * is important since this middleware function may be used in non-GET routes.
--- a/test/date_descriptor.test.js
+++ b/test/date_descriptor.test.js
@@ -10,7 +10,7 @@
 /**
  * Example dev use:
  * run the server in a separate process via:
- * RESOURCE_INDEX="./devData/index.json" node server.js --verbose
+ * RESOURCE_INDEX="./test/date_descriptor_index.json" node server.js --verbose
  * Then run this test file via:
  * NODE_OPTIONS='--experimental-vm-modules' npx jest test/date_descriptor.test.js
  */
--- a/test/date_descriptor.test.js
+++ b/test/date_descriptor.test.js
@@ -301,6 +301,54 @@ describe("Valid narratives", () => {
   }
 })
 
+const redirects = [
+  ["WNV", "WNV/NA"],
+  /* Note that WNV is not in the index, so this versioned test ensures we are
+  not checking the existence/validity of any version descriptor against the WNV
+  resource, rather we are redirecting and deferring those checks */
+  ["WNV@2020-01-01", "WNV/NA@2020-01-01"],
+]
+
+describe("Paths redirect with version descriptors", () => {
+  /* See <https://github.com/node-fetch/node-fetch#manual-redirect> for how
+  fetch stores the redirect location when {redirect: 'manual'} */
+  
+  redirects.forEach(([fromUrl, toUrl]) => {
+    /* test RESTful API */
+    (['html', 'json']).forEach((type) => {
+      it(`REST API: ${fromUrl} goes to ${toUrl} (${type})`, async () => {
+        const res = await fetchType(`${BASE_URL}/${fromUrl}`, type);
+        expect(res.status).toEqual(307);
+        expect(res.headers.get('location')).toEqual(`${BASE_URL}/${toUrl}`);
+      })
+    })
+
+
+    /* test Charon API */
+    const charonUrl = (path, sidecar) =>
+      `${BASE_URL}/charon/getDataset?prefix=${path}${sidecar ? `&type=root-sequence` : ''}`;
+
+    /* note that the redirect URL of charon requests is constructed by creating
+    a URL object (the RESTful API does not) which results in '/' and '@' being
+    escaped in the query. This isn't necessary, but it's also fine to do so. */
+
+    it(`Charon main dataset: ${fromUrl} goes to ${toUrl} `, async () => {
+      const res = await fetch(charonUrl(fromUrl, false), {redirect: 'manual'});
+      expect(res.status).toEqual(307);
+      expect(decodeURIComponent(res.headers.get('location'))).toEqual(charonUrl(toUrl, false));
+    })
+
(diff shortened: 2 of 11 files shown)


resourceIndexer/core-url-remapping.js
```diff
@@ -0,0 +1,49 @@
...
+ * resolve the entire path for every partial path, for instance
+ *      flu → flu/seasonal/h3n2/h2/2y
+ *      flu/seasonal → flu/seasonal/h3n2/h2/2y
+ * @param {Object} defaults 
+ * @returns {Object}
+ */
+function resolveAll(defaults) {
+  const urlMap = {};
+  const defaultKeys = new Set(Object.keys(defaults));
+  Object.entries(defaults).forEach(([partialPath, nextPathPart]) => {
```
tsibley, line 40: _non-blocking_

 `.forEach()` with side-effects is always odd to me...

```suggestion
  for (const [partialPath, nextPathPart] of Object.entries(defaults)) {
```

would be what I think is the more typical way to write this imperative code.

resourceIndexer/core-url-remapping.js
```diff
@@ -0,0 +1,49 @@
...
+    if (partialPath==='') return; // manifest has a base default of zika!
+    let resolvedPath = `${partialPath}/${nextPathPart}`;
+    while (defaultKeys.has(resolvedPath)) {
+      resolvedPath += `/${defaults[resolvedPath]}`;
+    }
```
tsibley, line 45: I see this sort of dual variable separation of the keys (in `defaultKeys`) from the actual values (in `defaults`)—where you check variable A but then access variable B—to be an anti-pattern. If what you're reaching for is the `.has()` method, you can use a Map instead and keep the data in a single variable, e.g.:

```js
const defaultPaths = new Map(Object.entries(defaults));
// …
  while (defaultPaths.has(resolvedPath)) {
    resolvedPath += `/${defaultPaths.get(resolvedPath)}`;
  }
```

You can also directly iterate over a Map, so combined with my `for (…) {` suggestion above, you'd do:

```js
for (const [partialPath, nextPathPart] of defaultPaths) {
```

tsibley (changes requested):

src/endpoints/sources.js
```diff
@@ -54,10 +55,12 @@ export const canonicalizeDataset = (pathBuilder) => (req, res, next) => {
 
   if (dataset === resolvedDataset) return next();
 
-  const canonicalPath = pathBuilder.length >= 2
+  let canonicalPath = pathBuilder.length >= 2
     ? pathBuilder(req, resolvedDataset.pathParts.join("/"))
     : pathBuilder(resolvedDataset.pathParts.join("/"));
 
+  if (dataset.versionDescriptor) canonicalPath+=`@${dataset.versionDescriptor}`;
```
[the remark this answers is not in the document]
jameshadfield: Fixed up & tests added.

resourceIndexer/core-url-remapping.js
```diff
@@ -0,0 +1,49 @@
...
+ * @returns {string}
+ */
+export const remapCoreUrl = (path) => {
+  if (!resolvedPath) {
+    /* map only computed the first time this function is run */
+    const serverManifest = JSON.parse(readFileSync("./data/manifest_core.json"));
+    resolvedPath = resolveAll(
+      convertManifestJsonToAvailableDatasetList(serverManifest).defaults
+    )
+  }
```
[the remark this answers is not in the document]
jameshadfield: Refactored the server code to avoid this behaviour. 

> If we leave the file-reading code here, then the path should be made portable

Done.

resourceIndexer/core-url-remapping.js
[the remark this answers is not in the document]
jameshadfield: Done.

resourceIndexer/core-url-remapping.js
jameshadfield: Ok. Changed to checking membership of `defaults` without maintaining a parallel `defaultKeys`.

resourceIndexer/core-url-remapping.js
jameshadfield: Thanks. `.forEach()` without side-effects would seem odd to me - it would be doing nothing! I'll keep this advice in mind for future work.

tsibley: Right. It can sometimes be convenient in an interactive context for inspection, but outside a REPL I find there's generally ~no upside to the indirection of the `.forEach()` methods (only downside) vs. the typical imperative `for` loops.

test/date_descriptor.test.js
```diff
@@ -301,6 +301,54 @@ describe("Valid narratives", () => {
...
+      `${BASE_URL}/charon/getDataset?prefix=${path}${sidecar ? `&type=root-sequence` : ''}`;
+
+    /* note that the redirect URL of charon requests is constructed by creating
+    a URL object (the RESTful API does not) which results in '/' and '@' being
+    escaped in the query. This isn't necessary, but it's also fine to do so. */
+
+    it(`Charon main dataset: ${fromUrl} goes to ${toUrl} `, async () => {
+      const res = await fetch(charonUrl(fromUrl, false), {redirect: 'manual'});
+      expect(res.status).toEqual(307);
+      expect(decodeURIComponent(res.headers.get('location'))).toEqual(charonUrl(toUrl, false));
```
tsibley, line 338: _non-blocking_

These are only tests, so non-blocking, but this is the wrong way 'round in terms of URL encoding/decoding. The Location header is a URL, not a single URL component, so the more appropriate function is `decodeURI` not `decodeURIComponent`. Instead of decoding the Location URL at all though, the right way 'round fix is making `charonUrl` properly encode the URL it generates.

tsibley (approved):

Issue #792, opened 2024-02-06
Preserve queries across (dataset resolved) redirects


jameshadfield:
We redirect URL paths via dataset-dependent express middleware `canonicalizeDataset` however this does not preserve URL queries.

As an example, https://nextstrain.org/flu/seasonal?c=cTiter redirects to https://nextstrain.org/flu/seasonal/h3n2/ha/2y (the query is discarded).

This should be easy to fix - the following proof of principle patch works but needs a bit more testing:

```diff
--- a/src/endpoints/sources.js
+++ b/src/endpoints/sources.js
@@ -64,7 +64,7 @@ export const canonicalizeDataset = (pathBuilder) => (req, res, next) => {
   /* 307 Temporary Redirect preserves request method, unlike 302 Found, which
    * is important since this middleware function may be used in non-GET routes.
    */
-  return res.redirect(307, canonicalPath);
+  return res.redirect(307, url.format({pathname: canonicalPath, query: req.query}));
 };
 ```

PR #791, opened 2024-02-01
[resource indexer] include dataset redirects in index

jameshadfield:
Recent work surfaced previous dataset versions across URL redirects¹ by
mirroring the dataset-name resolution process we use for requests on the
server. However it neglected to consider the redirects which are handled
prior to this in the server. This commit adds that functionality as
well. This situation was recently discussed in slack².

To use mpox as an example: the dataset (URL) path "mpox/all-clades" now includes previous versions which were named "monkeypox_mpxv.json", thus extending the snapshot history of this dataset from 2023-09-23 to 2022-06-12.

Our usage of `ncov_global.json` (similarly for other regions) is a lot more complex, because we didn't make clean dataset name switches like monkeypox/mpox.

Looking at the current live index (i.e. before this PR) the `ncov.json` dataset stops being uploaded 2020-04-23 and then starts being uploaded again on 2020-09-03, leaving a large gap in the snapshot history. When we do consider `ncov_global.json` (this PR), we fill in this gap with 99 snapshots. Great!

However `ncov.json` continued to be uploaded through 2021-02-15. In cases such as this the indexer will pick the last uploaded version in a given day. However looking at the data it's clear `ncov.json` was not being rebuilt, simply re-uploaded. For instance, here are nextstrain.org URLs you can view the data in:

* 2020-09-03: [ncov.json](https://nextstrain.org/fetch/nextstrain-data.s3.amazonaws.com/ncov.json%3FversionId=U74n6nPfuAYtFoJZDy5mTffmRL9k.91f), [ncov_global.json](https://nextstrain.org/fetch/nextstrain-data.s3.amazonaws.com/ncov_global.json%3FversionId=y9aBoREy41se0iYGH1.y7i7BtSoGleN.)

* 2021-02-15: [ncov.json](https://nextstrain.org/fetch/nextstrain-data.s3.amazonaws.com/ncov.json%3FversionId=WiHRklbcFMB.WpgFMh2PG7g69PK3_Xcl), [ncov_global.json](https://nextstrain.org/fetch/nextstrain-data.s3.amazonaws.com/ncov_global.json%3FversionId=ANBYqAHGJWpFuzusK2gD6e5X_VdTZF6B)

Looking at this data it's clear that we should drop any `ncov.json` datasets after 2020-04-23, which ~I'll add to this PR now~ _update: no need to programmatically drop, see next message in this PR_

Closes #784

¹ <https://github.com/nextstrain/nextstrain.org/pull/783>
[...]

PR #793, opened 2024-02-07
Preserve URL queries across redirects

jameshadfield:
Closes https://github.com/nextstrain/nextstrain.org/issues/792

Old behaviour: flu/seasonal/h3n2?c=cTiter → flu/seasonal/h3n2/h1/2y
New behaviour: flu/seasonal/h3n2?c=cTiter → flu/seasonal/h3n2/h1/2y?c=cTiter

### DIFF
--- a/src/endpoints/sources.js
+++ b/src/endpoints/sources.js
@@ -57,9 +57,7 @@ export const canonicalizeDataset = (pathBuilder) => (req, res, next) => {
 
   const version = dataset.versionDescriptor ? `@${dataset.versionDescriptor}` : '';
 
-  const canonicalPath = pathBuilder.length >= 2
-    ? pathBuilder(req, resolvedDataset.pathParts.join("/") + version)
-    : pathBuilder(resolvedDataset.pathParts.join("/") + version);
+  const canonicalPath = pathBuilder(req, resolvedDataset.pathParts.join("/") + version);
 
   /* 307 Temporary Redirect preserves request method, unlike 302 Found, which
    * is important since this middleware function may be used in non-GET routes.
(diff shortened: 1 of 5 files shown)


src/routing/core.js
```diff
@@ -52,7 +54,10 @@ export function setup(app) {
   app.use([coreBuildRoutes, "/narratives/*"], setSource(req => new CoreSource())); // eslint-disable-line no-unused-vars
 
   app.routeAsync(coreBuildRoutes)
-    .all(setDataset(req => req.path), canonicalizeDataset(path => `/${path}`))
+    .all(
+      setDataset(req => req.path),
+      canonicalizeDataset((req, path) => `/${path}` + (Object.keys(req.query).length ? `?${querystring.stringify(req.query)}` : ''))
```
tsibley, line 59: _non-blocking_

I think something like the following reads cleaner:

```suggestion
      canonicalizeDataset((req, path) => url.format({pathname: `/${path}`, query: req.query}))
```

and it should have the same behaviour.

(Ideally we'd be using the modern `URL` and `URLSearchParams` interfaces like we do elsewhere, but those are a bit more verbose in this situation without helper funcs. Also, ideally we'd be preserving the exact query string as-is without roundtripping thru a deserialize/serialize transform.)

tsibley, line 59: Separately, the use of `req.query` here with `querystring.stringify` brings up an asymmetry we should fix.¹ `req.query` is produced by parsing the request URL's query string with [`qs`](https://www.npmjs.com/package/qs) since don't set Express' `query parser` option. The parsing behaviour of `qs` is a bit different from `querystring` and leads to not roundtripping, e.g.

```console
> qs.parse("a[b]=c")
{ a: { b: 'c' } }
> querystring.stringify(qs.parse("a[b]=c"))
'a='
```

The best solution is not roundtripping, but if we're going to roundtrip, we should be consistent. As `qs` has behaviour I don't think we want nor expect, I think best choice here is to add

    app.set("query parser", "simple"); // querystring.parse()

to our other app settings (like `json replacer`). This would let us roundtrip `req.query` with `querystring.stringify()` and thus also `url.format()` with peace of mind.

See the [`req.query` docs](https://expressjs.com/en/api.html#req.query) and the [`query parser` setting docs](https://expressjs.com/en/api.html#app.settings.table).

¹ This asymmetry was sort of pre-existing, with the use of `url.format({…, query: req.query})` in one spot, which you've also recently touched in #791.

tsibley (changes requested):
Good to preserve this!

I wonder about a slightly different interface for `canonicalizeDataset()` where it preserves the query by default (and thus only would need to be explicitly handled differently in `canonicalizeDatasetPrefix()`), but I think this interface is just fine for now.

src/routing/core.js
jameshadfield: Switched to parsing with `querystring` in [847282a](https://github.com/nextstrain/nextstrain.org/pull/793/commits/847282a19b386b82d7e8cb1532f4809caa32f680)

jameshadfield: Switched to `url.format` which nicely matches our other usage across the codebase.

src/endpoints/sources.js
```diff
@@ -57,9 +57,7 @@ export const canonicalizeDataset = (pathBuilder) => (req, res, next) => {
 
   const version = dataset.versionDescriptor ? `@${dataset.versionDescriptor}` : '';
 
-  const canonicalPath = pathBuilder.length >= 2
-    ? pathBuilder(req, resolvedDataset.pathParts.join("/") + version)
-    : pathBuilder(resolvedDataset.pathParts.join("/") + version);
+  const canonicalPath = pathBuilder(req, resolvedDataset.pathParts.join("/") + version);
```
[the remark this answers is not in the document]
jameshadfield: Yup. Done. I've also updated the type to indicate it returns a path + any query.

PR #797, opened 2024-02-12
endpoints/cli: Support PEP 440/508 version range constraints for downloads

tsibley:
This allows URLs like:

    https://nextstrain.org/cli/download/>=7.4.0/standalone-x86_64-unknown-linux-gnu.tar.gz

which will let us use the standalone installer, e.g.

    curl -fsSL --proto '=https' https://nextstrain.org/cli/installer/linux | bash -s '>=7.4.0'

in automated contexts where we want to be able to declare constraints like lower version bounds or incompatible versions.

Note that the standalone installer on macOS on aarch64 hardware does very rudimentary version comparison¹ to decide if the requested version is older or newer than the first release with actual aarch64 support (8.2.0), and it won't compare version range constraints correctly: they'll always be considered greater than (newer/later) than 8.2.0.  But I expect this to be fine in practice and not matter for actual usage.

We depend on an older version of @renovatebot/pep440 (<3) because >=3 adds a dependency on Node >=18 and this codebase is still on Node 16.  I reviewed the changelog for newer versions of the package and nothing else substantial seems to have changed anyhow.

Related-to: <https://github.com/nextstrain/.github/pull/55>

¹ <https://github.com/nextstrain/cli/pull/358>
  <https://github.com/nextstrain/cli/blob/af976b06/bin/standalone-installer-unix#L146-L154>

## Checklist

- [x] Checks pass


tsibley:
I'm going to merge this pre-review so I get a little chance to see how it works in production in practice. It'll also unblock https://github.com/nextstrain/.github/pull/55.

I definitely still welcome post-merge review! :pray:

victorlin (commented):
👍

Skimmed through code and tested:

```
$ curl \
  --location \
  --remote-name \
  --remote-header-name \
  --write-out "%{filename_effective}" \
  "https://nextstrain.org/cli/download/>=7.1.0,<8.0.0/standalone-x86_64-apple-darwin.tar.gz"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   138  100   138    0     0    217      0 --:--:-- --:--:-- --:--:--   219
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 37.1M  100 37.1M    0     0  16.4M      0  0:00:02  0:00:02 --:--:-- 33.7M
nextstrain-cli-7.4.0-standalone-x86_64-apple-darwin.tar.gz
```

Issue #794, opened 2024-02-07
[CI] False positive build failures


jameshadfield:
Our CI [monitors & reports the status of heroku builds for pushes to master](https://github.com/nextstrain/nextstrain.org/actions/runs/7821163186/workflow#L110-L129). Today's CI (triggered by the merge of #791) [reported a build failure](https://github.com/nextstrain/nextstrain.org/actions/runs/7821163186/job/21337674779), however the build succeeded ([heroku build log](https://dashboard.heroku.com/apps/nextstrain-canary/activity/builds/10967800-d10e-4612-b839-e340cb3ae077)).

The CI observed `"status": "pending"` which is not "succeeded" and thus the CI failed. I guess (?) this is caused by `curl` command returning before the build was actually finished. I haven't investigated further.

tsibley:
Yeah, I saw this the other day too. I think we're racing Heroku's API here and inserting a short delay should probably make it Just Work again.

(That said, I do still want to #485 which in many incarnations would render this race non-existent.)

PR #800, opened 2024-02-15
dev: Wait up to 60s for the Heroku build status to change to "succeeded"

tsibley:
It used to be that once the Heroku build log stream response completed, their API would return a build status of "succeeded".  Recently however it seems like now there's a delay between the log stream response completing and their API returning the up-to-date build status.  Build in a 60s buffer wait time to account for that.

Resolves <https://github.com/nextstrain/nextstrain.org/issues/794>.

## Checklist

- [x] Checks pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,16 +115,27 @@ jobs:
       - name: Check Heroku build
         run: |
           id="$(jq -r .id build.json)"
+          SECONDS=0
+          maxwait=60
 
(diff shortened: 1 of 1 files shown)


tsibley:
This worked in my local testing in isolation. I'll merge it and it will then test itself.

tsibley:
The [CI for merge](https://github.com/nextstrain/nextstrain.org/actions/runs/7922754889/attempts/1) wasn't a full test of the new behaviour, as the stochastic `"status": "pending"` wasn't present on the first check.

PR #801, opened 2024-02-21
upgrade auspice

jameshadfield:


victorlin:
(Noted on Slack) If anyone finds this looking for how to trigger this automatically, the instructions are in [dependabot.yml](https://github.com/nextstrain/nextstrain.org/blob/-/.github/dependabot.yml).

PR #810, opened 2024-04-01
Replace gatsby with next.js

jameshadfield:
This PR replaces the engine behind the frontend (client) pages within nextstrain.org from Gatsby (v2.32) to Next.JS (v14).

[Working review app](https://nextstrain-s-james-next-zggv34.herokuapp.com/)

The reasons for wanting to move away from Gatsby are many and have been discussed elsewhere. For me the greatest motivation was the high bar for development work within the static site which is (one of) the reasons the frontend site has seen little development recently. We currently have a situation where some people can only develop nextstrain.org by using the heroku review app to test changes as they cannot install/run the frontend locally.

People's ideal implementations for the frontend have been [discussed elsewhere](https://github.com/nextstrain/private/issues/88) but this PR takes a different approach: it keeps the overall structure and design and simply replaces the code which interacted with Gatsby to code which now interacts with Next.js. Importantly we are still using static (build-time) compilation of pages¹. If you would like to implement server-side rendering then this PR can be seen as a step along that path as Next.JS has extensive support for server rendering, however I don't think this is right direction². Similarly, improvements to styling, design, image optimization (etc) are all good ideas which were not in scope for this PR.

Development times and complexity have been significantly improved by this PR. Whilst there is undoubtedly some Next.JS specific concepts to learn, their [docs](https://nextjs.org/docs) are pretty good and I've tried to highlight relevant concepts in the added documentation here. [clades.nextstrain.org](https://clades.nextstrain.org) uses it moving more of our sites to a shared architecture will hopefully assist in wider development efforts. Compared to the difficulties in maintaining and improving our previous frontend it is night and day. While user performance was not a motivating factor for this PR, first-contentful paint and other metrics have improved.

While this PR touches many files most changes are trivial. Probably the biggest change is the code to generate the blog pages where we move away from Gatsby's GraphQL approach. Along the way there are quite a few bug fixes and removal of unused code and pages. I noticed many more rough edges and little bugs; hopefully with a more pleasant dev experience we can attend to these. 

There are still some outstanding tasks which I'll attend to when I get a chance this week:

- [x] Fix failing tests (they are asserting 404 responses match the old Gatsby 404 page)
- [x] Test interaction with groups app (hopefully with help from @victorlin)
- [x] Re-enable linting in the static-site
- [x] Re-enable the map on /sars-cov-2
- [x] Re-enable the table listing datasets/narratives (whilst my resources UI will replace some of them shortly, the tables will remain in some form on some pages for a while at least)
- [x] Re-enable plausible analytics

¹ There are some differences, with our implementation of Next.JS using a more minimal entrypoint HTML which allows reuse and caching of shared components. But the server is not rendering pages in production mode.

² There are some exceptions such as rendering error pages with context about the actual error, however this PR specifically avoids changing our existing approach of generating pages at build time.

---

Closes #698


static-site/package.json
```diff
@@ -1,79 +0,0 @@
...
-    "gatsby-plugin-offline": "^3.0.31",
-    "gatsby-plugin-plausible": "0.0.7",
-    "gatsby-plugin-react-helmet": "^3.1.21",
-    "gatsby-plugin-sharp": "^2.3.13",
-    "gatsby-plugin-sitemap": "^2.2.25",
-    "gatsby-plugin-styled-components": "^3.1.17",
-    "gatsby-plugin-twitter": "^2.1.18",
-    "gatsby-remark-autolink-headers": "^2.1.24",
-    "gatsby-remark-copy-linked-files": "^2.1.36",
-    "gatsby-remark-images": "^3.1.42",
```
victorlin, line 36: James mentioned this on a call and indirectly in the PR description, but it's worth noting explicitly that image optimization is removed (previously done by `gatsby-remark-images`). Most notable on pages such as [/blog/2021-07-08-ncov-open-announcement](https://next.nextstrain.org/blog/2021-07-08-ncov-open-announcement) where the size of the image served changes from 32kB to 1.8MB.

It doesn't seem like much of the rest of the site uses large images so this is a minor thing that we can live with.

static-site/src/sections/sars-cov-2-page.jsx
```diff
@@ -22,9 +21,6 @@ import GenericPage from "../layouts/generic-page";
 import { ErrorBanner } from "../components/splash/errorMessages";
 import { withRouter } from 'next/router'
 
-const nextstrainLogoPNG = "/favicon.png";
```
victorlin, line 21: Suggestion: remove the favicon.png file along with this change. I don't see any other references.

victorlin (commented):
Reviewed most changes. There's still some that I'll want to look more into such as the interaction between `PRODUCTION` and `NEXT_JS_PRODUCTION`.

Most comments are small nitpicks. I think the overall direction is good. It's not an ideal setup between Next.js and Express.js, but that can be improved later.

static-site/package.json
jameshadfield: > Most notable on pages such as [/blog/2021-07-08-ncov-open-announcement](https://next.nextstrain.org/blog/2021-07-08-ncov-open-announcement)

I think the blog's the _only_ place -- everywhere else we were (and still are) directly importing / requiring the image so Gatsby wasn't optimising it. I'll double check this and report back.

(P.S. this was included in the commit message for [Move blog to next.js](https://github.com/nextstrain/nextstrain.org/pull/810/commits/36a79a9b656de9be3c0753442363f05c305b043b))

static-site/pages/ncov/[[...ncov]].jsx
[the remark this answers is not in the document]
Comment (jameshadfield) [reply static-site/pages/ncov/[[...ncov]].jsx:1]: Yeah, exactly. There's one difference (I think) relating to the response code (404 vs 200) that we send. I'll dig a little more here tomorrow when I address the failing tests related to this.

P.S. This concept (an auspice-like route going to our gatsby/next.js frontend and displaying an error banner) exists for `/flu/*`, `/influenza/*`, `/ncov/*`, `/sars-cov-2/*` and `/staging/*`.

package.json
```diff
@@ -8,19 +8,22 @@
...
+    "lint:server": "DEBUG=eslint:cli-engine npx eslint --ext .js,.jsx .",
+    "lint:static-site": "cd static-site && DEBUG=eslint:cli-engine npx eslint --ext .js,.jsx .",
     "server": "node server.js",
     "groups": "node server.js --app ./src/groupsApp.js",
     "start": "npm run server",
     "//test": "echo 'NODE_OPTIONS is required until this issue is resolved: https://github.com/facebook/jest/issues/9430 '",
     "test": "NODE_OPTIONS='--experimental-vm-modules' jest --roots ./test/",
     "test:ci": "start-server-and-test 'RESOURCE_INDEX=test/date_descriptor_index.json node server.js --verbose >./test/server.log 2>&1' http://localhost:5000 test",
-    "dev": "./develop.sh"
+    "dev": "NODE_ENV=development ./node_modules/.bin/nodemon --verbose -- server.js --verbose \"$@\"",
```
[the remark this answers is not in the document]
jameshadfield: Yup - forgot about this when I copied over the code from `develop.sh`. Changed (as part of a force push which included a rebase onto master to pick up some recently merged PRs)

static-site/src/styles/globals.css
```diff
@@ -9,19 +9,7 @@ html {
     box-sizing: border-box;
     margin: 0;
     padding: 0;
-    font-family: 'Lato', 'Helvetica Neue', sans-serif;
+    /* We use nextjs' next/font to manage our primary font (Lato). `var(--lato)` is defined in pages/_document.tsx */
```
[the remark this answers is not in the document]
jameshadfield: Thanks!  Changed (as part of a force push which included a rebase onto master to pick up some recently merged PRs)

static-site/src/sections/sars-cov-2-page.jsx
jameshadfield: Thanks for picking this up. I've decided to move the favicons from the gatsby public folder (`./static-site/static`) to the next.js public folder (`./static-site/public`) which allows us to drop the `serve-favicon` dependency I originally added to this PR and use the implicit approach we were using with Gatsby.

README.md
```diff
@@ -73,12 +71,17 @@ For running locally, you should ensure
   Alternatively, these may be configured using `~/.aws/credentials`.
   If you add a new profile to `~/.aws/credentials`, you can then tell the server to use it by setting `AWS_PROFILE=...`.
 
+If you have built the next.js part of the site (`./static-site`), via `npx next build static-site` or similar, then setting `NEXT_JS_PRODUCTION=1` will use these built assets rather than server-side compilation.
```
jameshadfield, line 74: @victorlin - adding a thread here as you wanted to look into `NEXT_JS_PRODUCTION` - this variable is only in play for local development (it allows you to use the next.js statically generated assets). When `NODE_ENV=production`, e.g. the heroku review apps, this variable isn't used. We could avoid this if it were simple to run the server in production mode locally, but I find that near impossible (ssl redirect, session secret keys, redis keys etc etc); and I didn't want to complicate this PR with making it easier to run in production locally.

static-site/package.json
jameshadfield: Comparing the [140kb ncov splash tile image](https://github.com/nextstrain/nextstrain.org/blob/master/static-site/static/splash_images/ncov.png), both Gatsby + next.js are currently sending the unoptimised image for any screen resolution. This is because we are using `require` rather than just using a URL path for ``, an approach we use throughout the site:

https://github.com/nextstrain/nextstrain.org/blob/f1106c1f438ddcc836b1421d2387d7d0c44e28f8/static-site/src/components/Cards/index.jsx#L22

(I'm not saying this is the best way, but it's the way it's been for a long time and I don't think this PR is the place to change it.)

For the blog we do use a URL path for the `` tags (somewhat indirectly, it's part of the HTML generated from parsing the markdown). So to use the blog post in your example, it links to the 1.7Mb (!) `gisaid_vs_open_map.png`; when viewed on mobile Gatsby serves us a 32kb file whereas on next.js we get the full sized image. 

I think image optimisation is important for us to implement, and the images in the blog are a regression in this PR (although the post you picked is one of the worst!), but I'm hoping it's acceptable to leave this for future work.

static-site/src/styles/globals.css
```diff
@@ -9,19 +9,7 @@ html {
...
+    font-family: var(--lato), 'Helvetica Neue', sans-serif;
```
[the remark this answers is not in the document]
jameshadfield: Nice find! I _think_ this is because we're using `font-family: lato` not `font-family: var(--lato)`. I'll get to this tomorrow.

static-site/package.json
victorlin: Thanks for the clarification!

> although the post you picked is one of the worst

Yes, that was specially hand-picked 😅 I think totally fine to leave this for later.

static-site/src/sections/sars-cov-2-page.jsx
victorlin: Looks like it's still unused though? At least on `sars-cov-2-page.jsx`, the reference changed from `favicon.png` to `nextstrain-logo-small.png`.

static-site/src/styles/globals.css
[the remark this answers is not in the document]
victorlin: Thanks! Relatedly, I also noticed that `var(--lato)` renders a different weight variant in a few places, like the blog sidebar:

diff

static-site/src/sections/sars-cov-2-page.jsx
jameshadfield: It's unused in our code, but browsers will (unless instructed otherwise) access it to use as a favicon. This [post](https://webmasters.stackexchange.com/questions/123109/can-a-favicon-at-the-root-path-be-in-a-different-image-format-than-ico) is a good summary, and you can see it in the network requests too:

image

static-site/src/styles/globals.css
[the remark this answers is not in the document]
jameshadfield: I dug into this a bunch and there are some aspects of fonts I still don't understand but I am happy with the current state.

> Somehow I'm not getting Lato in a few places on the preview build:

This was fixed by using `var(--lato)` in our styled-components theme. Added by fixing up 
[Move static-site to use Next.JS](https://github.com/nextstrain/nextstrain.org/pull/810/commits/0a56fd9d94e2449fb4652bc75c9cdde0281a77d9)

> Relatedly, I also noticed that var(--lato) renders a different weight variant in a few places, like the blog sidebar:

I've written a bit about what font-weights were asking for vs what weights are being used for rendering in [Remove unused font weight 100](https://github.com/nextstrain/nextstrain.org/pull/810/commits/c4182b4af4d5e6fbd33963f4d19e9a43b5e725f7) but I still can't tell you _why_ the blog sidebar entries were rendered as `font-weight: 100` in your screenshot. They appeared like that for me too, but then I refreshed and they changed to 300: 
image

Interestingly, there are two `<h6>` styles which apply to the element in question:

https://github.com/nextstrain/nextstrain.org/blob/6eb8f3a3d3906d4b96e96e35878578810f6eacbe/static-site/src/components/Sidebar/index.jsx#L110-L112

https://github.com/nextstrain/nextstrain.org/blob/6eb8f3a3d3906d4b96e96e35878578810f6eacbe/static-site/src/components/layout.jsx#L67

In your screenshot your browser is using the former, while for me my browser is using the latter as higher priority. Weird. 

Anyways, I've removed the request for 100 (aka "hairline", aka "thin") from the codebase, as it's too thin to be useful anywhere and we never explicitly request it. So the situation you encountered shouldn't happen again. (But can you still test for me?!)

package.json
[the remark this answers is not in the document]
jameshadfield: (this got re-applied somewhere along the line, but It'll be fixed on the next force-push)

package.json
```diff
@@ -8,19 +8,22 @@
...
+    "lint:server": "DEBUG=eslint:cli-engine npx eslint --ext .js,.jsx .",
+    "lint:static-site": "cd static-site && DEBUG=eslint:cli-engine npx eslint --ext .js,.jsx .",
     "server": "node server.js",
     "groups": "node server.js --app ./src/groupsApp.js",
     "start": "npm run server",
     "//test": "echo 'NODE_OPTIONS is required until this issue is resolved: https://github.com/facebook/jest/issues/9430 '",
     "test": "NODE_OPTIONS='--experimental-vm-modules' jest --roots ./test/",
-    "test:ci": "start-server-and-test 'RESOURCE_INDEX=test/date_descriptor_index.json node server.js --verbose >./test/server.log 2>&1' http://localhost:5000 test",
-    "dev": "./develop.sh"
+    "test:ci": "next build static-site && start-server-and-test 'NEXT_JS_PRODUCTION=1 RESOURCE_INDEX=test/date_descriptor_index.json node server.js --verbose >./test/server.log 2>&1' http://localhost:5000 test",
```
jameshadfield, line 24: The (GitHub actions) test `test/cryptography.test.js` has been consistently failing. I'll do some digging as obviously this is blocking. I don't think anything in this PR is at all related to the code paths it uses, except for the maybe the node upgrade, but I cannot recreated this failure locally (and it's not dependent on any AWS environment variables as far as I can tell).

static-site/src/styles/globals.css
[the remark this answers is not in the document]
victorlin: Wow, nice digging! 

I'm not sure why my browser picks 200 and yours 300. They have the same specificity of parent class + h6. With static CSS files, it would be resolved by order of appearance. styled-components is different with injection at runtime, and there doesn't seem to be explanation on what's resolved with equal specificity. All I can find is an [FAQ entry](https://styled-components.com/docs/faqs#how-can-i-override-styles-with-higher-specificity) on their alternative to `!important` which doesn't explain our browsers' difference in behavior.

Maybe "nearby weight" picks the lower one when there are equidistant ties? I.e. since mine was using 200, it picked 100 instead of 300?

```
100, 200 -> rendered using 100 (lato-hairline)
300      -> rendered using 300 (lato-light)
```

Anyways, it uses 300 now that you've removed 100, so we're all good here.

static-site/src/sections/sars-cov-2-page.jsx
victorlin: Are you confusing `favicon.png` with `favicon.ico`? I see that the ICO is used by default unless instructed otherwise to use PNG. I only get a network request to PNG after adding the line in the post you linked using 203440f334e2c1c3336b1942528da67a8d777e01.

package.json
jameshadfield: `cryptography.test.js` passes on nodejs 20.9.0 (which is what I was developing on, as it's the latest available in conda) but consistently failed when [run on 20.12.0](https://github.com/nextstrain/nextstrain.org/actions/runs/8566685893/job/23476917786#step:11:257) and [20.11.1](https://github.com/nextstrain/nextstrain.org/actions/runs/8564336483/job/23470717176) in GitHub Actions. Pinning GitHub actions to 20.9.0 allowed the [test to pass](https://github.com/nextstrain/nextstrain.org/actions/runs/8575901351/job/23505846732).

For the purposes of this PR pinning to 20.9.0 is fine and I'll create a separate issue to track this bug.

Separately, while debugging this I identified that the version we've elected to use for CI (`actions/setup-node`) is slightly different from the version Heroku will use (via `package.json`). I'll either fix this here or create a separate issue.

static-site/src/styles/globals.css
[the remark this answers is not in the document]
jameshadfield: As an aside, I'm all for removing styled components from the frontend where possible. Next.js supports CSS modules out of the box, and I use them in one or two places in this PR, as typical CSS imports aren't allowed except for at the top level. There are also other styling frameworks we can reach for. But to keep this PR limited in scope I explicitly wanted to modify the styling as little as possible.

package.json
jameshadfield: > I'll either fix this here or create a separate issue.

I've done this, and as expected GitHub actions is using 20.9.0 and [tests pass](https://github.com/nextstrain/nextstrain.org/actions/runs/8583144975/job/23522029992). I'll double check the Heroku review app is also pulling in 20.9.0, but I'd be very surprised if it's not.

jameshadfield: > I'll double check the Heroku review app is also pulling in 20.9.0, but I'd be very surprised if it's not.

Yes, it is.

static-site/src/sections/sars-cov-2-page.jsx
jameshadfield: > unless instructed otherwise to use PNG

I thought some browsers accessed PNG by default but looking into it I think that's not correct. I'll remove the PNG version now.

jameshadfield: [Removed](https://github.com/nextstrain/nextstrain.org/compare/e54146066179d2a6a52a8a554c5310ca99b89802..9668e99897fadc9d8f2a761bc0229755f1deba40)

package.json
jameshadfield: https://github.com/nextstrain/nextstrain.org/issues/814

README.md
jameshadfield: @victorlin are you still looking into this part? Happy to jump on a call and answer any questions if so

victorlin: I just tried the npm scripts locally and understand better now.

Having a 2nd "production" flag was a bit confusing to me. How about renaming to something that describes the behavior more literally, such as `USE_PREBUILT_STATIC_SITE`?

There's a slight change in that `server.js`, `npm run server`, and `npm run groups` will no longer use prebuilt static-site files by default. I think this preferable with static-site effectively adhering to the values `NODE_ENV=development` vs. `NODE_ENV=production` unless this new flag is set.

victorlin (commented):
I haven't reviewed latest commits in detail but clicked around the review app - things look good except one minor bug.

README.md
jameshadfield: 👍 [Updated](https://github.com/nextstrain/nextstrain.org/compare/a237a3fe289d1df150a88b3f3c005c55c4f54c51..c078420a277bc15c78aacd8fd04c52fda9c79293) to use `USE_PREBUILT_STATIC_SITE`

> There's a slight change in that server.js, npm run server, and npm run groups will no longer use prebuilt static-site files by default. 

Right, but since [node.js runs in "development" mode by default](https://nodejs.org/en/learn/getting-started/nodejs-the-difference-between-development-and-production#nodejs-the-difference-between-development-and-production) I'd argue that this is more in line with expectations and the previous situation was strange. I do think we should make it easier to run in "production" mode locally (not with the same secrets we actually use in production of course!).

victorlin: Agreed - sorry if it wasn't clear in the last sentence. Thanks for the rename!

One more small request related to this: could you move this snippet from the README into infrastructure.md's [Environment (or config) variables section](https://docs.nextstrain.org/projects/nextstrain-dot-org/en/latest/infrastructure.html#environment-or-config-variables)? I realize that `GATSBY_DEV_URL` didn't have an entry but I think it should've and this new env var should too.

https://github.com/nextstrain/nextstrain.org/blob/c078420a277bc15c78aacd8fd04c52fda9c79293/README.md?plain=1#L74

static-site/pages/blog.jsx
```diff
@@ -0,0 +1,32 @@
...
+ * post. Note that we could achieve the same end by adding redirects to
+ * `next.config.mjs` but we run into issues (performance & technical) when
+ * running getBlogPosts there.
+ */
+export default function Index({redirectTo}) {
+  const router = useRouter();
+  useEffect(() => {
+    if (redirectTo) router.push(redirectTo);
+  }, [redirectTo, router])
+  return null;
```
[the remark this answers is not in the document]
jameshadfield: Good find! Fixed (in forthcoming force push) by using `router.replace`. (The buggy behaviour was because we would redirect by _adding_ a new page onto the navigation stack so when you go "back" you then hit the same code which adds the page you were just on.)

This also occurred in the groups app (`/` to `/groups`) which is harder to test but has also been fixed.

README.md
jameshadfield: Totally! [Done as part of this force-push](https://github.com/nextstrain/nextstrain.org/compare/c078420a277bc15c78aacd8fd04c52fda9c79293..c673b1be652297ab0c112c755f8ba140569c5214)

.github/workflows/ci.yml
```diff
@@ -6,31 +6,21 @@ on:
...
     if: github.repository == 'nextstrain/nextstrain.org'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
         with:
-          node-version: '16'
+          node-version-file: 'package.json'
       - run: node --version
       - run: npm ci
```
jameshadfield, line 18: [not related to this line, just using it to start a thread]

We are getting a lot of transient test failures (as well as some legitimate ones!). Currently the failure is related to tests which hit GitHub endpoints (and unrelated to any changes in this PR):

```console
FAIL test/get_available.test.js
  ● getAvailable community URLs correctly interpret the GitHub branch › Explicit non-default branch

    expect(received).toEqual(expected) // deep equality

    Expected: 1
    Received: 0

      44 |       }
      45 |       const branches = new Set(requests.map((r) => parseBranchFromAddress(r)));
    > 46 |       expect(branches.size).toEqual(1);
         |                             ^
      47 |       expect(branches).toContain(testCase.branch);
      48 |     });
      49 |   }

      at Object.<anonymous> (test/get_available.test.js:46:29)
```

```console
FAIL test/auspice_client_requests.test.js
  ● smoke testing URLs described in auspice_client_requests.json › Check getSourceInfo API for a community (GitHub) URL

    expect(received).toEqual(expected) // deep equality

    Expected: 200
    Received: 404

      13 |       const res = await fetch(url, {redirect: 'manual'});
      14 |
    > 15 |       expect(res.status).toEqual(testCase.expectStatusCode || 200);
         |                          ^
      16 |
      17 |       if (testCase.responseIsJson) await testResponseIsJson(res);
      18 |

      at Object.<anonymous> (test/auspice_client_requests.test.js:15:26)

```

My guess (haven't looked into it) is that tests are running without any GitHub credentials and we're running up against rate limits. It's not the remit of this PR, but going forward we should try hard to make our test suites less prone to stochastic failures like these.

victorlin: > My guess (haven't looked into it) is that tests are running without any GitHub credentials and we're running up against rate limits.

Seems likely. Rate limit for unauthenticated requests is [60 per hour per IP](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-unauthenticated-users). Searching for `prefix=/community` in the code shows that there's 12 requests per test run, so 5 back-to-back test runs will do it.

I wonder if it's as easy as using the [default `GITHUB_TOKEN`](https://docs.github.com/en/actions/security-guides/automatic-token-authentication):

```yaml
- run: npm run test:ci
  env:
    GITHUB_TOKEN: ${{ github.token }}
```

README.md
```diff
@@ -41,7 +41,7 @@ from this directory (the "nextstrain.org" directory).
 
 
 ### 3. Build the site
-`npm run build` runs `./build.sh` to build both the static site & an auspice client with customisations.
+`npm run build` runs `./build.sh` to build both the static site & an Auspice client with customisations.
 The following section details the different ways to serve these pages on a local server after building the site.
 
 ### 4. Run server
```
victorlin, line 47: Noting possibly the biggest dev change here: "server" now includes the static site and can't run without it.

Previously, `npm run server` would run just the server with Express.js routes, whereas now it will run along with a hot-reloading version of the static site. If you try to bypass that with `USE_PREBUILT_STATIC_SITE=1` and haven't built the static site, you'll get an error:

```
$ USE_PREBUILT_STATIC_SITE=1 node server.js
…
node:internal/process/promises:289
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[Error: ENOENT: no such file or directory, open '…/static-site/.next/BUILD_ID'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '…/static-site/.next/BUILD_ID'
}
```

I'm all good with the change, just wanted to note it. My guess for why the separation had existed with Gatsby is because building that version of the static site was somewhat slow, didn't allow hot-reloading, and needed workarounds on some machines. The Next.js build seems much faster and robust.

victorlin (approved):
Nice work - excited to no longer need to use `NODE_OPTIONS=--openssl-legacy-provider` when building the static site locally.

.github/workflows/ci.yml
jameshadfield: > Searching for prefix=/community in the code shows that there's 12 requests per test run

If you look at the server logs after a single test run (e.g. `npm run test:ci` then look at `test/server.log`) we make 127 GET requests to https://api.github.com, but some of them are cache hits and so (maybe?) aren't counted in the limits.

victorlin: [Looks promising](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-github_token-in-github-actions):

> You can use the built-in `GITHUB_TOKEN` to authenticate requests in GitHub Actions workflows.
> The rate limit for `GITHUB_TOKEN` is 1,000 requests per hour per repository

I'll make a separate PR.

victorlin: Oh I haven't looked at server.log, that's useful.

I see 85 GET requests. Without any `GITHUB_TOKEN`, I ran `npm run test:ci` 4 times in a row successfully, then the amount of failures varied:

1. all passed
2. all passed
3. all passed
4. all passed
5. 4 failed
6. 16 failed
7. 4 failed
8. 4 failed

Not sure what's going on, but using the GitHub Actions provided token should be an improvement: #817

static-site/src/components/Sidebar/index.jsx
```diff
@@ -1,83 +1,38 @@
...
+      <div style={{paddingBottom: "30px"}}/>
+    </SidebarContainer>
+  );
+}
 
-class Sidebar extends React.Component {
-  generateItems(selectedSlugInfo) {
-    const [hasChapters, data] = structureEdges.parseEdges(this.props.allNodes, selectedSlugInfo.section);
-    const renderListOfPosts = (listOfPosts, chapterNameOfPost = undefined) => listOfPosts.map((post) => {
-      let highlightPost = selectedSlugInfo.title === post.title;
```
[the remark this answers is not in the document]
jameshadfield: It's GraphQL in the sense that Gatsby used that API language and here we're consuming a response which ultimately came from parsing the markdown files. But conceptually this part is assuming a structure to our blog posts which doesn't exist. Basically it's overkill for how we used it in practice, and the GraphQL aspect made it a minefield. The new approach is simpler both in terms of the concept (no chapters etc) and the external functions etc we call.

static-site/src/styles/globals.css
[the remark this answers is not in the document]
jameshadfield: Yeah, they seem really appealing (they're "just" CSS but scoped), but this was my first time using them and I only ported over one or two files. If we ever want to do true SSG they'll be a much nicer approach than styled components. I'll write more about this in a bit, perhaps on the official "move away from gatsby" issue.

README.md
jameshadfield: 👍 I've added in [6e97d3c](https://github.com/nextstrain/nextstrain.org/pull/810/commits/6e97d3cc3cf36aa3362c8629edc88b0957bfecba) which'll catch this (but only this) error and print a helpful message explaining that the static-site must be built.

jameshadfield:
Thanks for the review @victorlin! I'm going to merge this and create a few issues for improvements / future work we _could_ do. Next.js should make it possible to take the site in many different directions, some of which I think are sensible. The terminology is somewhat overwhelming but there's a rough heirarchy available for us if we want to do more here:
* build-time compiled JS bundles + client rendering. This is essentially what we are doing here, and what we did in gatsby 95% of the time (we had few pages with some areas of SSG in gatsby).
* True SSG + hydration. Removing `styled-components` is going to make life much easier for this. There may be an argument to invest in this for the splash page at least.
* SSR - depending on the caching used this can be very similar to SSG. Would make custom error pages possible.
* Server components.

For me functionality improvements are more important than these, and hopefully this PR makes such work easier for all of us.

static-site/src/components/Sidebar/index.jsx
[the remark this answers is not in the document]
victorlin: Yes, the new approach is much easier to understand!

static-site/src/sections/sars-cov-2-page.jsx
jameshadfield: We both missed that Auspice requests the PNG version. Unlucky! The file is restored in  https://github.com/nextstrain/nextstrain.org/pull/825

victorlin: Oops, thanks for catching!

src/routing/staticSite.js
```diff
@@ -1,27 +1,6 @@
...
-     * returns a 200 Ok for missing pages.
-     */
-    app.use(createProxyMiddleware(app.locals.gatsbyDevUrl));
-  } else {
-    app.use(endpoints.static.gatsbyAssets);
-  }
-}
+  app.routeAsync("*")
+    .getAsync(endpoints.nextJsApp.handleRequest);
+}
```
victorlin, line 6: In trying to understand this setup better, I searched for examples of Express servers using Next.js. I didn't find any, but found [this example](https://github.com/vercel/next.js/blob/e969d226999bb0fcb52ecc203b359f3715ff69bf/examples/custom-server-express/server.js) in the official repo of a Next.js server using Express (though there are some [unanswered questions](https://github.com/vercel/next.js/pull/24814#pullrequestreview-652292574)).

jameshadfield: One way to think about it is we have two largely independent layers:

* Layer 1 is our express server with defined routes and handlers.
* Layer 2 is our nextjs server which handles any route route not handled by layer 1.

In the previous implementation is was largely similar except Layer 2 was an [express static-file server](https://expressjs.com/en/starter/static-files.html) which would access the pre-built Gatsby assets. There are a few other differences for sure, including 404 handling¹. Let me know if you have specific questions!

¹ _One of the strangest parts of the previous set up was when the server would explicitly send a 404 page which would then use client-side routing and show the user another page altogether!_

victorlin: Thanks for the overview, that makes sense. I'm just wondering if there is any reason to switch the entrypoint from Express to Next.js (example linked above is set up like that).

jameshadfield: I tried that initially but switched to the current implementation for a few reasons. Firstly, for development mode using a nextjs server is very compelling (hot reloading etc), and using a similar approach in production seemed less prone to bugs. Secondly, for routes such as `/ncov/*` (where `ifDatasetExists` is false, so we aren't using Auspice), I wanted to move us away from the previous approach of sending a 404 page and relying on client-side routing to route us to the right page (there are other ways to solve this, but they felt brittle). Finally, using a next.js server opens the door to different forms of SSR should we want them, although we should be able to do true SSG without it.

Happy to explore this more if you find that it's clearer how the routing works under that model?

> example linked above is set up like that

The example is essentially the same as our current Layer 2. (Wasn't clear if you were claiming that the example was similar to our current approach or was like our previous approach.)

victorlin: > The example is essentially the same as our current Layer 2

Forgot to update here - had a quick call with @jameshadfield where he gave a good explanation. The combination of `const prepared = nextJsApp.prepare()` + `await prepared` in [src/endpoints/nextjs.js](https://github.com/nextstrain/nextstrain.org/blob/1b0df7748055bf8525d503e70c33a5204ef1c3e2/src/endpoints/nextjs.js) is functionally equivalent to `app.prepare().then(…` in the Next.js example.

PR #811, opened 2024-04-02
Avian flu restructure

lmoncla:
## Description of proposed changes

This pull request separates `avian-flu` from `seasonal-flu` and sets them to the be at the same hierarchy. 

Original structure: `flu/seasonal/` and `flu/avian` 
New structure: `seasonal-flu/` and `avian-flu/`

This was accomplished by editing 3 files: 
1. `data/manifest_core.json`: `seasonal-flu` and `avian-flu` now exist at the same header level. `h5nx` is set as the default for `avian-flu`. 
2. `redirects.js`: I added in redirects for the seasonal flu and avian flu urls. These should be double checked. 
3. `static-site/src/components/Cards/pathogenCards.js`: added in a pathogen card for Avian Influenza. 

## Checklist

I would like help with testing out these changes to make sure that all the necessary edits have been made, and that the changes work as intended. Specific things that should be double checked are the redirects, `manifest_core.json` structure, and that the splash card works as intended.

PR #812, opened 2024-04-03
Redirect with regex

joverlee521:

Use `path-to-regexp` for dataset redirects. This allows us to use regex patterns to match datasets and redirect them
to a new pattern without having to list out all of the existing paths.

Motivated by https://github.com/nextstrain/nextstrain.org/pull/811#issuecomment-2033012211

## Checklist

- [x] Checks pass


src/redirects.js
```diff
@@ -162,6 +162,12 @@ export function updateDatasetUrl(originalUrl) {
     ['/monkeypox/hmpxv1', '/mpox/clade-IIb'],
     ['/monkeypox/hmpxv1/big', '/mpox/lineage-B.1'],
     ["/monkeypox", "/mpox"],
+    /**
+     * We shifted avian-flu and seasonal-flu to top level to stop nesting them
+     * both under flu/ so that avian-flu is more discoverable
+     */
+    ['/flu/avian(.*)', '/avian-flu(.*)'],
+    ['/flu/seasonal(.*)', '/seasonal-flu(.*)'],
```
joverlee521, line 170: I'm including this mostly to demonstrate how we would redirect the flu paths.
I can drop it if we need to coordinate change with https://github.com/nextstrain/nextstrain.org/pull/811

src/redirects.js
```diff
@@ -128,6 +127,58 @@ export const setup = (app) => {
 
 };
 
+/**
+ * Checks if the provided originalUrl matches are any of our datasetRedirectPatterns
+ * If the there is match, then return the new URL for the dataset.
+ * If there is not match, then return the original URL.
+ *
+ * @param {string} originalUrl
```
jameshadfield, line 135: This docstring should be changed to indicate that the input must be the pathname only (as the calling code does at the moment) - the function will return or destroy queries depending on if there's a `:version` match.

src/redirects.js
jameshadfield: Yeah - nice to see how well these work! I'd say drop this commit from this PR and we can add it to a subsequent PR depending on the migration path we choose to take.

jameshadfield (changes requested):
Looks like a good solution - thanks @joverlee521. Left a few comments inline. I'd favour merging this soonish as it'll be necessary for #811.

src/redirects.js
joverlee521: Dropped https://github.com/nextstrain/nextstrain.org/commit/3e8319440a77b1ab7206f34c3fa7085546e33c21

src/redirects.js
```diff
@@ -128,6 +127,58 @@ export const setup = (app) => {
...
+     * e.g. /monkeypox/hmpxv1/2022-09-04
+     */
+    ['/monkeypox/mpxv', '/mpox/all-clades'],
+    ['/monkeypox/hmpxv1', '/mpox/clade-IIb'],
+    ['/monkeypox/hmpxv1/big', '/mpox/lineage-B.1'],
+    ["/monkeypox", "/mpox"],
+  ];
+
+  // Include an optional versioned dataset pattern for all redirect patterns
+  const versionDatasetPattern = ":version(@{1}.*)?"
```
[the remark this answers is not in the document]
joverlee521: Thanks! Updated in https://github.com/nextstrain/nextstrain.org/pull/812/commits/989b393ad1bca02ef53db6e420076fa801824911 to `:version(@.*)?`

src/redirects.js
```diff
@@ -128,6 +127,58 @@ export const setup = (app) => {
...
+
+  for (const [originalPattern, redirectPattern] of datasetRedirectPatterns) {
+    const urlMatch = match(`${originalPattern}${versionDatasetPattern}`);
```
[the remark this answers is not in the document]
joverlee521: Thanks! I need to get better at recognizing these patterns. 

Moved out of `updateDatasetUrl` in https://github.com/nextstrain/nextstrain.org/pull/812/commits/443e132a2e8958b03db8409e23aa72f8a8cd6d9c

src/redirects.js
joverlee521: Updated docstring in https://github.com/nextstrain/nextstrain.org/pull/812/commits/989b393ad1bca02ef53db6e420076fa801824911

(I also updated the param to `originalUrlPathname` to be clearer)

jameshadfield: 👍

src/redirects.js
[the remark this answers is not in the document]
jameshadfield: 👍

src/redirects.js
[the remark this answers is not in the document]
jameshadfield: Thanks!

jameshadfield (approved):

PR #813, opened 2024-04-04
Update `data/manifest_core.json` with `genome` and `N450` for measles

kimandrews:

Update `data/manifest_core.json` to visualize the new `genome` and `N450` trees for measles

## Checklist

- [ ] Checks pass

### DIFF
--- a/data/manifest_core.json
+++ b/data/manifest_core.json
@@ -230,7 +230,13 @@
           "default": "s"
         }
       },
-      "measles": "",
+      "measles": {
+        "resolution": {
+          "genome": "",
+          "N450": "",
+          "default": "genome"
+        }
+      },
       "mpox": {
         "resolution": {
           "all-clades": "",


jameshadfield (approved):

joverlee521:
@jameshadfield Just to confirm, after merging this manifest change, we should trigger the [index resources](https://github.com/nextstrain/nextstrain.org/actions/workflows/index-resources.yml) action to update the index? 

Would that automatically group past versioned `/measles` datasets with the new default `/measles/genome` on the pathogens page?

jameshadfield:
> after merging this manifest change, we should trigger the [index resources](https://github.com/nextstrain/nextstrain.org/actions/workflows/index-resources.yml) action to update the index?

> I think that should automatically group past versioned /measles datasets with the new default /measles/genome on the pathogens page?

We carried on this discussion [in Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1712269046319059) but the summary is that yes, the index will get out-of-sync with the different versions of the site (next.nextstrain.org, nextstrain.org) with changes such as these. We're not going to address this for mesles, but we will for the bigger flu to seasonal-flu + avian-flu changes.

PR #816, opened 2024-04-09
[auspice] Add link out modal

jameshadfield:
See https://github.com/nextstrain/auspice/pull/1763 for context.


jameshadfield:
I've dropped the auspice update from this PR, so the changes here should be a no-op. I plan to merge this (which will have no changes to the site) so that once a new Auspice release is made nextstrain.org can take advantage of the new link-out features.

PR #817, opened 2024-04-09
CI: Use GitHub Actions token for higher rate limit

victorlin:

Previously, the unauthenticated requests were limited to [60 per hour per IP address](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api#primary-rate-limit-for-unauthenticated-users) which are likely shared by other jobs using the same GitHub-hosted runners.

## Related issue(s)

Prompted by https://github.com/nextstrain/nextstrain.org/pull/810#discussion_r1556421873

## Checklist

- [x] Updated repo's [Actions settings](https://github.com/nextstrain/nextstrain.org/settings/actions) to use read-only permissions on `GITHUB_TOKEN`
- [x] Checks pass
- [x] Not testing the new rate limit - will monitor CI on other builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,11 @@ jobs:
           aws-secret-access-key: ${{ secrets.DEV_SERVER_AWS_SECRET_ACCESS_KEY }}
       - run: aws sts get-caller-identity
       - run: npm run test:ci
+        env:
+          # Tests make GitHub API requests which are rate limited.
+          # Use the GitHub Actions token for a limit of 1,000 requests per hour.
+          # <https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api#primary-rate-limit-for-github_token-in-github-actions>
+          GITHUB_TOKEN: ${{ github.token }}
 
       - if: always()
         uses: actions/upload-artifact@v4


jameshadfield (commented):
I think adding a Github token to the tests is almost essential at this point, so I'm all for it. 

From what I can tell, we're using "Default access (permissive)" rather than "Default access (restricted)" for the permissions granted by [the default token](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token), which would give the server and test runners more privileges than they need. But maybe this is just fine. @joverlee521 what do you think?

victorlin:
@jameshadfield good point. All the testing operations are read-only, so the restricted access should be sufficient. Since this is the only usage of `GITHUB_TOKEN` in this repo's CI, I've updated it to be restricted in the repo settings.

jameshadfield (commented):
> Since this is the only usage of GITHUB_TOKEN in this repo's CI, I've updated it to be restricted in the repo settings.

Thanks! This PR looks good to go from me, but... tests are failing:

```console
Run aws-actions/configure-aws-credentials@v4
Error: Input required and not supplied: aws-region
```

I haven't kept track of the recent (org-wide?) changes to to aws-region, but I presume it's related to this.

victorlin:
Rebased onto latest `master` which should fix the issue with aws-region.

The org-wide change is to store `AWS_REGION` in an organization variable instead of per-repo secrets.

jameshadfield (approved):

PR #818, opened 2024-04-09
CI: Use GitHub org variable for AWS region

victorlin:
Organization variable was recently added in https://github.com/nextstrain/zika/pull/54.

## Checklist

- [ ] Checks pass
- [ ] Post-merge: Remove `AWS_DEFAULT_REGION` repo secret

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
       # configure AWS to run dev server necessary for `npm run test:ci`
       - uses: aws-actions/configure-aws-credentials@v4
         with:
-          aws-region: ${{ secrets.AWS_DEFAULT_REGION }}
+          aws-region: ${{ vars.AWS_DEFAULT_REGION }}
           aws-access-key-id: ${{ secrets.DEV_SERVER_AWS_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.DEV_SERVER_AWS_SECRET_ACCESS_KEY }}
       - run: aws sts get-caller-identity


.github/workflows/ci.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       # configure AWS to run dev server necessary for `npm run test:ci`
       - uses: aws-actions/configure-aws-credentials@v4
         with:
-          aws-region: ${{ secrets.AWS_DEFAULT_REGION }}
+          aws-region: ${{ vars.AWS_DEFAULT_REGION }}
```
victorlin, line 37: On one hand, this reduces redundancy in having multiple `AWS_DEFAULT_REGION`s defined. On the other hand, this moves `AWS_DEFAULT_REGION` away from other related variables, some of which must be kept as secrets:

image

Thoughts? Based on [code search](https://github.com/search?q=org%3Anextstrain+secrets.AWS_DEFAULT_REGION&type=code), this applies to nextstrain.org and forecasts-ncov.

joverlee521: We've had to remove `AWS_DEFAULT_REGION` from secrets for any repo that uses the `pathogen-repo-build` workflow to prevent the AWS batch link from being borked (e.g. https://github.com/nextstrain/rsv/issues/53#issuecomment-1913717467). I think this move will match expectations that only "real" secrets should be kept as secrets and other shared values can be the unmasked variables.

victorlin: Thanks for the added context! I agree better to properly scope usage of the secrets store. I'll follow through with the move here and in forecasts-ncov.

Issue #814, opened 2024-04-08
`cryptography.test.js` fails on recent node versions


jameshadfield:
During development of #810, which didn't modify any code specifically relating to `cryptography.test.js`, test failures were consistenly observed [when run on nodejs 20.12.0](https://github.com/nextstrain/nextstrain.org/actions/runs/8566685893/job/23476917786#step:11:257) and [20.11.1](https://github.com/nextstrain/nextstrain.org/actions/runs/8564336483/job/23470717176) in GitHub Actions. The tests passed on nodejs 20.9.0 both locally and via GitHub actions. PR #810 pinned the version to 20.9.0.

I haven't explored _why_, as 20.9.0 is the latest available on conda, but the [node.js changelog](https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V20.md) has a few crypto-related changes between those versions.

<details>
  <summary>In case those action logs are gone by the time someone reads this, here's the error log:</summary>
  
```console
FAIL test/cryptography.test.js (15.536 s)
  ● encrypt/decrypt roundtrip

    TypeError: Cannot read properties of null (reading 'length')

      at getStateLength (node_modules/stream-shift/index.js:16:28)
      at shift (node_modules/stream-shift/index.js:6:99)
      at Duplexify.Object.<anonymous>.Duplexify._forward (node_modules/@aws-crypto/encrypt-node/node_modules/duplexify/index.js:170:35)
      at SignatureStream.onreadable (node_modules/@aws-crypto/encrypt-node/node_modules/duplexify/index.js:136:10)

    Use jest.setTimeout(newTimeout) to increase the timeout value, if this is a long-running test."

      87 |
      88 |
    > 89 | test("session id must match", async () => {
         | ^
      90 |   const {getTokens, setTokens} = await import("../src/authn/session");
      91 |
      92 |   const session = {

      at test/authn.test.js:89:1

  ● benchmark

    TypeError: Cannot read properties of null (reading 'length')

      at getStateLength (node_modules/stream-shift/index.js:16:28)
      at shift (node_modules/stream-shift/index.js:6:99)
      at Duplexify.Object.<anonymous>.Duplexify._forward (node_modules/@aws-crypto/encrypt-node/node_modules/duplexify/index.js:170:35)
      at SignatureStream.onreadable (node_modules/@aws-crypto/encrypt-node/node_modules/duplexify/index.js:136:10)

  ● benchmark

[...]

PR #826, opened 2024-04-11
Improve server startup messages

### LINKED ISSUE #685: [dev only] explain configuration settings in server start up message
The nextstrain.org server has conditional logic to change certain features into a "testing" / "development" mode. There are a few different (overlapping) concepts of testing/development, and it would be great to say up-front what the server is using. The logic to choose different modes is quite flexible -- environment variables, server configuration files, local aws configuration files etc, and it can be hard to keep track of what's being used! As an example of the kind of start-up message I would find really useful:

```
  -------------------------------------------------------------------------

  Nextstrain is an open-source project to harness the scientific and public
  health potential of pathogen genome data.

  This is the server behind nextstrain.org.
  See https://github.com/nextstrain/nextstrain.org for more.

  Server listening at http://localhost:5000
  Running in testing mode:
    Cognito user pool: nextstrain.org-testing
    Server knows about 3 groups (see ./env/testing/groups.json)
    Charon API requests may be made to localhost:8000
  Authorization is via IAM user nextstrain.org
  -------------------------------------------------------------------------
```
(I'm not sure the IAM user name is knowable, but if it's not we could at least say `via env variables $... and $...` or `via ./aws/config` etc.)

There are other things we could include here (perhaps GITHUB_TOKEN) but the above would be the most useful

jameshadfield:
Closes #685

@victorlin could you take a look at this, especially with how it may pertain to someone running the server as a groups app?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,6 +10,7 @@
         "@aws-crypto/client-node": "^3.1.1",
         "@aws-sdk/client-cognito-identity-provider": "^3.53.0",
         "@aws-sdk/client-s3": "^3.431.0",
+        "@aws-sdk/client-sts": "^3.552.0",
         "@aws-sdk/s3-request-presigner": "^3.431.0",
         "@keyv/redis": "^2.8.0",
         "@keyv/sqlite": "^3.6.6",
@@ -87,7 +88,7 @@
         "start-server-and-test": "^1.11.4"
       },
       "engines": {
-        "node": "^20",
+        "node": "20.9.0",
         "npm": "^10"
       }
     },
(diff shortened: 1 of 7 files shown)


victorlin (changes requested):

src/cognito.js
```diff
@@ -97,3 +98,21 @@ export async function removeUserFromGroup(username, group) {
...
+
+/**
+ * Return the name of the cognito user pool. Note that this requires elevated
+ * privileges for the AWS user the server is running under (specifically
+ * `cognito-idp:DescribeUserPool action`), but if you do have these it can be
+ * very useful for debugging.
+ */
+export async function getCognitoUserPoolName() {
+  if (!COGNITO_USER_POOL_ID) return `No cognito pool in use`
+  try {
```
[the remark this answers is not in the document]
jameshadfield: I'll read #731 next week, but my aim for including this info is because I have a set of cognito (nextstrain.org) usernames (as do a few of us) and you need to remember which pool the server is accessing to know which subset of users are valid. Of the commits added here it's not the most important, and since our common server IAM users don't have the permissions to even see the user pool name I may end up dropping the commit.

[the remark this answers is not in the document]
victorlin: For sure, my point is that there are effectively two entrypoints to Cognito and the server could theoretically be using two different user pools at the same time if `config.json` isn't configured properly. If you're using this info to decide which user to use on the login page, `COGNITO_USER_POOL_ID` isn't the right variable to inspect - it should be the OIDC/OAuth2 variables.

victorlin:
Force-pushed to account for conflicts from #834 – notably d9f7e225311a20218698f024c2198676c8980ec2 changed to 3fda8863d944e93ee9960030ecfd47dc6105d9e1

src/cognito.js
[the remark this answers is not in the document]
jameshadfield: I've dropped this commit for now. I'd still like to include the functionality however I don't have time right now to invest in understanding the recent changes here. I'll revisit it some time in the future.

src/startupMessage.js
```diff
@@ -0,0 +1,31 @@
...
+  console.log(nextstrainAbout);
+  console.log(`  Server listening on port ${server.address().port}, accessible at http://localhost:${server.address().port}`);
+  console.log(`  Server is running in ${app.locals.production ? 'production' : 'development'} mode`);
+  if (app.locals.STATIC_SITE_PRODUCTION) {
+    console.log(`  Next.js frontend is running in production mode (statically generated assets)`)
+  } else {
+    console.log(`  Next.js frontend is running in development mode (hot reloading etc)`)
+  }
+  console.log(`  AWS credentials: ${awsId}`)
+  console.log(`  Cognito user pool name: ${cognitoName}`)
```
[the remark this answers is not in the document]
jameshadfield: Thanks! I've taken some of the ideas suggested here and applied them in a repush. Would you mind re-reviewing when you get a chance @victorlin ?

src/cognito.js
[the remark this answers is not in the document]
victorlin: Dropped commit for future reference: 3ef40529c311e9d7a3d27faccfdcf60b997ed8b3

src/startupMessage.js
```diff
@@ -0,0 +1,35 @@
...
+  Nextstrain is an open-source project to harness the scientific and public
+  health potential of pathogen genome data.
+  See https://github.com/nextstrain/nextstrain.org for more.
+`;
+
+export async function startupMessage(server, app, groupsApp) {
+
+  const [awsId] = (await Promise.allSettled([
+    groupsApp ? _false() : getAwsId(), // omit AWS user for groups app
+  ])).map(({status, value, reason}) => value || reason); // eslint-disable-line no-unused-vars
```
victorlin, line 14: This can be simplified and `_false()` removed:

```suggestion
  // omit AWS user for groups app
  const awsId = groupsApp ? false : await getAwsId();
```

server.js
```diff
@@ -55,17 +51,8 @@ const server = http
...
-    console.log(`  Accessible at https://nextstrain.org or http://localhost:${server.address().port}`);
-    console.log(`  Server is running in ${app.locals.production ? 'production' : 'development'} mode`);
-    if (app.locals.STATIC_SITE_PRODUCTION) {
-      console.log(`  Next.js frontend is running in production mode (statically generated assets)`)
-    } else {
-      console.log(`  Next.js frontend is running in development mode (hot reloading etc)`)
-    }
-    console.log("\n  -------------------------------------------------------------------------\n\n");
+    utils.log(`Server now listening on port ${server.address().port} (startup message will appear shortly)`);
+    startupMessage(server, app, args.app!==DEFAULT_ENTRYPOINT)
```
victorlin, line 55: _(non-blocking)_

The mapping of `args.app!==DEFAULT_ENTRYPOINT` to `groupsApp` (from the function signature) feels indirect and implicit based on knowledge outside this file that there is only one other possible entrypoint which is for the groups app. Why not change the option from `--app` to `--groupsApp` as done in c125524212d5091802aa6888f691edf33aef7590?

victorlin (changes requested):

server.js
jameshadfield: Because that's a breaking change for external users. But you're involved in these use cases (I'm not!), so I'm happy to take your advice and make this change here.

src/startupMessage.js
jameshadfield: Yeah, I contemplated making this simplification here, but chose to leave the machinery in place for a future where we make multiple async requests for data (e.g. as I did for the cognito info, which has now been dropped). I'll make the simplification now, and restore this code when I add subsequent async calls.

server.js
victorlin: People running the groups app should be following [groupsApp.md](https://github.com/nextstrain/nextstrain.org/blob/2b12ed80595a1bdba7a4e8c4cb045cb84470f618/groupsApp.md?plain=1) which uses `npm run groups`. As long as we change the command used by `npm run groups`, it shouldn't be a breaking change.

src/startupMessage.js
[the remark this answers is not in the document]
victorlin: Continued in https://github.com/nextstrain/nextstrain.org/pull/826#discussion_r1576922375

server.js
tsibley: I'd consider this blocking. `args.app!==DEFAULT_ENTRYPOINT` mapped to "Groups app or not" is sloppy. How about reporting the entrypoint instead? And if you really want to do keep wording like `NEXTSTRAIN ' + (groupsApp?'GROUPS ':'') + 'SERVER'`, then have a property on `app` itself (i.e. set by each app constructor) with the desired title, e.g.

```diff
diff --git a/src/app.js b/src/app.js
index dcb8844a..634ef8e3 100644
--- a/src/app.js
+++ b/src/app.js
@@ -29,7 +29,7 @@ const {
  *   /logged-in
  *   /logout
  */
-const app = setupApp();
+const app = setupApp("Nextstrain Server");
 
 
 /* Canary.
diff --git a/src/groupsApp.js b/src/groupsApp.js
index 0430c729..fb5275a3 100644
--- a/src/groupsApp.js
+++ b/src/groupsApp.js
@@ -21,7 +21,7 @@ const {
  *   /logged-in
  *   /logout
  */
-const app = setupApp();
+const app = setupApp("Nextstrain Groups Server");
 
 
 /* Charon API used by Auspice.
diff --git a/src/routing/setup.js b/src/routing/setup.js
index 95f024f7..70eef7b7 100644
--- a/src/routing/setup.js
+++ b/src/routing/setup.js
@@ -11,7 +11,7 @@ import { replacer as jsonReplacer } from '../json.js';
 import * as middleware from '../middleware.js';
 
 
-export function setupApp() {
+export function setupApp(name) {
 
   /* Express boilerplate.
    */
@@ -19,6 +19,7 @@ export function setupApp() {
 
   app.set("json replacer", jsonReplacer);
 
+  app.locals.name = name;
   app.locals.production = PRODUCTION;
   app.locals.STATIC_SITE_PRODUCTION = STATIC_SITE_PRODUCTION;
 
diff --git a/src/startupMessage.js b/src/startupMessage.js
index b7334a1e..11b14018 100644
--- a/src/startupMessage.js
+++ b/src/startupMessage.js
@@ -7,9 +7,9 @@ export const nextstrainAbout = `
   See https://github.com/nextstrain/nextstrain.org for more.
 `;
 
-export async function startupMessage(server, app, groupsApp) {
+export async function startupMessage(server, app) {
 
   console.log("  -------------------------------------------------------------------------");
-  console.log('                         NEXTSTRAIN ' + (groupsApp?'GROUPS ':'') + 'SERVER');
+  console.log('                         ' + app.name.toUpperCase());
 
   console.log(nextstrainAbout);
   console.log(`  Server listening on port ${server.address().port}, accessible at http://localhost:${server.address().port}`);
```

I'd keep `--app`, i.e. not switch it to `--groupsApp`. I'm not inclined to break the invocation for others for what feels like just churn.

src/aws.js
```diff
@@ -32,3 +33,22 @@ export const clientConfig = {
...
+ */
+export async function getAwsId() {
+  const client = new STSClient({ region: AWS_REGION });
+  const command = new GetCallerIdentityCommand({});
+  try {
+    const data = await client.send(command);
+    if (!data.Arn.startsWith('arn:aws:iam::')) throw new Error("Identity is not an IAM user")
+    return (`IAM user '${data.Arn.replace(/^.+user\//, '')}'`)
+  } catch (error) {
+    if (String(error).includes('Could not load credentials from any providers')) {
```
tsibley, line 49: `error` should be an object you can type test against rather than testing against the stringified form which is likely to be more brittle.

src/startupMessage.js
tsibley: `Promise.allSettled()` will also, IIRC, handle bare values, e.g. `false`, just fine. And `new Promise((resolve) => resolve(false))` is more idiomatically spelled `Promise.resolve(false)`.

src/startupMessage.js
```diff
@@ -7,6 +7,11 @@ export const nextstrainAbout = `
 `;
 
 export async function startupMessage(server, app, groupsApp) {
+
+  const [awsId] = (await Promise.allSettled([
+    groupsApp ? _false() : getAwsId(), // omit AWS user for groups app
```
tsibley, line 13: AWS credentials will still be configured for the Groups app, so I'm not sure why we're skipping this? The credentials may not be true AWS credentials, but credentials for an S3-compatible service (like Cloudian Hyperstore, Minio, etc). Some of those will also support an STS-compatible service for GetCallerIdentity, but some won't.

src/groups.js
```diff
@@ -6,10 +6,21 @@ import * as cognito from "./cognito.js";
...
 import { readFile } from 'fs/promises';
 const GROUPS_DATA = JSON.parse(await readFile(GROUPS_DATA_FILE));
 
+/**
+ * Summary string of where the groups known to the server come from
+ * (useful for debugging purposes)
+ */
+export function groupsSummary() {
+  const nPublic = GROUPS_DATA.filter((g)=>g.isPublic).length;
+  const groupsFile = GROUPS_DATA_FILE.replace(rootDirFullPath, '.'); // shown relative to server.js
```
tsibley, line 20: Prefer path-handling functions for paths instead of string-handling functions.

```suggestion
  const groupsFile = path.relative(rootDirFullPath, GROUPS_DATA_FILE); // shown relative to server.js
```

tsibley (changes requested):

src/startupMessage.js
jameshadfield: > AWS credentials will still be configured for the Groups app, so I'm not sure why we're skipping this? 

Skipped due to (PR feedback)[https://github.com/nextstrain/nextstrain.org/pull/826#discussion_r1562968002]

src/startupMessage.js
jameshadfield: > `Promise.allSettled()` will also, IIRC, handle bare values, e.g. `false`, just fine. 

Cool. Indeed it seems it does. The [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled#parameters) state "An iterable of promises".

tsibley: The behaviour is supported by [the spec](https://tc39.es/ecma262/multipage/control-abstraction-objects.html#sec-promise.allsettled) too, but it's a bit harder to read.

server.js
tsibley: @victorlin 

> People running the groups app should be following [groupsApp.md](https://github.com/nextstrain/nextstrain.org/blob/2b12ed80595a1bdba7a4e8c4cb045cb84470f618/groupsApp.md?plain=1) which uses `npm run groups`. As long as we change the command used by `npm run groups`, it shouldn't be a breaking change.

I dunno, that seems short-sighted to me. That document is a guide/how-to, not a mandate to run things exactly a certain way…

src/startupMessage.js
victorlin: I didn't realize that it would be useful beyond AWS. If that's the case, I'm fine with it being included as long as the wording doesn't assume AWS

tsibley: Nod. I think we'd do well to be precise with wording here. CDC, for example, runs the Groups app against a Cloudian Hyperstore's S3-compatible service. But they still set `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` (and some other things) in order to authenticate with it, because we're still using an AWS library to talk to it.

server.js
victorlin: That's a fair point. I don't like the idea of setting a new property `app.name` for it to only be used in the startup message. Reporting the entrypoint file seems the most direct with minimal changes.

src/startupMessage.js
jameshadfield: Resolved in repush. Now using:
```js
const awsId = await getAwsId();
```

server.js
jameshadfield: I've gone back to my original implementation. We now attempt to report the AWS credentials for all use cases, thus avoiding the conditional logic being debated here.

src/groups.js
jameshadfield: Thanks. Fixed in repush.

src/startupMessage.js
jameshadfield: Repushed to go back to the original implementation (i.e. we always attempt to get the AWS(-like) ID)

src/aws.js
jameshadfield: Yeah, we can get this from `import { CredentialsProviderError } from "@smithy/property-provider";` but I think this is used more widely than the case where credentials aren't provided. I've gone a different way and now just print the strigified error message, which seems to be appropriate from my testing.

src/aws.js
```diff
@@ -32,3 +33,22 @@ export const clientConfig = {
...
+/**
+ * Attempt to identify the IAM user name behind the current AWS credentials
+ * (useful for debugging / server start-up messages).
+ */
+export async function getAwsId() {
+  const client = new STSClient({ region: AWS_REGION });
+  const command = new GetCallerIdentityCommand({});
+  try {
+    const data = await client.send(command);
+    if (!data.Arn.startsWith('arn:aws:iam::')) throw new Error("Identity is not an IAM user")
```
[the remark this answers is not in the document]
jameshadfield: Sure. Changed to
```js
    if (data.Arn.startsWith('arn:aws:iam::')) {
      return (`IAM user '${data.Arn.replace(/^.+user\//, '')}'`)
    }
    return data.Arn; // report the ARN for non-IAM credentials
```

src/aws.js
```diff
@@ -32,3 +33,23 @@ export const clientConfig = {
...
 };
+
+/**
+ * Attempt to identify the IAM user name behind the current AWS credentials
+ * (useful for debugging / server start-up messages). Some external users of
+ * the nextstrain server may be using non-AWS services, but these may still be
+ * compatible with STS so we make a best-effort attempt here.
+ */
+export async function getAwsId() {
+  const client = new STSClient({ region: AWS_REGION });
```
tsibley, line 44: Sorry for not noticing on the first round, but this should be using the shared global client config defined immediately above, e.g.

```suggestion
  const client = new STSClient(clientConfig);
```

package-lock.json
```diff
@@ -88,7 +89,7 @@
         "start-server-and-test": "^1.11.4"
       },
       "engines": {
-        "node": "^20",
+        "node": "20.9.0",
```
tsibley, line 92: This doesn't seem intentional?

tsibley (changes requested):

src/aws.js
jameshadfield: Changed in repush.

package-lock.json
jameshadfield: Must have arrived via a rebase and has now been disappeared by a rebase.

jameshadfield: Hmm. The current `package.json` engine is `"node": "20.9.0"` (due to #814) however this didn't get added to `package-lock.json` as part of 2d39dca12901bd3c1b46a33b2a0032c215ae21e5. So it's not intentional - it arrived as part of `npm i @aws-sdk/client-sts` - but it is now correct.

jameshadfield: I should get in the habit of reading `package-lock.json` more. In the old days it was very chaotic, but it seems to be a little more stable with newer node versions.

tsibley: Yeah, it's pretty stable and diffs tend to be understandable, esp. with Git's [moved-lines highlighting](https://git-scm.com/docs/git-diff#Documentation/git-diff.txt---color-movedltmodegt) enabled. I like to read it since it's the actual things we'll get installed.

PR #828, opened 2024-04-15
WIP resourc listing for core data files

jameshadfield:
A quick adaption of the current resource listing UI to display [data files](https://docs.nextstrain.org/en/latest/reference/data-files.html)

Code is not production ready. I'd want to spend some time and tidy it up and implement proper composition/inheritance in the code for the different sources & resource types. 

The UI likewise is rough around the edges, but it's _working_ and it'd be good to get some feedback.

URL: TKTK

### DIFF
--- a/src/resourceIndex.js
+++ b/src/resourceIndex.js
@@ -132,6 +132,9 @@ async function updateResourceVersions() {
  * ListResources is intended to respond to resource listing queries. The current
  * implementation only handles a single source Id and single resource type, but
  * this will be extended as needed.
+ * 
+ * There's definitely an inheritance structure here, but I haven't spent time to
+ * really draw it out. So instead of polymorphism we use conditionals.
  */
 class ListResources {
   constructor(sourceIds, resourceTypes) {
@@ -148,63 +151,87 @@ class ListResources {
     this.resourceType = resourceTypes[0];
   }
 
-  coreDatasetFilter([name, ]) {
-    /* Consult the manifest to and restrict our listed resources to those whose
-    _first words_ appear as a top-level key the manifest. Subsequent words
-    aren't checked, so datasets may be returned which aren't explicitly defined
-    in the manifest.
+  filterFn() {
 
-    This is very similar to restricting based on the routing rules (e.g. using
-    `coreBuildPaths`) however the manifest is a subset of those and is used here
-    as the listed resources should be those for which we have added the pathogen
-    name to the manifest.
-    */
-    if (!this._coreDatasetFirstWords) {
-      this._coreDatasetFirstWords = new Set(
-        global?.availableDatasets?.core?.map((path) => path.split("/")[0]) || []
-      );
-    }
-    return this._coreDatasetFirstWords.has(name.split("/")[0])
+    // TODO XXX
+    const _coreDatasetFirstWords = new Set(
+      global?.availableDatasets?.core?.map((path) => path.split("/")[0]) || []
+    );
+
+    const fn = ({
+      dataset: {
+        core([name, ]) {
+          /* Consult the manifest to and restrict our listed resources to those whose
+          _first words_ appear as a top-level key the manifest. Subsequent words
+          aren't checked, so datasets may be returned which aren't explicitly defined
+          in the manifest.
+      
+          This is very similar to restricting based on the routing rules (e.g. using
+          `coreBuildPaths`) however the manifest is a subset of those and is used here
+          as the listed resources should be those for which we have added the pathogen
+          name to the manifest.
+          */
+          return _coreDatasetFirstWords.has(name.split("/")[0])
+        },
+        staging() {return true;},
+      },
+      intermediate: {
+        core() {return true;},
+      },
+    })[this.resourceType][this.sourceId];
+    if (fn!==undefined) return fn;
+    throw new InternalServerError(`Source "${this.sourceId}" + resource type "${this.resourceType} does not have a corresponding filter function`);
   }
 
-  pathPrefixBySource(name) {
+  pathPrefix() {
     /**
      * We separate out the "source part" from the "pathParts" part in our
      * routing logic, creating corresponding Source and Resource objects. Here
      * we go in the other direction. We could link the two approaches in the
      * future if it's felt this duplication is too brittle.
+     * 
+     * Returns string | undefined
      */
-    switch (name) {
-      case "core":
-        return ""
-      case "staging":
-        return "staging/"
-      default:
-        throw new InternalServerError(`Source "${name}" does not have a corresponding prefix`)
-    }
+    const prefix = ({
+      dataset: {
+        core() {return "";},
+        staging() {return "staging/";},
+      },
+    })?.[this.resourceType]?.[this.sourceId]?.();
+    return prefix;
   }
 
+  pathVersions(_resources) {
+    const fn = ({
+      dataset() {
+        return Object.entries(_resources).map(([name, data]) => {
+          return [name, data.versions.map((v) => v.date)];
+        })
+      },
+      intermediate() {
+        return Object.entries(_resources).map(([name, data]) => {
+          return [name, Object.fromEntries((data.versions).map(({date, fileUrls}) => [date, fileUrls]))] // FIXME XXX
+        });
+      },
+    })[this.resourceType];
+    if (!fn) throw new InternalServerError(`Resource type "${this.resourceType} does not have a path version extractor`);
+    return fn();
+  }
 
   get data() {
     const _resources = resources?.[this.sourceId]?.[this.resourceType];
     if (!_resources) {
       throw new NotFound(`No resources exist for the provided source-id / resource-type`);
     }
-    if (this.resourceType !== 'dataset') {
-      throw new InternalServerError(`Resource listing is currently only implemented for datasets`);
-    }
-    const pathVersions = Object.fromEntries(
-      Object.entries(_resources).map(([name, data]) => {
-        return [name, data.versions.map((v) => v.date)];
-      })
-      .filter((d) => this.sourceId==='core' ? this.coreDatasetFilter(d) : true)
-    )
-    const d = {}
-    d[this.resourceType] = {}
+    const d = {};
+    d[this.resourceType] = {};
     d[this.resourceType][this.sourceId] = {
-      pathVersions,
-      pathPrefix: this.pathPrefixBySource(this.sourceId)
-    }
+      pathPrefix: this.pathPrefix(),
+      pathVersions: Object.fromEntries(
+        this.pathVersions(_resources)
+          .filter(this.filterFn())
+      )
+    };
     return d;
   }
 }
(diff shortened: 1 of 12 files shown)


huddlej:
I'm sorry I missed this when you first posted about it, @jameshadfield, but this is a super useful feature/resource! I wasn't sure which pathogens we actually had public data for (I forgot that I pushed cartography files up to S3) and I didn't realize the extent of the different files we produce for different pathogens (e.g., insertions, NDJSONs, and translations for mpox!). This seems like a major use case which is "I want to find out what pathogen datasets Nextstrain provides".

Like you mentioned above, there some UI ideas to discuss about handling longer-lived datasets like RSV A below where the links scroll off to the right.

image

I've always liked/preferred the "long page" approach to these types of data listings where I can scroll from the most recent results at the top to the first results at the bottom. What if you placed the dot plot at the top of the modal window as the visual representation of the data updates and then list the available datasets below that by data (sort of like a change log). This would look more like [the SARS-CoV-2 remote inputs list](https://docs.nextstrain.org/projects/ncov/en/latest/reference/remote_inputs.html#remote-inputs-open-files) but replicated by date. The idea there being that the other primary use case is probably (but we should check!) finding the latest or just-before-the-latest version of a specific file (like I want to use the latest metadata/alignments or the ones just before something broke in my workflow).

I'd be curious to hear other ideas for use cases for this page, though.

trvrb:
Looking at this again, I'm definitely a fan and don't think much needs to change. Mainly just fixing the overflow issue above etc... making the red x's a bit less strong, fixing the dates to not wrap. Basically just little UI cleanup rather than large new directions.

My only more major suggestion would be to just start the info box with the latest results (without version ID) rather than trying to show a long table. And then you click on bubbles to replace these latest results with specific dated versions. Usually people will want latest files and starting with the big tables is a bit overwhelming.

jameshadfield:
Thanks! Given John & Trevor's feedback on this prototype I'll find some time to bring the code & UI up to scratch and get this pushed out.

jameshadfield:
**[Review app](https://nextstrain-s-james-file-v7poga.herokuapp.com)**

This PR has been reworked / rewritten (on top of #1213). There are 3 main changes and one todo:

### 1. Addition of [pathogens/files](https://nextstrain-s-james-file-v7poga.herokuapp.com/pathogens/files) page

This lists the _latest versions_ of [data files](https://docs.nextstrain.org/en/latest/reference/data-files.html) for pathogens. Some files have been excluded (see [this code](https://github.com/nextstrain/nextstrain.org/pull/828/files#diff-8d23a42ad8201e6cafbc7882f552d24ee1a10ff96e4ab577f34e5b32b40b7381R238-R248) for more). The listing of past versions is left for subsequent work (see comment below). This necessitated updates to the resource indexer (with the index now bumped to v6) and the server 'list-resources' routes. 

### 2. URL queries save filter state

This applies to anywhere `<ListResources>` is used - /pathogens, /groups, /staging and /pathogens/files. For instance, here's the [/pathogens page showing H3N2 only](https://nextstrain-s-james-file-v7poga.herokuapp.com/pathogens?filter=h3n2&filter=seasonal-flu). I haven't added the "open modal" as a URL query, but we can (should) do so one day.

### 3. Separate out data parsing & visualisation

The visualisation logic of the `<ListResources>` component is now fully decoupled from the fetching / parsing / munging logic to prepare data for visualisation. This is done by providing an async function `fetchResourceGroups: () => Promise<Group[]>` to the rendering component. This is much cleaner and easier to reason with as the component is used across more pages. 

### TODO: list past versions of files

I've done the server-side part, just not the rendering. It's quite different from Auspice datasets, as there we have a single unit (the dataset) which we can track back in time. Here we have a set of units per pathogen (workflow?) and those differ as we go back in time. E.g. The latest Rubella file is from 2025-08-15 and only includes `metadata.tsv.zst`, but previous updates have included `ncbi.ndjson.zst` and `sequences.fasta.zst`. I am imagining a modal for each card (e.g. Rubella) which is essentially a scrollable changelog showing the files uploaded on different days. I don't consider this work as blocking for this PR but it may get done as part of it.

## Open questions

* Is picking the most recent day with anything uploaded (see rubella example above) the right move, or should we instead track individual files back in time? E.g. show all 3 rubella datasets as the "latest" update, even though they're different upload dates. This may affect how we choose to show past versions of files. I thought the answer was _no_ but I'm now leaning towards _yes_.

jameshadfield:
We can now list all previous versions of a group (where group is a separate card in the UI) via a changelog-style modal UI. If any filters have been applied then the listed versions will also be filtered.


image

---

image

src/resourceIndex.js
```diff
@@ -186,26 +193,40 @@ class ListResources {
...
+            .map(_formatIntermediates)
+        );
+      } else {
+        result[this.resourceType][this.sourceId].latestVersions = Object.fromEntries(_resources
+          .map(_formatIntermediates)
+          .map(_onlyLatestIntermediates)
+        );
+      }
+    }
+    return result;
```
genehack, line 229: Suggested refactor:

```suggestion
    // Use conditional logic to select appropriate behaviour as it's simpler to understand
    // than splitting behaviour across multiple classes (in this case, anyway)
    let key;
    let entries;
    if (this.resourceType === 'dataset') {
      key = "pathVersions";
      entries = _resources.map(_formatDatasets);
    } else if (this.resourceType === 'intermediate') {
      if (this.groupHistory) {
        key = "pathVersions";
        entries = _resources
          .filter(([name,]) => name.split('/')[0]===this.groupHistory)
          .map(_formatIntermediates);
      } else {
        key = "latestVersions"
        entries = _resources
          .map(_formatIntermediates)
          .map(_onlyLatestIntermediates);
      }
    }

    result[this.resourceType][this.sourceId][key] = Object.fromEntries(entries);

    return result;
```

This extracts all (most?) of the duplicated boilerplate into the single line at the end, and IMO makes it easier to understand what the conditional logic is switching across.

static-site/app/pathogens/[[...pathogens]]/resources.tsx
```diff
@@ -0,0 +1,29 @@
+"use client";
+// Note: this is only in a separate file as it needs to be run client side
+// and we want to run as much of the parent page server-side as possible.
+
+import React from "react";
+import ListResources from "../../../components/list-resources";
+import { listResourcesAPI } from "../../../components/list-resources/listResourcesApi";
+import * as coreResources from "../../../content/resource-listing.yaml";
+
+export default function CorePathogenResourceListing(): React.ReactElement {
```
Comment (genehack) [inline static-site/app/pathogens/[[...pathogens]]/resources.tsx:10]: This needs a JSDoc description.

static-site/app/pathogens/[[...pathogens]]/resources.tsx
```diff
@@ -0,0 +1,29 @@
...
+      quickLinks={coreResources["coreQuickLinks"]}
+      resourceType="dataset"
+      tileData={coreResources["coreTiles"]}
+      versioned
+      fetchResourceGroups={coreDatasetResourceGroups}
+    />
+  )
+}
+
+async function coreDatasetResourceGroups() {
```
Comment (genehack) [inline static-site/app/pathogens/[[...pathogens]]/resources.tsx:22]: as does this.

static-site/app/pathogens/files/page.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const title = "Workflow data files for core pathogens";
+
+export const metadata: Metadata = {
+  title,
+};
+
+export default function PathogenWorkflowFiles(): React.ReactElement {
+  return (
+    <div className="pageContent">
```
genehack, line 18: I ended up undoing this, based on feedback on one of my PRs, so this can just be

```suggestion
    <div>
```

static-site/components/list-resources/index.tsx
```diff
@@ -117,7 +115,9 @@ export default function ListResources(
   return (
     <ErrorBoundary>
       <div ref={ref}>
-        <ListResourcesContent {...props} elWidth={elWidth} />
+        <Suspense>
```
genehack, line 118: not sure what this accomplishes without a `fallback` prop?

static-site/components/list-resources/listResourcesApi.ts
```diff
@@ -0,0 +1,335 @@
...
+    /** Report prior versions of each resource.
+     * TODO: infer this from the API data itself
+     */
+    versioned: boolean,
+    groupDisplayNames?: Record<string, string>,
+    groupUrl?: (groupName: string) => string
+  }
+): Promise<Group[]> {
+  const requestPath = `/list-resources/${sourceId}/${resourceType}`;
+  const response = await fetch(requestPath, {
```
genehack, line 35: why not use the `fetchAndParseJson` method here? It'll handle all the error checking and you can inject the type param and avoid the cast on L44.

static-site/components/list-resources/listResourcesApi.ts
```diff
@@ -0,0 +1,335 @@
...
+    throw new Error(
+      `fetching application/json data from "${requestPath}" returned status code ${response.status}`,
+    );
+  }
+  // eslint-disable-next-line @typescript-eslint/consistent-type-assertions
+  const data = ((await response.json())[resourceType][sourceId]) as ResourceListingDatasets |  ResourceListingIntermediates;
+  const urlBuilder = (name) => `/${data.pathPrefix}/${name}`;
+  const areDatasets = (x: ResourceListingDatasets |  ResourceListingIntermediates): x is ResourceListingDatasets => {
+    return Object.hasOwn(x, 'pathVersions');
+  }
```
genehack, line 48: Strong, non-blocking preference: use a function unless there's an actual reason for storing an anonymous function in a variable. (AFAICT, here, there isn't.)

static-site/components/list-resources/modal-contents-group-history.tsx
```diff
@@ -0,0 +1,179 @@
...
+    return [[date, Object.fromEntries(filePairs)]]
+  })
+}
+
+
+type GroupHistory = [
+  date: string,
+  {[nameInclFilename: string]: string}
+][]
+
```
genehack, line 133: since there's a `./types` file for this component, probably better to put this there and import it.

static-site/components/list-resources/types.ts
```diff
@@ -16,13 +16,30 @@ export interface Group {
...
   groupUrl?: string;
   groupDisplayName?: string;
+  fetchHistory?: FetchGroupHistory;
 }
 
+/* helper function / type guard */
+export function isGroup(x: Group|Resource): x is Group {
+  return Object.hasOwn(x, 'resources');
+}
+
```
genehack, line 26: Strong, non-blocking opinion: `types.ts` files should not export functions, only types or interfaces.

genehack (commented):
I left comments on some, but not all, of the undocumented functions.

One of the things I was very careful to do, with the port to the App Router, was to write a "proper" comment on every "public" method (i.e., anything that didn't start with a leading `_`), explaining what the function does, if it was a React Component whether it was Server, Client, or helper, documented all the params, etc. 

It would be nice if that consistency of documentation could be maintained.

PR #830, opened 2024-04-15
index-resources.yml: Create separate indexes for Heroku apps

### LINKED ISSUE #829: [RFC] Maintain different resource indexes
**background**

Review apps and {dev, canary, production} servers use the same resource index file:
https://github.com/nextstrain/nextstrain.org/blob/bb40806bf54473b179cdcf0f8a647119ae422ce9/env/production/config.json#L113

The resource index is generated on a schedule using a [node.js program](https://github.com/nextstrain/nextstrain.org/blob/master/resourceIndexer/main.js) from the master branch and uploading the result:

https://github.com/nextstrain/nextstrain.org/blob/bb40806bf54473b179cdcf0f8a647119ae422ce9/.github/workflows/index-resources.yml#L39

AWS lifecycles expire the indexes after a month.

**out of sync errors**

Because everything references the same index file, our typical development and deployment cycle is broken if it requires changes to the resource indexer. Using [Avian flu restructure](https://github.com/nextstrain/nextstrain.org/pull/811) as an example, changes in that PR to the indexer will result in "flu/..." datasets being removed and "avian-flu/..." and "seasonal-flu/..." datasets added. The server & frontend code are updated accordingly. Upon merge the following would happen:
* The resource indexer would (within 24h) rebuild using the newly merged code, then:
* next.nextstrain.org would work well
* nextstrain.org would no longer be able to access `flu/...@version` URLs, and the `/pathogens` page would show avian-flu and seasonal-flu datasets which would all 404.

A similar story exists for review apps, unless you do [something like this](https://github.com/nextstrain/nextstrain.org/pull/828/commits/4c9379b2c5c2d2901ed8302e47e4948986675efb)

I didn't anticipate we would be making so many structural changes to the index. But we are, so we should improve the system.

**proposed solution**

Whenever changes to the indexer are made we increment some identifier¹  in the index filename, so it looks like `resources_mk1.json.gz` or similar and adjust the `config.json` values accordingly so the server is pulling in the appropriate file. IAM policies would be ammended so that we can access the index for all identifiers. Lifecycle methods would be likewise adjusted to expire all index files after a month.

For review apps, if you triggered an index rebuild (GitHub actions or locally) then the app would work. The index would become stale, but that seems fine for review purposes.

For merges into master, the new index file would now be rebuilt daily, and so next.nextstrain.org would work well. Prior to promotion, nextstrain.org would continue to work but the index it refers to would no longer be being updated, which isn't great.

Are there better ideas? Would be great to work through this with others.

¹ Or generate a random hex, which would better guard against collisions.

joverlee521:

We are running into out-of-sync issues between the resources index and the production site when canary is not immediately promoted.

This change creates separate indexes for the nextstrain-server and nextstrain-canary Heroku apps by checking out the commit of their latest released slug.

The separate resources JSONs are uploaded to separate paths prefixed by the app name, which required updating the permissions for `GitHubActionsRoleResourceIndexer` outside of this commit.

## Related issue(s)

Resolves #829 

## Checklist

- [x] Checks pass
- [x] [Test run of indexer ](https://github.com/nextstrain/nextstrain.org/actions/runs/8696800252)

### DIFF
--- a/.github/workflows/index-resources.yml
+++ b/.github/workflows/index-resources.yml
@@ -19,8 +19,40 @@ jobs:
     defaults:
       run:
         shell: bash
+    env:
+      HEROKU_TOKEN: ${{ secrets.HEROKU_TOKEN_READ_PROTECTED }}
+      HEROKU_PIPELINE: 38f67fc7-d93c-40c6-a182-501da2f89d9d
+      APP_NAME: ${{ matrix.app_name }}
+    strategy:
+      matrix:
+        app_name: ["nextstrain-server", "nextstrain-canary"]
     steps:
+      - name: Get latest slug ID for ${{ env.APP_NAME }}
+        run: |
+          slug_id=$(
+            curl -n https://api.heroku.com/pipelines/"$HEROKU_PIPELINE"/latest-releases \
+              --fail --silent --show-error \
+              -H "Accept: application/vnd.heroku+json; version=3" \
+              -H "Authorization: Bearer $HEROKU_TOKEN" \
+              | jq -r '.[] | select(.app | .name == env.APP_NAME) | .slug.id')
+
+          echo "SLUG_ID=$slug_id" | tee -a "$GITHUB_ENV"
+
+      - name: Get latest commit hash for ${{ env.app_name }}
+        run: |
+          commit_hash=$(
+            curl -n https://api.heroku.com/apps/"$APP_NAME"/slugs/"$SLUG_ID" \
+              --fail --silent --show-error \
+              -H "Accept: application/vnd.heroku+json; version=3" \
+              -H "Authorization: Bearer $HEROKU_TOKEN" \
+              | jq -r '.commit')
+
+          echo "COMMIT_HASH=$commit_hash" | tee -a "$GITHUB_ENV"
+
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ env.COMMIT_HASH }}
+
       - uses: actions/setup-node@v4
         with:
           node-version-file: 'package.json'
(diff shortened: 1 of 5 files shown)


joverlee521:
There's no difference in the test run because I just promoted canary to production earlier today.

.github/workflows/index-resources.yml
```diff
@@ -36,4 +68,4 @@ jobs:
             --resourceTypes dataset --collections core staging
       - name: Upload the new index, overwriting the existing index
         run: |
-          aws s3 cp resources.json.gz s3://nextstrain-inventories/resources.json.gz
+          aws s3 cp resources.json.gz s3://nextstrain-inventories/"$APP_NAME"/resources.json.gz
```
jameshadfield, line 71: You'll need to update the IAM users ([e.g. this one for production](https://github.com/nextstrain/nextstrain.org/blob/master/aws/iam/policy/NextstrainDotOrgServerInstance.tftpl.json#L49-L57)) to allow the server to read these files

joverlee521: Ah, thank you! I was wondering what other permissions might need to be updated!

joverlee521: Updated in https://github.com/nextstrain/nextstrain.org/pull/830/commits/00ab2c9c04ae99f794dc0be1953d1c7ef58754b7, although I have not deployed the changes via Terraform yet.

joverlee521: Applied the Terraform changes to the testing environment. 
Now, I can locally start the server with the testing IAM users credentials, pointing to the new resource.json.gz

```shell
RESOURCE_INDEX="s3://nextstrain-inventories/nextstrain-canary/resources.json.gz" \
    envdir ~/Repos/env.d/nextstrain.org-testing/ \
    npm run server
```

joverlee521:
Closing this PR for now since we will be going in a slightly different direction after [team discussion](https://github.com/nextstrain/nextstrain.org/issues/829#issuecomment-2065182925).

PR #831, opened 2024-04-19
Update tests now that zika root sequence is inline

jameshadfield:
https://github.com/nextstrain/zika/pull/57 removed the zika root sequence sidecar (as it's inlined in the main JSON). Note that the staging/zika dataset still has the root sequence sidecar and we assert its existence in tests here!


joverlee521 (approved):
Ah, thanks for fixing this! I did not realize we tested the existence of sidecar files here!

Issue #832, opened 2024-04-19
Understand nextstrain.org caching


jameshadfield:
As part of writing up this issue I fixed it, but it's still _really weird_ to me, and I'd like to understand _why_ this happened.

---

**Description of bug**

As part of #811 I uploaded new datasets to our staging bucket, by downloading then uploading the flu/seasonal/h3n2/ha/2y JSONs and renaming them to seasonal-flu, so we could update our tests to use the latter. I initially uploaded them using `aws s3 cp` (so the data wasn't zipped), and then re-uploaded them using `nextstrain remote upload ...` (which magically zips the data). The strangeness is to do with the associated content-encoding metadata and/or the actual encoding of the file. The file is gzipped (as it's 1.1MB, and uncompressed it's 37MB), and the content encoding is gzip (can see this in [AWS console](https://us-east-1.console.aws.amazon.com/s3/object/nextstrain-staging?region=us-east-1&bucketType=general&prefix=seasonal-flu_h3n2_ha_2y.json)). 

Using a small script to access this file (cf. the staging/zika dataset) via various APIs gives me weird results.

<details>
  <summary>Test script</summary>

Run the server locally at the default port 5000 and then run the following script: 

  ```js
import fetch from 'node-fetch';

async function main() {
  const paths = [
    "staging/seasonal-flu/h3n2/ha/2y",
    "staging/zika",
  ]

  /* Make a GET request to our REST API asking for JSON data */
  console.log("\nREST API")
  for (const path of paths) {
    const mediaType = "application/json"
    const url = `http://localhost:5000/${path}`
    console.log("\tfetch:", url, mediaType)
    const req = fetch(url, accept(mediaType));
    const res = await req;
    console.log("\t\tresponse content encoding:", res.headers.get('content-encoding'))

    /* print out the start of the response. Note that this is after decompression, if any,
    and I don't know how to truly know if decompression is happening. If content-encoding is gzip
    then does this mean the response has for sure been decompressed? */
    const text = await res.text();
    console.log("\t\tResponse:", text.slice(0,50));
  }

  /* now do the same but hitting the Charon endpoint */
  console.log("\nCharon API")
  for (const path of paths) {
    const url = `http://localhost:5000/charon/getDataset?prefix=${path}`
    console.log("\tfetch:", url)
    const req = fetch(url)
    const res = await req;
    console.log("\t\tresponse content encoding:", res.headers.get('content-encoding'))
    const text = await res.text();
[...]

PR #835, opened 2024-04-22
upstream: Forward Content-Type from upstream if we don't already have one

tsibley:
For example, our RESTful endpoints perform content negotiation and use the result to determine Content-Type (and also what to Accept in the request to upstream), but our Charon endpoints do not set Content-Type at all.  Set it based on the upstream response in that situation.

I noticed this bug (while investigating another issue¹) because the automatic compression middleware wasn't applying to the /charon/getDataset endpoint due to the lack of Content-Type.

¹ <https://github.com/nextstrain/nextstrain.org/issues/832>

## Checklist

- [x] Checks pass

### DIFF
--- a/src/upstream.js
+++ b/src/upstream.js
@@ -234,6 +234,11 @@ async function proxyResponseBodyFromUpstream(req, res, upstreamReq) {
      * client can decode the body itself.
      */
     ...(!upstreamReq.compress ? ["Content-Encoding"] : []),
+
+    /* Forward Content-Type if our response doesn't already have a preferred
+     * type set (e.g. from prior content negotiation).
+     */
+    ...(!res.get("Content-Type") ? ["Content-Type"] : []),
   ];
 
   res.set(copyHeaders(upstreamRes.headers, forwardedUpstreamResHeaders));


src/upstream.js
```diff
@@ -234,6 +234,11 @@ async function proxyResponseBodyFromUpstream(req, res, upstreamReq) {
      * client can decode the body itself.
      */
     ...(!upstreamReq.compress ? ["Content-Encoding"] : []),
+
+    /* Forward Content-Type if our response doesn't already have a preferred
+     * type set (e.g. from prior content negeotiation).
```
victorlin, line 239: Fix typo (also in commit message):

```suggestion
    /* Forward Content-Type if our response doesn't already have a preferred
     * type set (e.g. from prior content negotiation).
```

victorlin (approved):
- [x] [`/charon/getDataset?prefix=/groups/blab/sars-like-cov`](https://nextstrain-s-trs-charon-bfi0tp.herokuapp.com/charon/getDataset?prefix=/groups/blab/sars-like-cov) shows `Content-Type:
application/json; charset=utf-8` on preview app (it doesn't on next.nextstrain.org)

src/upstream.js
tsibley: Ha, thanks.

tsibley: Fixed in repush.

src/upstream.js
```diff
@@ -234,6 +234,11 @@ async function proxyResponseBodyFromUpstream(req, res, upstreamReq) {
      * client can decode the body itself.
      */
     ...(!upstreamReq.compress ? ["Content-Encoding"] : []),
+
+    /* Forward Content-Type if our response doesn't already have a preferred
+     * type set (e.g. from prior content negotiation).
+     */
+    ...(!res.get("Content-Type") ? ["Content-Type"] : []),
```
jameshadfield, line 241: The upstream content-type is `'application/json'` however we are converting this to `'application/json; charset=utf-8'`. In practice, I don't think clients care about charsets for JSON, but my understanding is that it shouldn't be there (see [page 11 of rfc8259](https://www.rfc-editor.org/rfc/rfc8259)).

This is automagically added by express so I'm guessing it's something we will live with:
```js
  res.set({'Content-Type': 'application/json'})
  console.log(res.get("Content-Type")) // application/json; charset=utf-8
```

but I want to point this out here as our code to copy upstream headers over to the response will not just copy them but modify them.

jameshadfield (approved):
Thanks - small changes like this are important and help keep the behaviour of the two APIs closer together.

PR #837, opened 2024-04-23
adding in time resolution for avian influenza builds to allow 2y and …

lmoncla:
…all-time builds for h5nx and h5n1, and all-time builds only for h9n2 and h7n9

We recently made a series of updates to the avian-flu builds to provide a more in-depth view of the ongoing H5Nx outbreak of 2.3.4.4b viruses. We now have 2 different time resolutions for H5Nx and H5N1: 2y and all-time. This pull request updates `manifest_core.json` to reflect these changes.

### DIFF
--- a/data/manifest_core.json
+++ b/data/manifest_core.json
@@ -231,7 +231,7 @@
               "default": "ha"
             }
           },
-          "default": "h5nx"
+          "default": "h5n1"
         }
       },
       "dengue": {
--- a/data/manifest_core.json
+++ b/data/manifest_core.json
@@ -5,55 +5,231 @@
         "subtype": {
           "h5n1": {
             "segment": {
-              "ha": "",
-              "mp": "",
-              "na": "",
-              "ns": "",
-              "np": "",
-              "pa": "",
-              "pb1": "",
-              "pb2": "",
+              "ha": {
+                "resolution": {
+                  "all-time": "",
+                  "2y": "",
+                  "default": "2y"
+                }
+              },
+              "mp": {
+                "resolution": {
+                  "all-time": "",
+                  "2y": "",
+                  "default": "2y"
+                }
+              },
+              "na": {
+                "resolution": {
+                  "all-time": "",
+                  "2y": "",
+                  "default": "2y"
+                }
+              },
+              "ns": {
+                "resolution": {
+                  "all-time": "",
+                  "2y": "",
+                  "default": "2y"
+                }
+              },
+              "np": {
+                "resolution": {
+                  "all-time": "",
+                  "2y": "",
+                  "default": "2y"
+                }
+              },
+              "pa": {
+                "resolution": {
+                  "all-time": "",
+                  "2y": "",
+                  "default": "2y"
+                }
+              },
+              "pb1": {
+                "resolution": {
+                  "all-time": "",
+                  "2y": "",
+                  "default": "2y"
+                }
+              },
+              "pb2": {
+                "resolution": {
+                  "all-time": "",
+                  "2y": "",
+                  "default": "2y"
+                }
+              },
               "default": "ha"
             }
           },
           "h5nx": {
             "segment": {
-              "ha": "",
-              "mp": "",
-              "na": "",
-              "ns": "",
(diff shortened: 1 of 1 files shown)


data/manifest_core.json
```diff
@@ -5,55 +5,199 @@
...
-              "np": "",
-              "pa": "",
-              "pb1": "",
-              "pb2": "",
+              "ha": {
+                "resolution": {
+                  "all-time": "",
+                  "2y": "",
+                  "default": "2y"
+                }
```
tsibley, line 13: Bunch of missing closing braces, e.g.

```suggestion
                }
              },
```

here and in other changed places.

data/manifest_core.json
```diff
@@ -5,55 +5,231 @@
...
+              },
+              "pb2": {
+                "resolution": {
+                  "all-time": "",
+                  "2y": "",
+                  "default": "2y"
+                }
```
tsibley, line 20: The ordering of segments is different with this change. Instead of being alphabetical (ha, mp, na, …) it's now something else (ha, pb2, pb1, …). Was that intentional? (It wasn't called out in the commit message, so not clear to me.)

data/manifest_core.json
```diff
@@ -5,55 +5,231 @@
...
+                "resolution": {
+                  "all-time": "",
+                  "default": "all-time"
+                }
+              },
               "default": "ha"
             }
           },
-          "h9n2": {
+          "h7n9": {
```
tsibley, line 181: Is this an intentional change? It looks like an error. A duplication of the above.

data/manifest_core.json
tsibley: I fixed and [repushed](https://github.com/nextstrain/nextstrain.org/compare/fc223a7e3cf8d8a6cf4a66c1eb53a4a079a1b80a..dfb2af56d8ffbc00faecf909b67cb43ffaa97774) so we can preview.

data/manifest_core.json
tsibley: Fixed and [repushed](https://github.com/nextstrain/nextstrain.org/compare/dfb2af56d8ffbc00faecf909b67cb43ffaa97774..297712220408e207adf7552a764eb53ecdf639c8) so this doesn't drop h9n2.

data/manifest_core.json
tsibley: Since this affects the order of the dropdowns, I'd think we'd want to keep the previous alphabetical ordering. (That's also what we force on https://nextstrain.org/pathogens.)

tsibley: [Restored](https://github.com/nextstrain/nextstrain.org/compare/297712220408e207adf7552a764eb53ecdf639c8..d899712136387b5628a7b785c30dbf33c13654d4) the not-quite-alphabetical order to:

```
ha
mp
na
ns
np
pa
pb1
pb2
```

trvrb: Yes. Let's keep the original ordering.

trvrb:
@lmoncla: Can we update default from `h5nx` to `h5n1`? It seems like this is more expected by the community and the vast majority of recent H5 viruses are H5N1 viruses. And all the viruses we care about in the recent 2.3.4.4b outbreak are H5N1.

lmoncla:
@trvrb now that we've switched over to the 2y build as default, switching back to h5n1 as default is fine by me. The all-time h5n1 tree gives a very incomplete picture of h5 evolution because it is missing a long stretch of time during which the ancestral 2.3.4.4 H5 HA was reassorting with other NAs. But I think that given these updates, fine to revert to H5N1 as default.

data/manifest_core.json
lmoncla: This was me trying to make sure I didn't forget a particular segment by using flu segment ordering, which goes by longest to shortest. I don't have a preference on ordering at all!

trvrb:
Thank you @lmoncla! I imagine these defaults to be something that we can update frequently as situation changes. I like H5N1 / 2y as default given that this is probably the best place to take someone who clicks on the "avian flu" tile because they're interested in what's going on.

PR #839, opened 2024-04-25
Fix /whoami

victorlin:

#780 broke /whoami for users that are members of old test Cognito groups. This is due to unhandled errors in the case that a Cognito group does not map to a valid group in groups.json. This PR addresses the issue.

## Related issue(s)

- prompted by https://github.com/nextstrain/nextstrain.org/pull/780#discussion_r1578485451

## Checklist

- [x] Tested locally. Provisioned a `fake-group` in test Cognito user pool and added my user to it. Logging in reproduced the error and these changes fix the issue.
- [x] Checks pass


src/authn/index.js
```diff
@@ -601,8 +602,17 @@ function parseCognitoGroups(cognitoGroups) {
...
+      // Just the Nextstrain Group names (i.e. excluding the "/role" suffix)
+      ...new Set(nextstrainGroupsRoles.map(g => splitGroupRole(g).group))
+    ].map(name => {
+      try {
+        return new Group(name);
+      } catch (error) {
+        // Nextstrain Group does not exist for the Cognito group prefix
+        return null;
+      }
+    }).filter(group => group !== null),
```
tsibley, line 615: This type change from `Array[String]` → `Array[Group]` will break the old-school session serialization/deserialization. I don't think we _need_ that older way of handling sessions any more, but it's currently still present.

tsibley: I'll look into ripping it out.

tsibley: https://github.com/nextstrain/nextstrain.org/pull/840 (but needs testing first)

victorlin: Rebased onto latest `master` to bring in changes from #840

tsibley (approved):

PR #841, opened 2024-04-26
Separate indexer

joverlee521:

This PR starts the use of versioned resource index JSONs so that we no longer run into out-of-sync issues between the canary and production servers. 

Major changes:
- the scheduled index resources workflow updates both canary and production resource index JSONs if they are different versions
- the CI workflow rebuilds the resource index before deployment of canary
- the CI workflow builds a `/tmp/<branch>` prefixed resource index for review apps

## Related issue(s)

Related to https://github.com/nextstrain/nextstrain.org/issues/829

## Checklist

- [x] Checks pass
- [x] [index resources workflow for canary/production](https://github.com/nextstrain/nextstrain.org/actions/runs/8840814930)
- [x] [CI run builds/uploads `/tmp/<branch>` resource index](https://github.com/nextstrain/nextstrain.org/actions/runs/8914280686/job/24481649899#step:8:28)  
- [x] Deploy env/production Terraform changes
- ~[ ] Tick the wait for CI box in Heroku~
- [ ] post-merge: verify CI rebuilds index for deployment
- [ ] post-merge: [update S3 lifecycle](https://github.com/nextstrain/nextstrain.org/pull/841/commits/4c7a293231ecded2e7b72eaa9460d7eb7c2dd39c#r1587035296)
- [x] Write up changes that require bumping the resource index version

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,16 +95,28 @@ jobs:
           name: test-logs
           path: test/server.log
 
+  index-resources:
+    if: |2
+         github.repository == 'nextstrain/nextstrain.org'
+      && github.event_name == 'push'
+      && github.ref == 'refs/heads/master'
+    needs: test
+    permissions:
+      id-token: write # needed to interact with GitHub's OIDC Token endpoint
+      contents: read
+    uses: ./.github/workflows/index-resources.yml
+
   deploy:
     if: |2
          github.repository == 'nextstrain/nextstrain.org'
       && (   (github.event_name == 'push' && github.ref == 'refs/heads/master')
           || (github.event_name == 'workflow_dispatch' && inputs.heroku-app) )
 
-    # Wait for "build" and "test" jobs above to pass.
+    # Wait for "build", "test", and "index-resources" jobs above to pass.
     needs:
       - build
       - test
+      - index-resources
 
     # Only one "deploy" job per Heroku app at a time.
     concurrency: deploy:${{ inputs.heroku-app || 'nextstrain-canary' }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,7 +108,11 @@ jobs:
 
   deploy:
     if: |2
-         github.repository == 'nextstrain/nextstrain.org'
+         !cancelled()
+      && needs.build.result == 'success'
+      && needs.test.result == 'success'
+      && contains(fromJSON('["success", "skipped"]'), needs.index-resources.result)
+      && github.repository == 'nextstrain/nextstrain.org'
       && (   (github.event_name == 'push' && github.ref == 'refs/heads/master')
           || (github.event_name == 'workflow_dispatch' && inputs.heroku-app) )
 
--- a/.github/workflows/index-resources.yml
+++ b/.github/workflows/index-resources.yml
@@ -34,6 +34,21 @@ jobs:
           node resourceIndexer/main.js \
             --gzip --output resources.json.gz \
             --resourceTypes dataset --collections core staging
-      - name: Upload the new index, overwriting the existing index
+      - name: Extract RESOURCE_CONFIG from config module
         run: |
-          aws s3 cp resources.json.gz s3://nextstrain-inventories/resources.json.gz
+          # This conditional is to continue to support servers that are still
+          # using versions of this code that do not include the get-resource-index
+          # script. We can remove it once all the servers have been updated.
+          #   -Jover, 01 May 2024
+          if [[ -f ".scripts/get-resource-index.js" ]]; then
+            resource_index=$(node scripts/get-resource-index.js)
+          else
+            resource_index=$(jq -r '.RESOURCE_INDEX' ./env/production/config.json)
+          fi
+
+          echo "RESOURCE_INDEX=$resource_index" | tee -a "$GITHUB_ENV"
+
+      - name: Upload the new index, overwriting the existing index
+        if: ${{ startsWith(env.RESOURCE_INDEX, 's3://') }}
+        run: |
+          aws s3 cp resources.json.gz "$RESOURCE_INDEX"
--- a/.github/workflows/index-resources.yml
+++ b/.github/workflows/index-resources.yml
@@ -12,9 +12,6 @@ on:
 
   workflow_call:
 
-# Only allow one run of the workflow per branch that triggers the workflow
-concurrency:
-  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}
 
 defaults:
   run:
@@ -46,7 +43,12 @@ jobs:
     needs: [build-ref-matrix]
     strategy:
       matrix:
-        ref: ${{ fromJson(needs.build-ref-matrix.outputs.ref-matrix) }}
+        include: ${{ fromJson(needs.build-ref-matrix.outputs.ref-matrix) }}
+    # Only allow one run of the job per resource index
+    concurrency:
+      group: ${{ github.workflow }}-${{ matrix.resource_index }}
+    env:
+        RESOURCE_INDEX: ${{ matrix.resource_index }}
     runs-on: ubuntu-latest
     permissions:
       id-token: write # needed to interact with GitHub's OIDC Token endpoint
@@ -71,20 +73,6 @@ jobs:
           node resourceIndexer/main.js \
             --gzip --output resources.json.gz \
             --resourceTypes dataset --collections core staging
-      - name: Extract RESOURCE_CONFIG from config module
-        run: |
-          # This conditional is to continue to support servers that are still
-          # using versions of this code that do not include the get-resource-index
-          # script. We can remove it once all the servers have been updated.
-          #   -Jover, 01 May 2024
-          if [[ -f ".scripts/get-resource-index.js" ]]; then
-            resource_index=$(node scripts/get-resource-index.js)
-          else
-            resource_index=$(jq -r '.RESOURCE_INDEX' ./env/production/config.json)
-          fi
-
-          echo "RESOURCE_INDEX=$resource_index" | tee -a "$GITHUB_ENV"
-
       - name: Upload the new index, overwriting the existing index
         if: ${{ startsWith(env.RESOURCE_INDEX, 's3://') }}
         run: |
--- a/.github/workflows/index-resources.yml
+++ b/.github/workflows/index-resources.yml
@@ -12,6 +12,10 @@ on:
 
   workflow_call:
 
+# Only allow one run of the workflow per branch that triggers the workflow
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref_name }}
+
 defaults:
   run:
     # This is the same as GitHub Action's `bash` keyword as of 20 June 2023:
--- a/.github/workflows/index-resources.yml
+++ b/.github/workflows/index-resources.yml
@@ -12,8 +12,47 @@ on:
 
   workflow_call:
 
+defaults:
+  run:
+    # This is the same as GitHub Action's `bash` keyword as of 20 June 2023:
+    # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell
+    #
+    # Completely spelling it out here so that GitHub can't change it out from under us
+    # and we don't have to refer to the docs to know the expected behavior.
+    shell: bash --noprofile --norc -eo pipefail {0}
+
 jobs:
+  build-ref-matrix:
+    runs-on: ubuntu-latest
+    outputs:
+      ref-matrix: ${{ steps.ref-matrix.outputs.ref-matrix }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: 'package.json'
+      - id: ref-matrix
+        name: Create ref matrix
+        env:
+          HEROKU_TOKEN: ${{ secrets.HEROKU_TOKEN_READ_PROTECTED }}
+        run: |
+          if [[ "$GITHUB_EVENT_NAME" == 'workflow_call' || \
+               ("$GITHUB_EVENT_NAME" == 'workflow_dispatch' && "$GITHUB_REF" != 'refs/heads/master') ]]; then
+            # This the commit SHA that triggered the workflow.
+            # For the workflow_call, this is in the context of the calling workflow.
+            # <https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows>
+            ref_matrix=$(jq -c --null-input '[ env.GITHUB_SHA ]')
+          else
+            ref_matrix=$(./scripts/get-resource-index-ref-matrix)
+          fi
+
+          echo "ref-matrix=$ref_matrix" | tee -a "$GITHUB_OUTPUT"
+
   rebuild-index:
+    needs: [build-ref-matrix]
+    strategy:
+      matrix:
+        ref: ${{ fromJson(needs.build-ref-matrix.outputs.ref-matrix) }}
     runs-on: ubuntu-latest
     permissions:
       id-token: write # needed to interact with GitHub's OIDC Token endpoint
@@ -24,10 +63,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          # This the commit SHA that triggered the workflow.
-          # For the workflow_call, this is in the context of the calling workflow.
-          # <https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows>
-          ref: ${{ github.sha }}
+          ref: ${{ matrix.ref }}
       - uses: actions/setup-node@v4
         with:
           node-version-file: 'package.json'
--- a/.github/workflows/index-resources.yml
+++ b/.github/workflows/index-resources.yml
@@ -10,6 +10,8 @@ on:
   # Manually triggered using GitHub's UI
   workflow_dispatch:
 
+  workflow_call:
+
 jobs:
   rebuild-index:
     runs-on: ubuntu-latest
@@ -21,6 +23,11 @@ jobs:
         shell: bash
     steps:
       - uses: actions/checkout@v4
+        with:
+          # This the commit SHA that triggered the workflow.
+          # For the workflow_call, this is in the context of the calling workflow.
+          # <https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows>
+          ref: ${{ github.sha }}
       - uses: actions/setup-node@v4
         with:
           node-version-file: 'package.json'
--- a/.github/workflows/index-resources.yml
+++ b/.github/workflows/index-resources.yml
@@ -40,17 +40,7 @@ jobs:
         env:
           HEROKU_TOKEN: ${{ secrets.HEROKU_TOKEN_READ_PROTECTED }}
         run: |
-          if [[ "$GITHUB_EVENT_NAME" == 'workflow_call' || \
-               ("$GITHUB_EVENT_NAME" == 'workflow_dispatch' && "$GITHUB_REF" != 'refs/heads/master') ]]; then
-            # This the commit SHA that triggered the workflow.
-            # For the workflow_call, this is in the context of the calling workflow.
-            # <https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows>
-            ref_matrix=$(jq -c --null-input '[ env.GITHUB_SHA ]')
-          else
-            ref_matrix=$(./scripts/get-resource-index-ref-matrix)
-          fi
-
-          echo "ref-matrix=$ref_matrix" | tee -a "$GITHUB_OUTPUT"
+          echo "ref-matrix=$(./scripts/get-resource-index-ref-matrix)" | tee -a "$GITHUB_OUTPUT"
 
   rebuild-index:
     needs: [build-ref-matrix]
--- a/aws/iam/policy/NextstrainDotOrgServerInstance.tftpl.json
+++ b/aws/iam/policy/NextstrainDotOrgServerInstance.tftpl.json
@@ -52,7 +52,8 @@
         "s3:GetObject"
       ],
       "Resource": [
-        "arn:aws:s3:::nextstrain-inventories/resources.json.gz"
+        "arn:aws:s3:::nextstrain-inventories/resources.json.gz",
+        "arn:aws:s3:::nextstrain-inventories/resources/*.json.gz"
       ]
     }
   ]
--- a/docs/resource-collection.rst
+++ b/docs/resource-collection.rst
@@ -10,11 +10,42 @@ resources (and their versions) to the user.
 The index location is set by the env/config variable ``RESOURCE_INDEX``. The
 ``RESOURCE_INDEX`` must be either a "s3://" address or a local file path. If the
 file is located on S3 it must be gzipped. The server loads this index at start
-time and refreshes it hourly. The nextstrain.org testing & production configs
-currently set this to ``s3://nextstrain-inventories/resources.json.gz``.
+time and refreshes it hourly. Resource collections can be ignored by the server
+by setting the env variable ``RESOURCE_INDEX="false"`` or by omitting it from
+your configuration JSON.
 
-Resource collections can be ignored by the server by setting the env variable
-``RESOURCE_INDEX="false"`` or by omitting it from your configuration JSON.
+
+Resource index revisions
+========================
+
+The nextstrain.org testing & production configs currently set this to
+``s3://nextstrain-inventories/resources/v<revision_number>.json.gz``.
+
+If you make any updates that changes the structure or the contents of the resource
+index JSON, then bump the ``<revision_number>`` within the configs (``env/production/config.json``)
+so that any uploads to S3 does not disrupt the production server.
+
+These updates include changes to:
+
+* any scripts within ``resourceIndexer/*``
+* ``data/manifest_core.json``
+* ``convertManifestJsonToAvailableDatasetList`` in ``src/endpoints/charon/parseManifest.js``
+* ``datasetRedirectPatterns`` in ``src/redirects.js``
+
+If you are ever unsure, it's better to just bump the revision number!
+
+Testing new revisions
+---------------------
+
+The handling of new revision numbers in Heroku review apps is currently still
+handled manually.
+
+Once you've pushed up the changes to the revision number in the config, run
+the `index-resource.yml workflow <https://github.com/nextstrain/nextstrain.org/actions/workflows/index-resources.yml>__`
+using your branch.
+
+After the new index resources JSON has been uploaded to S3, then open the PR for
+your branch and the Heroku review app should use the new revision.
 
 
 Local index generation
@@ -36,7 +67,7 @@ This will create ``./devData/core.manifest.json`` and
 ``./devData/core.inventory.csv.gz`` for the core (nextstrain-data) source and
 ``./devData/staging.manifest.json`` and ``./devData/staging.inventory.csv.gz``
 for the staging source. Alternately you can manually obtain a suitable manifest
-and inventory from ``s3://nextstrain-inventories`` 
+and inventory from ``s3://nextstrain-inventories``
 
 To generate the index using these local files run the indexer with the ``--local`` flag.
 
@@ -115,7 +146,7 @@ certain prefixes in that bucket - for instance ``nextstrain-data/config-v1`` and
 respectively.
 
 To upload the index you will need write access for
-s3://nextstrain-inventories/resources.json.gz. Note that if your aims are
+``s3://nextstrain-inventories/resources/*.json.gz``. Note that if your aims are
 limited to local development purposes this is not necessary (see `Local development`_).
 
 
@@ -123,7 +154,7 @@ Index backups
 -------------
 
 The ``nextstrain-inventories`` bucket is version enabled so past versions of
-``s3://nextstrain-inventories/resources.json.gz`` are available.
+``s3://nextstrain-inventories/resources/*.json.gz`` are available.
 
 A lifecycle rule on the s3://nextstrain-inventories bucket (`console link
 <https://s3.console.aws.amazon.com/s3/management/nextstrain-inventories/lifecycle/view?region=us-east-1&id=delete+old+versions+of+the+index>`__)
@@ -136,4 +167,4 @@ Index access by the server
 
 IAM users ``nextstrain.org`` and ``nextstrain.org-testing``, which are under
 terraform control, have read access to
-s3://nextstrain-inventories/resources.json.gz via their associated policies.
+``s3://nextstrain-inventories/resources/*.json.gz`` via their associated policies.
--- a/env/production/.terraform.lock.hcl
+++ b/env/production/.terraform.lock.hcl
@@ -5,6 +5,7 @@ provider "registry.terraform.io/hashicorp/aws" {
   version     = "4.32.0"
   constraints = "~> 4.32"
   hashes = [
+    "h1:7vBuXLVLAnkcLDxIb8QN6O5pD13KtFDTqcnF0hFnraM=",
     "h1:8AKJChT1Sgqjfdn16BayH5DonF3B9g7qQ6N+IKPulP4=",
     "zh:062c30cd8bcf29f8ee34c2b2509e4e8695c2bcac8b7a8145e1c72e83d4e68b13",
     "zh:1503fabaace96a7eea4d73ced36a02a75ec587760850e58162e7eff419dcbb31",
@@ -24,6 +25,7 @@ provider "registry.terraform.io/hashicorp/aws" {
 provider "registry.terraform.io/hashicorp/random" {
   version = "3.5.1"
   hashes = [
+    "h1:IL9mSatmwov+e0+++YX2V6uel+dV6bn+fC/cnGDK3Ck=",
     "h1:VSnd9ZIPyfKHOObuQCaKfnjIHRtR7qTw19Rz8tJxm+k=",
     "zh:04e3fbd610cb52c1017d282531364b9c53ef72b6bc533acb2a90671957324a64",
     "zh:119197103301ebaf7efb91df8f0b6e0dd31e6ff943d231af35ee1831c599188d",
--- a/scripts/get-resource-index-ref-matrix
+++ b/scripts/get-resource-index-ref-matrix
@@ -1,5 +1,6 @@
 #!/bin/bash
-# Builds a ref matrix of commit hashes for the `index-resources.yml` GH Action workflow
+# Builds a ref matrix of commit hashes and resource index URLs for the
+# `index-resources.yml` GH Action workflow
 #
 # Uses the GITHUB_SHA if the workflow was called as a reusable workflow or
 # if the workflow was manually run on a branch that is not the default branch.
@@ -22,7 +23,10 @@ main () {
         # This the commit SHA that triggered the workflow.
         # For the workflow_call, this is in the context of the calling workflow.
         # <https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows>
-        ref_matrix=$(jq -c --null-input '[ env.GITHUB_SHA ]')
+        resource_index=$(get_resource_index_at_commit "$GITHUB_SHA")
+        ref_matrix=$(jq -c --null-input \
+                        --arg RESOURCE_INDEX $(get_resource_index_at_commit "$GITHUB_SHA") \
+                        '[ {"ref": env.GITHUB_SHA, "resource_index": $RESOURCE_INDEX} ]')
     else
         ref_matrix=$(build_prod_and_canary_ref_matrix)
     fi
@@ -42,14 +46,19 @@ build_prod_and_canary_ref_matrix() {
     canary_resource_index=$(get_resource_index_at_commit "$canary_commit")
 
     if [[ "$prod_resource_index" == "$canary_resource_index" ]]; then
-        jq_array='[ $PROD_COMMIT ]'
+        jq_array='[ {"ref": $PROD_COMMIT, "resource_index": $PROD_RESOURCE_INDEX} ]'
     else
-        jq_array='[ $PROD_COMMIT, $CANARY_COMMIT ]'
+        jq_array='[
+            {"ref": $PROD_COMMIT, "resource_index": $PROD_RESOURCE_INDEX},
+            {"ref": $CANARY_COMMIT, "resource_index": $CANARY_RESOURCE_INDEX}
+        ]'
     fi
 
     ref_matrix=$(jq -c --null-input \
                     --arg PROD_COMMIT "$prod_commit" \
+                    --arg PROD_RESOURCE_INDEX "$prod_resource_index" \
                     --arg CANARY_COMMIT "$canary_commit" \
+                    --arg CANARY_RESOURCE_INDEX "$canary_resource_index" \
                     "$jq_array")
 
     echo "$ref_matrix"
--- a/scripts/get-resource-index-ref-matrix
+++ b/scripts/get-resource-index-ref-matrix
@@ -23,7 +23,6 @@ main () {
         # This the commit SHA that triggered the workflow.
         # For the workflow_call, this is in the context of the calling workflow.
         # <https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows>
-        resource_index=$(get_resource_index_at_commit "$GITHUB_SHA")
         ref_matrix=$(jq -c --null-input \
                         --arg RESOURCE_INDEX $(get_resource_index_at_commit "$GITHUB_SHA") \
                         '[ {"ref": env.GITHUB_SHA, "resource_index": $RESOURCE_INDEX} ]')
--- a/scripts/get-resource-index-ref-matrix
+++ b/scripts/get-resource-index-ref-matrix
@@ -44,23 +44,16 @@ build_prod_and_canary_ref_matrix() {
     canary_commit=$(get_heroku_slug_commit "$CANARY_APP_NAME")
     canary_resource_index=$(get_resource_index_at_commit "$canary_commit")
 
-    if [[ "$prod_resource_index" == "$canary_resource_index" ]]; then
-        jq_array='[ {"ref": $PROD_COMMIT, "resource_index": $PROD_RESOURCE_INDEX} ]'
-    else
-        jq_array='[
+    jq -c --null-input \
+        --arg PROD_COMMIT "$prod_commit" \
+        --arg PROD_RESOURCE_INDEX "$prod_resource_index" \
+        --arg CANARY_COMMIT "$canary_commit" \
+        --arg CANARY_RESOURCE_INDEX "$canary_resource_index" \
+        '[
             {"ref": $PROD_COMMIT, "resource_index": $PROD_RESOURCE_INDEX},
             {"ref": $CANARY_COMMIT, "resource_index": $CANARY_RESOURCE_INDEX}
-        ]'
-    fi
-
-    ref_matrix=$(jq -c --null-input \
-                    --arg PROD_COMMIT "$prod_commit" \
-                    --arg PROD_RESOURCE_INDEX "$prod_resource_index" \
-                    --arg CANARY_COMMIT "$canary_commit" \
-                    --arg CANARY_RESOURCE_INDEX "$canary_resource_index" \
-                    "$jq_array")
-
-    echo "$ref_matrix"
+         ]
+        | unique_by(.resource_index)'
 }
 
 get_heroku_slug_commit() {
--- a/scripts/get-resource-index-ref-matrix
+++ b/scripts/get-resource-index-ref-matrix
@@ -1,17 +1,36 @@
 #!/bin/bash
-# Checks the Nextstrain production and canary Heroku apps to determine
+# Builds a ref matrix of commit hashes for the `index-resources.yml` GH Action workflow
+#
+# Uses the GITHUB_SHA if the workflow was called as a reusable workflow or
+# if the workflow was manually run on a branch that is not the default branch.
+# Otherwise, checks the Nextstrain production and canary Heroku apps to determine
 # if they are using the same `config.RESOURCE_INDEX` and returns a JSON array
 # of the the commit hashes for the different RESOURCE_INDEX.
-#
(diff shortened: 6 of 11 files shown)


joverlee521:
[CI workflow](https://github.com/nextstrain/nextstrain.org/actions/runs/8841241789/job/24278095843#step:8:28) uploaded to the expected tmp S3 path 🎉 

However there was an error on the Heroku app side:
```
2024-04-26T00:34:32.072898+00:00 heroku[web.1]: Starting process with command `RESOURCE_INDEX=$(RESOURCE_INDEX_PREFIX="separate-indexer" ./get-resource-index) node server.js --verbose`

2024-04-26T00:34:32.591834+00:00 app[web.1]: ./get-resource-index: line 4: jq: command not found
```

And of course I didn't deploy the Terraform changes yet, so seeing Unauthorized errors:
```
2024-04-26T00:34:34.345480+00:00 app[web.1]: [verbose]	[fetch] 403 Forbidden https://nextstrain-inventories.s3.us-east-1.amazonaws.com/resources/v1.json.gz
```

joverlee521:
We could use third-party [buildpacks for jq](https://elements.heroku.com/buildpacks/szgupta/heroku-buildpack-jq) or just install jq within get-source-index if it's not available.

joverlee521:
Hmm, weird, I'm still seeing the `jq` not found error:

```
2024-04-30T03:03:27.552547+00:00 heroku[web.1]: Starting process with command `RESOURCE_INDEX=$(RESOURCE_INDEX_PREFIX="separate-indexer" ./get-resource-index) node server.js --verbose`
2024-04-30T03:03:28.407208+00:00 app[web.1]: ./get-resource-index: line 4: jq: command not found
```
Does Heroku do some sort of caching between releases?

get-resource-index
```diff
@@ -1,14 +1,44 @@
...
+    path="$(mktemp -t "jq.XXXXXX")"
+    system_name="$(uname -s | tr A-Z a-z)"
+
+    if [[ "$system_name" == "darwin" ]]; then
+        platform="osx-amd64"
+    else
+        platform="linux64"
+    fi
+
+    curl -fsSL --proto '=https' "https://github.com/jqlang/jq/releases/download/jq-1.6/jq-$platform" \
```
jameshadfield, line 37: Instead of downloading `jq` because it's not available in our Heroku environment, why not rewrite `get-resource-index` as a node script and utilise the native JSON parsing in node?

.github/workflows/ci.yml
```diff
@@ -41,6 +41,19 @@ jobs:
           name: logs
           path: test/server.log
 
+  index_resources:
+    if: |2
```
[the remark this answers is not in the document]
joverlee521: It's a [YAML block indentation indicator](https://yaml.org/spec/1.2.2/#8111-block-indentation-indicator) that honestly I just copied from the `deploy` job below. 
```yaml
  deploy:
    if: |2
         github.repository == 'nextstrain/nextstrain.org'
      && github.event_name == 'push'
      && github.ref == 'refs/heads/master'
```

It allows the first line of the block to start with extra spaces for readability.

get-resource-index
joverlee521: That was going to be my next step if my latest changes didn't work, but the Heroku logs show the app using the correct tmp resource index now:

```
2024-05-01T00:14:11.886398+00:00 app[web.1]: [verbose]  [RESOURCE INDEX UPDATE S3] Fetching index from s3://nextstrain-inventories/resources/tmp/separate-indexer/v1.json.gz
```

---
I can still rewrite it as a node script if that's preferred, jq was just simpler for me than a node script 😅

.github/workflows/index-resources.yml
```diff
@@ -34,6 +119,18 @@ jobs:
           node resourceIndexer/main.js \
             --gzip --output resources.json.gz \
             --resourceTypes dataset --collections core staging
+      - name: Extract RESOURCE_INDEX from config
+        env:
+          RESOURCE_INDEX_PREFIX: ${{ inputs.resource_index_prefix }}
+        run: |
+          if [[ -f "./get-resource-index" ]]; then
```
jameshadfield, line 126: (Just skimming this PR - will do a proper review later) - this conditional is just to allow the action to work while the canary/production servers are still using versions of the code prior to the introduction of `get-resource-index`?

get-resource-index
jameshadfield: I would prefer it to be in node. 

Could we also move it into `/scripts/`?

env/production/.terraform.lock.hcl
tsibley, line 1: > we can use `terraform providers lock` to pre-populate hashes for different platforms.¹

Looks like the added hashes are for `darwin_arm64` (and existing ones are for `linux_amd64`), so we're just missing `darwin_amd64` if any devs are running that combo. I think fine to leave it as-is for now, but good to know about `terraform providers lock`!

.github/workflows/index-resources.yml
```diff
@@ -21,6 +28,8 @@ jobs:
         shell: bash
     steps:
       - uses: actions/checkout@v4
+        with:
+          ref: ${{ inputs.ref || github.sha }}
```
tsibley: Why in a `workflow_call` invocation is `github.sha` not what we want (it should be the `github.sha` of the caller¹)? i.e. why do we need `inputs.ref`? i.e. why would we want to run CI for ref A but reindex with ref B?

¹ "When a reusable workflow is triggered by a caller workflow, the `github` context is always associated with the caller workflow." (https://docs.github.com/en/actions/using-workflows/reusing-workflows)

.github/workflows/ci.yml
```diff
@@ -41,6 +41,19 @@ jobs:
           name: logs
           path: test/server.log
 
+  index_resources:
```
tsibley, line 59: _non-blocking, minor nit_

We tend to use hyphens in our job names (and workflow names), not underscores.

.github/workflows/ci.yml
```diff
@@ -41,6 +41,19 @@ jobs:
...
+         github.repository == 'nextstrain/nextstrain.org'
+      && github.event_name == 'push'
+      && github.ref == 'refs/heads/master'
+    needs: test
+    permissions:
+      id-token: write # needed to interact with GitHub's OIDC Token endpoint
+      contents: read
+    uses: ./.github/workflows/index-resources.yml
+    with:
+      ref: ${{ github.sha }}
```
tsibley, line 69: _follow-up to prior comment_

I think `github.sha` will be the same inside the called workflow without having to pass it thru like this.

.github/workflows/index-resources.yml
joverlee521: Yup, I'll add a comment.

.github/workflows/index-resources.yml
```diff
@@ -10,6 +10,13 @@ on:
   # Manually triggered using GitHub's UI
   workflow_dispatch:
 
+  workflow_call:
+    inputs:
+      ref:
+        description: Ref of the repo to use to generate the resource index.
+        type: string
+        required: true
```
tsibley, line 31: With automated calls happening more frequently than once a day or an ad-hoc, manual basis, I think we want to add a `concurrency` restriction to this workflow (or maybe just the job?) so that multiple invocations aren't stepping on each other's toes.

tsibley (changes requested):
This behaviour will be good to have! Checkpointing my review commentary for now. Will finish up tomorrow.

Procfile
```diff
@@ -1 +1 @@
-web: RESOURCE_INDEX=$(RESOURCE_INDEX_PREFIX="$HEROKU_BRANCH" ./get-resource-index) node server.js --verbose
+web: RESOURCE_INDEX=$(RESOURCE_INDEX_PREFIX="$HEROKU_BRANCH" node ./scripts/get-resource-index.js) node server.js --verbose
```
jameshadfield, line 1: A more canonical way would be `node ./scripts/get-resource-index.js $HEROKU_BRANCH` and then access this via `process.argv[2]` in the script. But it's working fine as is, so no need to change.

jameshadfield, line 1: Do you have some docs on `HEROKU_BRANCH`? I can only find [this page](https://devcenter.heroku.com/articles/github-integration-review-apps) which implies it's only for review apps, but also implies that it's not stable and is likely to change in the future.

It'd be clearer to prepend the server call with a env-setting function, like `set-review-app-variables` which could self-document the steps that are going on, and perhaps query other env variables to ensure we are indeed running on a review app.

jameshadfield (changes requested):
This is going to be a nice improvement. I got a bit lost in the second half of the commits as the interconnectedness ramps up, so I'll try to re-read them tomorrow if I get a change. 

One missing piece seems to be documentation on incrementing the resource id¹ (e.g. `v1.json` to `v2.json`), and the relationship between the test/production environments vs local indexes once you do increment it.


¹ I [still think](https://github.com/nextstrain/nextstrain.org/issues/829#issuecomment-2065182925) it shouldn't be called `v*` as it's overloading the word "version". But I won't keep pressing on this.

.github/workflows/get-heroku-app-commit-index.yml
```diff
@@ -0,0 +1,69 @@
...
+            curl -n https://api.heroku.com/pipelines/"$HEROKU_PIPELINE"/latest-releases \
+              --fail --silent --show-error \
+              -H "Accept: application/vnd.heroku+json; version=3" \
+              -H "Authorization: Bearer $HEROKU_TOKEN" \
+              | jq -r '.[] | select(.app | .name == env.APP_NAME) | .slug.id')
+
+          echo "SLUG_ID=$slug_id" | tee -a "$GITHUB_ENV"
+
+      - id: commit_hash
+        name: Get latest commit hash for ${{ env.app_name }}
```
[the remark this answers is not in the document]
joverlee521: Ah, that would be a typo! Thanks for catching it.

aws/iam/policy/NextstrainDotOrgServerInstance.tftpl.json
```diff
@@ -52,7 +52,8 @@
         "s3:GetObject"
       ],
       "Resource": [
-        "arn:aws:s3:::nextstrain-inventories/resources.json.gz"
```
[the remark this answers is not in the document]
joverlee521: OH good to know! I added this to the PR checklist.

.github/workflows/index-resources.yml
joverlee521: Good idea! I'll probably use the [concurrency groups](https://docs.github.com/en/actions/using-jobs/using-concurrency#example-concurrency-groups) to restrict by `github.head_ref` since each PR will be uploading different to different tmp S3 paths.

.github/workflows/index-resources.yml
joverlee521: Ah, yeah looking at this now, this was unnecessary for this particular commit. It's clearer in https://github.com/nextstrain/nextstrain.org/pull/841/commits/238db45f13995856cef8701cd5b530cbcaff1da7 where I've conditionalized jobs on the presence of the `input.ref` rather than a list of workflow events.

.github/workflows/index-resources.yml
```diff
@@ -17,19 +17,76 @@ on:
...
+        run: |
+          if [[ "$INPUT_REF" != '' ]]; then
+            jq_filter='[ env.INPUT_REF ]'
+          elif [[ "$PRODUCTION_RESOURCE_INDEX" == "$CANARY_RESOURCE_INDEX" ]]; then
+            jq_filter='[ env.PRODUCTION_COMMIT ]'
+          else
+            jq_filter='[ env.PRODUCTION_COMMIT, env.CANARY_COMMIT ]'
+          fi
+
+          ref_matrix=$(jq -c --null-input "$jq_filter")
```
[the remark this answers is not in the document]
joverlee521: Ah, bad naming on my part! Like you said, this is just my way of ensuring I am building a properly formatted JSON array.

.github/workflows/index-resources.yml
```diff
@@ -16,13 +16,25 @@ on:
           If not provided, will use commits of the latest slug of nextstrain-server and nextstrain-canary Heroku apps.
         type: string
         required: false
+      resource_index_prefix:
```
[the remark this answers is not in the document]
joverlee521: > What's the use case for running this manually? 

The one use case I can think of is if we create a review app directly within Heroku without a PR that triggers the automated CI workflow rebuild of the index. We would manually run the index-resources.yml to upload the tmp index.

Procfile
```diff
@@ -1 +1 @@
-web: node server.js --verbose
+web: RESOURCE_INDEX=$(RESOURCE_INDEX_PREFIX="$HEROKU_BRANCH" ./get-resource-index) node server.js --verbose
```
tsibley: I find this sort of weird. Especially since for production apps (e.g. nextstrain.org, next.nextstrain.org) we need `RESOURCE_INDEX` to be calculated here to match the `config.json`. And the `config.json` in use is hardcoded everywhere, instead of respecting `CONFIG_FILE`.

What if instead of doing this just-in-time overriding of `RESOURCE_INDEX` we set it as a config var on the review app at creation (or very shortly afterwards)? This could be by moving to creating review apps ourselves as part of CI (using the [API endpoint](https://devcenter.heroku.com/articles/platform-api-reference#review-app-create)) or using the [post-deploy script](https://devcenter.heroku.com/articles/github-integration-review-apps#the-postdeploy-script) to update the app with the new config var.

scripts/get-resource-index.js
```diff
@@ -0,0 +1,26 @@
...
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+const __basedir = path.join(__dirname, "..");
+const productionConfig = path.join(__basedir, 'env/production/config.json');
+
+const resourceIndexPrefix = process.env.RESOURCE_INDEX_PREFIX;
+
+function main() {
+  const config = JSON.parse(fs.readFileSync(productionConfig));
+  var resourceIndex = new URL(config['RESOURCE_INDEX']);
```
tsibley, line 15: This could be accessing `RESOURCE_INDEX` from our `config` module directly, e.g.

```js
import { RESOURCE_INDEX } from './src/config.js';
```

Procfile
tsibley: Note that we already use one of those variables, `HEROKU_PR_NUMBER`, to detect review app-ness:

https://github.com/nextstrain/nextstrain.org/blob/31ecbc4a7b9d5ab72341040a1e06371551cec659/src/config.js#L36-L46

 and adjust config:

https://github.com/nextstrain/nextstrain.org/blob/31ecbc4a7b9d5ab72341040a1e06371551cec659/src/config.js#L391-L412

tsibley (changes requested):
It will be helpful to have separate indexes!

I wonder though about approaching the machinery for setting them up a bit differently, as noted in my comments.

.github/workflows/index-resources.yml
tsibley: Right, I saw that once I got further. The current conditionals are like:

```yaml
if: ${{ inputs.ref == '' }}
```

and I think that's quite a bit less clear than:

```yaml
if: github.event_name == 'schedule'
```

aws/iam/policy/NextstrainDotOrgServerInstance.tftpl.json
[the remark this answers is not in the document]
tsibley: It'd be nice to have that sort of thing in Terraform for exactly this reason.

.github/workflows/index-resources.yml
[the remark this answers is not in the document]
tsibley: In case it's clarifying, the difference between:

```bash
ref_matrix=$(jq -c --null-input "[ env.PRODUCTION_COMMIT, env.CANARY_COMMIT ]")
```

and something like:

```bash
ref_matrix="[ \"$PRODUCTION_COMMIT\", \"$CANARY_COMMIT\" ]"
```

is that the former ensures a valid JSON value and the latter does not.

.github/workflows/index-resources.yml
joverlee521: Ah, I wasn't going to use 
```yaml
if: github.event_name == 'schedule'
```
because sometimes we want to manually trigger index rebuilds via `workflow_dispatch`. 

I think checking `input.ref` is able to catch all cases where we don't want to run the default production/canary updates.

.github/workflows/get-heroku-app-commit-index.yml
```diff
@@ -0,0 +1,69 @@
...
+        description: Heroku app name
+        type: string
+        required: true
+    outputs:
+      commit_hash:
+        description: The commit hash of the latest Heroku slug for the app.
+        value: ${{ jobs.heroku_app_info.outputs.commit_hash }}
+      resource_index:
+        description: The RESOURCE_INDEX extracted from the env/production/config.json
+        value: ${{ jobs.heroku_app_info.outputs.resource_index }}
```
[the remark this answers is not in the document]
joverlee521: > Or even better, perhaps, what if we exposed this deployment/version metadata via our own API endpoint?

I like this idea for the scheduled updates of the `RESOURCE_INDEX`. We'd still have to worry about the initial deployment of the server though. I'm leaning towards deploying the review apps ourselves (as you [suggested](https://github.com/nextstrain/nextstrain.org/pull/841#discussion_r1588367534)) rather than depending on the `config.json`.

.github/workflows/index-resources.yml
tsibley: Hmm. I think it's fairly confusing/tangled. AIUI, `input.ref` could be provided two places:

1. `workflow_call` (i.e. invoked by another workflow) where a `ref` value was passed. The only invoking workflow is CI, which passes `ref: ${{ github.sha }}`.

2. `workflow_dispatch` (i.e. manual invocation) where the `input.ref` textbox is filled in. Note that this is **different** than choosing the ref at which to run the workflow, which is an implicit input that's required. When `input.ref` is provided, it should ~always match the implicit ref input, but nothing guarantees that.

It didn't initially seem to me we'd ever want to do (2), but [you proposed a use case](https://github.com/nextstrain/nextstrain.org/pull/841#discussion_r1588355716).

So it seems like the spelled-out logic would be something like:

- If `github.event_name == 'schedule'`, then update for our two production Heroku apps.
- If `github.event_name == 'workflow_dispatch' && github.ref == 'refs/heads/master'`, then update for our two production Heroku apps.
- If `github.event_name == 'workflow_dispatch' && github.ref != 'refs/heads/master'`, then update for `github.sha`.
- If `github.event_name == 'workflow_call'`, then update for `github.sha`.

Procfile
tsibley: I was hesitant to suggest that because it means that the value passed in as `RESOURCE_INDEX` won't be what's actually used depending on unseen other values (i.e. `HEROKU_PR_NUMBER`), and that seemed undesirable/confusing and a first for the config. I think `RESOURCE_INDEX` for review apps should still be set outside of the app, like it is for production.

.github/workflows/get-heroku-app-commit-index.yml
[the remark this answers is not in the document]
tsibley: Or instead of provisioning the index in CI, can we provision the index during the [Heroku release phase](https://devcenter.heroku.com/articles/release-phase)? This would, for both production apps and review apps, solve the "which code to run, which index to use" issues, and I think overall makes the coupling pretty clear.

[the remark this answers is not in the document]
joverlee521: > Or instead of provisioning the index in CI, can we provision the index during the [Heroku release phase](https://devcenter.heroku.com/articles/release-phase)?

(Hah, we've [come full circle](https://github.com/nextstrain/nextstrain.org/issues/829#issuecomment-2062219926) back to the Heroku release phase 😆)
I'm also fine with using the release phase if we are open to updating the server IAM user permissions.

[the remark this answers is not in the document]
tsibley: Oh boy, I'd missed/forgotten that!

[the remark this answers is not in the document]
jameshadfield: > This would, for both production apps and review apps, solve the "which code to run, which index to use" issues

The resource indexer is built under fail-fast design principles. For heroku review apps this is fine. For production apps less so. 

Reading through this PR the complexity really ramps up when we start dealing with review apps. I'd support removing that complexity from actions and changing the procfile to something like `web: create-local-index-if-review-app && node server.js ...`, but keeping the canary/production site updated via github actions as done in this PR (with suggested changes)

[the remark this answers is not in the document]
tsibley: > The resource indexer is built under fail-fast design principles. For heroku review apps this is fine. For production apps less so.

If the release phase errors, then the deploy is aborted. Nothing breaks, the site keeps working. We would have to address the release phase errors before deploying again.

> changing the procfile to something like `web: create-local-index-if-review-app && node server.js ...`,

But in that case we'd still need to grant permissions that the actual server doesn't need in order for review apps to read the inventories?

I think we can keep index generation in GitHub Actions for both production and review apps but simplify the workflows a lot by moving automatic deploy of review apps into CI. This then parallels the production process too, and brings them more in line.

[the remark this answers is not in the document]
joverlee521: A lot of the complexity comes from keeping things in sync between GH Actions and Heroku, so to reduce complexity, we should just do everything in one or the other. Since we rely on GH Actions for CI (see https://github.com/nextstrain/nextstrain.org/issues/850), it makes sense to move everything into GH Actions.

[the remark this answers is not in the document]
tsibley: That sounds good to me. But let's talk about what that looks like for review apps?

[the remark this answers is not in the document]
joverlee521: Yeah, I would like to scope this PR to just supporting separate indexes for prod/canary. The work on the review apps can be done separately.

.github/workflows/index-resources.yml
joverlee521: Spelled-out the cases to use `GITHUB_SHA` in https://github.com/nextstrain/nextstrain.org/pull/841/commits/a90c1c35da5265a1680ae513bcb701615feaba94.

.github/workflows/ci.yml
joverlee521: Updated to hyphens in https://github.com/nextstrain/nextstrain.org/pull/841/commits/cd51343ad370085e39f25f738318962bab328f61.

.github/workflows/ci.yml
joverlee521: Thanks for talking through this in https://github.com/nextstrain/nextstrain.org/pull/841#discussion_r1586923809. Now the workflow just directly references `GITHUB_SHA`.

.github/workflows/index-resources.yml
joverlee521: Added concurrency restriction in https://github.com/nextstrain/nextstrain.org/pull/841/commits/33e8b990b51bd096d27d5cb14305e8b4c15c78b4, but I'm not entirely happy with it since different branches can still be using the same RESOURCE_INDEX. 

Hmm, I guess move the `aws cp` to a separate job and put a concurrency restriction on that job based on the `RESOURCE_INDEX` value.

.github/workflows/get-heroku-app-commit-index.yml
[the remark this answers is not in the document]
joverlee521: No longer relevant since I moved this piece into a separate shell script in https://github.com/nextstrain/nextstrain.org/pull/841/commits/a90c1c35da5265a1680ae513bcb701615feaba94

.github/workflows/index-resources.yml
```diff
@@ -16,13 +16,25 @@ on:
...
 
   workflow_call:
     inputs:
       ref:
         description: Ref of the repo to use to generate the resource index.
         type: string
         required: true
+      resource_index_prefix:
+        description: |
+          A prefix to add to the uploaded resource index.
```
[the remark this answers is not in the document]
joverlee521: No longer relevant for this PR since I'm going to do the review-app stuff separately.

.github/workflows/ci.yml
```diff
@@ -41,18 +41,33 @@ jobs:
...
-      && github.ref == 'refs/heads/master'
-    needs: test
+      && contains(fromJSON('["push", "pull_request"]'), github.event_name)
     permissions:
       id-token: write # needed to interact with GitHub's OIDC Token endpoint
       contents: read
     uses: ./.github/workflows/index-resources.yml
     with:
       ref: ${{ github.sha }}
+      resource_index_prefix: ${{ needs.get_branch.outputs.github_branch }}
```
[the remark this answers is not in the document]
joverlee521: No longer relevant for this PR since I'm going to do the review-app stuff separately.

.github/workflows/index-resources.yml
```diff
@@ -17,19 +17,76 @@ on:
...
+          echo "ref_matrix=$ref_matrix" | tee -a "$GITHUB_OUTPUT"
+
   rebuild-index:
+    needs: [build_ref_matrix]
+    # build_ref_matrix's optional dependency on production_app and canary_app
+    # requires an additional condition to run rebuild-index. Otherwise,
+    # rebuild-index will be skipped if production_app/canary_app are skipped.
+    # <https://github.com/actions/runner/issues/2205>
+    #   -Jover 25 April 2024
+    if: ${{ !cancelled() && needs.build_ref_matrix.result == 'success' }}
```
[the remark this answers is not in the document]
joverlee521: Yay, this went away since I moved to a shell script in https://github.com/nextstrain/nextstrain.org/pull/841/commits/a90c1c35da5265a1680ae513bcb701615feaba94

scripts/get-resource-index.js
joverlee521: Updated in https://github.com/nextstrain/nextstrain.org/pull/841/commits/538c63c3cb10797e463e6d56e8af91d0da146d34

.github/workflows/get-heroku-app-commit-index.yml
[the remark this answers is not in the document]
joverlee521: To keep things simple, I've opted for a separate bash script to grab the Heroku/deployment metadata. This ~fine for now since we not setting any `RESOURCE_INDEX` envvars in Heroku. 

I still really like the idea of surfacing the deployment metadata through our own API endpoint, so tracking that separately in https://github.com/nextstrain/nextstrain.org/issues/853

Procfile
joverlee521: Tracking creating Heroku apps through CI in https://github.com/nextstrain/nextstrain.org/issues/854

joverlee521:
Thanks for all of the feedback @jameshadfield and @tsibley! 

To reduce complexity of this PR, I've scoped it down to just separating the index resources JSON and kicking the can on supporting review apps. I've [documented](https://github.com/nextstrain/nextstrain.org/pull/841/commits/b7d050ca60981d0c1f7d7b67d83e891967092dfc) the manual steps to create a new revision of the resources JSON for future branches for now. It's definitely not foolproof, but I'd like to separate the index resources JSON earlier rather than later.

This PR is ready for re-review when you get the chance!

scripts/get-resource-index-ref-matrix
```diff
@@ -0,0 +1,91 @@
...
+# of the the commit hashes for the different RESOURCE_INDEX.
+#
+# Intended to be used within the `index-resources.yml` GH Action workflow.
+
+: "${HEROKU_TOKEN:?The HEROKU_TOKEN environment variable is required.}"
+: "${HEROKU_PIPELINE:=38f67fc7-d93c-40c6-a182-501da2f89d9d}"
+
+: "${PROD_APP_NAME:=nextstrain-server}"
+: "${CANARY_APP_NAME:=nextstrain-canary}"
+
```
genehack, line 13: Maybe y'all don't do this in GH Actions for reasons, but IMO for `bash` stuff it's always good to have a `set -euo pipefail` near the top of the script, so errors actually count.

joverlee521: Ah, my bad, we do usually do set that at the top! Thanks for catching that!

joverlee521:
Rebased to fix small merge conflict in CI workflow.

docs/resource-collection.rst
```diff
@@ -10,11 +10,42 @@ resources (and their versions) to the user.
...
+so that any uploads to S3 does not disrupt the production server.
+
+These updates include changes to:
+
+* any scripts within ``resourceIndexer/*``
+* ``data/manifest_core.json``
+* ``convertManifestJsonToAvailableDatasetList`` in ``src/endpoints/charon/parseManifest.js``
+* ``datasetRedirectPatterns`` in ``src/redirects.js``
+
+If you are ever unsure, it's better to just bump the revision number!
```
tsibley, line 35: _non-blocking_

How much simpler would it be if we used the commit id as the "index version" instead of a monotonically-increasing number? (I know we floated both ideas, but I don't recall discussing which to use…)

.github/workflows/index-resources.yml
tsibley: Nod. Restricting [`concurrency` at the job level](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idconcurrency) by `RESOURCE_INDEX` could work without the separate job if the [job matrix grew to include `RESOURCE_INDEX` too](https://github.com/nextstrain/nextstrain.org/pull/841#discussion_r1599128769) instead of just `ref`.

.github/workflows/ci.yml
```diff
@@ -95,16 +95,28 @@ jobs:
...
          github.repository == 'nextstrain/nextstrain.org'
       && (   (github.event_name == 'push' && github.ref == 'refs/heads/master')
           || (github.event_name == 'workflow_dispatch' && inputs.heroku-app) )
 
-    # Wait for "build" and "test" jobs above to pass.
+    # Wait for "build", "test", and "index-resources" jobs above to pass.
     needs:
       - build
       - test
+      - index-resources
```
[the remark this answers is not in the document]
joverlee521: Oy vey, [GH Action docs on `needs`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds):

> If a job fails or is skipped, all jobs that need it are skipped unless the jobs use a conditional expression that causes the job to continue.

So the deploy conditional expression would need to be updated:
```diff
diff --git a/.github/workflows/ci.yml b/.github/workflows/ci.yml
index a7d67a28..ff76f91b 100644
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -108,7 +108,11 @@ jobs:
 
   deploy:
     if: |2
-         github.repository == 'nextstrain/nextstrain.org'
+         !cancelled()
+      && needs.build.result == 'success'
+      && needs.test.result == 'success'
+      && contains(fromJSON('["success", "skipped"]'), needs.index-resource.result)
+      && github.repository == 'nextstrain/nextstrain.org'
       && (   (github.event_name == 'push' && github.ref == 'refs/heads/master')
           || (github.event_name == 'workflow_dispatch' && inputs.heroku-app) )
```

scripts/get-resource-index-ref-matrix
```diff
@@ -0,0 +1,92 @@
...
+}
+
+get_heroku_slug_commit() {
+    local app_name="$1"
+    local slug_id commit_hash
+    slug_id=$(curl -n https://api.heroku.com/pipelines/"$HEROKU_PIPELINE"/latest-releases \
+                --fail --silent --show-error \
+                -H "Accept: application/vnd.heroku+json; version=3" \
+                -H "Authorization: Bearer $HEROKU_TOKEN" \
+                | jq -r --arg APP_NAME "$app_name" '.[] | select(.app | .name == $APP_NAME) | .slug.id')
```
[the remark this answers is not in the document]
joverlee521: Ah, good to know!

scripts/get-resource-index-ref-matrix
```diff
@@ -0,0 +1,92 @@
...
+    # This conditional is to continue to support servers that are still
+    # using versions of this code that do not include the get-resource-index
+    # script. We can remove it once all the servers have been updated.
+    #   -Jover, 01 May 2024
+    if [[ -f "$repo/scripts/get-resource-index.js" ]]; then
+        npm ci --prefix "$repo"
+        resource_index=$(node "$repo"/scripts/get-resource-index.js)
+    else
+        resource_index=$(jq -r '.RESOURCE_INDEX' "$repo"/env/production/config.json)
+    fi
```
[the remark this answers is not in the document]
joverlee521: Hah, yup! Not sure why I didn't do this to begin with!

.github/workflows/index-resources.yml
```diff
@@ -12,8 +12,44 @@ on:
...
+        run: |
+          if [[ "$GITHUB_EVENT_NAME" == 'workflow_call' || \
+               ("$GITHUB_EVENT_NAME" == 'workflow_dispatch' && "$GITHUB_REF" != 'refs/heads/master') ]]; then
+            # This the commit SHA that triggered the workflow.
+            # For the workflow_call, this is in the context of the calling workflow.
+            # <https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows>
+            ref_matrix=$(jq -c --null-input '[ env.GITHUB_SHA ]')
+          else
+            ref_matrix=$(./scripts/get-resource-index-ref-matrix)
+          fi
```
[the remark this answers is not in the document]
joverlee521: I wanted the condition of GitHub events to be immediately obvious within the GH Action workflow, but I can also move everything to within `./scripts/get-resource-index-ref-matrix`.

docs/resource-collection.rst
joverlee521: Hmm, if we use commit id, we wouldn't run into potential version number clashes between branches. 

I thought using commit id comes with it's own weirdness if we continue to store the `RESOURCE_INDEX` in the config.json. E.g. the commit id in the `RESOURCE_INDEX` would be pointing to the parent commit id. Also, instead of updating it manually, we would probably want to set up a commit hook to automatically update the `RESOURCE_INDEX` on every commit?

.github/workflows/index-resources.yml
joverlee521: Updated concurrency to job level with `RESOURCE_INDEX` in https://github.com/nextstrain/nextstrain.org/pull/841/commits/4da1e8c8340a379b8a32daeaa65a80834482b6f1

Procfile
joverlee521: No longer relevant for this PR since I'm going to do the review-app stuff separately.

scripts/get-resource-index-ref-matrix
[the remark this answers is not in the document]
joverlee521: Updated in https://github.com/nextstrain/nextstrain.org/pull/841/commits/bc6dfea0fe9fea4d0a8296ca08219ee175f57eee

scripts/get-resource-index-ref-matrix
```diff
@@ -0,0 +1,92 @@
...
+                    | jq -r '.commit')
+
+    echo "$commit_hash"
+}
+
+
+get_resource_index_at_commit() {
+    local commit_hash="$1"
+    local repo repo_archive resource_index
+    repo="$(mktemp -d nextstrain-dot-org-repo-$commit_hash-XXXXXX)"
```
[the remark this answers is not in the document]
joverlee521: Thanks for tip! Fixed up in https://github.com/nextstrain/nextstrain.org/pull/841/commits/a31877c1cec116228cf0ce19d907484f65877cb2

scripts/get-resource-index-ref-matrix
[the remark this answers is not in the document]
joverlee521: Updated in https://github.com/nextstrain/nextstrain.org/pull/841/commits/4da1e8c8340a379b8a32daeaa65a80834482b6f1

.github/workflows/index-resources.yml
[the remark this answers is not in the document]
joverlee521: Updated in https://github.com/nextstrain/nextstrain.org/pull/841/commits/2866d9cf7fe3ac7f7e9240e291776c09855fed90

scripts/get-resource-index-ref-matrix
```diff
@@ -0,0 +1,92 @@
...
+                --fail --silent --show-error \
+                -H "Accept: application/vnd.heroku+json; version=3" \
+                -H "Authorization: Bearer $HEROKU_TOKEN" \
+                | jq -r --arg APP_NAME "$app_name" '.[] | select(.app | .name == $APP_NAME) | .slug.id')
+
+    commit_hash=$(curl -n https://api.heroku.com/apps/"$app_name"/slugs/"$slug_id" \
+                    --fail --silent --show-error \
+                    -H "Accept: application/vnd.heroku+json; version=3" \
+                    -H "Authorization: Bearer $HEROKU_TOKEN" \
+                    | jq -r '.commit')
```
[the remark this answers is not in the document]
joverlee521: Fixed up in https://github.com/nextstrain/nextstrain.org/pull/841/commits/968afbe20419a9f07bcf7d49fb5e414b1f24945f

.github/workflows/ci.yml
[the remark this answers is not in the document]
joverlee521: Updated in https://github.com/nextstrain/nextstrain.org/pull/841/commits/784b56c36ea5d30aea96f9b88eb3477b0109b0c3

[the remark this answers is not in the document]
joverlee521: [Testing a manual run of CI to deploy to nextstrain-dev](https://github.com/nextstrain/nextstrain.org/actions/runs/9084575496)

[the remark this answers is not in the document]
joverlee521: Hmm, the [`deploy` job still got skipped](https://github.com/nextstrain/nextstrain.org/actions/runs/9084575496/job/24966030064) 🤔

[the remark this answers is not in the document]
joverlee521: Ah my mistake...[fixed up](https://github.com/nextstrain/nextstrain.org/compare/784b56c36ea5d30aea96f9b88eb3477b0109b0c3..159ccf767b619600b74eefb489ed947e92f5ab58) and [testing again](https://github.com/nextstrain/nextstrain.org/actions/runs/9084716969)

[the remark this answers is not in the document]
joverlee521: Aaand found another typo...[fixed up](https://github.com/nextstrain/nextstrain.org/compare/159ccf767b619600b74eefb489ed947e92f5ab58..cf5035f0d655abba959006f7a841e8a9ef4751c1) and [testing again](https://github.com/nextstrain/nextstrain.org/actions/runs/9085210452)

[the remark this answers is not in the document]
joverlee521: Yay, [deployment to `nextstrain-dev` worked](https://github.com/nextstrain/nextstrain.org/actions/runs/9085210452) after I fixed all of my dumb mistakes.

docs/resource-collection.rst
joverlee521: Chatted more about this with @tsibley in our 1:1 today.

If we do use the commit id as the version, then we would probably stop hard-coding the `RESOURCE_INDEX` within the config.json and just set it within the Heroku slug. It would be less obvious of what `RESOURCE_INDEX` the deployed site is using unless you download the slug or we implement https://github.com/nextstrain/nextstrain.org/issues/853. 

---

Leaving this as a future decision as we can make do with the number versions for now.

tsibley: Ah, yeah. Let's not rewrite files on every single commit! We could instead do something similar as part of the build process (e.g. bake in an updated `RESOURCE_INDEX`, or bake in a commit id in a file that's used at runtime to construct the right path, etc.). This could even be done via [Git's `ident` munging](https://git-scm.com/docs/gitattributes#_ident) (à la RCS). But let's leave that for the future, if ever.

scripts/get-resource-index-ref-matrix
```diff
@@ -22,7 +23,10 @@ main () {
         # This the commit SHA that triggered the workflow.
         # For the workflow_call, this is in the context of the calling workflow.
         # <https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows>
-        ref_matrix=$(jq -c --null-input '[ env.GITHUB_SHA ]')
+        resource_index=$(get_resource_index_at_commit "$GITHUB_SHA")
+        ref_matrix=$(jq -c --null-input \
+                        --arg RESOURCE_INDEX $(get_resource_index_at_commit "$GITHUB_SHA") \
+                        '[ {"ref": env.GITHUB_SHA, "resource_index": $RESOURCE_INDEX} ]')
```
tsibley, line 29: _minor, non-blocking_

```suggestion
        resource_index=$(get_resource_index_at_commit "$GITHUB_SHA")
        ref_matrix=$(jq -c --null-input \
                        --arg RESOURCE_INDEX "$resource_index" \
                        '[ {"ref": env.GITHUB_SHA, "resource_index": $RESOURCE_INDEX} ]')
```

or 

```suggestion
        ref_matrix=$(jq -c --null-input \
                        --arg RESOURCE_INDEX $(get_resource_index_at_commit "$GITHUB_SHA") \
                        '[ {"ref": env.GITHUB_SHA, "resource_index": $RESOURCE_INDEX} ]')
```

scripts/get-resource-index-ref-matrix
```diff
@@ -42,14 +46,19 @@ build_prod_and_canary_ref_matrix() {
...
+            {"ref": $CANARY_COMMIT, "resource_index": $CANARY_RESOURCE_INDEX}
+        ]'
     fi
 
     ref_matrix=$(jq -c --null-input \
                     --arg PROD_COMMIT "$prod_commit" \
+                    --arg PROD_RESOURCE_INDEX "$prod_resource_index" \
                     --arg CANARY_COMMIT "$canary_commit" \
+                    --arg CANARY_RESOURCE_INDEX "$canary_resource_index" \
                     "$jq_array")
```
tsibley: _minor, non-blocking_

This could be simplified further a bit by doing the comparison within jq, e.g. something like:

```jq
  [
    {"ref": $PROD_COMMIT, "resource_index": $PROD_RESOURCE_INDEX},
    {"ref": $CANARY_COMMIT, "resource_index": $CANARY_RESOURCE_INDEX}
  ]
| unique_by(.resource_index)
```

tsibley: Also, instead of storing the jq output in `ref_matrix` and then immediately echoing it, the jq command can just be run outside of the process interpolation.

tsibley (approved):

scripts/get-resource-index-ref-matrix
joverlee521: Hah, not sure how that sneaked in. Thanks!

scripts/get-resource-index-ref-matrix
joverlee521: Updated, thanks!

joverlee521:
I'm planning to merge this tomorrow ~morning, so I can monitor the first CI deployment.

.github/workflows/index-resources.yml
```diff
@@ -10,8 +10,45 @@ on:
...
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version-file: 'package.json'
+      - id: ref-matrix
+        name: Create ref matrix
+        env:
+          HEROKU_TOKEN: ${{ secrets.HEROKU_TOKEN_READ_PROTECTED }}
+        run: |
+          echo "ref-matrix=$(./scripts/get-resource-index-ref-matrix)" | tee -a "$GITHUB_OUTPUT"
```
tsibley, line 40: This job unfortunately ends up [not failing](https://github.com/nextstrain/nextstrain.org/actions/runs/9116694184/job/25066044977#step:4:7) even when `./scripts/get-resource-index-ref-matrix` errors because the error exit code is ignored within the process substitution, e.g.

```console
$ if (set -euo pipefail; echo "x=$(exit 1)"); then echo ok; else echo error; fi
x=
ok
```

```bash
(echo -n "ref-matrix="; ./scripts/get-resource-index-ref-matrix) | tee -a "$GITHUB_OUTPUT"
```

would work as a substitute.

joverlee521:
The first merge/deploy failed because I missed the secrets weren't being passed from CI to index-resources. Fixed in #864. 

However still running into some issues:

1. The second merge/deploy succeeded, however, the build-index-matrix job pointed to the [nextstrain-server ref](https://github.com/nextstrain/nextstrain.org/actions/runs/9116872377/job/25066632054#step:4:7).

2. I manually triggered the index-resources workflow, but the [created matrix hash](https://github.com/nextstrain/nextstrain.org/actions/runs/9117045417/job/25067165140#step:4:7) for the nextstrain-canary is the shortened version instead of the full hash.

aws/iam/policy/NextstrainDotOrgServerInstance.tftpl.json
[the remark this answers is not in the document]
joverlee521: From what I understand, the lifecycle only accepts a prefix and not regex/pattern matching. So I've updated the lifecycle [delete old versions of the index](https://us-east-1.console.aws.amazon.com/s3/management/nextstrain-inventories/lifecycle/view?region=us-east-1&bucketType=general&id=delete+old+versions+of+the+index) to the prefix `resources` to capture both the old `resources.json.gz` and the new `resources/*.json.gz` files.

.github/workflows/index-resources.yml
joverlee521: Thanks for flagging that, didn't realize this would be an issue. 
I think might clearer to just assign to a variable then:

```shell
ref_matrix=$(./scripts/get-resource-index-ref-matrix)
echo "ref_matrix=$ref_matrix" | tee -a "$GITHUB_OUTPUT"
```

docs/resource-collection.rst
joverlee521: We can revisit this idea in https://github.com/nextstrain/nextstrain.org/issues/867

.github/workflows/index-resources.yml
joverlee521: Updating in https://github.com/nextstrain/nextstrain.org/pull/866

joverlee521:
Last night's scheduled [index resources workflow](https://github.com/nextstrain/nextstrain.org/actions/runs/9122774061) ran as expected, updating both the nextstrain-server `resources.json.gz` and the nextstrain-canary `resources/v1.json.gz` 🎉

PR #847, opened 2024-05-01
config: Use our own fetch() for logging and proxy (and caching) support

tsibley:

## Checklist

- [x] Checks pass

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,1 +1,1 @@
-web: node server.js --verbose
+web: DEBUG="nextstrain:fetch${DEBUG:+,}${DEBUG:-}" node server.js --verbose
--- a/src/fetch.js
+++ b/src/fetch.js
@@ -1,5 +1,7 @@
+import debugFactory from 'debug';
 import __fetch from 'make-fetch-happen';
-import * as utils from './utils/index.js';
+
+const debug = debugFactory("nextstrain:fetch");
 
 const FETCH_OPTIONS = Symbol("Request options for make-fetch-happen");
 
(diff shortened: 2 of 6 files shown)


Procfile
```diff
@@ -1 +1 @@
-web: node server.js --verbose
+web: DEBUG=nextstrain:fetch node server.js --verbose
```
tsibley, line 1: > A behaviour change is that --verbose no longer includes fetch logging,
> so I adjust the Heroku process invocation to preserve logging of fetches
> in production.  In development, a similar DEBUG value will be necessary
> if fetch logging is still desired.

Should I make a similar change to `package.json` for fetch logging by default for `npm run test:ci`, `… dev`, and `… dev:ssg`?

genehack: I don't have full context to spot any surprise implications of doing that, but +1 for keeping environments as consistent as possible about things like logging.

src/fetch.js
```diff
@@ -1,5 +1,7 @@
+import debugFactory from 'debug';
 import __fetch from 'make-fetch-happen';
```
genehack, line 2: 🤣

genehack (approved):

Procfile
jameshadfield: > Should I make a similar change to package.json for fetch logging by default for npm run test:ci, … dev, and … dev:ssg?

I'd suggest enabling verbose logging¹ for `test:ci` - in part because the logs are sent to a file and we (I) only look at it when things have gone wrong and when things go wrong the more info the better. For `dev` and `dev:ssg` I would skip it. 

¹ I'd personally suggest adding _more_ DEBUG flags - e.g. `DEBUG=nextstrain:*,compression`

jameshadfield (changes requested):
Looks good. I personally find the fetch statements too chatty so I'm happy to hide these behind an opt-in debug flag.

I probably missed 3cf94a7e154b7cc70d837f5baf0bb050d6738a88 when it landed, but we should add documentation around the `DEBUG` variable (it's notably absent). I'd suggest this as a short subsection in the README under `Environment variables` - probably mentioning `DEBUG=nextstrain:*` and `DEBUG=*` would be enough.

Procfile
tsibley: I enabled `DEBUG=nextstrain:fetch` for `test:ci` as part of [afd9069](https://github.com/nextstrain/nextstrain.org/pull/847/commits/afd90693f655e06da592e27c189b0f7b3febd0d8) and then `DEBUG=nextstrain:*` in a separate commit ([2426cf3](https://github.com/nextstrain/nextstrain.org/pull/847/commits/2426cf3106cf943805faa6966598e215e4db6945)).

tsibley:
@jameshadfield Documented `DEBUG` briefly in the README and a little more extensively in `docs/infrastructure.md` with other vars (in [1e6ffd1](https://github.com/nextstrain/nextstrain.org/pull/847/commits/1e6ffd1b150357d205b31c4e5bf1b8acc07f67fb)).

jameshadfield (approved):

PR #848, opened 2024-05-02
[resource indexer] Use debug library for verbose logging

jameshadfield:
Following on from #847 and 3cf94a7, use the 'debug' logging library which is already in use (explicitly and implicitly) within the nextstrain.org codebase. See #511 for more context.

## Checklist

- [ ] Checks pass


resourceIndexer/inventory.js
```diff
@@ -40,14 +42,14 @@ const fetchInventoryRemote = async ({bucket, prefix, name, save}) => {
       resolve(orderedKeys[0])
     });
   });
-  logger.info(`inventory for ${name} - manifest key: ${manifestKey}`)
+  console.log(`inventory for ${name} - manifest key: ${manifestKey}`)
```
genehack, line 45: From reviewing the `debug` docs (new library to me, looks useful!), it's possible to rebind the output of a debugger to STDOUT instead of STDERR.

I wonder if a pattern like this would be useful? Being able to control debug and info output with the same env var feels like a win to me. 

```suggestion
  const info = debugFactory("resource-indexer:inventory:info");
  info.log = console.log.bind(console);
  // this goes to STDOUT
  info(`inventory for ${name} - manifest key: ${manifestKey}`)
```

tsibley: I'm skeptical of that approach. It's essentially a reinvention of log levels but with log levels that aren't a first-class concept but instead smuggled into the logger name (`…:info`). Controlling an `info` log level via an env var named `DEBUG` is also a bit fishy to me.

If we want this level of logging control, we should use a proper logging library (as in #511).

resourceIndexer/inventory.js
```diff
@@ -4,12 +4,14 @@ import zlib from 'zlib';
 import {writeFileSync} from 'fs';
 import { promisify } from 'util';
 import AWS from 'aws-sdk';
-import {logger} from './logger.js';
 import { DateTime } from 'luxon';
 import escapeStringRegexp from 'escape-string-regexp';
 import { ResourceIndexerError } from './errors.js';
+import debugFactory from 'debug';
+const debug = debugFactory("resource-indexer:inventory");
```
tsibley, line 11: Let's keep all of our own codebase's debug logging namespaced under `nextstrain:`? That makes it easy to enable everything of our own with `DEBUG=nextstrain:*`.

```suggestion
const debug = debugFactory("nextstrain:resource-indexer:inventory");
```

tsibley (changes requested):

resourceIndexer/inventory.js
genehack: Fair enough. `console.log()` gets zero granular control (other than a `> /dev/null` or whatever...) and I was trying to think of ways to have a more uniform interface. (_Maybe_ also a visceral reaction to `console.log` 😁 )

resourceIndexer/inventory.js
jameshadfield: Yup. Changed in repush (and updated help message also)

Issue #850, opened 2024-05-06
Use same build in CI for both testing and deploy to Heroku


tsibley:
[Previous commentary](https://github.com/nextstrain/nextstrain.org/blob/3f862f95953dadb2a2ceb2bf5046a10881790a8c/.github/workflows/ci.yml#L44-L51):

https://github.com/nextstrain/nextstrain.org/blob/3f862f95953dadb2a2ceb2bf5046a10881790a8c/.github/workflows/ci.yml#L44-L51

and [more recently](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1714586868169559):

> I just noticed that our nextstrain.org deploy times are now half what they used to be (8m23s vs. 16m) as a nice side effect of the Gatsby → Next.js switch. Still too long, IMO, but much more bearable. We could chop that in half again if we didn't build in CI and then build again on Heroku, instead re-using the CI build for Heroku. So this is still something I'd like to do something about one day, but it's a bit less painful now.

So how to do this?

Our options boil down to:

  1. Build on Heroku, then download to GitHub Actions for testing, then deploy from Heroku
  2. Build on GitHub Actions, test, and then upload slug to Heroku for deploy

Option 1 is slow—both the build itself and the downloading of the resulting slug—but means that the Heroku build process (buildpacks, stacks, etc.) remains entirely owned and managed by Heroku.

Option 2 is fast—GH Actions runners are much faster than Heroku build dynos—but it pulls aside the curtain a bit on how Heroku works and takes some of it into our own hands.  Less magical maybe, but greater understanding of how things actually work (and that seems useful for deploy infra).

tsibley:
Option 2 seems nicer to me.  It gives us precise control of the build/deploy process and sets us up better for future deploy improvements (or even someday moving off Heroku).  It's more involved than what we do now, but the machinery is pretty straightforward.

It starts with building the app using the Heroku Node.js buildpack (the [classic one](https://github.com/heroku/heroku-buildpack-nodejs), not the [Cloud Native one](https://github.com/heroku/buildpacks-nodejs)) and packing the result into an archive:

```bash
git clone https://github.com/nextstrain/nextstrain.org app

mkdir build
git clone https://github.com/heroku/heroku-buildpack-nodejs build/pack
mkdir build/cache # or, restore cache from previous build.
mkdir build/env   # *not* config vars/env

export STACK=heroku-20
export SOURCE_VERSION="$(git -C app describe --always)"
[...]

Issue #854, opened 2024-05-07
Create Heroku review apps through GH Actions CI workflow


joverlee521:
Motivated by https://github.com/nextstrain/nextstrain.org/pull/841#discussion_r1589437154

> A lot of the complexity comes from keeping things in sync between GH Actions and Heroku, so to reduce complexity, we should just do everything in one or the other. Since we rely on GH Actions for CI (see https://github.com/nextstrain/nextstrain.org/issues/850), it makes sense to move everything into GH Actions.

This would involve creating review apps through the [Heroku API](https://devcenter.heroku.com/articles/platform-api-reference#review-app-create) instead of using Heroku's automatic creation but a lot of implementation details TBD.

PR #857, opened 2024-05-10
Build, test, and deploy a Heroku slug instead of re-building on Heroku

tsibley:
See commit messages.

Resolves: <https://github.com/nextstrain/nextstrain.org/issues/850>

## Checklist

- [x] Checks pass
- [ ] Post-merge: Delete ["heroku" GitHub environment](https://github.com/nextstrain/nextstrain.org/settings/environments/482814264/edit)
- [ ] Post-merge: Delete Heroku API token for bot@nextstrain.org that was associated with the "heroku" GitHub environment:
  ```
  curl -n -X DELETE https://api.heroku.com/oauth/authorizations/6716fa11-cb1e-443a-be34-f9051cbb33b3 \
    -H 'Accept: application/vnd.heroku+json; version=3'
  ```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,9 +5,17 @@ on:
       - master
   pull_request:
   workflow_dispatch:
+defaults:
+  run:
+    # This is the same as GitHub Action's `bash` keyword as of 20 June 2023:
+    # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell
+    #
+    # Completely spelling it out here so that GitHub can't change it out from under us
+    # and we don't have to refer to the docs to know the expected behavior.
+    shell: bash --noprofile --norc -eo pipefail {0}
 jobs:
-  test:
-    if: github.repository == 'nextstrain/nextstrain.org'
+  # Lint regardless of build or test status and lint early.
+  lint:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
@@ -16,18 +24,59 @@ jobs:
           node-version-file: 'package.json'
       - run: node --version
       - run: npm ci
-      # building the static site (Next.js) will fail if linting fails, so explicitly run linting
-      # first to make the test output clearer
       - run: npm run lint:server
       - run: npm run lint:static-site
-      - run: npm run build
+
+  # Build into Heroku slug so we can deploy the same build that we test.
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - run: ./scripts/heroku-build
+
+      - if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: build
+          path: |
+            build/slug.tar.gz
+            build/slug.tar.gz.sha256sum
+            build/slug.json
+
+  # Run tests in Heroku runtime environment against our built slug.
+  test:
+    if: github.repository == 'nextstrain/nextstrain.org'
+    needs: build
+    runs-on: ubuntu-latest
+    container:
+      image: heroku/heroku:20
+    env:
+      # Override NODE_ENV=production default in production build (slug)
+      NODE_ENV: development
+
+      # Set up Heroku runtime environment.  See bash(1) or
+      # <https://www.gnu.org/software/bash/manual/bash.html#Invoked-non_002dinteractively>.
+      BASH_ENV: .heroku/BASH_ENV
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          name: build
+          path: build/
+      - run: sha256sum --check build/slug.tar.gz.sha256sum
+      - run: tar --extract --file build/slug.tar.gz --xform 's,^[.]/app/,,'
+      - run: node --version
+
+      # (Re-)install dev deps which got pruned from production build (slug)
+      - run: npm ci
+
       # configure AWS to run dev server necessary for `npm run test:ci`
       - uses: aws-actions/configure-aws-credentials@v4
         with:
           aws-region: ${{ vars.AWS_DEFAULT_REGION }}
           aws-access-key-id: ${{ secrets.DEV_SERVER_AWS_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.DEV_SERVER_AWS_SECRET_ACCESS_KEY }}
-      - run: aws sts get-caller-identity
+
       - run: npm run test:ci
         env:
           # Tests make GitHub API requests which are rate limited.
@@ -38,25 +87,19 @@ jobs:
       - if: always()
         uses: actions/upload-artifact@v4
         with:
-          name: logs
+          name: test-logs
           path: test/server.log
 
-  # XXX TODO: It'd be nice to avoid the rebuild on Heroku and instead deploy
-  # the artifacts (source code + generated files + node_modules/) already built
-  # above.  This would dramatically reduce deploy times and move us closer to
-  # "deploy what you tested", but it may also run into platform compatibility
-  # issues given CI is building on a different platform (arch + OS + sys libs)
-  # than Heroku's dynos and some deps are compiled.  But should try it and see!
-  # Or do our build above inside a Heroku buildpack…
-  #   -trs, 2 May 2022
   deploy:
     if: |2
          github.repository == 'nextstrain/nextstrain.org'
       && github.event_name == 'push'
       && github.ref == 'refs/heads/master'
 
-    # Wait for "test" job above to pass.
-    needs: test
+    # Wait for "build" and "test" jobs above to pass.
+    needs:
+      - build
+      - test
 
     # Only one "deploy" job at a time.
     concurrency: deploy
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -16,18 +29,59 @@ jobs:
           node-version-file: 'package.json'
       - run: node --version
       - run: npm ci
-      # building the static site (Next.js) will fail if linting fails, so explicitly run linting
-      # first to make the test output clearer
```
victorlin, line 20: From c305e916d03d794f35d6483e0c7b40703ba560aa:

> it doesn't make sense anymore to lint _after_ build

I assume you meant _before_ build?

scripts/heroku-build
```diff
@@ -0,0 +1,112 @@
...
+  --user $(id -u):$(id -g) \
+  --volume "$(realpath build/app)":/build/app:rw \
+  --volume "$(realpath build/pack)":/build/pack:ro \
+  --volume "$(realpath build/cache)":/build/cache:rw \
+  --volume "$(realpath build/env)":/build/env:ro \
+  --env STACK=heroku-20 \
+  --env SOURCE_VERSION \
+  --env SOURCE_DESCRIPTION \
+  --env NODE_MODULES_CACHE=false \
+  heroku/heroku:20-build bash -c '
```
victorlin, line 60: *note*

`heroku/heroku:20-build` (and `22-build`) are single-platform `linux/amd64` - no `linux/arm64` variant. This means `scripts/heroku-build` runs slower on Apple silicon, but not terribly slow - 7m17s on my M1 just now.

`heroku/heroku:24-build` is multi-platform with a `linux/arm64` variant. Just noting a reason to upgrade in the future.

scripts/heroku-exec
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
```
victorlin, line 1: Ancient macOS `/bin/bash` gives an error:

```
./scripts/heroku-exec: line 42: docker_args[@]: unbound variable
```

Homebrew-installed bash is 👍 but needs this shebang:

```suggestion
#!/usr/bin/env bash
```

victorlin (approved):
Didn't test much, just ran the two scripts locally and skimmed commits. Looking forward to faster build times!

.github/workflows/ci.yml
tsibley: Ah, this was poor wording on my part. I'll fix it up with this:

```diff
-    Lint is moved out of test, because it doesn't make sense anymore to lint
-    _after_ build (since, as a previous comment noted, the build will fail
-    if linting fails).
+    Lint is moved out of test, because test now happens after build and it
+    doesn't make sense to lint _after_ build (since, as a previous comment
+    noted, the build will fail if linting fails).
```

tsibley:
> Looking forward to faster build times!

Yeah! For CI runs that deploy, it'll mean faster times. For CI runs that don't deploy, it may be a bit slower. Need to look into that more closely. OTOH, if we're going to start [deploying review apps ourselves](https://github.com/nextstrain/nextstrain.org/issues/854) (and I think we probably should?) then _all_ CI runs will do deploys.

scripts/heroku-build
tsibley, line 1: TODO for myself: Discuss somewhere how much simpler this will be once Heroku supports [Cloud Native buildpacks](https://blog.heroku.com/heroku-cloud-native-buildpacks) (or we move to somewhere that already does?).

scripts/heroku-build
tsibley: Good to note! Upgrading should be pretty painless, but I wanted to save that for another PR.

scripts/heroku-exec
tsibley: 👀 https://github.com/nextstrain/ingest/pull/17#discussion_r1312013780

victorlin: Thanks, I forgot about that conversation (bad memory)! If there's a simple workaround that works for Bash 3.2 and above, better to use that.

tsibley: 17a85d7e

.github/workflows/ci.yml
tsibley: Rewording in [re-push](https://tsibley.net/paste/2024-05-13mFnDJvko-diff.html).

scripts/heroku-build
tsibley: 0f76b4b7

tsibley:
> For CI runs that deploy, it'll mean faster times. For CI runs that don't deploy, it may be a bit slower.

One thing we could do here (even if we're going to move review apps into CI and make all CI runs do deploys) is run tests against _both_ the "deployment-style" build as is done in this PR currently _and_ a "development-style" build as we used to. They could run concurrently and wouldn't change the overall duration of the CI run while still giving us faster deploys. Put another way, we'd make the current `lint` job in this PR also continue to run tests (and rename the job).

victorlin (approved):
> run tests against both the "deployment-style" build as is done in this PR currently and a "development-style" build as we used to

Is the only benefit to have a slightly faster build result? If so, I don't think it's worth the added complexity.

We could always merge as-is and reconsider later if build results seem painfully slower.

scripts/heroku-build
```diff
@@ -0,0 +1,112 @@
...
+  echo "---> Using buildpack in build/pack/"
+fi
+
+echo
+echo "---> Buildpack version is $(git -C build/pack describe --always --dirty)"
+
+echo
+echo "---> Running buildpack"
+docker run \
+  --rm --interactive $(tty --quiet && echo --tty) \
```
joverlee521, line 60: When running locally, I see an error for the ancient macOS tty
```
tty: illegal option -- -
usage: tty [-s]
```

Updating to `tty -s` works as expected

scripts/heroku-build
```diff
@@ -0,0 +1,112 @@
...
+      HOME="$(realpath "$(dirname "$BASH_SOURCE")"/..)" source "$f"
+  done
+'
+
+# <https://devcenter.heroku.com/articles/platform-api-deploying-slugs>
+echo
+echo "---> Packing build/app/ into build/slug.tar.gz"
+tar --create --gzip --file build/slug.tar.gz -C build/ ./app/
+sha256sum build/slug.tar.gz > build/slug.tar.gz.sha256sum
+du --human-readable --si --summarize build/app/ build/slug.tar.gz
```
joverlee521, line 107: When running locally, I see an error the ancient macOS du
```
du: unrecognized option `--human-readable'
usage: du [-Aclnx] [-H | -L | -P] [-g | -h | -k | -m] [-a | -s | -d depth] [-B blocksize] [-I mask] [-t threshold] [file ...]
```

Updating to `du -h` works as expected.

tsibley:
> Is the only benefit to have a slightly faster build result?

Slightly faster time-to-test-failure for failing CI runs. Yeah, wasn't suggesting I do it now in this PR but as an option for the future.

joverlee521 (commented):
Hooray for faster deploys! 

Will need to read more on buildspacks...I'm assuming this will inform the implementation details for [creating Heroku review apps](https://github.com/nextstrain/nextstrain.org/issues/854).

scripts/heroku-build
joverlee521: Ah, `--summarize` causes an error too 🤦‍♀️ 

Updating to `du -h --si -s ...` works as expected.

scripts/heroku-build
tsibley: Ah, I'll fix this up. My preference is use to long options in code so that they're more self-describing, but that runs up against OS/version compatibility where short options are often more widely supported.

scripts/heroku-build
tsibley: [Ditto](https://github.com/nextstrain/nextstrain.org/pull/857#discussion_r1599065517).

tsibley:
First deploy to canary with this PR went smoothly. Note the lack of a Heroku build before the Heroku deploy:

image

[GitHub CI run](https://github.com/nextstrain/nextstrain.org/actions/runs/9069460615) and [GitHub deployment record](https://github.com/nextstrain/nextstrain.org/deployments/nextstrain-canary).

image

I've done the post-merge tidy ups I'd previously identified.

scripts/heroku-build
tsibley: https://github.com/nextstrain/nextstrain.org/pull/859

scripts/heroku-build
tsibley: https://github.com/nextstrain/nextstrain.org/pull/859

.github/workflows/ci.yml
```diff
@@ -70,69 +118,59 @@ jobs:
     #
     # ¹ https://github.com/nextstrain/nextstrain.org/settings/environments
     environment:
-      name: heroku
-      url: https://next.nextstrain.org
+      name: ${{ inputs.heroku-app || 'nextstrain-canary' }}
+      url: ${{ inputs.heroku-app && format('https://{0}.herokuapp.com', inputs.heroku-app) || 'https://next.nextstrain.org' }}
```
victorlin, line 122: On the [deployments](https://github.com/nextstrain/nextstrain.org/deployments) page, `nextstrain-canary` maps to `next.nextstrain.org` and `nextstrain-dev` to `nextstrain-dev.herokuapp.com` (as opposed to `dev.nextstrain.org`). I thought that was weird, but it makes sense after looking back at this.

tsibley: Yeah, we could improve that various ways later.

PR #859, opened 2024-05-13
dev: Swap out a few long options for equivalent but more available short options

tsibley:
The long options are from GNU coreutils, but macOS ships a BSD tty and du.  Only one of the swapped options is POSIX.

Noted by @joverlee521 in post-merge review.¹

¹ <https://github.com/nextstrain/nextstrain.org/pull/857#pullrequestreview-2053748042>

## Checklist

- [x] Checks pass


joverlee521 (approved):
Thanks for following up!

PR #860, opened 2024-05-14
Add SAB to /team

victorlin:
_[preview](https://nextstrain-s-add-sab-u6dfbs051.herokuapp.com/team)_

Based on #862

## Checklist

- [x] Replace filler text with SAB member info
- [x] Add member images
- [x] Get 👍 from Vítor
- [x] Get 👍 from Sarah
- [x] Get 👍 from Peter
- [x] Get 👍 from Zam
- [x] Get 👍 from Nicola
- [x] Get 👍 from Nídia
- [x] Checks pass

PR #862, opened 2024-05-15
Update people component

victorlin:

No functional changes – prepping for #860

## Checklist

- [x] Checks pass
- [x] [Preview](https://nextstrain-s-victorlin--ex1fdc.herokuapp.com) looks good

### DIFF
--- a/static-site/src/components/Footer/index.jsx
+++ b/static-site/src/components/Footer/index.jsx
@@ -3,7 +3,7 @@ import Link from 'next/link'
 import * as Styles from "../splash/styles";
 import { SmallSpacer, BigSpacer } from "../../layouts/generalComponents";
 import { Logos } from "../../components/logos";
-import { ListOfPeople } from "../People/avatars";
+import { FooterList } from "../People/avatars";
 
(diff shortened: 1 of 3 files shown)


tsibley (approved):
LGTM.

One minor thing I noted is that people without links are shown in a slightly different shade of black. Compare Pavel to Moira:

Screenshot from 2024-05-15 16-10-34

victorlin:
@tsibley thanks for noting that! I didn't realize there was already a member without link. I've updated the commit message from a8e591e418f39ed13a4d448595e642026ef20de7 to 4047edc198177367e5c87cf421b11a122441e2b7.

The color difference is because (1) links here are set to [a specific color](https://github.com/nextstrain/nextstrain.org/blob/4047edc198177367e5c87cf421b11a122441e2b7/static-site/src/components/People/avatars.jsx#L13) and (2) the non-links don't have any color set via CSS. It seems like both our browsers default to pure black `#000` in that case, so the link color `#333` is slightly lighter. There's probably a few options to address this - I'll open a separate issue.

PR #864, opened 2024-05-16
CI: fix use of index-resources.yml

joverlee521:
Follow up to #841 

Inherit secrets so that the called index-resource.yml workflow has access to repo secrets.

## Checklist

- [ ] Checks pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,6 +105,7 @@ jobs:
       id-token: write # needed to interact with GitHub's OIDC Token endpoint
       contents: read
     uses: ./.github/workflows/index-resources.yml
+    secrets: inherit
 
   deploy:
     if: |2


.github/workflows/ci.yml
```diff
@@ -105,6 +105,7 @@ jobs:
       id-token: write # needed to interact with GitHub's OIDC Token endpoint
       contents: read
     uses: ./.github/workflows/index-resources.yml
+    secrets: inherit
```
tsibley, line 108: There's not a `HEROKU_TOKEN` repo secret. It's attached to the `nextstrain-{canary,dev}` environments. I think what's really wanted here is something like:

```yaml
secrets:
  HEROKU_TOKEN: ${{ secrets.HEROKU_TOKEN_READ_PROTECTED }}
```

to use the read-only token available at the org-level.

joverlee521: Oh, the called index-resources workflow uses the repo secret: 
https://github.com/nextstrain/nextstrain.org/blob/1d83ffe85c68444f1fbd3cec72e6c56ed330ff01/.github/workflows/index-resources.yml#L38

Issue #849, opened 2024-05-03
splash page redesign


jameshadfield:
## Progress

- [x] https://github.com/nextstrain/nextstrain.org/pull/868
- [x] https://github.com/nextstrain/nextstrain.org/pull/878
- [x] https://github.com/nextstrain/nextstrain.org/pull/901
- [x] https://github.com/nextstrain/nextstrain.org/pull/902
- [x] https://github.com/nextstrain/nextstrain.org/issues/917
- [x] https://github.com/nextstrain/nextstrain.org/pull/916
- [x] https://github.com/nextstrain/nextstrain.org/pull/930
- [x] https://github.com/nextstrain/nextstrain.org/pull/914
- [x] https://github.com/nextstrain/nextstrain.org/pull/922
- [x] https://github.com/nextstrain/nextstrain.org/pull/951
- [x] https://github.com/nextstrain/nextstrain.org/pull/952
- [x] https://github.com/nextstrain/nextstrain.org/issues/934

## Initial issue text

**why?**

The current page is overly focused on surfacing builds and doesn't mention nextclade and requires reading the bottom of the page to learn about the bioinformatics tools etc. The layout of the builds makes it hard to surface recent builds because we have partitioned things into 5 sections with 2 tiles each, which both fails to give an overarching sense of importance and limits us to 2 tiles per section. Furthermore the current tile design doesn't provide any way to add contextual information about a build (i.e. a card is limited to a title).

**aim**

1. Expose a showcase selection of builds not partitioned by source/type, with an ordering which conveys how important we think each is. These don't have to be one build per pathogen, for instance for the current cattle-fly outbreak we may choose to expose a HA 2y build as well as the full genome build, and maybe even a tanglegram view.
2. Make it technically simple for us to update the showcase cards as needed. (Increasing the frequency of updates here is not only a technical problem, but we can ensure that it's simple to do.)
3. Surface relevant pages early on (e.g. nextclade, /pathogens, bioinformatics tools)

**proposed changes**

![Splash page design mk1](https://github.com/nextstrain/nextstrain.org/assets/8350992/c5612b54-930a-474a-bf27-855a8a488c6d)

**implementation details**

[...]

Issue #863, opened 2024-05-15
Slight color difference between links and non-links on /team


victorlin:
Noted by @tsibley in https://github.com/nextstrain/nextstrain.org/pull/862#pullrequestreview-2059192580:

> people without links are shown in a slightly different shade of black. Compare Pavel to Moira:

![Screenshot from 2024-05-15 16-10-34](https://github.com/nextstrain/nextstrain.org/assets/79913/cdaec8b5-297d-4b67-ab14-def53af06599)

The color difference is because (1) links here are set to [a specific color](https://github.com/nextstrain/nextstrain.org/blob/4047edc198177367e5c87cf421b11a122441e2b7/static-site/src/components/People/avatars.jsx#L13) and (2) the non-links don't have any color set via CSS. It seems like both our browsers default to pure black `#000` in that case, so the link color `#333` is slightly lighter.

victorlin:
I'm guessing intent of setting link color to `#333` is to match the non-link color.

The first thing to do is set the default `color` explicitly. [global.css](https://github.com/nextstrain/nextstrain.org/blob/d532048a6fd73526a14bf5f0b340493c3558c962/static-site/src/styles/globals.css#L12) seems to be most appropriate.

Next, those two color definitions can be synced. I don't think it's easily done with the current setup since global.css and styled components in JSX can't share a common variable (maybe I'm wrong). The lazy alternative is to add a comment next to each declaring the relationship.

victorlin:
Updated this issue after receiving email feedback on font colors for accessibility.

Another factor to consider alongside color might be font weight, which seems pretty thin to me.

genehack:
It should be much easier to accomplish this once everything is using App Router, because all the styled-components stuff will be gone and we can set this in `globals.css` and be done with it.

PR #868, opened 2024-05-18
Add sitemap to footer

victorlin:
_[preview on contact page](https://nextstrain-s-victorlin--rytbno.herokuapp.com/contact)_

Adds links to the footer.

## Related issue(s)

- #849

## Checklist

- [x] Checks pass

### DIFF
--- a/static-site/src/components/Footer/index.jsx
+++ b/static-site/src/components/Footer/index.jsx
@@ -4,6 +4,7 @@ import * as Styles from "../splash/styles";
 import { SmallSpacer, BigSpacer } from "../../layouts/generalComponents";
 import { Logos } from "../../components/logos";
 import { FooterList } from "../People/avatars";
+import { SiteMap } from "./sitemap";
 
 const SplashImagesCredit = () => (
   <div className="row">
@@ -55,6 +56,12 @@ class Footer extends React.Component {
   render() {
     return (
       <div>
+        <div className="row">
+          <SiteMap />
+        </div>
+
+        <BigSpacer />
+
         <div className="row">
           <div className="col-md-1"/>
           <div className="col-md-10">
--- a/static-site/src/components/Footer/sitemap.jsx
+++ b/static-site/src/components/Footer/sitemap.jsx
@@ -1,0 +1,106 @@
+import React from "react";
+import styled from "styled-components";
+import Link from 'next/link'
+import { FaExternalLinkAlt } from "react-icons/fa";
+import * as Styles from "../splash/styles";
+import { BigSpacer } from "../../layouts/generalComponents";
+import sections from "./sitemap.yaml";
+
+
+export const SiteMap = () => {
+  return  (
+    <SiteMapContainer>
+      {sections.map((section) => (
+        <SectionWrapper className="col-sm-3 col-xs-6" key={section.title}>
+          <Section>
+            <BigSpacer />
+            <SectionTitle $left>{section.title}</SectionTitle>
+            <SectionList entries={section.entries} />
+          </Section>
+        </SectionWrapper>
+        ))
+      }
+    </SiteMapContainer>
+  )
+}
+
+
+const SectionList = ({ entries }) => {
+  return (
+    <UnstyledList>
+      {entries.map((entry) => (
+        <li key={entry.name}>
+          <ListItemParagraph>
+            <UncoloredLink>
+              {entry.href.startsWith('http') ? (
+                <Link href={entry.href}  rel="noopener noreferrer" target="_blank">
+                  {entry.name} <FooterIcon><FaExternalLinkAlt/></FooterIcon>
+                </Link>
+              ) : (
+                <Link href={entry.href}>
+                  {entry.name}
+                </Link>
+              )}
+            </UncoloredLink>
+          </ListItemParagraph>
+        </li>
+      ))}
+    </UnstyledList>
+  )
+}
+
+const SiteMapContainer = styled.div`
+  // Center contents
+  margin: 0 auto;
+
+  // Ensure that, on wide screens, 4 sections renders to roughly align with the
+  // width of IconParagraph (rendered below this element)
+  max-width: 680px;
+`
+
+// Wrapper is used to center the actual Section while allowing it to be
+// left-aligned.
+const SectionWrapper = styled.div`
+  // Ensure the longest entry (Discussion forum ↗️) still fits on a single line
+  min-width: 170px;
+
+  // Center section contents when all sections are displayed in a single row
+  @media (min-width: 680px) {
+    text-align: center;
+  }
+
+  // Otherwise, when there are multiple rows, left-align so columns are aligned
+  text-align: left;
+`
+
+const Section = styled.div`
+  display: inline-block;
+  text-align: left;
+`
+
+// The global Styles.H3 has a min-width that is undesirable here
+const SectionTitle = styled.h3`
+`
+
+const UnstyledList = styled.ul`
+  list-style: none;
+`
+
+const ListItemParagraph = styled(Styles.FocusParagraph)`
+  line-height: 0.5;
+  font-weight: 300;
+  a {
+    font-weight: 300;
+  }
+`
+
+const UncoloredLink = styled.span`
+  & a {
+    color: #000 !important;
+  }
+`
+
+const FooterIcon = styled.span`
+  margin-left: 2px;
+  font-size: 12px;
+`
--- a/static-site/src/components/Footer/sitemap.yaml
+++ b/static-site/src/components/Footer/sitemap.yaml
@@ -1,0 +1,39 @@
+- title: Resources
+  entries:
+  - name: Core pathogens
+    href: /pathogens
+  - name: SARS-CoV-2
+    href: /sars-cov-2
+  - name: Community
+    href: /community
+  - name: Groups
+    href: /groups
+
+- title: Tools
+  entries:
+  - name: Nextclade
+    href: https://clades.nextstrain.org/
+  - name: Auspice.us
+    href: https://auspice.us/
+  - name: Augur
+    href: https://docs.nextstrain.org/projects/augur
+  - name: Auspice
+    href: https://docs.nextstrain.org/projects/auspice
+  - name: Nextstrain CLI
+    href: https://docs.nextstrain.org/projects/cli
+
+- title: Support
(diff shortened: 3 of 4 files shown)


static-site/src/components/Footer/sitemap.jsx
```diff
@@ -0,0 +1,183 @@
...
+`
+
+export const SiteMap = () => {
+  return  (
+    <>
+      <div className="col-md-3">
+        <BigSpacer />
+        <Styles.H3 $left>Resources</Styles.H3>
+        <UnstyledList>
+          <li>
```
jameshadfield, line 35: To reduce duplication like this we can define the links in an array and make use of [map within JSX](https://legacy.reactjs.org/docs/lists-and-keys.html#embedding-map-in-jsx), e.g.:

```js
// earlier in the file
const RESOURCES = [
  {name: 'Core pathogens', href: '/pathogens'},
  {name: 'SARS-CoV-2', href: '/sars-cov-2'},
  ...
];
```

and then swap `<UnstyledList>...</UnstyledList>` with
```js
<UnstyledList>
  {RESOURCES.map((resource) => (
    <li>
      <ListItem>
        <UncoloredLink>
          {resource.href.startswith('http') ? (
            <Link href={resource.href}  rel="noopener noreferrer" target="_blank">
              {resource.name} <FooterIcon><FaExternalLinkAlt/></FooterIcon>
            </Link>
          ) : (
            <Link href={resource.href}>
              {resource.name}
            </Link>
          )}
        </UncoloredLink>
      </ListItem>
    </li>
  ))}
</UnstyledList>
```

You can go further if you want and define everything together, e.g.:
```js
const LINKS = [
  {title: 'Resources', links: [{name: ...}]},
  {title: 'Tools', links: [{name: ...}]},
  ...
]
```
and then use nested `map`s to further simplify things.

static-site/src/components/Footer/sitemap.jsx
```diff
@@ -0,0 +1,183 @@
+import React from "react";
+import styled from "styled-components";
```
jameshadfield, line 2: FWIW, I'd like us to move away from styled components over time. This isn't a request to stop using it in this PR - we use it everywhere at the moment, so staying consistent is great - but just a heads up as to my thinking of where we'll take this over time.

static-site/src/components/Footer/sitemap.jsx
```diff
@@ -0,0 +1,183 @@
...
+import * as Styles from "../splash/styles";
+import { BigSpacer } from "../../layouts/generalComponents";
+
+
+const UnstyledList = styled.ul`
+  list-style: none;
+`
+
+const ListItem = styled(Styles.FocusParagraph)`
+  line-height: 0.7;
```
jameshadfield, line 14: [subjective] I'd go for a line-height of 0.5, font-weight of 300, and color of `var(--text-color)` personally. What do you think?

static-site/src/components/Footer/index.jsx
```diff
@@ -55,6 +56,12 @@ class Footer extends React.Component {
   render() {
     return (
```
jameshadfield, line 57: [minor] we typically render a horizontal line above the footer but don't do this on the main splash page. With the changes introduced here I think it'd be clarifying to also render the horizontal line on the main splash page

jameshadfield (commented):
Nice one @victorlin! I haven't thought too hard about the entries in each list but as a v1 it looks great.

static-site/src/components/Footer/sitemap.jsx
```diff
@@ -0,0 +1,183 @@
...
+
+const FooterIcon = styled.span`
+  margin-left: 2px;
+  font-size: 12px;
+`
+
+export const SiteMap = () => {
+  return  (
+    <>
+      <div className="col-md-3">
```
jameshadfield, line 31: I think these (bootstrap) parameters aren't quite right for the sitemap as it switches to a single column too soon:

image

tsibley:
This will be a nice improvement!

Aesthetically, one thing I'd point out (in addition to what James did) is that the alignment/justification of the sitemap as a whole and the columns within it feel a bit off to me. Our central layout element/block is a centered column, but we have a slew of widths for that column. Within the column, text and inline elements are either centered or left-justified.

image

I think the sitemap would "fit in" better with the rest of the page blocks with a reduced overall width (e.g. what the team block uses) and/or with the inner columns/blocks of the sitemap (Resources, Tools, etc) themselves centered (but keeping the text elements within them still left-justified). Something like the below.

alignment

static-site/src/components/Footer/sitemap.jsx
victorlin: Could you write up an issue for this? I remember you also brought it up somewhere else with more reasoning but I can't seem to find it.

static-site/src/components/Footer/sitemap.jsx
victorlin: Added in eca5ef6b.

I already see an inherited font-weight of 300, but good to be explicit.

Not sure what you mean by `var(--text-color)` - nothing defines `--text-color` currently. Also, setting it here has no effect with this in place (which I could change):

https://github.com/nextstrain/nextstrain.org/blob/eca5ef6b8a5c525f0942369f7c1b69abb22fdb3c/static-site/src/components/Footer/sitemap.jsx#L87-L91

static-site/src/components/Footer/index.jsx
victorlin, line 1: @tsibley re: [alignment/justification concerns](https://github.com/nextstrain/nextstrain.org/pull/868#issuecomment-2120956166)

Done in 86bb23e3 but noted that it doesn't work well on narrow screens for some reason.

static-site/src/components/Footer/index.jsx
victorlin: Added: 6b6f4520159c9aaa2bb437e41c22c4d08162d25d

static-site/src/components/Footer/sitemap.jsx
jameshadfield: Nice!

static-site/src/components/Footer/sitemap.jsx
jameshadfield: It's really part of that larger discussion issue that I can't find at present -- what frontend do we want? That discussion involves styles, how data is passed between client & server etc.

static-site/src/components/Footer/sitemap.jsx
jameshadfield: >  Also, setting it here has no effect...

Ah, I didn't mean to say this was the correct component to add all of these styles, I was using it as a hook to describe what styles I thought we should end up with. Sorry for any confusion!

> I already see an inherited font-weight of 300, but good to be explicit.

As of 6b6f4520 I still get a font-weight of 500 in Firefox 125 via this definition:

https://github.com/nextstrain/nextstrain.org/blob/08db0bd73841c8e0ef798e679167fb11ae014664/static-site/src/components/layout.jsx#L39-L44

> Not sure what you mean by var(--text-color) - nothing defines --text-color currently

Not sure where I got that from. I'm sure it was set when I last tested but you're right I cannot find it anywhere. In fact I don't even think we set a page-default `color`! I believe our original intention was `#333`.

static-site/src/components/Footer/index.jsx
jameshadfield: > it doesn't work well on narrow screens for some reason.

This is because of the `min-width: 780px`, so essentially all screen widths above that are fine (and for them it's much improved with your recent changes!). Changing to `min-width: min(100%, 780px);` doesn't fix all the problems, but it's _better_. I'd also switch to `col-xs-12`; again not _perfect_ but _better_.

Finally, I wasn't sure if you were intending to match up with width of the sitemap with the team avatars? Because of the centered text it's hard to get them to line up perfectly but for screen widths between ~700 and 1200 it's pretty nice and then >1200px the avatars container increases to 911px wide. (This is not a request for changes.)

victorlin: > This is because of the `min-width: 780px`

Oh good catch! Fixed in 339f968d35488d7d199d16a2ef1a10fbfb2390c1 with reasons described in the next response.

> Finally, I wasn't sure if you were intending to match up with width of the sitemap with the team avatars?

That was @tsibley's original idea in https://github.com/nextstrain/nextstrain.org/pull/868#issuecomment-2120956166, but I think it looks better when it's narrower. I added a comment describing my intent:

https://github.com/nextstrain/nextstrain.org/blob/339f968d35488d7d199d16a2ef1a10fbfb2390c1/static-site/src/components/Footer/sitemap.jsx#L56-L58

victorlin: > I'd also switch to `col-xs-12`; again not _perfect_ but _better_

Hmm, I prefer `col-xs-6`. Compare the two at a width of 600px:

`col-xs-6`:

col-xs-6

`col-xs-12`:

col-xs-12

static-site/src/components/Footer/sitemap.jsx
victorlin: > I still get a font-weight of 500

Oh you're right, I was looking in the wrong place. Fixed in b4727b1f4c3382019f7801df3b3e13f0913ffd06

> Ah, I didn't mean to say this was the correct component to add all of these styles, I was using it as a hook to describe what styles I thought we should end up with. Sorry for any confusion!
> I don't even think we set a page-default `color`! I believe our original intention was `#333`.

Gotcha, no worries! I'll defer this to an existing issue on global text colors: #863

static-site/src/components/Footer/index.jsx
jameshadfield: > Hmm, I prefer col-xs-6. Compare the two at a width of 600px:

With the new changes you've just pushed up I agree - `col-xs-6` is good 👍 

> I think it looks better when it's narrower. I added a comment describing my intent:

Sounds good!

jameshadfield (approved):
Thanks @victorlin! (A rebase onto master might be nice before merge, but not required)

victorlin:
Thanks for reviewing! There's no merge conflicts and #869 is based on this, so I'll just merge as-is to avoid extra rebase churn on that PR.

static-site/src/components/Footer/sitemap.jsx
victorlin: Issue created: #963

PR #873, opened 2024-05-21
Add carousel to /pathogens showcase

victorlin:
[_preview_](https://nextstrain-s-victorlin--phhwcc.herokuapp.com/pathogens)

Previously, the showcase would drop cards as the width decreases. This is presumably to prevent the showcase from taking up too much height. Instead of hiding the unused cards, allow them to be accessed in a carousel via scroll/arrow clicking.

## Related issue(s)

- Prompted by https://github.com/nextstrain/nextstrain.org/issues/849#issuecomment-2121064824

## Checklist

- [x] Checks pass

### DIFF
--- a/static-site/src/components/ListResources/Showcase.jsx
+++ b/static-site/src/components/ListResources/Showcase.jsx
@@ -1,5 +1,7 @@
 /* eslint-disable react/prop-types */
 import React, { useCallback, useEffect, useState } from 'react';
+import Carousel from "react-multi-carousel";
+import "react-multi-carousel/lib/styles.css";
 import styled from 'styled-components';
 import {CardInner, CardImg, CardTitle} from "../Cards/styles";
 import { theme } from "../../layouts/theme";
@@ -10,18 +12,35 @@ const cardWidthHeight = 160; // pixels
 
 export const Showcase = ({cards, setSelectedFilterOptions}) => {
   if (!cards.length) return null;
+
+  // Generate options for carousel
+  let responsiveOptions = {};
+
+  for (let i = 0; i < cards.length; i++) {
+    const breakpointMin = i * cardWidthHeight;
+    let breakpointMax = (i + 1) * cardWidthHeight;
+
+    // Don't limit max otherwise carousel won't render at all on larger screens
+    if (i == cards.length - 1) {
+      breakpointMax = 999999
+    }
+
+    responsiveOptions[i] = {
+      breakpoint: { max: breakpointMax, min: breakpointMin },
+      items: i
+    };
+  }
+
   return (
     <div>
       <Byline>
         Showcase resources: click to filter the resources to a pathogen
       </Byline>
-      <SingleRow>
-        <ShowcaseContainer>
-          {cards.map((el) => (
-            <ShowcaseTile data={el} key={el.name} setSelectedFilterOptions={setSelectedFilterOptions}/>
-            ))}
-        </ShowcaseContainer>
-      </SingleRow>
+      <Carousel responsive={responsiveOptions} renderButtonGroupOutside={true}>
+        {cards.map((el) => (
+          <ShowcaseTile data={el} key={el.name} setSelectedFilterOptions={setSelectedFilterOptions}/>
+          ))}
+      </Carousel>
       <Spacer/>
     </div>
   )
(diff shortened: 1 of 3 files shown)


jameshadfield (commented):
Tested out briefly and looks great at a glance. The arrows currently obscure the titles at the moment (or vice versa?)
image

tsibley:
There's some non-ideal behaviour going on with the responsive widths of the tiles at some viewport widths:

image

If I make the window a little bit smaller or little bit wider it looks fine.

static-site/src/components/ListResources/Showcase.jsx
```diff
@@ -10,18 +12,35 @@ const cardWidthHeight = 160; // pixels
...
         Showcase resources: click to filter the resources to a pathogen
       </Byline>
-      <SingleRow>
-        <ShowcaseContainer>
-          {cards.map((el) => (
-            <ShowcaseTile data={el} key={el.name} setSelectedFilterOptions={setSelectedFilterOptions}/>
-            ))}
-        </ShowcaseContainer>
-      </SingleRow>
+      <Carousel responsive={responsiveOptions} renderButtonGroupOutside={true}>
```
victorlin, line 39: @jameshadfield re: https://github.com/nextstrain/nextstrain.org/pull/873#pullrequestreview-2069756947

> The arrows currently obscure the titles at the moment (or vice versa?)

There's props to customize the arrows which I'll look into. I tried briefly with `renderButtonGroupOutside` but looks like more options are necessary to get it working properly.

static-site/src/components/ListResources/Showcase.jsx
```diff
@@ -10,18 +12,35 @@ const cardWidthHeight = 160; // pixels
...
+    // Don't limit max otherwise carousel won't render at all on larger screens
+    if (i == cards.length - 1) {
+      breakpointMax = 999999
+    }
+
+    responsiveOptions[i] = {
+      breakpoint: { max: breakpointMax, min: breakpointMin },
+      items: i
+    };
+  }
```
victorlin, line 32: @tsibley re: https://github.com/nextstrain/nextstrain.org/pull/873#issuecomment-2123601423

> There's some non-ideal behaviour going on with the responsive widths of the tiles at some viewport widths

Unfortunately react-multi-carousel takes a fixed number of items opposed to a fixed width per item, which I would prefer avoids the behavior you'r seeing. With a fixed number of items, there's bound to be some stretching and contracting. I did a quick and dirty calculation here – will tweak it to contract a bit less.

jameshadfield: Can you use a flexbox to evenly space them? That seems better than stretching / contracting cards

jameshadfield: I presume the bigger issue behind this is that we're not observing / responding to viewport changes (this is from memory... we do observe this elsewhere in the resource listing UI).

jameshadfield: Scratch that, the viewport observation is flowing into the carousel. But if we've scrolled (?) to the end of the carousel and then expand the page it doesn't respond appropriately. I'd consider this minor / don't need to fix:

image


A separate issue is the number of cards displayed for a given width, which is causing them to become rectangles:

image

trvrb:
Thanks for diving in here @victorlin! This might be personal bias, but I'm not a big fan of carousel UI on almost any website. In this case especially, you have to do a lot of clicking to reveal the additional tiles one at a time.

My guess is that a nicer UI would be a dropdown carrot that expands (and collapses back) from a single row of tiles to a grid of tiles showing everything. This way initial view of tiles doesn't eclipse the cards, but you can expand to see everything if you'd like.

victorlin:
@trvrb thanks for the feedback, I'll explore that idea with a separate PR so we can compare the two. Note that we also plan to use this on the splash page at https://nextstrain.org.

tsibley:
> My guess is that a nicer UI would be a dropdown carrot that expands (and collapses back) from a single row of tiles to a grid of tiles showing everything. This way initial view of tiles doesn't eclipse the cards, but you can expand to see everything if you'd like.

I like that idea. Something like the following (bad) mockup?

expand

victorlin:
Closing in favor of #878

Issue #871, opened 2024-05-20
Start using TypeScript


victorlin:
It would be nice to allow usage of TypeScript files in this project similar to how it was introduced in Auspice: https://github.com/nextstrain/auspice/pull/1450

This would affect the way the site is built, but given the benefits of TypeScript, I think we're all fine with the added complexity (mentioned briefly in https://github.com/nextstrain/private/issues/88)

genehack:
*big flaming +1*

jameshadfield:
One salient difference between this project and Auspice is that Auspice has a bundler/transpiler layer whereas this repo runs directly through nodejs. In other words there's no "build stage" for the server. 

Is this issue proposing using `ts-node` (or similar), or proposing adding a build stage, or something else?

victorlin:
> there's no "build stage" for the server

Previously there wasn't, but now there is with the addition of Next.js, at least based on my understanding in https://github.com/nextstrain/nextstrain.org/pull/810#discussion_r1558304436.

I just noticed 2eedca4d7eba4b88012eb63fa694f9d44f2fff1b already introduced traces of TypeScript: a tsconfig file and 2 TSX files. So it looks like it's already used behind the scenes by Next.js - I think we just need to make our own configurations to use it elsewhere.

victorlin:
I see – I forgot that there is still a big distinction between frontend and server even though they share the same package.json now. Yeah it would be worth trying a few options and comparing the complexities here - I haven't used any of them before.

> since it's no longer siloed into `static-site/` quite like it used to be we have the top-level tsconfigs

It's still under `static-site/tsconfig.json` and can't be moved up to top-level with the way the site is built currently.

genehack:
> I see – I forgot that there is still a big distinction between frontend and server even though they share the same package.json now. 

FWIW, $JOB[-1] had a monorepo with an Express-based backend app / API server and a Svelte-based front-end. They shared a package.json (IIRC anyway) and there wasn't a perceptively painful build step for the backend during normal development, only when we built a specific artifact for deployment (to AWS, as a "lambda-lith"). 

Happy to mock up a small demo repo with what that looked like if folks would like to kick the tires. (Note: this is not a proposal to convert the front-end to Svelte!)

jameshadfield:
> This would affect the way the site is built, but given the benefits of TypeScript

Upon re-reading this today I realise there may be two separate conversations going on in this thread (and that may be on me!)

[...]

PR #874, opened 2024-05-21
Add types to ListResources

victorlin:
## Description

Taking a first stab at applying TypeScript to a set of components. Suggestions and discussion welcome for how I've used TypeScript in this PR, as it's likely to shape the way we use TypeScript in this project going forwards.

## Related issues

- Based on #877
- #871

## Discussions

- function style ([ref](https://github.com/nextstrain/nextstrain.org/pull/874#discussion_r1613813999))
- `type` vs. `interface` ([ref](https://github.com/nextstrain/nextstrain.org/pull/874#discussion_r1613822159))
- type/non-null assertions ([ref](https://github.com/nextstrain/nextstrain.org/pull/874#discussion_r1623070378))

## Checklist

- [x] [Preview](https://nextstrain-s-victorlin--wzhki5.herokuapp.com/pathogens) looks good, no functional changes
- [x] Checks pass

### DIFF
--- a/static-site/src/components/ListResources/IndividualResource.tsx
+++ b/static-site/src/components/ListResources/IndividualResource.tsx
@@ -3,6 +3,8 @@ import React, {useState, useRef, useEffect, useContext} from 'react';
 import styled from 'styled-components';
 import { MdHistory } from "react-icons/md";
 import { SetModalContext } from './Modal';
+import { ResourceDisplayName, Resource } from './types';
+import { IconType } from 'react-icons';
 
 export const LINK_COLOR = '#5097BA'
 export const LINK_HOVER_COLOR = '#31586c'
@@ -17,7 +19,7 @@ export const LINK_HOVER_COLOR = '#31586c'
 const [resourceFontSize, namePxPerChar, summaryPxPerChar] = [16, 10, 9];
 const iconWidth = 20; // not including text
 const gapSize = 10;
-export const getMaxResourceWidth = (displayResources) => {
+export const getMaxResourceWidth = (displayResources: Resource[]) => {
   return displayResources.reduce((w, r) => {
     /* add the pixels for the display name */
     let _w = r.displayName.default.length * namePxPerChar;
@@ -29,7 +31,7 @@ export const getMaxResourceWidth = (displayResources) => {
   }, 200); // 200 (pixels) is the minimum
 }
 
-export const ResourceLink = styled.a`
+export const ResourceLink = styled.a<{$hovered?: boolean}>`
   font-size: ${resourceFontSize}px;
   font-family: monospace;
   white-space: pre; /* don't collapse back-to-back spaces */
@@ -37,7 +39,14 @@ export const ResourceLink = styled.a`
   text-decoration: none !important;
 `;
 
-function Name({displayName, $hovered, href, topOfColumn}) {
+interface NameProps {
+  displayName: ResourceDisplayName
+  $hovered?: boolean
+  href: string
+  topOfColumn: boolean
+}
+
+function Name({displayName, $hovered = false, href, topOfColumn}: NameProps) {
   return (
     <ResourceLink href={href} target="_blank" rel="noreferrer" $hovered={$hovered}>
       {'• '}{($hovered||topOfColumn) ? displayName.hovered : displayName.default}
@@ -70,7 +79,15 @@ export function TooltipWrapper({description, children}) {
   )
 } 
 
-export function IconContainer({Icon, text, handleClick=undefined, color=undefined, hoverColor=undefined}) {
+interface IconContainerProps {
+  Icon: IconType
+  text: string
+  handleClick?: () => void
+  color?: string
+  hoverColor?: string
+}
+
+export function IconContainer({Icon, text, handleClick, color, hoverColor}: IconContainerProps) {
   const [hovered, setHovered] = useState(false);
   const defaultColor = '#aaa';
   const defaultHoverColor = "rgb(79, 75, 80)";
@@ -89,14 +106,14 @@ export function IconContainer({Icon, text, handleClick=undefined, color=undefine
 }
 
 
-/**
- * 
- * @param {*} param0 
- * @returns 
- */
-export const IndividualResource = ({data, isMobile}) => {
+interface IndividualResourceProps {
+  data: Resource
+  isMobile: boolean
+}
+
+export const IndividualResource = ({data, isMobile}: IndividualResourceProps) => {
   const setModal = useContext(SetModalContext);
-  const ref = useRef(null);
+  const ref = useRef<HTMLDivElement>(null);
   const [topOfColumn, setTopOfColumn] = useState(false);
   useEffect(() => {
     /* The column CSS is great but doesn't allow us to know if an element is at
--- a/static-site/src/components/ListResources/IndividualResource.tsx
+++ b/static-site/src/components/ListResources/IndividualResource.tsx
@@ -21,11 +21,13 @@ const iconWidth = 20; // not including text
 const gapSize = 10;
 export const getMaxResourceWidth = (displayResources: Resource[]) => {
   return displayResources.reduce((w, r) => {
+    if (!r.displayName || !r.updateCadence) return w
+
     /* add the pixels for the display name */
     let _w = r.displayName.default.length * namePxPerChar;
     if (r.nVersions) {
       _w += gapSize + iconWidth;
-      _w += ((r?.updateCadence?.summary?.length || 0) + 5 + String(r.nVersions).length)*summaryPxPerChar;
+      _w += ((r.updateCadence.summary.length || 0) + 5 + String(r.nVersions).length)*summaryPxPerChar;
     }
     return _w>w ? _w : w;
   }, 200); // 200 (pixels) is the minimum
@@ -113,15 +115,41 @@ interface IndividualResourceProps {
 
 export const IndividualResource = ({resource, isMobile}: IndividualResourceProps) => {
   const setModalResource = useContext(SetModalResourceContext);
+  if (!setModalResource) throw new Error("Context not provided!")
+
   const ref = useRef<HTMLDivElement>(null);
   const [topOfColumn, setTopOfColumn] = useState(false);
   useEffect(() => {
+    // don't do anything if the ref is undefined or the parent is not a div (IndividualResourceContainer)
+    if (!ref.current
+     || !ref.current.parentNode
+     || ref.current.parentNode.nodeName != 'DIV') return;
+
     /* The column CSS is great but doesn't allow us to know if an element is at
     the top of its column, so we resort to JS */
-    if (ref.current.offsetTop===ref.current.parentNode.offsetTop) {
+    if (ref.current.offsetTop===(ref.current.parentNode as HTMLDivElement).offsetTop) {
       setTopOfColumn(true);
     }
   }, []);
+
+  // don't show anything if display name is unavailable
+  if (!resource.displayName) return null
+
+  // add history if mobile and resource is versioned
+  let history: React.JSX.Element | null = null
+  if (!isMobile && resource.versioned && resource.updateCadence && resource.nVersions) {
+    history = (
+      <TooltipWrapper description={resource.updateCadence.description +
+        `<br/>Last known update on ${resource.lastUpdated}` +
+        `<br/>${resource.nVersions} snapshots of this dataset available (click to see them)`}>
+        <IconContainer
+          Icon={MdHistory}
+          text={`${resource.updateCadence.summary} (n=${resource.nVersions})`}
+          handleClick={() => setModalResource(resource)}
+        />
+      </TooltipWrapper>
+    )
+  }
 
   return (
     <Container ref={ref}>
@@ -134,17 +162,7 @@ export const IndividualResource = ({resource, isMobile}: IndividualResourceProps
           </ResourceLinkWrapper>
         </TooltipWrapper>
 
-        {resource.versioned && !isMobile && (
-          <TooltipWrapper description={resource.updateCadence.description +
-            `<br/>Last known update on ${resource.lastUpdated}` +
-            `<br/>${resource.nVersions} snapshots of this dataset available (click to see them)`}>
-            <IconContainer
-              Icon={MdHistory}
-              text={`${resource.updateCadence.summary} (n=${resource.nVersions})`}
-              handleClick={() => setModalResource(resource)}
-            />
-          </TooltipWrapper>
-        )}
+        {history}
 
       </FlexRow>
 
--- a/static-site/src/components/ListResources/IndividualResource.tsx
+++ b/static-site/src/components/ListResources/IndividualResource.tsx
@@ -2,7 +2,7 @@
 import React, {useState, useRef, useEffect, useContext} from 'react';
 import styled from 'styled-components';
 import { MdHistory } from "react-icons/md";
-import { SetModalContext } from './Modal';
+import { SetModalResourceContext } from './Modal';
 import { ResourceDisplayName, Resource } from './types';
 import { IconType } from 'react-icons';
 
@@ -107,12 +107,12 @@ export function IconContainer({Icon, text, handleClick, color, hoverColor}: Icon
 
 
 interface IndividualResourceProps {
-  data: Resource
+  resource: Resource
   isMobile: boolean
 }
 
-export const IndividualResource = ({data, isMobile}: IndividualResourceProps) => {
-  const setModal = useContext(SetModalContext);
+export const IndividualResource = ({resource, isMobile}: IndividualResourceProps) => {
+  const setModalResource = useContext(SetModalResourceContext);
   const ref = useRef<HTMLDivElement>(null);
   const [topOfColumn, setTopOfColumn] = useState(false);
   useEffect(() => {
@@ -128,20 +128,20 @@ export const IndividualResource = ({data, isMobile}: IndividualResourceProps) =>
 
       <FlexRow>
 
-        <TooltipWrapper description={`Last known update on ${data.lastUpdated}`}>
-          <ResourceLinkWrapper onShiftClick={() => setModal(data)}>
-            <Name displayName={data.displayName} href={data.url} topOfColumn={topOfColumn}/>
+        <TooltipWrapper description={`Last known update on ${resource.lastUpdated}`}>
+          <ResourceLinkWrapper onShiftClick={() => setModalResource(resource)}>
+            <Name displayName={resource.displayName} href={resource.url} topOfColumn={topOfColumn}/>
           </ResourceLinkWrapper>
         </TooltipWrapper>
 
-        {data.versioned && !isMobile && (
-          <TooltipWrapper description={data.updateCadence.description +
-            `<br/>Last known update on ${data.lastUpdated}` +
-            `<br/>${data.nVersions} snapshots of this dataset available (click to see them)`}>
+        {resource.versioned && !isMobile && (
+          <TooltipWrapper description={resource.updateCadence.description +
+            `<br/>Last known update on ${resource.lastUpdated}` +
+            `<br/>${resource.nVersions} snapshots of this dataset available (click to see them)`}>
             <IconContainer
               Icon={MdHistory}
-              text={`${data.updateCadence.summary} (n=${data.nVersions})`}
-              handleClick={() => setModal(data)}
+              text={`${resource.updateCadence.summary} (n=${resource.nVersions})`}
+              handleClick={() => setModalResource(resource)}
             />
           </TooltipWrapper>
         )}
--- a/static-site/src/components/ListResources/Modal.tsx
+++ b/static-site/src/components/ListResources/Modal.tsx
@@ -247,9 +247,11 @@ function _draw(ref, resource: Resource) {
       .attr("cy", (d) => heights.height - heights.marginBelowAxis - heights.marginAboveAxis - radius - padding - d.y)
       .attr("r", radius)
       .attr('fill', color)
+      // @ts-expect-error
       .on("mouseover", function(e, d) {
         /* lower opacity of non-hovered, increase radius of hovered circle */
         beeswarm.join(
+          // @ts-expect-error
           (enter) => {}, /* eslint-disable-line */
           (update) => selectSnapshot(update, d)
         )
@@ -269,6 +271,7 @@ function _draw(ref, resource: Resource) {
       })
       .on("mouseleave", function() {
         beeswarm.join(
+          // @ts-expect-error
           (enter) => {}, /* eslint-disable-line */
           (update) => resetBeeswarm(update)
         )
@@ -276,6 +279,7 @@ function _draw(ref, resource: Resource) {
         selectedVersionGroup.selectAll("*")
           .style("opacity", 0)
       })
+      // @ts-expect-error
       .on("click", function(e, d) {
         window.open(`/${resource.name}@${d.data.version}`,'_blank'); // TEST!
       })
@@ -296,6 +300,7 @@ function _draw(ref, resource: Resource) {
       .on("mousemove", function(e) {
         const { datum, hoveredDateStr } = getVersion(e);
         beeswarm.join(
+            // @ts-expect-error
             (enter) => {}, /* eslint-disable-line */
             (update) => selectSnapshot(update, datum)
         )
@@ -316,6 +321,7 @@ function _draw(ref, resource: Resource) {
       })
       .on("mouseleave", function() {
         beeswarm.join(
+          // @ts-expect-error
           (enter) => {}, /* eslint-disable-line */
           (update) => resetBeeswarm(update)
         )
--- a/static-site/src/components/ListResources/Modal.tsx
+++ b/static-site/src/components/ListResources/Modal.tsx
@@ -6,18 +6,18 @@ import { MdClose } from "react-icons/md";
 import { dodge } from "./dodge";
 import { Resource } from './types';
 
-export const SetModalContext = createContext<React.Dispatch<React.SetStateAction<Resource | undefined>> | null>(null);
+export const SetModalResourceContext = createContext<React.Dispatch<React.SetStateAction<Resource | undefined>> | null>(null);
 
 export const RAINBOW20 =   ["#511EA8", "#4432BD", "#3F4BCA", "#4065CF", "#447ECC", "#4C91BF", "#56A0AE", "#63AC9A", "#71B486", "#81BA72", "#94BD62", "#A7BE54", "#BABC4A", "#CBB742", "#D9AE3E", "#E29E39", "#E68935", "#E56E30", "#E14F2A", "#DC2F24"];
 const lightGrey = 'rgba(0,0,0,0.1)';
 
 
 interface ResourceModalProps {
-  data?: Resource
+  resource?: Resource
   dismissModal: () => void
 }
 
-export const ResourceModal = ({data, dismissModal}: ResourceModalProps) => {  
+export const ResourceModal = ({resource, dismissModal}: ResourceModalProps) => {  
   const [ref, setRef] = useState(null); 
   const handleRef = useCallback((node) => {setRef(node)}, [])
 
@@ -37,32 +37,32 @@ export const ResourceModal = ({data, dismissModal}: ResourceModalProps) => {
   }, []);
 
   useEffect(() => {
-    if (!ref || !data) return;
-    _draw(ref, data)
-  }, [ref, data])
+    if (!ref || !resource) return;
+    _draw(ref, resource)
+  }, [ref, resource])
 
-  if (!data) return null;
+  if (!resource) return null;
 
-  const summary = _snapshotSummary(data.dates);
+  const summary = _snapshotSummary(resource.dates);
   return (
     <div ref={scrollRef}>
     
       <Background onClick={dismissModal}/>
       <ModalContainer onClick={(e) => {e.stopPropagation()}}>
         <CloseIcon onClick={dismissModal}/>
-        <Title>{data.name.replace(/\//g, "│")}</Title>
+        <Title>{resource.name.replace(/\//g, "│")}</Title>
 
         <div style={{paddingBottom: '5px'}}>
           <Bold>
-            {`${data.dates.length} snapshots spanning ${summary.duration}: ${summary.first} - ${summary.last}`}
+            {`${resource.dates.length} snapshots spanning ${summary.duration}: ${summary.first} - ${summary.last}`}
           </Bold>
           <a style={{fontSize: '1.8rem', paddingLeft: '10px'}}
-            href={`/${data.name}`}  target="_blank" rel="noreferrer noopener">
+            href={`/${resource.name}`}  target="_blank" rel="noreferrer noopener">
             (click to view the latest available snapshot)
           </a>
         </div>
         <div>
-          {data.updateCadence.description}
+          {resource.updateCadence.description}
         </div>
 
         <div ref={handleRef} /> {/* d3 controlled div */}
@@ -142,11 +142,11 @@ function _snapshotSummary(dates: string[]) {
   return {duration, first: d[0], last:d.at(-1)};
 }
 
-function _draw(ref, data: Resource) {
+function _draw(ref, resource: Resource) {
   /* Note that _page_ resizes by themselves will not result in this function
   rerunning, which isn't great, but for a modal I think it's perfectly
   acceptable */  
-  const sortedDateStrings = [...data.dates].sort();
+  const sortedDateStrings = [...resource.dates].sort();
   const flatData = sortedDateStrings.map((version) => ({version, 'date': new Date(version)}));
 
   const width = ref.clientWidth;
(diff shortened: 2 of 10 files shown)


static-site/src/components/ListResources/types.ts
```diff
@@ -0,0 +1,45 @@
...
+    groupName: any
+    nResources: number
+    nVersions?: number
+    lastUpdated: any
+    resources: Resource[]
+    groupUrl?: any
+    groupDisplayName?: any
+}
+
+export type Resource = {
```
jameshadfield, line 17: If TS were available this would have made my life a lot easier when implementing the UI. So +1 for this effort!

victorlin: types = easier to implement + easier to understand!

static-site/src/components/ListResources/Showcase.tsx
```diff
@@ -28,10 +34,15 @@ export const Showcase = ({cards, setSelectedFilterOptions}) => {
   )
 }
 
+type ShowcaseTileProps = {
+    data: Card
```
genehack, line 38: <nitpick> `data` is pretty generic to start with; having it typed as `Card` increases the cognitive dissonance. Consider? 

```suggestion
    card: Card
```

static-site/src/components/ListResources/Showcase.tsx
```diff
@@ -115,11 +126,11 @@ const Byline = styled.div`
  * which the filters are valid given the resources known to the resource listing
  * UI
  */
-export const useShowcaseCards = (cards, resources) => {
-  const [restrictedCards, setRestrictedCards] = useState([]);
+export const useShowcaseCards = (cards: Card[], groups: Group[]) => {
```
genehack, line 129: why no `*Props` type for the method arguments here? The lack of deconstruction?

static-site/src/components/ListResources/types.ts
```diff
@@ -0,0 +1,45 @@
+// FIXME: is this in react-select?
+export type FilterOption = {
+    value: string
+    label: string
+}
+
+export type Group = {
+    groupName: any
```
genehack, line 8: `any` types, IME, are timebombs. If you really can't type it any more precisely than this, use `unknown` -- that will force you to deal with things the compiler doesn't like. An `any` type is just a way of doing a lot of extra work to get zero benefit.

genehack (commented):
++1 for getting started on this! Left some more or less nitpicky commentary; the `any` types are the thing I feel strongest about.

static-site/src/components/ListResources/Showcase.tsx
victorlin: Yeah, the `*Props` definitions was something I used in another project that seems to work well for deconstruction and also reusability between functions that had an overlap in props. It can't be applied here. The closest equivalent I can think of would be this, which I think is more redundant and less readable:

```ts
type useShowcaseCardsFunction = (cards?: Card[], groups?: Group[]) => Card[]

export const useShowcaseCards: useShowcaseCardsFunction = (cards?, groups?) => {
  const [restrictedCards, setRestrictedCards] = useState<Card[]>([]);
  // …
```

static-site/src/components/ListResources/Showcase.tsx
victorlin: Good suggestion, I'll add this to deb9e0c22da15b8c05994e84abacd4917cd0f67c

static-site/src/components/ListResources/types.ts
victorlin: Agreed. I put `any` for things that I couldn't deduce quickly. I'll take a closer look at each and consult @jameshadfield if I still can't figure it out.

genehack: In the past, I've had decent luck looking at what VS Code (or any LSP-capable editor) infers based on where the type'd thing is being used — not 100% of the time, but it's sometimes a short cut.

static-site/src/components/ListResources/Showcase.tsx
victorlin: Added as part of e03c7b5f9a9bbbfaa82b816c03de00e17c5e73b1

static-site/src/components/ListResources/types.ts
victorlin: Yeah I did that where it was helpful - all the `any`s were already detected as `any` by VS Code. I've gone through most but there's a few left in 6198eef3a51c07fdf31778377afe68d80810b700. Might get to them tomorrow but I'm signing off for the day.

ivan-aksamentov:
Not strictly Typescript related and very subjective, but as you are going through the code conversion, how do you guys feel about also reducing the punctuation noise?

Functions don't have to be anonymous closures assigned to named variables:

```diff
-export const useShowcaseCards = (cards: Card[], groups: Group[]) => {
+export function useShowcaseCards(cards: Card[], groups: Group[]) {
```

P.S. Watch out for `this` and other runtime quirks, especially in d3 code.

I believe there exist a transformer tool or a eslint rule with autofix to apply this en masse.


Also, interfaces don't have to be aliases to anonymous object types:

```diff
-type ShowcaseTileProps = {
+interface ShowcaseTileProps {
```

P.S. Watch out for differences as well, notably when extending.

This one is a bit longer to type, but there's 1 less token (`=`) to parse with your eyes and we all use autocomplete and snippets nowadays, [right](https://github.com/nextstrain/nextstrain.org/assets/9403403/87ba3ce5-ec11-4779-9428-ffa36a52ca28)?

<details>

<summary>An optional "old man yells at a cloud" rant/tantrum</summary>

I believe both are abuses of the language, coming from the days when ES6 and TS were new and trendy, lacking in features and in permanent wild west of development and experimentation. Oh and also the threadoors: 

> See, I can do cool things like `= () => { }`. It's not your grandpa's outdated `function() {}` anymore. Buy my ES6 course! ($69.99 only this week)"

There was a legitimate use in React classes, and arrow functions are still useful when anonymous, but now that the React classes are mostly gone and things are well shaken out, this syntax for named free functions and functional React components feels redundant and a bit 2015-ish.

</details>

static-site/src/components/ListResources/Showcase.tsx
```diff
@@ -28,26 +34,31 @@ export const Showcase = ({cards, setSelectedFilterOptions}) => {
   )
 }
 
+type ShowcaseTileProps = {
```
victorlin, line 37: Re: [suggestion 2/2](https://github.com/nextstrain/nextstrain.org/pull/874#issuecomment-2129012940)

> [@ivan-aksamentov] Also, interfaces don't have to be aliases to anonymous object types:
> 
> ```diff
> -type ShowcaseTileProps = {
> +interface ShowcaseTileProps {
> ```
> 
> P.S. Watch out for differences as well, notably when extending.

Differences are listed in the [TypeScript Handbook](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces), and answers to this [StackOverflow question](https://stackoverflow.com/q/37233735) provide some clear examples.

I have a preference for `type`. Reasons:

1. It ensures all fields are declared in a single declaration. I don't see any reason to allow adding new fields via multiple `interface` declarations and find that behavior unintuitive when looking at the code in the TypeScript Handbook example.
2. It allows renaming primitives such as `type Name = string`. I just pushed an example of how this can be useful and self-documenting: f2e4113dd3e7c472cbcdcbd65b73b32ed0258972
3. The name `type` aligns with what it's doing and what it's used for (**Type**Script)
4. I don't mind the extra `=`, I actually find it intuitive.

That said, I haven't used `interface` much to make other comparisons so I'd be curious to know if there are other reasons you prefer it over `type`.

static-site/src/components/ListResources/Showcase.tsx
```diff
@@ -115,19 +126,19 @@ const Byline = styled.div`
  * which the filters are valid given the resources known to the resource listing
  * UI
  */
-export const useShowcaseCards = (cards, resources) => {
-  const [restrictedCards, setRestrictedCards] = useState([]);
+export const useShowcaseCards = (cards?: Card[], groups?: Group[]) => {
```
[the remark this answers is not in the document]
tsibley: > Functions don't have to be anonymous closures assigned to named variables:

In general I'm +1 for using named functions instead of anonymous funcs assigned to named variables. (All functions are closures in JS, right? Not just the arrow funcs.)

But it's funny, because the arrow func syntax reminds me of Haskell and in the context of [doing functional composition in JS](https://github.com/nextstrain/nextstrain.org/blob/08db0bd73841c8e0ef798e679167fb11ae014664/src/endpoints/sources.js#L83-L120), the arrow syntax feels way more natural to me. So "use named functions" isn't an absolute thing for me.

static-site/src/components/ListResources/Showcase.tsx
tsibley: > Is there any downside to thinking/pretending that it does create a new type?

Giving the appearance that a new type is created but not actually creating one means that (to reiterate Ivan's example) it's very easy to put a `type NucSequence = string` value into a `type Address = string` slot and TS will be happy as a clam. An actual new type wouldn't let you mix them up, and can be a useful guardrail against programming errors.

That said, `type` in [Haskell](https://wiki.haskell.org/Type) and [Python](https://docs.python.org/3/library/typing.html#type-aliases) both create aliases too…

static-site/src/components/ListResources/types.ts
jameshadfield: Gradual adoption of TS to a codebase is a balancing act as judicial use of `any`/`unknown` is necessary to stop you introducing types to the whole codebase! But for this case, `groupName` is within the boundary of `ListResources` so it should be typed. Looking at 6198eef there's 7 `any` types left, and all of them should be typeable. I can help with this next week.

static-site/src/components/ListResources/Showcase.tsx
jameshadfield: > That said, type in [Haskell](https://wiki.haskell.org/Type) and [Python](https://docs.python.org/3/library/typing.html#type-aliases) both create aliases too…

Yeah, this is (overly) common and is a common source of errors. Rust does it as well - the following is just fine!

```rust
type CdsName = String;
type GeneName = String;

fn print_gene(name: GeneName) {
    print!("GENE: {name}")
}

fn main() {
    let cds: CdsName = String::from("this is a CDS");
    print_gene(cds);
}
```

victorlin: Thanks for the examples, I understand the issue better now. But it's specific the point on renaming primitives and I don't see a clear suggestion. Should we avoid usage such as f2e4113dd3e7c472cbcdcbd65b73b32ed0258972 by avoiding `type` entirely and using `interface` for object types?

tsibley: My 2¢ would be to generally avoid aliases except in (hopefully rare) cases where a) they make a complex type used lots of places easier to write repeatedly and b) we really don't want a new specific type. Meeting both (a) and (b) would be fairly rare, I think. Aliases like `type GroupName = string` wouldn't meet that bar and we'd instead use the `string` type directly or actually declare a new `GroupName` type.

static-site/src/components/ListResources/Modal.tsx
```diff
@@ -125,8 +132,8 @@ const Title = styled.div`
 
 function _snapshotSummary(dates) {
```
genehack, line 133: Might make some of the need for subsequent casts go away?

```suggestion
function _snapshotSummary(dates: [number]) {
```

static-site/src/components/ListResources/Showcase.tsx
genehack: +1 for not bothering with aliases to simple built-in types

victorlin: Ok, I'll drop f2e4113dd3e7c472cbcdcbd65b73b32ed0258972 to get rid of aliases.

The discussion on `type` vs. `interface` for object types is still up in the air – I'll keep `type` in this PR unless I hear another vote for `interface`. I've also started some work (35e004b0a05d437033c880a4f5be743a5ff23975...98fdcc9d9d49e0371237dcb1c7ebe6adf39e544f) which might influence the decision here.

jameshadfield: > The discussion on type vs. interface for object types is still up in the air 

I don't feel as strongly as Ivan does about the "2-step procedure of (a) creating an anonymous object type and then (b) assigning it to a name, when there's a native construct describing object types exists - interface", but he's written more TS than me so I'd listen to him! As far as I understand, they still don't guard against the mistakes we've been talking about in this thread, e.g. both allow you to do the following bad thing:

```js
{
    interface Gene {name: string};
    interface Cds {name: string};
    const gene1:Gene = {name: "gene1value"}
    const cds1:Cds = {name: "cds1value"}
    const printGene = (gene: Gene): void => console.log(`GENE: ${gene.name}`)
    printGene(gene1)
    printGene(cds1)
}
{    
    type Gene = {name: string};
    type Cds = {name: string};
    const gene1:Gene = {name: "gene1value"}
    const cds1:Cds = {name: "cds1value"}
    const printGene = (gene: Gene): void => console.log(`GENE: ${gene.name}`);
    printGene(gene1)
    printGene(cds1)
}

```

Tangential question: is the above situation always and necessarily the case in a [structural type system](https://en.wikipedia.org/wiki/Structural_type_system)?

> +1 for not bothering with aliases to simple built-in types

I'd bother with them when it makes things clarifying for development (i.e. increases self-documentation), even though the type system doesn't distinguish between them. As in, we don't have to use them systematically but when it's clarifying to do so we can.

genehack: > > +1 for not bothering with aliases to simple built-in types
> 
> I'd bother with them when it makes things clarifying for development (i.e. increases self-documentation), even though the type system doesn't distinguish between them. As in, we don't have to use them systematically but when it's clarifying to do so we can.

I find 

```typescript
type Gene = string;

// some potentially large amount of semantic distance away, perhaps in another file
const gene: Gene = "ampR";
```

to be significantly less clear than:

```typescript
const gene = "ampR";
```

because I don't have to wonder what's going on inside `Gene`. I think the "clarifying for development" part can generally be conveyed via good variable names (`gene` being a good example of a horrible one...) 

Aliasing simple types adds an additional layer of abstraction and complexity with (to me) minimal benefit over what it replaces.

victorlin: Switched to `interface` in 213995c9b8a189285ec62094009204e0d981f70d

static-site/src/components/ListResources/Modal.tsx
victorlin: Updated to:

https://github.com/nextstrain/nextstrain.org/blob/213995c9b8a189285ec62094009204e0d981f70d/static-site/src/components/ListResources/Modal.tsx#L133-L137

static-site/src/components/ListResources/types.ts
victorlin: Replaced explicit `any`s in 8e5e5daab7b489ef5368d881f45f11930ef07de7

static-site/src/components/ListResources/IndividualResource.tsx
```diff
@@ -111,20 +129,20 @@ export const IndividualResource = ({data, isMobile}) => {
 
       <FlexRow>
 
-        <TooltipWrapper description={`Last known update on ${data.lastUpdated}`}>
-          <ResourceLinkWrapper onShiftClick={() => setModal(data)}>
-            <Name displayName={data.displayName} href={data.url} topOfColumn={topOfColumn}/>
+        <TooltipWrapper description={`Last known update on ${resource.lastUpdated}`}>
+          <ResourceLinkWrapper onShiftClick={() => setModalResource && setModalResource(resource as VersionedResource)}>
```
ivan-aksamentov, line 133: I find amount of casting (`as` operator) excessive. It is not uncommon for projects in transition from js to ts to have some, in particularly uncertain places and where typed code needs to interact with yet untyped code. But upon a quick examination I found at least one serious typing issue which is being hidden by the cast.

Here we have a cast object `resource`, an instance of type `ResourceWithDisplayName` to  type `VersionedResource`. This is needed because `VersionedResource` the underlying type of the context, but `ResourceWithDisplayName` extends `Resource`, not `VersionedResource`.

The `Resource` interface contains all fields of the `VersionedResource` interface. So does it mean every `Resource` is also a `VersionedResource`? In this case this makes  little sense:

```ts
// types.ts
interface VersionedResource extends Resource
```

because it says "VersionedResource is a Resouce". So we just deduced that:

 - Resouce is VersionedResource
 - VersionedResource is Resouce

Meaning they are the same type.


I think (looking at names of the types) that the intent was different. Here is my proposal:

 - remove the 4 fields from the `Resource` and keep them only in `VersionedResource`: firstUpdated, dates, nVersions, updateCadence. Consider also moving versioned and lastUpdated (I don't know the meaning or concrete use, but from names they sound like they belong to something that's "versioned")

 - extend concrete resources from `VersionedResource`, rather than from `Resource`:

    ```ts
    export interface ResourceWithDisplayName extends VersionedResource {
    ```

 - obtain the following inheritance chain:

	```
	Resource -> VersionedResource -> ResourceWithDisplayName
	```

 - returning to the code snippet in review, the `resource`, which is still an instance of `ResourceWithDisplayName`, but now it is also an instance of `VersionedResource`, so no cast is needed.

I don't know if this proposal is compatible with the rest of the code, but my impression from the names of interfaces and fields is that this was the idea.

It might be worth to review all other occurrences of casting. It is better to keep casting to minimum, because it can lead to runtime errors if incompatible types are cast and then accessed incorrectly. It undermines the protection mechanism which is the whole point of adding ts. In C++ community they like to say "Casting is lying to your compiler, and compiler is your friend. You are lying to your friend who tries to help you." (citing in my words, but close to the original)

victorlin: @ivan-aksamentov thanks for flagging this - the `Resource`/`VersionedResource`/`ResourceWithDisplayName` thing was something I struggled with and could not come up with anything better on a Friday afternoon.

I went through all the type casts/assertions and replaced all but one: 7b24562d68ae048c949d0d1746b23ace37ea3941

Instead of extending `Resource` I figured it might be better to use it strictly and handle the optional attributes explicitly. What do you think of this approach?

static-site/src/components/ListResources/IndividualResource.tsx
```diff
@@ -111,20 +129,20 @@ export const IndividualResource = ({data, isMobile}) => {
...
+        {resource.versioned && !isMobile && (
+          <TooltipWrapper description={(resource as VersionedResource).updateCadence.description +
+            `<br/>Last known update on ${resource.lastUpdated}` +
+            `<br/>${resource.nVersions} snapshots of this dataset available (click to see them)`}>
             <IconContainer
               Icon={MdHistory}
-              text={`${data.updateCadence.summary} (n=${data.nVersions})`}
-              handleClick={() => setModal(data)}
+              text={`${(resource as VersionedResource).updateCadence.summary} (n=${resource.nVersions})`}
+              handleClick={() => setModalResource && setModalResource(resource as VersionedResource)}
```
ivan-aksamentov: Stylistic suggestion:

To increase signal-to-noise ratio, you could use JavaScript's Elvis operator ([optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)). It works for function calls too:

```ts
setModalResource?.(resource)
```

Interestingly, the null check was not needed in js version. It's either that the function (context) can be typed more strictly (i.e. the function actually cannot be nil) or that this has been a bug in js version. To know, need to check context docs and how it is initialized in this particular case. But I think this does not matter much to be worth a hassle.

static-site/src/components/ListResources/IndividualResource.tsx
ivan-aksamentov: @victorlin Not sure. Time will tell. I don't quite like the proliferation of non-null assertions now. It's really the same as cast, but which is harder to find.

In order to decide the granullarity of typings, I'd usualy look at the origin of these objects (e.g. incoming HTTP request JSON, a file, a constant etc.). Where do they come from and in what shape?

If it comes as one monolithic object, then do one interface. If you find that in some places the null fields make no sense or become too tedious, then split the interface up and then combine the parts with `extend` and/or `&`. If you find that some function only wants a part of the object, then you could also split up.

One important hint:

If you's think about a program as of a black box with inputs and outputs:

```
input -> business logic -> output
```

then in TS you only need to type the input and the output - to make sure they have the shape that's expected on entry to and exit from the program. Within business logic, most types should be inferred by the compiler pretty well, so explicit types might not be needed at all. Notable exception is React, because you need props as all-in-one input, so you have to create new types there. For no reason really. But it's just a weird animal.

So if I'd convert a codebase to Typescript, I'd start with the inputs. For example, let's say for Auspice codebase, you could take Augur's JSON schema and generate TS types from it (I like using [quicktype](https://github.com/glideapps/quicktype)). And this will be like 80% of the typings for the whole program.

ivan-aksamentov:
Looking at the code a little more closely I see that there are some cases of unsound js code ([1](https://github.com/nextstrain/nextstrain.org/commit/7b24562d68ae048c949d0d1746b23ace37ea3941#r142683167), [2](https://github.com/nextstrain/nextstrain.org/commit/7b24562d68ae048c949d0d1746b23ace37ea3941#r142684106)). It is difficult or impossible to provide sound typings for unsound code. The usual problems (and both of the examples I highlighted) are related to accessing potentially a null pointer.

Another example I saw is the `ref.current.parentNode` type casting to `HTMLElement`. This is an assumption ("we know") which can be broken, making this potentially a null pointer access undetectable by the compiler.


Hence, in terms of methodology, in the process of conversion, there is need for either:

 - fixing the code
 - adding a typing hack until later

One should decide which direction to chose on a case-by-case basis. If a hack direction is taken (cast, bang etc.), it is better to clearly document why it is needed with one of the loud comments, e.g. `FIXME(ts)` or `HACK(types)` and explain why types could not be correctly written here, **without blaming the silly compiler**, e.g.

```ts
// FIXME(ts): `dates` can be empty,
// but this code assumes at least 2 elements.
// We need explicit checks here (or a utility function).
```

not

```ts
// Silly compiler... Such a sweet summer child...
// Let's make it happy. Where is candies?
dates.at(999999999)!.explode
```

This way it's easy to grep and fix it later. By contrast, solitary bangs and casts are hard to find if there is no additional marking.


There is a trade-off:

 - try to add typings with only minimal changes to the codebase (makeing it a refactoring)

 - make typings sound (which requires fixes to the runtime behavior)


One of the things I like the most about TS is that when you add it, it highlights all the bad code immediately. And once TS is in place, it prevents from writing whole classes of new bad code - they will simply not compile. And if type hacks don't pass code reviews, then the codebase slowly becomes magically better.

static-site/src/components/ListResources/IndividualResource.tsx
jameshadfield: I haven't written all that much TS, but I found that implementing something in TS involved a different mindset compared with a JS implementation. I found it very clarifying to be forced to define structures and not resort to conditional assertions & object modification which I tend to use a lot in pure JS. 

Looking at this code, there's no real difference in my JS implementation between `Resource`, `VersionedResource` and `ResourceWithDisplayName` objects - it's just that some have `undefined` values and we perform conditional checks on that in the consuming code.

This brings me to a more general point, and one I don't have a simple answer for. Converting JS to TS isn't a matter of simply adding type annotations to existing code, it is a process which should change the structure of the code as well. 

> If you's think about a program as of a black box with inputs and outputs... then in TS you only need to type the input and the output 

In this situation I think `<ListResources>` is a pretty good starting point - we have an input (JSON data sent from the server) and the output is react components.

victorlin: Thanks both for the input.

> There is a trade-off:
> 
> * (1) try to add typings with only minimal changes to the codebase (makeing it a refactoring)
> * (2) make typings sound (which requires fixes to the runtime behavior)

My goal with this PR is to do both sequentially - (1) with 8e5e5daab7b489ef5368d881f45f11930ef07de7 then (2) with 5151710b54c19ca4f86a3d7f039c5cadef2f0423.

---

> Looking at this code, there's no real difference in my JS implementation between `Resource`, `VersionedResource` and `ResourceWithDisplayName` objects - it's just that some have `undefined` values and we perform conditional checks on that in the consuming code.

This is now my sense of things too after removing the distinction in 7b24562d68ae048c949d0d1746b23ace37ea3941. I think it's cleaner this way and will squash most of that commit into 5151710b54c19ca4f86a3d7f039c5cadef2f0423.

---

The non-null assertions brought up in https://github.com/nextstrain/nextstrain.org/commit/7b24562d68ae048c949d0d1746b23ace37ea3941#r142683167 and https://github.com/nextstrain/nextstrain.org/commit/7b24562d68ae048c949d0d1746b23ace37ea3941#r142684106 are separate from the `Resource`/`VersionedResource`/`ResourceWithDisplayName` split - I'll think about and address those differently.

static-site/src/components/ListResources/IndividualResource.tsx
victorlin: Done:

https://github.com/nextstrain/nextstrain.org/blob/adc666a4ce56df6011a2ad3c076406512810bf38/static-site/src/components/ListResources/IndividualResource.tsx#L146-L147

victorlin: TypeScript is correct and it's possible for `setModalResource` to be `null` if the `SetModalResourceContext.Provider` parent is removed. The child component has no way of knowing if it's there or not.

https://github.com/nextstrain/nextstrain.org/blob/adc666a4ce56df6011a2ad3c076406512810bf38/static-site/src/components/ListResources/index.tsx#L77-L78

Such a case would definitely be an internal error though, not user error. I've updated to raise an error in this case (see [force-push](https://github.com/nextstrain/nextstrain.org/compare/adc666a4ce56df6011a2ad3c076406512810bf38..2ad11e7bc97f3a54c4a819096b9a270adfc58eb7)).

victorlin:
I've applied suggestions and cleaned up the commits – this is now ready for re-review.

ivan-aksamentov (approved):

static-site/src/components/ListResources/useDataFetch.ts
```diff
@@ -56,55 +62,56 @@ export function useDataFetch(sourceId, versioned, defaultGroupLinks, groupDispla
...
  */
-function groupsFrom(partitions, pathPrefix, defaultGroupLinks, groupDisplayNames) {
-  return Object.entries(partitions).map(makeGroup)
-
-  function makeGroup([groupName, resources]) {
-    const groupInfo = {
+function groupsFrom(partitions: Partitions, pathPrefix: string, defaultGroupLinks: boolean, groupDisplayNames: GroupDisplayNames) {
+  return Object.keys(partitions).map(groupName => {
+    // groupName will always be a valid key based on map() above
+    const resources = partitions[groupName]!;
```
jameshadfield, line 109: [not a request for changes] I guess the more ergonomic way is something like the following?
```js
Object.entries(partitions).map(([groupName, resources]) => {
```
Is this assertion needed because the compiler isn't smart enough to know that group name is taken from the valid keys of partitions, or because mutability means that a key may have been removed from `partitions` before the map function gets that key?

jameshadfield (approved):
Adding types without changing too much code isn't easy, and I've no doubt the types will help development, so thanks for this work @victorlin!

No blocking changes, just questions. I'm surprised how many bangs were needed (I gave up making in-line comments!) and for my own learning I'd be interested to know if that's just the way it is or if it's an indication that the style of code used wasn't great and a better design is wanted. (I'm not saying a refactor / rewrite is worth our effort for this code at this point.)

static-site/src/components/ListResources/Modal.tsx
```diff
@@ -123,22 +131,28 @@ const Title = styled.div`
...
 `
 
-function _snapshotSummary(dates) {
+function _snapshotSummary(dates: string[]) {
   const d = [...dates].sort()
-  const [d1, d2] = [d[0], d.at(-1)].map((di) => new Date(di));
-  const days = (d2-d1)/1000/60/60/24;
+  if (d.length < 1) throw new Error("Missing dates.")
+
+  const d1 = new Date(d.at( 0)!).getTime();
```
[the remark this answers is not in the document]
ivan-aksamentov: > Why is this bang needed? the preceding line ensures that there is at lease one element in d.

Compiler is not smart enough to figure this out from this kind of the conditional. Connecting `.length` to `.at()` is tough for a type system.

What it can easily figure out is null checks. So a cleaner way would be to check values and throw after the non-fallible array access:

```ts
const t0 = d.at(0)?.getTime()
if(isNil(t0)) { // careful to not exclude the legit value `0`
    throw ...
}
// `t0` is guaranteed to be `number` in this branch
```

I would go further. Finding first and last value is a common enough "algorithm" that I would introduce a utility function:

```ts
export function firstLastOrThrow<T>(a: T[]): [T, T] {
    // TODO: use .at() or lodash head() and tail() along with null checks 
}
```

This would make the code very pretty and safe, with all dirty details hidden:

```ts
const [t0, t1] = firstLastOrThrow(dates) // guaranteed to return a pair of numbers or to  throw
```

static-site/src/components/ListResources/Modal.tsx
```diff
@@ -164,7 +178,8 @@ function _draw(ref, data) {
 
   /* Create the x-scale and draw the x-axis */
   const x = d3.scaleTime()
-    .domain([flatData[0].date, new Date()]) // the domain extends to the present day
+    // presence of dates on resource has already been checked so this assertion is safe
```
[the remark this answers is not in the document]
ivan-aksamentov: Bracket notation returns `undefined` if accessing out of bounds (IIRC), so the type of `a` is potentially `undefined`.

The problem with comments like this:
```ts
// presence of dates on resource has already been checked so this assertion is safe
```
is that 3 months from now a fresh intern comes and inserts new code between the check and the assertion and then :boom:. Though, to be fair, it did not happen with js previously (due to lack of interns?)

Again, a small wrapper would make it safe and clean.

If you can find a fallback value, then something like this might work:

```ts
flatData[0]?.date ?? new Date()
``` 

This cannot fail and requires no hacks.

static-site/src/components/ListResources/useDataFetch.ts
ivan-aksamentov: It's simply because operator [] returns either `T` or `undefined`. It does not care about any other conditions.

static-site/src/components/ListResources/Modal.tsx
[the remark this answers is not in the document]
ivan-aksamentov: Another thing is that this entire calculation is wrong. For example, not all years have 365 days. Date-time calculations is a very hard topic, and it is somewhat unreasonable to try and do on the spot. There are specialized high-quality libraries exist to do this correctly, e.g. [luxon](https://moment.github.io/luxon/#/?id=luxon) (continuation of `moment.js`). And then for human durations there are also small libs (e.g. [this](https://github.com/EvanHahn/HumanizeDuration.js))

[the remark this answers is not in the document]
jameshadfield: > Another thing is that this entire calculation is wrong.

(Let's break this out to a separate conversation if we want to continue discussing it -- I am aware how fraught date parsing is but I'll argue that the data this function returns is (purposefully) vague such that the overall function is correct.)

static-site/src/components/ListResources/ResourceGroup.tsx
```diff
@@ -225,14 +247,14 @@ function NextstrainLogo() {
...
-function _setDisplayName(resources) {
+function _setDisplayName(resources: Resource[]) {
   const sep = "│"; // ASCII 179
   resources.forEach((r, i) => {
     let name;
     if (i===0) {
       name = r.nameParts.join(sep);
     } else {
-      let matchIdx = r.nameParts.map((word, j) => word === resources[i-1].nameParts[j]).findIndex((v) => !v);
+      let matchIdx = r.nameParts.map((word, j) => word === resources[i-1]?.nameParts[j]).findIndex((v) => !v);
```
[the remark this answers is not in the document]
victorlin: From [4.1 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#checked-indexed-accesses---nouncheckedindexedaccess) which added `noUncheckedIndexedAccess`:

> One consequence of using `noUncheckedIndexedAccess` is that indexing into an array is also more strictly checked, even in a bounds-checked loop.

static-site/src/components/ListResources/useDataFetch.ts
victorlin: @jameshadfield I got confused and didn't realize that [groupName, resources] is properly typed as long as the function is defined inline within `Object.entries().map()`. Changed 2aeb155851e225c5039fa172d96f681f6c665fba to 34a314070e2d1326555a5df78323f8e8086c9dc2 which removes the need for the bang (reflected in [force-push](https://github.com/nextstrain/nextstrain.org/compare/aae874c06cd6d17616710c501ad73104a7c5e051..7c750e278da27459f54a91ee373580e33089777d)).

victorlin:
Sorry that c9788e758c14413fa792de429427f545c3bcde9f left you with so many (valid) questions @jameshadfield. Ideally that should've been broken into multiple commits with more detailed reasoning. At this point I'll just merge – happy to discuss other details with you later.

Also thanks @ivan-aksamentov for your advise here.

TypeScript incoming!

PR #877, opened 2024-05-24
Configure TypeScript for static-site

victorlin:

See commit messages

## Related issue(s)

- #871

## Checklist

- [x] `npm run dev` doesn't overwrite TSConfig file locally
- [x] Checks pass
- [ ] Post-merge: Create issue to enable `noImplicitAny`

### DIFF
--- a/static-site/tsconfig.json
+++ b/static-site/tsconfig.json
@@ -1,27 +1,53 @@
+/*
+Much of Auspice was written in JavaScript. There is an effort to convert
+existing JavaScript files to TypeScript in an iterative, file-by-file approach.
+
+TypeScript is used in this project for type-checking only, not for its compiled
+output files - Babel is used for that. Config options should reflect this usage
+pattern.
+
+Visit https://aka.ms/tsconfig.json for a detailed list of options.
+*/
+
 {
   "compilerOptions": {
-    "lib": [
-      "dom",
-      "dom.iterable",
-      "esnext"
-    ],
-    "allowJs": true,
-    "skipLibCheck": true,
-    "strict": false,
-    "forceConsistentCasingInFileNames": true,
-    "noEmit": true,
-    "incremental": true,
-    "esModuleInterop": true,
-    "module": "esnext",
-    "moduleResolution": "node",
-    "resolveJsonModule": true,
-    "isolatedModules": true,
-    "jsx": "preserve"
+    /* Language and Environment */
+    "jsx": "react",                                      /* Specify what JSX code is generated. */
+
+    /* Modules */
+    "moduleResolution": "node",                          /* Specify how TypeScript looks up a file from a given module specifier. */
+    "types": ["node"],                                   /* Specify type package names to be included without being referenced in a source file. */
+
+    /* JavaScript Support */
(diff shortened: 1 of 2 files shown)


static-site/tsconfig.json
jameshadfield, line 1: Did you find a way to get this working with VS-Code @victorlin?

I can get VS-Code to use the workspace's `tsc` just fine (v4.9.3) but not to use this config file. According to https://github.com/microsoft/vscode/issues/12463 this is a known bug and won't be fixed.

Specifically, in the current code I get errors in `_app.tsx` because the default config doesn't use `"jsx": "preserve"`:

image

jameshadfield (approved):
Code passes type checks - both `npx next build static-site` and `cd static-site && npx tsc --noEmit`. Running the server and/or building the static-site doesn't result in modifications to the tsconfig from my testing. Mirroring the Auspice config is a nice move.

static-site/tsconfig.json
victorlin: What you saw is because that file wasn't included in the TSConfig:

https://github.com/nextstrain/nextstrain.org/blob/8d45cc352916cd837999bf40d3f910adcaba0dee/static-site/tsconfig.json#L58-L61

This should be fixed now with a7bd07cd8caf93844fc48d2e3649f3080855437f – can you try again?

jameshadfield: Yes - thanks! I was (literally) looking in the wrong place 👍 

(VS-Code's TypeScript output logs are pretty sparse, doesn't help debugging much...)

victorlin: Which output logs are you referring to? So far I'm using it to look for rule violations (i.e. "problems", red squiggles) which seems sufficient. For example, when enabling `noImplicitAny` on #874, I can see all the violations in `static-site/src/components/ListResources/Modal.tsx`.

static-site/tsconfig.json
```diff
@@ -1,27 +1,63 @@
...
+    "noEmit": true,                                      /* Do not emit compiler output files like JavaScript source code, source-maps or declarations. */
+
+    /* Interop Constraints */
+    "isolatedModules": true,                             /* Ensure that each file can be safely transpiled without relying on other imports. */
+    "allowSyntheticDefaultImports": true,                /* Allow 'import x from y' when a module doesn't have a default export. */
+    "forceConsistentCasingInFileNames": true,            /* Ensure that casing is correct in imports. */
+
+    /* Type Checking */
+    "strict": true,                                      /* Enable all strict type-checking options. */
+    "noImplicitAny": false,                              /* Allow implicit any to make incremental TypeScript adoption easier. */
```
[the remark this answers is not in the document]
victorlin: #879

static-site/tsconfig.json
jameshadfield: Oh, I mean the debug/logging output from the typescript checker VS-Code is running itself, which is how I've debugged other things not working (linters, syntax highlighting etc)

image

victorlin: Oh gotcha! It looks like that log only contains info on the TS Server process. There is a separate log that you can enable/view using the command **TypeScript: Open TS Server log**. It's much more verbose (6MB large from just a couple minutes of usage) but not sure if it would've been helpful in this case.

PR #878, opened 2024-05-24
Make showcase expandable

victorlin:
[_preview_](https://nextstrain-s-victorlin--ptoszr.herokuapp.com/pathogens)

Previously, the showcase would drop cards as the width decreases. This is presumably to prevent the showcase from taking up too much height. Instead of hiding the unused cards, allow them to be accessed by making the showcase expandable.

## Related issue(s)

- Based on #874
- Prepping for #849
- Alternative to (closes #873)
    - [Suggested](https://github.com/nextstrain/nextstrain.org/pull/873#issuecomment-2125430386) by @trvrb, [designed](https://github.com/nextstrain/nextstrain.org/pull/873#issuecomment-2125606373) by @tsibley

## Checklist

- [x] Checks pass

### DIFF
--- a/static-site/src/components/ListResources/Showcase.tsx
+++ b/static-site/src/components/ListResources/Showcase.tsx
@@ -11,6 +11,8 @@ import { Card, FilterOption, Group } from './types';
 
 const cardWidthHeight = 160; // pixels
 const expandPreviewHeight = 50 //pixels
+const transitionDuration = "0.3s"
+const transitionTimingFunction = "ease"
 
 type ShowcaseProps = {
     cards: Card[]
@@ -20,6 +22,7 @@ type ShowcaseProps = {
 export const Showcase = ({cards, setSelectedFilterOptions}: ShowcaseProps) => {
 
   const [numRows, setNumRows] = useState<number>(0);
+  const [expandedHeight, setExpandedHeight] = useState<number>(0);
   const [isExpanded, setIsExpanded] = useState<boolean>(false);
 
   const toggleExpand = () => {
@@ -32,6 +35,10 @@ export const Showcase = ({cards, setSelectedFilterOptions}: ShowcaseProps) => {
    */
   function cardsContainerRef(cardsContainer: HTMLDivElement) {
     if (!cardsContainer) return;
+
+    if(expandedHeight != cardsContainer.clientHeight) {
+      setExpandedHeight(cardsContainer.clientHeight)
+    }
 
     const cards = cardsContainer.children;
 
@@ -57,13 +64,13 @@ export const Showcase = ({cards, setSelectedFilterOptions}: ShowcaseProps) => {
       <Byline>
         Showcase resources: click to filter the resources to a pathogen
       </Byline>
-      <ShowcaseContainer $isExpanded={isExpanded} $isExpandable={isExpandable}>
+      <ShowcaseContainer className={!isExpandable ? "" : isExpanded ? "expanded" : "collapsed"} $expandedHeight={expandedHeight}>
         <CardsContainer ref={cardsContainerRef}>
           {cards.map((el) => (
             <ShowcaseTile card={el} key={el.name} setSelectedFilterOptions={setSelectedFilterOptions}/>
             ))}
         </CardsContainer>
-        {isExpandable && !isExpanded && <PreviewOverlay onClick={toggleExpand} />}
+        <PreviewOverlay onClick={toggleExpand} className={!isExpandable || isExpanded ? "hidden" : "visible"} />
       </ShowcaseContainer>
       {isExpandable && <>
         <ArrowButton onClick={toggleExpand}>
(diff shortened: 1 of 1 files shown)


static-site/src/components/ListResources/Showcase.tsx
```diff
@@ -9,26 +10,66 @@ import { LIST_ANCHOR } from "./index";
...
+      for (let i = 0; i < cards.length; i++) {
+        if ((cards[i] as HTMLElement).offsetLeft == leftBorder) {
+          newNumRows++;
+        }
+      }
+      if (numRows != newNumRows) {
+        setNumRows(newNumRows);
+      }
+    }
+  }
```
tsibley: _non-blocking_

Why do these layout calculations ourselves instead of using flexbox? I'd think it'd be simpler to implement and maintain with flexbox.

tsibley (commented):
The spacing of the last row looks too much to me, aesthetically. What about centering the tiles so the spacing between them is always the same in every row?

image

tsibley:
> This does feel like something CSS Grid could help with

I think flexbox is a more natural fit for what we want (which is not a fixed grid, unless I'm misunderstanding intent).

genehack:
> > This does feel like something CSS Grid could help with
> 
> I think flexbox is a more natural fit for what we want (which is not a fixed grid, unless I'm misunderstanding intent).

Yeah, either way I think; the big point is not re-inventing a wheel the browser should hand us. :grin:

static-site/src/components/ListResources/Showcase.tsx
victorlin: I did it this way following recommendations from online sources. Not sure what you mean by "using flexbox", but your comment realized that the calculations were unnecessary and I've simplified it in d60cf39011a03479c40fdbc442d030e22c4a4af3.

static-site/src/components/ListResources/Showcase.tsx
victorlin, line 1: Re: spacing concerns ([1](https://github.com/nextstrain/nextstrain.org/pull/878#pullrequestreview-2078138490), [2](https://github.com/nextstrain/nextstrain.org/pull/878#issuecomment-2130487036)):

> [@tsibley] The spacing of the last row looks too much to me, aesthetically. What about centering the tiles so the spacing between them is always the same in every row?
> [@genehack] right now the grid can get real funky depending on viewport sizing and zoom level

I added a simple fix 940209688387f17c5ccd4da85a1afbcd4a1b8e18, let me know what you think.

victorlin: The downsides are most noticeable at width ~670px where there is a prominent imbalance:

670px width preview

I'll explore alternatives described in the answers [here](https://stackoverflow.com/q/32802202).

jameshadfield: The spacing I always want (but CSS can't give me, as far as I know) is for `justify-content: space-between;` but if the final row is incomplete then its cards are left-aligned and positioned vertically below the preceding row.

victorlin: CSS grid seems to be better: 13e890682388084c7ee5a9fad9179a1f61adff12

670px width preview with CSS grid

tsibley: The CSS Grid solution doesn't result in centered tiles for me?

Screenshot from 2024-05-29 14-25-52

tsibley: Oh, I guess the containing grid is centered, but because it's a grid the last row will appear left-aligned when not full. (That's why I was thinking flexbox > grid here in my earlier comments, but I see from the linked examples that that's not as possible as I thought it was.)

tsibley (approved):

trvrb (approved):

static-site/src/components/ListResources/Showcase.tsx
victorlin: Centered tiles + centered container is possible with flexbox using `justify-content: center`, but I think it looks weird:

centered tiles and centered container

Left-aligned tiles + centered container looks better to me and is consistent with the layout of https://nextstrain.org/groups

genehack (approved):

static-site/src/components/ListResources/Showcase.tsx
victorlin, line 1: Re: [1-tile overflow](https://github.com/nextstrain/nextstrain.org/pull/878#issuecomment-2138309348)

> [@trvrb] The single tile overhang feels off. I'd either add a couple additional tiles or just drop one. If adding a couple, I'd add mumps and EV-D68.

Agreed. Going forwards we're more likely to add than remove, and with this being expandable now there's not much reason to prefer a shorter list. I'll make a PR to add the additional tiles.

genehack: 2¢ — I like that version with the centered terminal row better — assuming that it's properly responsive and would go to 3x3 and 5+4 at narrower and wider viewports, respectively. 

(No _strong_ feels about this over the left-flush final row.)

tsibley: I'm also +1 for the centered last row, but it's minor.

victorlin:
Whoops, I forgot this was based on #874, so these changes won't make it into `master` until that PR is merged.

static-site/src/components/ListResources/Showcase.tsx
victorlin: #883

victorlin: Continued in https://github.com/nextstrain/nextstrain.org/pull/930#discussion_r1651742019

PR #885, opened 2024-05-30
Replace Next.js Links with HTML anchor

victorlin:

Consistently use server-side navigation links across the codebase.

## Checklist

- [x] Checks pass

### DIFF
--- a/static-site/src/components/splash/index.jsx
+++ b/static-site/src/components/splash/index.jsx
@@ -1,6 +1,5 @@
 import React, { useEffect } from "react";
 import ScrollableAnchor, { configureAnchors } from '../../../vendored/react-scrollable-anchor/index';
-import Link from 'next/link'
 import Cards from "../Cards";
 import nCoVCards from "../Cards/nCoVCards";
 import coreCards from "../Cards/coreCards";
@@ -58,7 +57,7 @@ const Splash = () => {
           potential of pathogen genome data. We provide a continually-updated view of publicly
           available data alongside powerful analytic and visualization tools for use by the
           community. Our goal is to aid epidemiological understanding and improve outbreak
-          response. If you have any questions, please <Link href="/contact">contact us</Link>.
+          response. If you have any questions, please <a href="/contact">contact us</a>.
         </Styles.CenteredFocusParagraph>
       </FlexCenter>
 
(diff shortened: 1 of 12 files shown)


static-site/src/components/splash/index.jsx
```diff
@@ -28,13 +28,33 @@ const Section = ({id, title, abstract, cards, buttonText, buttonLink}) => (
...
-      <Styles.Button to={buttonLink}>
+      <Button to={buttonLink}>
         {buttonText}
-      </Styles.Button>
+      </Button>
     </div>
   </div>
 );
 
+const Button = ({to, children}) => {
```
genehack, line 38: would it be better to pull this out into a dedicated component file?

genehack (commented):
Wrapping next/link because we _might_ want to switch to plain old HTML in the future has a strong YAGNI vibe, I think.

tsibley:
What @genehack said... I think I'd prefer to either decide _now_ to use `<Link>` or `<a>` (and in general I'd prefer the latter) rather than do a refactor which avoids making that decision.

genehack:
> (and in general I'd prefer the latter) 

FWIW I lean in that direction too, but I noticed from the [next/link docs](https://nextjs.org/docs/app/api-reference/components/link) that `<Link>` does some opportunistic pre-fetching, and I wondered how removing that might affect perceived site responsiveness.

tsibley:
Yeah, it's not just pre-fetching, it's also the difference between client-side navigation and round trips to the server.

victorlin:
I don't have a strong opinion here, but is this part of the broader discussion at https://github.com/nextstrain/private/issues/88?

tsibley:
Sorta, yeah. "Do we want to have client-side navigation between pages? Or is it overkill for nextstrain.org?"

I personally see client-side nav for page-oriented sites like nextstrain.org as effectively an optimization layer, and find that it's often not worth the added complexity/quirks it brings (even when that complexity is wrapped up and tied with a bow by a framework like Next.js).

victorlin:
This should really be discussed on an issue which I should've created before PR, but I think we've enough conversation to continue here rather than starting fresh with a new issue. I also don't want to close this PR or block on https://github.com/nextstrain/private/issues/88 because there's no good reason for the current mix of client and server side navigation.

Given the votes here, I'll rework this PR to replace all client side navigation with server side navigation and drop the wrapper.

static-site/src/components/splash/index.jsx
victorlin: Dropped this change from the PR

Issue #882, opened 2024-05-29
`<Link>` components don't work well with browser history navigation


victorlin:
### Description

Clicking on a`<Link>` then using the browser to navigate to the previous page updates the URL but page contents remain unchanged.

### How to reproduce
Steps to reproduce the current behavior:
1. Go to https://next.nextstrain.org
2. Click on "Groups" or "Please see the team page" in the footer
4. Attempt to navigate back in browser history

### Possible solutions

1. Workaround: #886
2. #887

### References

Noted on [Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1717024682181459)

victorlin:
This was noted by @jameshadfield a while back [on Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1717028679922599?thread_ts=1717024682.181459&cid=C01LCTT7JNN).

I did some searching and it seems like this is a common problem with [client-side navigation](https://nextjs.org/learn-pages-router/basics/navigate-between-pages/client-side), but no clear solution.

I'm thinking to apply the workaround of replacing `<Link>` with `<a>` across `static-site/*` as a quick fix. The server-side navigation with `<a>` is a bit slower, but we don't really need the client-side navigation anywhere in particular.

ivan-aksamentov:
@victorlin Here is a little experiment:

 - open dev console 
 - open new tab, paste to URL bar https://next.nextstrain.org/ and press Enter. Do not scroll!
 - paste `window.history.state`, observe it's an object
 - scroll the page
 - paste `window.history.state` again, observe it's an empty string

I suspect it's this magic:

https://github.com/nextstrain/nextstrain.org/blob/64470d388c33a04ae8b83942bfaf828e4edfe885/static-site/vendored/react-scrollable-anchor/utils/hash.js#L14-L20

and also maybe this:

https://github.com/nextstrain/nextstrain.org/blob/64470d388c33a04ae8b83942bfaf828e4edfe885/static-site/src/components/Datasets/dataset-select.jsx#L52

When using a client-side router, any fiddling with history API will very likely break the router. I'd zap this magical solution and search for a package on NPM (if the usual anchor `id="foo"` and `<a href="#foo" />` is not enough).

(It might make sense to also review all other overly smart solutions, e.g. involving `window.`, `global.`, manual DOM manipulation, complex on-mount logic etc. Chances are most of them are well-known NPM packages or built-in into Next.js)

[...]

PR #886, opened 2024-05-30
Use server-side rendering on internal links

victorlin:

Client-side rendering is buggy at the moment, not working well with browser navigation features.

Server-side rendering is a bit slower but has no major issues.

## Related issue(s)

- Based on #885
- Workaround for #882

## Checklist

- [ ] Merge #885
- [x] Checks pass (doc building failure is unrelated)
- [x] [Bug](https://github.com/nextstrain/nextstrain.org/issues/882) is not reproducible in [preview](https://nextstrain-s-victorlin--73jmxp.herokuapp.com)


victorlin:
Closing in favor of #887 which addresses the bug directly.

Leaving #885 open to be considered as a separate improvement.

victorlin:
#885 has been reworked to effectively apply this PR's original intent.

Issue #870, opened 2024-05-20
Port new resource-listing “cards” UI for use to display datasets on /groups and /community


genehack:

victorlin:
FYI I'm planning to make some edits to [static-site/src/components/ListResources/Showcase.jsx](https://github.com/nextstrain/nextstrain.org/blob/61c0419da4dfb823ea081a65c2d4ad82b245b3cd/static-site/src/components/ListResources/Showcase.jsx) proposed in https://github.com/nextstrain/nextstrain.org/issues/849#issuecomment-2121064824.

We may run into conflicts here, but I suspect that as long as you use the `Showcase` component just like this in the other pages, it should be trivial to resolve.

https://github.com/nextstrain/nextstrain.org/blob/61c0419da4dfb823ea081a65c2d4ad82b245b3cd/static-site/src/components/ListResources/index.jsx#L64

jameshadfield:
@genehack happy to talk through any design questions you have here (as they arise). Perhaps a couple of points as to what I was thinking:
* The current `<ListResources>` component gets its data from [a hardcoded API call](https://github.com/nextstrain/nextstrain.org/blob/61c0419da4dfb823ea081a65c2d4ad82b245b3cd/static-site/src/components/ListResources/useDataFetch.js#L18). The current groups/community pages uses the `charon/getAvailable` API. My suggestion is to allow the `<ListResources>` component to use those `charon/getAvailable` APIs, either by expanding `useDataFetch` or by injecting custom data fetch functions to the component.
* Narratives aren't currently displayed in `<ListResources>`. They totally can be, but lots of UI decisions there. As an intermediate we could replace the existing narrative table with a simple HTML list, which I think would be just fine.

genehack:
Noting that Heroku deployment is currently blocked on [making `eslint` aware of TypeScript](https://github.com/nextstrain/nextstrain.org/issues/905).

jameshadfield:
> Noting that Heroku deployment is currently blocked on https://github.com/nextstrain/nextstrain.org/issues/905.

I mean, we could just fix the linting errors highlighted by the failing build step (which you can run locally as well), which'll let this PR be built as a Heroku review app and even merged. Not a great long term state, but doesn't seem like something that should block this work.

genehack:
> > Noting that Heroku deployment is currently blocked on #905.
> 
[...]

PR #890, opened 2024-06-03
Simplify hover detection

victorlin:
[_preview_](https://nextstrain-s-victorlin--kg9kva.herokuapp.com/pathogens)

- Use CSS hover selector instead of relying on a prop
- Limit the scope of ResourceLinkWrapper to just adding shift-click behavior

## Related issue(s)

N/A, thought of this while working on #874 

## Checklist

- [x] Checks pass
- [x] Hover still changes text color on quick links and individual resources
- [x] Shift+click still opens modal for quick links and individual resources

### DIFF
--- a/static-site/src/components/ListResources/IndividualResource.tsx
+++ b/static-site/src/components/ListResources/IndividualResource.tsx
@@ -184,10 +184,8 @@ export const ResourceLinkWrapper = ({children, onShiftClick}) => {
     }
   };
   return (
-    <div>
-      <div onMouseOver={() => setHovered(true)} onMouseOut={() => setHovered(false)} onClick={onClick}>
-        {React.cloneElement(children, { $hovered: hovered })}
-      </div>
+    <div onMouseOver={() => setHovered(true)} onMouseOut={() => setHovered(false)} onClick={onClick}>
+      {React.cloneElement(children, { $hovered: hovered })}
     </div>
   )  
 }
--- a/static-site/src/components/ListResources/IndividualResource.tsx
+++ b/static-site/src/components/ListResources/IndividualResource.tsx
@@ -33,25 +33,32 @@ export const getMaxResourceWidth = (displayResources: Resource[]) => {
   }, 200); // 200 (pixels) is the minimum
 }
 
-export const ResourceLink = styled.a<{$hovered?: boolean}>`
+export const ResourceLink = styled.a`
   font-size: ${resourceFontSize}px;
   font-family: monospace;
   white-space: pre; /* don't collapse back-to-back spaces */
-  color: ${(props) => props.$hovered ? LINK_HOVER_COLOR : LINK_COLOR} !important;
+  color: ${LINK_COLOR} !important;
   text-decoration: none !important;
+
+  &:hover {
+    color: ${LINK_HOVER_COLOR} !important;
+  }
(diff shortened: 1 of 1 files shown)


victorlin:
Tagged @jameshadfield as author of these lines and @genehack in case you're interested in front-end work related to #870

static-site/src/components/ListResources/IndividualResource.jsx
```diff
@@ -33,14 +33,22 @@ export const ResourceLink = styled.a`
...
+function Name({displayName, href, topOfColumn}) {
+  const [hovered, setHovered] = useState(false);
+
   return (
-    <ResourceLink href={href} target="_blank" rel="noreferrer" $hovered={$hovered}>
-      {'• '}{($hovered||topOfColumn) ? displayName.hovered : displayName.default}
+    <ResourceLink href={href} target="_blank" rel="noreferrer"
+      onMouseOver={() => setHovered(true)}
+      onMouseOut={() => setHovered(false)}>
+      {'• '}{(hovered||topOfColumn) ? displayName.hovered : displayName.default}
```
tsibley, line 51: _non-blocking_

This display logic could also be done in CSS, removing the need entirely for `hovered`, `setHovered`, and `onMouseOver` / `onMouseOut`.

jameshadfield (approved):

genehack (approved):

genehack (approved):

static-site/src/components/ListResources/IndividualResource.jsx
victorlin: Implemented in bfbf4da116563cdc83913db5ff2295dc53ef16b8.

I have a slight preference for the original React implementation since it keeps the parent+child relationship between component and content. I can keep this change though if others think it's better.

Thoughts @jameshadfield @genehack @tsibley?

static-site/src/components/ListResources/IndividualResource.tsx
```diff
@@ -43,6 +43,14 @@ export const ResourceLink = styled.a`
...
     color: ${LINK_HOVER_COLOR} !important;
   }
+
+  &:not(:hover)::before {
+    content: "${(props) => props.$content}";
+  }
+
+  &:hover::before {
+    content: "${(props) => props.$hoverContent || props.$content}";
+  }
```
tsibley, line 53: _non-blocking_

This isn't what I meant… but that's on me for not specifying it and assuming you'd understand. I meant to include the content in the HTML itself (rather than using pseudo-elements in CSS) and only show the last label/name until hovered.

Don't worry about doing anything different now. I wanted to point out that I think there's a simplification here that's not mixing runtime inspection of CSS states with React component rendering.

victorlin: Oh gotcha, thanks for clarifying. I'll just revert bfbf4da116563cdc83913db5ff2295dc53ef16b8 and merge this PR. Maybe will try out that idea later...

static-site/src/components/ListResources/IndividualResource.jsx
victorlin: Continued in https://github.com/nextstrain/nextstrain.org/pull/890#discussion_r1628576162

PR #894, opened 2024-06-05
Remove `versioned` attribute from Resource interface

victorlin:

This attribute only served as an alias to determine if the optional attributes `firstUpdated|dates|nVersions|updateCadence` are available. It's more direct (and preferred by the compiler) to check for presence of the individual attributes directly where needed.

`versioned` at the top level on ListResourcesResponsiveProps is still valuable and should be kept.

## Related issue(s)

- follow-up to https://github.com/nextstrain/nextstrain.org/pull/874#discussion_r1626764829

## Checklist

- [x] Preview: [/pathogens](https://nextstrain-s-victorlin--exbkjc.herokuapp.com/pathogens) and [/staging](https://nextstrain-s-victorlin--exbkjc.herokuapp.com/staging) look normal (apart from #896)
- [x] Checks pass

### DIFF
--- a/static-site/src/components/ListResources/IndividualResource.tsx
+++ b/static-site/src/components/ListResources/IndividualResource.tsx
@@ -135,9 +135,9 @@ export const IndividualResource = ({resource, isMobile}: IndividualResourceProps
   // don't show anything if display name is unavailable
   if (!resource.displayName) return null
 
-  // add history if mobile and resource is versioned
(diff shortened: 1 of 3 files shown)


jameshadfield (approved):
Haven't tested but looks good by inspection

victorlin:
Added testing notes in the PR checklist.

PR #897, opened 2024-06-05
Move updateCadence check

### LINKED ISSUE #896: Unversioned resource listings are truncated
### Current Behavior

When setting `versioned=false` on ListResources, the resources entries are truncated.

image

### Expected behavior

Resource entries should not be truncated.

image

### How to reproduce

1. Go to https://next.nextstrain.org/staging
2. Observe bug

### Scope

This was introduced in 54be4d60c17e62d236ad95d10aff3895ca0a4553.

This affects `/staging` but not `/pathogens`. It may also affect work on #870 that uses unversioned resources.

### Possible solution

Fix the condition here to not return `w` on lack of `updateCadence`:

https://github.com/nextstrain/nextstrain.org/blob/2e4844e6472ef4c1cf3aebf366fd6901758f64c1/static-site/src/components/ListResources/IndividualResource.tsx#L22-L34

victorlin:

Resources that lack this attribute should at least have a width set based on displayName, not the fixed default of 200.

## Related issue(s)

- Fixes #896

## Checklist

- [x] [Preview](https://nextstrain-s-victorlin--95mbmc.herokuapp.com/staging) looks good
- [x] Checks pass

### DIFF
--- a/static-site/src/components/ListResources/IndividualResource.tsx
+++ b/static-site/src/components/ListResources/IndividualResource.tsx
@@ -21,11 +21,11 @@ const iconWidth = 20; // not including text
 const gapSize = 10;
 export const getMaxResourceWidth = (displayResources: Resource[]) => {
   return displayResources.reduce((w, r) => {
-    if (!r.displayName || !r.updateCadence) return w
+    if (!r.displayName) return w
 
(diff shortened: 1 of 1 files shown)


jameshadfield (approved):
👍 

image

PR #904, opened 2024-06-06
Replace Datasets DatasetSelect component on groups page with ListResources [#870]

genehack:

Replaces the current "datasets" DatasetSelect component with a ListResources component that hits the same backend `charon` API. Also refactors in `useDataFetch` to change the hard-coded parser into a passed-in (mandatory) callback function, and updates two existing uses to the new pattern. Finally, adds in conditional display logic to the ListResources component to hide or otherwise not set UI elements that don't make sense with the `groups` dataset (e.g., last modified dates, since we don't have those). 

[Preview](https://nextstrain-s-group-reso-fafosr.herokuapp.com/groups)

## Related issue(s)

#870 

## Checklist

- [x] Checks pass

### DIFF
--- a/static-site/src/components/ListResources/IndividualResource.tsx
+++ b/static-site/src/components/ListResources/IndividualResource.tsx
@@ -86,7 +86,7 @@ export function TooltipWrapper({description, children}) {
       {children}
     </div>
   )
-} 
+}
 
 interface IconContainerProps {
   Icon: IconType
@@ -194,5 +194,5 @@ export const ResourceLinkWrapper = ({children, onShiftClick}) => {
     <div onClick={onClick}>
       {children}
     </div>
-  )  
+  )
 }
--- a/static-site/src/components/ListResources/IndividualResource.tsx
+++ b/static-site/src/components/ListResources/IndividualResource.tsx
@@ -144,7 +144,7 @@ export const IndividualResource = ({resource, isMobile}: IndividualResourceProps
 
   // add history if mobile and resource has version info
   let history: React.JSX.Element | null = null
-  if (!isMobile && resource.updateCadence && resource.nVersions) {
+  if (!isMobile && resource.updateCadence && resource.nVersions && resource.lastUpdated) {
     history = (
       <TooltipWrapper description={resource.updateCadence.description +
         `<br/>Last known update on ${resource.lastUpdated}` +
@@ -158,12 +158,14 @@ export const IndividualResource = ({resource, isMobile}: IndividualResourceProps
     )
   }
 
+  const description = resource.lastUpdated ? `Last known update on ${resource.lastUpdated}` : "";
+
   return (
     <Container ref={ref}>
 
       <FlexRow>
 
-        <TooltipWrapper description={`Last known update on ${resource.lastUpdated}`}>
+        <TooltipWrapper description={description}>
           <ResourceLinkWrapper onShiftClick={() => setModalResource(resource)}>
             <Name displayName={resource.displayName} href={resource.url} topOfColumn={topOfColumn}/>
           </ResourceLinkWrapper>
--- a/static-site/src/components/ListResources/ResourceGroup.tsx
+++ b/static-site/src/components/ListResources/ResourceGroup.tsx
@@ -49,7 +49,7 @@ const ResourceGroupHeader = ({group, isMobile, setCollapsed, collapsible, isColl
             <TooltipWrapper description={`The most recently updated datasets in this group were last updated on ${group.lastUpdated}` +
                 '<br/>(however there may have been a more recent update since we indexed the data)'}>
               <span>
-                {`Most recent snapshot: ${group.lastUpdated}`}
+                {group.lastUpdated && `Most recent snapshot: ${group.lastUpdated}`}
               </span>
             </TooltipWrapper>
           )}
--- a/static-site/src/components/ListResources/ResourceGroup.tsx
+++ b/static-site/src/components/ListResources/ResourceGroup.tsx
@@ -31,7 +31,7 @@ const ResourceGroupHeader = ({group, isMobile, setCollapsed, collapsible, isColl
       <NextstrainLogo/>
 
       <FlexColumnContainer>
-        
+
         <HeaderRow>
           {group.groupUrl ? (
             <TooltipWrapper description={`Click to load the default (and most recent) analysis for ${group.groupDisplayName || group.groupName}`}>
@@ -164,7 +164,7 @@ const ResourceGroupContainer = styled.div`
 `;
 
 const IndividualResourceContainer = styled.div<{$maxResourceWidth: number}>`
-  /* Columns are a simple CSS solution which works really well _if_ we can calculate the expected maximum 
+  /* Columns are a simple CSS solution which works really well _if_ we can calculate the expected maximum
   resource width */
   column-width: ${(props) => props.$maxResourceWidth}px;
   column-gap: 20px;
@@ -239,7 +239,7 @@ function NextstrainLogo() {
 /**
  * Adds the `displayName` property to each resource.
  * Given successive nameParts:
- *      [ seasonal-flu, h1n1pdm]   
+ *      [ seasonal-flu, h1n1pdm]
  *      [ seasonal-flu, h3n2]
  * We want to produce two names: a default name, which contains all parts,
  * and a displayName which hides the fields that match the preceding name.
@@ -255,7 +255,7 @@ function _setDisplayName(resources: Resource[]) {
       name = r.nameParts.join(sep);
     } else {
       let matchIdx = r.nameParts.map((word, j) => word === resources[i-1]?.nameParts[j]).findIndex((v) => !v);
-      if (matchIdx===-1) { // -1 means every word is in the preceding name, but we should display the last word anyway 
+      if (matchIdx===-1) { // -1 means every word is in the preceding name, but we should display the last word anyway
         matchIdx = r.nameParts.length-2;
       }
       name = r.nameParts.map((word, j) => j < matchIdx ? ' '.repeat(word.length) : word).join(sep);
--- a/static-site/src/components/ListResources/index.tsx
+++ b/static-site/src/components/ListResources/index.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useRef, useEffect, useCallback, createContext, useContext } from 'react';
+import React, { FormEvent, useState, useRef, useEffect, useCallback, createContext, useContext } from 'react';
 import styled from 'styled-components';
 import Select, { MultiValue } from 'react-select';
 import ScrollableAnchor, { goToAnchor } from '../../../vendored/react-scrollable-anchor/index';
@@ -170,8 +170,8 @@ export default ListResourcesResponsive
 
 
 function SortOptions({sortMethod, changeSortMethod}) {
-  function onChangeValue(event) {
-    changeSortMethod(event.target.value);
+  function onChangeValue(event:FormEvent<HTMLInputElement>): void {
+    changeSortMethod(event.currentTarget.value);
   }
   return (
     <SortContainer>
--- a/static-site/src/components/ListResources/index.tsx
+++ b/static-site/src/components/ListResources/index.tsx
@@ -76,17 +76,25 @@ function ListResources({
   return (
     <ListResourcesContainer>
 
-      <Byline>
-        Showcase resources: click to filter the resources to a pathogen
-      </Byline>
+      { showcaseCards.length > 0 && (
+        <>
+        <Byline>
+          Showcase resources: click to filter the resources to a pathogen
+        </Byline>
 
-      <SetSelectedFilterOptions.Provider value={setSelectedFilterOptions}>
-        <Showcase cards={showcaseCards} CardComponent={FilterShowcaseTile} />
-      </SetSelectedFilterOptions.Provider>
+        <SetSelectedFilterOptions.Provider value={setSelectedFilterOptions}>
+          <Showcase cards={showcaseCards} CardComponent={FilterShowcaseTile} />
+        </SetSelectedFilterOptions.Provider>
+        </>
+      )}
 
       <Filter options={availableFilterOptions} selectedFilterOptions={selectedFilterOptions} setSelectedFilterOptions={setSelectedFilterOptions}/>
 
-      <SortOptions sortMethod={sortMethod} changeSortMethod={changeSortMethod}/>
+      { groups?.[0]?.lastUpdated && (
+        <SortOptions sortMethod={sortMethod} changeSortMethod={changeSortMethod}/>
+      ) || (
+        <HugeSpacer/>
+      )}
 
       <SetModalResourceContext.Provider value={setModalResource}>
         <ScrollableAnchor id={LIST_ANCHOR}>
--- a/static-site/src/components/ListResources/index.tsx
+++ b/static-site/src/components/ListResources/index.tsx
@@ -12,7 +12,8 @@ import { ErrorContainer } from "../../pages/404";
 import { TooltipWrapper } from "./IndividualResource";
 import {ResourceModal, SetModalResourceContext} from "./Modal";
 import { CardImgWrapper, CardInner, CardOuter, CardTitle, Showcase } from "../Showcase";
-import { FilterCard, FilterOption, Group, GroupDisplayNames, QuickLink, Resource } from './types';
+import { FilterCard, FilterOption, Group, GroupDisplayNames, QuickLink, Resource, ResourceListingInfo } from './types';
+import { HugeSpacer } from "../../layouts/generalComponents";
 
 interface ListResourcesProps extends ListResourcesResponsiveProps {
   elWidth: number
@@ -23,23 +24,29 @@ const LIST_ANCHOR = "list";
 const SetSelectedFilterOptions = createContext<React.Dispatch<React.SetStateAction<readonly FilterOption[]>> | null>(null);
 
 /**
- * A React component to fetch data and display the available resources,
- * including past versions ("snapshots").
+ * A React component that uses a callback to fetch data about
+ * available resources, and then display them, including past versions
+ * ("snapshots"), if those exist.
  *
  * Note that currently this only uses 'dataset' resources. In the future this
  * will be expanded. Similarly, we define versioned: boolean here in the UI whereas
  * this may be better expressed as a property of the API response.
  */
 function ListResources({
-  sourceId,
   versioned=true,
   elWidth,
   quickLinks,
   defaultGroupLinks=false,
   groupDisplayNames,
   showcase,
+  resourceListingCallback: resourceListingCallback,
 }: ListResourcesProps) {
-  const {groups, dataFetchError} = useDataFetch(sourceId, versioned, defaultGroupLinks, groupDisplayNames);
+  const {groups, dataFetchError} = useDataFetch(
+    versioned,
+    defaultGroupLinks,
+    groupDisplayNames,
+    resourceListingCallback,
+  );
   const showcaseCards = useShowcaseCards(showcase, groups);
   const [selectedFilterOptions, setSelectedFilterOptions] = useState<readonly FilterOption[]>([]);
   const [sortMethod, changeSortMethod] = useState("alphabetical");
@@ -50,7 +57,7 @@ function ListResources({
 
   if (dataFetchError) {
     return (
-      <ErrorContainer>  
+      <ErrorContainer>
         {"Whoops - listing resources isn't working!"}
         <br/>
         {'Please '}<a href="/contact" style={{fontWeight: 300}}>get in touch</a>{" if this keeps happening"}
@@ -108,7 +115,6 @@ function ListResources({
 
 
 interface ListResourcesResponsiveProps {
-  sourceId: string
   versioned: boolean
   quickLinks: QuickLink[]
 
@@ -116,6 +122,7 @@ interface ListResourcesResponsiveProps {
   defaultGroupLinks: boolean
   groupDisplayNames: GroupDisplayNames
   showcase: FilterCard[]
+  resourceListingCallback: () => Promise<ResourceListingInfo>;
 }
 
 /**
@@ -160,7 +167,7 @@ function SortOptions({sortMethod, changeSortMethod}) {
   }
   return (
     <SortContainer>
-      Sort pathogens by: 
+      Sort pathogens by:
       <input id="alphabetical" type="radio" onChange={onChangeValue} value="alphabetical"
         checked={"alphabetical"===sortMethod} style={{cursor: "alphabetical"===sortMethod ? 'default' : 'pointer'}}/>
       <TooltipWrapper description={'Pathogen groups ordered alphabetically. ' +
--- a/static-site/src/components/ListResources/useDataFetch.ts
+++ b/static-site/src/components/ListResources/useDataFetch.ts
@@ -59,7 +59,7 @@ interface Partitions {
 /**
  * Groups the provided array of pathVersions into an object with keys
  * representing group names (pathogen names) and values which are arrays of
- * resource objects. 
+ * resource objects.
  */
 function partitionByPathogen(pathVersions: PathVersions, pathPrefix: string, versioned: boolean) {
   return Object.entries(pathVersions).reduce((store: Partitions, [name, dates]) => {
--- a/static-site/src/components/ListResources/useDataFetch.ts
+++ b/static-site/src/components/ListResources/useDataFetch.ts
@@ -1,53 +1,52 @@
 import { useState, useEffect } from 'react';
-import { Group, GroupDisplayNames, PathVersions, Resource } from './types';
+import { Group, GroupDisplayNames, PathVersions, Resource, ResourceListingInfo } from './types';
 
 
 /**
- * Fetches the datasets for the provided `sourceId` and parses the data into an
- * array of groups, each representing a "pathogen" and detailing the available
- * resources for each, and the available versions (snapshots) for each of those
- * resources.
+ * Uses the provided callback to fetch resources and parse those
+ * resources into the `pathVersions` and `pathPrefix` values. The
+ * callback is expected (and encouraged!) to throw() on any errors.
  *
- * The current implementation defines `versioned: boolean` for the entire API
- * response, however in the future we may shift this to the API response and it
- * may vary across the resources returned.
+ * Continues on to parse the `pathVersions`/`pathPrefix` data
+ * structures into an array of groups, each representing a "pathogen"
+ * and detailing the available resources for each, and the available
+ * versions (snapshots) for each of those resources. In the case of
+ * un-versioned resources, versions will be a zero-length array (i.e.,
+ * `[]`)
+ *
+ * The current implementation defines `versioned: boolean` for the
+ * entire API response, however in the future we may shift this to the
+ * API response and it may vary across the resources returned.
  */
-export function useDataFetch(sourceId: string, versioned: boolean, defaultGroupLinks: boolean, groupDisplayNames: GroupDisplayNames) {
+export function useDataFetch(
+  versioned: boolean,
+  defaultGroupLinks: boolean,
+  groupDisplayNames: GroupDisplayNames,
+  resourceListingCallback: () => Promise<ResourceListingInfo>,
+) : {groups: Group[] | undefined, dataFetchError: boolean} {
   const [groups, setGroups] = useState<Group[]>();
-  const [dataFetchError, setDataFetchError] = useState<boolean>();
-  useEffect(() => {
-    const url = `/list-resources/${sourceId}`;
+  const [dataFetchError, setDataFetchError] = useState<boolean>(false);
 
-    async function fetchAndParse() {
-      let pathVersions: PathVersions, pathPrefix: string;
+  useEffect((): void => {
+    async function fetchAndParse(): Promise<void> {
       try {
-        const response = await fetch(url, {headers: {accept: "application/json"}});
-        if (response.status !== 200) {
-          console.error(`ERROR: fetching data from "${url}" returned status code ${response.status}`);
-          return setDataFetchError(true);
-        }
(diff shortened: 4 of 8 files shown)


static-site/src/components/ListResources/index.tsx
```diff
@@ -109,6 +123,7 @@ interface ListResourcesResponsiveProps {
   defaultGroupLinks: boolean
   groupDisplayNames: GroupDisplayNames
   showcase: Card[]
+  parserCallBackFxn: (response: Response) => Promise<{ pathPrefix: string; pathVersions: PathVersions; }>;
```
victorlin, line 126: The object type `{pathPrefix: string, pathVersions: PathVersions}` is used in many places, I think enough to warrant some standardization.

Suggestion: Add an interface in `types.ts` and use this in place of the object type.

```ts
interface ResourceListingInfo {
  pathPrefix: string
  pathVersions: PathVersions
}
```

static-site/src/components/ListResources/index.tsx
```diff
@@ -37,8 +38,15 @@ function ListResources({
...
   showcase,
+  parserCallBackFxn,
 }: ListResourcesProps) {
-  const {groups, dataFetchError} = useDataFetch(sourceId, versioned, defaultGroupLinks, groupDisplayNames);
+  const {groups, dataFetchError} = useDataFetch(
+    sourceId,
+    versioned,
+    defaultGroupLinks,
+    groupDisplayNames,
+    parserCallBackFxn,
```
victorlin, line 48: ### Nitpick: naming the callback function

Currently the caller names it `parserCallBackFxn` while the function signature uses `parserCallBack`.

Suggestion 1: use the same name. Between the two, I'd pick `parserCallBack` – the `Fxn` abbreviation is new to this codebase and I think redundant with the type info.

Suggestion 2: Based on interface name suggested in based on https://github.com/nextstrain/nextstrain.org/pull/904#discussion_r1631592549, use the name `getResourceListingInfo`. Usage would look like this, which to me reads better than the current names.

```js
// static-site/src/sections/pathogens.jsx
const getResourceListingInfo = async (response) => (await response.json()).dataset[sourceId];

// static-site/src/components/ListResources/useDataFetch.ts
({ pathVersions, pathPrefix } = await getResourceListingInfo(response));

// static-site/src/pages/groups.jsx
async function getResourceListingInfo(response) {
    const datasets = (await response.json()).datasets;
    // ...
    return { pathPrefix: "groups/", pathVersions };
}
```

static-site/src/pages/groups.jsx
```diff
@@ -4,23 +4,39 @@ import { SmallSpacer, HugeSpacer, FlexCenter } from "../layouts/generalComponent
...
-    url: (d) => d.url
-  },
-  {
-    name: "Group Name",
-    value: (d) => d.request.split("/")[2],
-    url: (d) => `/groups/${d.request.split("/")[2]}`
-  }
+const sourceId = "charon/groups";
+async function parserCallBackFxn(response) {
+    const datasets = (await response.json()).datasets;
```
victorlin, line 16: This uses `charon/getAvailable` but it wasn't super obvious to me - there's a big distance between this line and [where the use of charon is defined](https://github.com/nextstrain/nextstrain.org/blob/bc73c3b9336fb48602d918ef41452b69611abd55/static-site/src/components/ListResources/useDataFetch.ts#L35).

Suggestion: define the URL and make the fetch request directly in the definition of each `parserCallBackFxn`, rather than shared across all invocations of `useDataFetch` and conditions written inside of that. This would be a larger refactor, but I think could be in scope of this PR if the goal is to make ListResources re-usable outside of `/pathogens` and `/staging` pages.

static-site/src/pages/groups.jsx
```diff
@@ -4,23 +4,39 @@ import { SmallSpacer, HugeSpacer, FlexCenter } from "../layouts/generalComponent
...
+
+    // Use "" instead of fake date because downstream components can
+    // conditionalize display choices based on `== ""` or `!== ""`
+    const pathVersions = Object.assign(
+        ...datasets.map((ds) => ({
+            [ds.request.replace(/^\/groups\//, "")]: [""]
+        })),
+    );
```
victorlin, line 24: This seems like a hack around fragile logic under `versioned=false` that requires `PathVersions` values to be `string[]` which doesn't mesh well with `/groups`. Ideally it can be `undefined`:

```ts
export interface PathVersions {
  [name: string]: string[] | undefined
}
```

This would be a larger refactor which also necessitates attributes like `Resource.lastUpdated` to be optional, but maybe worth doing now.

static-site/src/components/ListResources/IndividualResource.tsx
```diff
@@ -158,12 +158,14 @@ export const IndividualResource = ({resource, isMobile}: IndividualResourceProps
     )
   }
 
+  const description = resource.nVersions ? `Last known update on ${resource.lastUpdated}` : "";
```
victorlin, line 161: Suggestion: condition on `resource.lastUpdated` directly

```suggestion
  const description = resource.lastUpdated ? `Last known update on ${resource.lastUpdated}` : "";
```

static-site/src/components/ListResources/useDataFetch.ts
```diff
@@ -8,25 +8,44 @@ import { Group, GroupDisplayNames, PathVersions, Resource } from './types';
...
  * may vary across the resources returned.
  */
-export function useDataFetch(sourceId: string, versioned: boolean, defaultGroupLinks: boolean, groupDisplayNames: GroupDisplayNames) {
+export function useDataFetch(
+  sourceId: string,
+  versioned: boolean,
+  defaultGroupLinks: boolean,
+  groupDisplayNames: GroupDisplayNames,
+  parserCallBack: (response: Response) => Promise<{ pathPrefix:string, pathVersions:PathVersions }>
+) : {groups: Group[] | undefined, dataFetchError: boolean | undefined} {
```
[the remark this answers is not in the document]
genehack: yeah, i've frequently found problems by asserting explicit return types before; tiny bit of hassle, but worth it when it find something funky!

static-site/src/components/ListResources/index.tsx
genehack: I've got a lean for callbacks to be clearly identifiable by name; additionally, this callback isn't getting, it's only parsing.

How does `parseResourceListingInfoCb` grab you?

static-site/src/pages/groups.jsx
genehack: I structured this the way I did because it felt like the smallest effective change, and I wasn't sure we really wanted to effectively outsource the production of the ResourceListingInfo datastructure to the consumer of the <ListResources> component — which is where I think this suggestion ends up once the dust settles.

Anybody else have a strong opinion here?

static-site/src/pages/groups.jsx
genehack: I didn't like the empty string approach, but using `undefined` or `nil` didn't feel great either, and changing all the other code to deal with the value-side _not_ being an array felt like a *lot* of work.

static-site/src/components/ListResources/IndividualResource.tsx
genehack: I want to say I tried this and it didn't DTRT …but I don't remember what it did. Will give it a shot.

static-site/src/components/ListResources/index.tsx
victorlin: Better – not a huge fan of abbreviations (seems to go against "clearly identifiable by name"), but dealer's choice!

victorlin: If spelling it out, I would use `Callback` instead of `CallBack` (with `React.useCallback` in mind)

static-site/src/pages/groups.jsx
victorlin: > using `undefined` or `nil` didn't feel great either

`undefined` sounds right to me... e.g. "what are the dates associated with this path?"

1. `undefined` – there are no dates!
2. there are no dates, but here's a single date with the value `""`

> changing all the other code to deal with the value-side _not_ being an array felt like a _lot_ of work.

It might be nontrivial, but I would consider it in scope of making ListResources re-usable outside of `/pathogens` and `/staging` pages (same reasoning as https://github.com/nextstrain/nextstrain.org/pull/904#discussion_r1631608768). Otherwise, ListResources will remain tailored for those two "first-class" pages, with pages such as `/groups` considered "second-class" and relying on workarounds to match assumptions made for the first-class pages.

static-site/src/components/ListResources/useDataFetch.ts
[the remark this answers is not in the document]
jameshadfield: P.S. We could use [@typescript-eslint/explicit-function-return-type](https://typescript-eslint.io/rules/explicit-function-return-type/) if we want to enforce this

static-site/src/pages/groups.jsx
jameshadfield: >  I wasn't sure we really wanted to effectively outsource the production of the ResourceListingInfo datastructure to the consumer of the component — which is where I think this suggestion ends up once the dust settles.

I'm not sure we need to (nor should) go this far. But I do think adding the `fetch` call to the callback (and therefore the API address) is a better separation of concerns.

static-site/src/pages/groups.jsx
jameshadfield: > Otherwise, ListResources will remain tailored for those two "first-class" pages, with pages such as /groups considered "second-class" and relying on workarounds to match assumptions made for the first-class pages.

This is a compelling point for me. In the short-term future we will probably have:
* 1 page with versions (/pathogens)
* 1 page with last-updated date (/staging)
* lots of pages with no dates (/groups, `/groups/<NAME>`, community splash pages)

And that's not including [a files listing page](https://github.com/nextstrain/nextstrain.org/pull/828) which would itself require larger changes. 

For what it's worth, I'd use an empty array to represent no dates.

static-site/src/components/ListResources/index.tsx
genehack: Split the difference with `parseResourceListingCallback`.

static-site/src/pages/groups.jsx
genehack: I refactored things so that the `<ListResources>` component gets passed a URL instead of a `sourceId`. I left the `fetch()` functionality where it is. With the way `useDataFetch` is currently implemented and structured, pulling the `fetch` call out into the resource listing call back would be fairly ugly, IMO.

The way things are structured now, `<ListResources>` fetches a URL and hands the response for that fetch off to the provided callback, which is responsible for producing an object with this shape: 

```json
{ 
  "pathPrefix": string,
  "pathVersions": (string|undefined)[],
}
```

I'm just not seeing where the `fetch` in the callback is an improvement over the updates I just pushed.

static-site/src/pages/groups.jsx
genehack: > For what it's worth, I'd use an empty array to represent no dates.

To be explicit: will there _ever_ be a resource that has *multiple* undated datasets? 

If the answer to that is "no", then sure, an empty array will work to indicate "no date".

If that answer to that is "...maybe", then I think an array with one `undefined` value per dataset is a more future proof approach.

static-site/src/components/ListResources/IndividualResource.tsx
genehack: Nope, this works great, changed.

genehack:
Note: `CI / build` will be broken until [TS lint changes PR](https://github.com/nextstrain/nextstrain.org/pull/906) lands.

victorlin:
> Note: `CI / build` will be broken until [TS lint changes PR](https://github.com/nextstrain/nextstrain.org/pull/906) lands.

This blockage doesn't seem necessary/productive. Attempting removal with #909

genehack:
> > Note: `CI / build` will be broken until [TS lint changes PR](https://github.com/nextstrain/nextstrain.org/pull/906) lands.
> 
> This blockage doesn't seem necessary/productive. Attempting removal with #909

it's not _necessary_, in that i could add lines to disable the couple lint failures -- but then those would need to be removed once #906 lands, and that feels like unproductive churn to me. 

(i don't have a strong opinion about the direction #909 moves things in; i think linting being mildly annoying when broken is maybe a good thing…)

victorlin:
@genehack continuing the tangential discussion in https://github.com/nextstrain/nextstrain.org/pull/909#issuecomment-2159141269

genehack:
> @genehack continuing the tangential discussion in [#909 (comment)](https://github.com/nextstrain/nextstrain.org/pull/909#issuecomment-2159141269)

👍 

I merged #906 so that should get this unblocked soon (he says, as he discovers _another_ linting problem...) — I will link a preview link once it builds.

genehack:
[Preview](https://nextstrain-s-group-reso-fafosr.herokuapp.com/groups) — note that only `blab` group data is being returned by the `charon/getAvailable` call in the staging/preview environment, and I'm not sure why that isn't reflected in the group tiles in the preview…

static-site/src/pages/groups.jsx
jameshadfield: If we had such data then the internal structure would have to change anyway because the UI would need a way to distinguish between them, ultimately by surfacing different URLs for each.

The difference between `[]` and `[undefined]` in JS is subtle and I don't know if I'd write code that relied on that subtlety.

genehack: > If we had such data then the internal structure would have to change anyway because the UI would need a way to distinguish between them, ultimately by surfacing different URLs for each.

okay, good to know. 

> The difference between `[]` and `[undefined]` in JS is subtle and I don't know if I'd write code that relied on that subtlety.

the main reason i would reach for `[undefined]` instead of `[]` is because all the existing code expects to iterate over a list, and `[undefined]` _can_ be iterated over. 

i will try to see how ugly an empty list will make this; i expect a couple parts will get more funky than they currently are — we'll see.

jameshadfield: `[undefined]` is fine by me

genehack: > i will try to see how ugly an empty list will make this; i expect a couple parts will get more funky than they currently are — we'll see.

was not horrible; pushed updates.

static-site/src/pages/groups.jsx
victorlin: > I'm just not seeing where the `fetch` in the callback is an improvement over the updates I just pushed.

The new updates are better, but I still see some benefits to having `fetch` in callback. Something like this, feel free to take or leave: c842d4d84c96af8eba3cb4c109f36e40c99ccca0

static-site/src/pages/groups.jsx
```diff
@@ -92,31 +108,22 @@ class GroupsPage extends React.Component {
...
-          </splashStyles.CenteredFocusParagraph>
-        </FlexCenter>
-        <HugeSpacer />
-        {this.state.dataLoaded && (
-          <DatasetSelect
-            datasets={this.state.datasets}
-            columns={datasetColumns({isNarrative: false})}
-          />
-        )}
+        <ListResources
```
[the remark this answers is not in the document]
genehack: yep, good catch; should be fixed.

static-site/src/pages/groups.jsx
```diff
@@ -4,23 +4,39 @@ import { SmallSpacer, HugeSpacer, FlexCenter } from "../layouts/generalComponent
...
-    value: (d) => d.request.split("/")[2],
-    url: (d) => `/groups/${d.request.split("/")[2]}`
-  }
+const sourceUrl = "/charon/getAvailable?prefix=/groups/";
+async function parseResourceListingCallback(response) {
+    const datasets = (await response.json()).datasets;
+
+    // Use "" instead of fake date because downstream components can
+    // conditionalize display choices based on `== ""` or `!== ""`
+    const pathVersions = Object.assign(
```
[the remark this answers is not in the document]
genehack: fixed.

static-site/src/components/ListResources/ResourceGroup.tsx
```diff
@@ -49,7 +49,7 @@ const ResourceGroupHeader = ({group, isMobile, setCollapsed, collapsible, isColl
             <TooltipWrapper description={`The most recently updated datasets in this group were last updated on ${group.lastUpdated}` +
                 '<br/>(however there may have been a more recent update since we indexed the data)'}>
               <span>
-                {`Most recent snapshot: ${group.lastUpdated}`}
+                {group.lastUpdated !== undefined && `Most recent snapshot: ${group.lastUpdated}`}
               </span>
```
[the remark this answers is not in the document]
genehack: 👍

static-site/src/components/ListResources/index.tsx
```diff
@@ -67,11 +75,17 @@ function ListResources({
...
-      <Showcase cards={showcaseCards} setSelectedFilterOptions={setSelectedFilterOptions}/>
+      { showcaseCards.length > 0 && (
+        <Showcase cards={showcaseCards} setSelectedFilterOptions={setSelectedFilterOptions}/>
+      )}
 
       <Filter options={availableFilterOptions} selectedFilterOptions={selectedFilterOptions} setSelectedFilterOptions={setSelectedFilterOptions}/>
 
-      <SortOptions sortMethod={sortMethod} changeSortMethod={changeSortMethod}/>
+      { typeof groups !== "undefined" && groups[0]?.lastUpdated !== "" && (
+        <SortOptions sortMethod={sortMethod} changeSortMethod={changeSortMethod}/>
```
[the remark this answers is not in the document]
genehack: 👍

static-site/src/components/ListResources/index.tsx
```diff
@@ -152,7 +167,7 @@ function SortOptions({sortMethod, changeSortMethod}) {
   }
   return (
     <SortContainer>
-      Sort pathogens by: 
+      Sort pathogens by:
```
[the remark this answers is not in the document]
genehack: fair! will breakout.

genehack:
Replicating a f2f chat I had (separately) with both Victor and James:

> The new updates are better, but I still see some benefits to having fetch in callback. 

Originally I was thinking all the callbacks would be using `fetch()`, not understanding some would be static data structures. Knowing this, I agree there's some benefit to internalizing more responsibility into the callback. 

Reviewing Victor's draft version of this, the callbacks relying on importing a fetch helper/wrapper out of useDataFetch seemed a bit too pretzel-y to me. I'm going to try a version where the pathogens and staging pages duplicate the ~4-6 lines of fetch() logic, and where the general signature of the callback is it gets passed the `setDataError` state callback. I think that will ultimately loosen the coupling between the caller and the callback in a way that allows sufficient flexibility to cover all current, and many future, use cases.

static-site/src/pages/groups.jsx
genehack: folded those changes into the "replace datasets..." commit, thanks.

genehack:
Okay, updated: callbacks now handle the fetch(); general other cleanups and fixups.

static-site/src/pages/groups.jsx
```diff
@@ -92,31 +119,21 @@ class GroupsPage extends React.Component {
...
-          </splashStyles.CenteredFocusParagraph>
-        </FlexCenter>
-        <HugeSpacer />
-        {this.state.dataLoaded && (
-          <DatasetSelect
-            datasets={this.state.datasets}
-            columns={datasetColumns({isNarrative: false})}
-          />
-        )}
+        <ListResources
```
jameshadfield, line 122: [haven't reviewed, just noting something while playing with the review app]

Looks like the resource collapsing functionality isn't working (as in, the carot rotates and text changes, but the number of resources shown is unchanged). Or is this some conditional I added if number of cards is 1 or similar?

victorlin: It works for me (Vivaldi using Chrome/124.0.0.0):

https://github.com/nextstrain/nextstrain.org/assets/13424970/832c8db2-5eda-478a-994f-25f130e32410

jameshadfield: Oh. You've got a bigger screen than me ;) Maybe I set too big of a "show this many when collapsed" count - I remember I made it a function of the number of cards in play. Not a bug in this PR then.

victorlin: > You've got a bigger screen than me

Nah, I just zoomed out a bunch haha

static-site/src/components/ListResources/useDataFetch.ts
```diff
@@ -1,42 +1,39 @@
...
-          console.error(`ERROR: fetching data from "${url}" returned status code ${response.status}`);
-          return setDataFetchError(true);
-        }
-        ({ pathVersions, pathPrefix } = (await response.json()).dataset[sourceId]);
-      } catch (err) {
-        console.error(`Error while fetching data from "${url}"`);
-        console.error(err);
-        return setDataFetchError(true);
-      }
+        const { pathPrefix, pathVersions } = await resourceListingCallback(setDataFetchError);
```
jameshadfield, line 32: At the moment, if the callback _fails_ it's setting an error state and returning `undefined`, which is then causing a runtime error (because `const { pathPrefix, pathVersions } = undefined`) which gets caught and then... sets  the error state again¹ to `true`. 

Would it be simpler to just allow the callback to throw, and catch that here (which we are doing anyway)? If we want to set custom error messages then we need more sweeping changes which seem unnecessary at this stage.

¹ Conceptually speaking. In reality, I think the two `setDataFetchError` calls will be batched so you only see the last one. Maybe.

static-site/src/pages/groups.jsx
```diff
@@ -4,23 +4,50 @@ import { SmallSpacer, HugeSpacer, FlexCenter } from "../layouts/generalComponent
...
 import { UserContext } from "../layouts/userDataWrapper";
 import { DataFetchErrorParagraph } from "../components/splash/errorMessages";
 import { groupsTitle, GroupsAbstract } from "../../data/SiteConfig";
 
-const datasetColumns = ({isNarrative}) => [
+const resourceListingCallback = async (setDataFetchError) => {
+  const sourceUrl = "/charon/getAvailable?prefix=/groups/";
+
+  try {
+    const response = await fetch(sourceUrl, {headers: {accept: "application/json"}});
```
jameshadfield, line 18: FYI - the charon API doesn't use content negotiation. No need to change, but just so you know this header isn't doing anything ;)

static-site/src/components/ListResources/index.tsx
```diff
@@ -67,11 +74,17 @@ function ListResources({
...
 
-      <Showcase cards={showcaseCards} setSelectedFilterOptions={setSelectedFilterOptions}/>
+      { showcaseCards.length > 0 && (
+        <Showcase cards={showcaseCards} setSelectedFilterOptions={setSelectedFilterOptions}/>
+      )}
 
       <Filter options={availableFilterOptions} selectedFilterOptions={selectedFilterOptions} setSelectedFilterOptions={setSelectedFilterOptions}/>
 
-      <SortOptions sortMethod={sortMethod} changeSortMethod={changeSortMethod}/>
+      { groups && groups[0]?.lastUpdated && (
```
jameshadfield, line 83: [not a request for changes - I just wanted to see if this was possible and am noting it here]

So you can use the elvis operator for array indexing too! `groups?.[0]?.lastUpdated && ...`

jameshadfield (commented):
The overloading of the word "group" must have made this a confusing first experience!

The error handling seems a little convoluted to me and it seems like it could be simplified in this PR, but that's the only thing that stands out to me. Functionally it's working really well.

static-site/src/components/ListResources/useDataFetch.ts
genehack: yeah, that's fair; i've ripped out the try/catch bits in the callbacks and it's clearer. will push update soon.

static-site/src/pages/groups.jsx
genehack: fixed since i was in there anyway.

static-site/src/components/ListResources/index.tsx
```diff
@@ -147,12 +160,12 @@ export default ListResourcesResponsive
 
 
 function SortOptions({sortMethod, changeSortMethod}) {
-  function onChangeValue(event) {
-    changeSortMethod(event.target.value);
+  function onChangeValue(event:FormEvent<HTMLInputElement>): void {
+    changeSortMethod(event.currentTarget.value);
```
victorlin, line 164: The change from `target` to `currentTarget` doesn't have any functional impact?

static-site/src/components/ListResources/useDataFetch.ts
```diff
@@ -1,53 +1,52 @@
...
+ * The current implementation defines `versioned: boolean` for the
+ * entire API response, however in the future we may shift this to the
+ * API response and it may vary across the resources returned.
  */
-export function useDataFetch(sourceId: string, versioned: boolean, defaultGroupLinks: boolean, groupDisplayNames: GroupDisplayNames) {
+export function useDataFetch(
+  versioned: boolean,
+  defaultGroupLinks: boolean,
+  groupDisplayNames: GroupDisplayNames,
+  resourceListingCallback: () => Promise<ResourceListingInfo>,
```
victorlin, line 25: > I'm going to try a version … where the general signature of the callback is it gets passed the `setDataError` state callback

I noticed you kept `setDataFetchError` in `useDataFetch` instead of the propsed implementation – not sure if you're planning to do that still but I think it looks good as-is!

victorlin (approved):

static-site/src/components/ListResources/useDataFetch.ts
victorlin: Oh I didn't realize this was already discussed in https://github.com/nextstrain/nextstrain.org/pull/904#discussion_r1637489359, sorry

static-site/src/components/ListResources/index.tsx
genehack: Not that I've noticed. The Typescript compiler tells me `event.target` is not defined, too.

tsibley:
> note that only `blab` group data is being returned by the `charon/getAvailable` call in the staging/preview environment, and I'm not sure why that isn't reflected in the group tiles in the preview…

Maybe this was addressed in a meeting off-thread you had, but since I didn't see mention of it I wanted to clarify this for you.

Heroku review apps run in production mode with production config, _but_ we've intentionally configured them (on the Heroku side) to use the `nextstrain.org-testing` AWS IAM user, which has [a much more restrictive policy](https://github.com/nextstrain/nextstrain.org/blob/e70f5304ce0969d9e0214a043d9c3438df91ee2c/aws/iam/policy/NextstrainDotOrgServerInstance-testing.tftpl.json#L1) attached to it than the actual production IAM user (`nextstrain.org`)'s [attached policy](https://github.com/nextstrain/nextstrain.org/blob/e70f5304ce0969d9e0214a043d9c3438df91ee2c/aws/iam/policy/NextstrainDotOrgServerInstance.tftpl.json#L1).

genehack:
> > note that only `blab` group data is being returned by the `charon/getAvailable` call in the staging/preview environment, and I'm not sure why that isn't reflected in the group tiles in the preview…
> 
> Maybe this was addressed in a meeting off-thread you had, but since I didn't see mention of it I wanted to clarify this for you.
> 
> Heroku review apps run in production mode with production config, _but_ we've intentionally configured them (on the Heroku side) to use the `nextstrain.org-testing` AWS IAM user, which has [a much more restrictive policy](https://github.com/nextstrain/nextstrain.org/blob/e70f5304ce0969d9e0214a043d9c3438df91ee2c/aws/iam/policy/NextstrainDotOrgServerInstance-testing.tftpl.json#L1) attached to it than the actual production IAM user (`nextstrain.org`)'s [attached policy](https://github.com/nextstrain/nextstrain.org/blob/e70f5304ce0969d9e0214a043d9c3438df91ee2c/aws/iam/policy/NextstrainDotOrgServerInstance.tftpl.json#L1).

nope, nobody else has shared this; thanks, this makes sense.

trvrb:
Makes sense that this is just showing blab on the preview app, but the wider blab dataset list is a bit funny in terms of collapsing. The amount of content shown should be relative to the how many lines there are. Currently the blab dataset list is definitely too long even in collapsed view.

genehack:
> Makes sense that this is just showing blab on the preview app, but the wider blab dataset list is a bit funny in terms of collapsing. The amount of content shown should be relative to the how many lines there are. Currently the blab dataset list is definitely too long even in collapsed view.

I will look at what's going on here. Interestingly, if I point my local dev instance at the production data endpoint (i.e., if I fetch the data fram `https://nextstrain.org/charon/getAvailable?prefix=/groups/`), this does not happen: 

Screenshot 2024-06-14 at 10 10 06

I wonder if the oddness in the display is because of the smaller data set size available in the preview environment?

genehack:
> I wonder if the oddness in the display is because of the smaller data set size available in the preview environment?

Experimenting further, if I take the production data but filter it down to just 2 cards worth of data, the cards I get are large. Expanding that to 3 cards, the layout is as expected. 

Two groups:

Screenshot 2024-06-14 at 10 23 55

Three groups: 

Screenshot 2024-06-14 at 10 24 09

genehack:
> > I wonder if the oddness in the display is because of the smaller data set size available in the preview environment?
> 
> Experimenting further, if I take the production data but filter it down to just 2 cards worth of data, the cards I get are large. Expanding that to 3 cards, the layout is as expected.

Digging ever further, [this seems to be intentional behavior](https://github.com/nextstrain/nextstrain.org/blob/master/static-site/src/components/ListResources/ResourceGroup.tsx#L267-L279). @jameshadfield could you speak to this? 

I'm going to say I don't think this should be something that blocks this PR; if we do want to change this behavior, it can be a distinct thing.

jameshadfield:
>Digging ever further, [this seems to be intentional behavior](https://github.com/nextstrain/nextstrain.org/blob/master/static-site/src/components/ListResources/ResourceGroup.tsx#L267-L279). @jameshadfield could you speak to this?

The intention is that if few cards are in view (commonly achieved via filtering) then we should show more entries per card. The exact thresholds are (as always) open to change. We could (probably should) also remove the expand/contract toggle if there is only 1 card.

Issue #905, opened 2024-06-06
[static-site] we're not linting TS files


jameshadfield:
Our static-site linting rule needs to consider `ts` and `tsx` files:

https://github.com/nextstrain/nextstrain.org/blob/5f758195b16654e48e4d3b8057c407af429915ec/package.json#L18

Observed because the nextjs build failed (due to linting errors) in https://github.com/nextstrain/nextstrain.org/pull/904

genehack:
Note: this is also going to require extending/upgrading the ESLint ecosystem to get it to be able to parse Typescript ...

jameshadfield:
[Here's how auspice does it](https://github.com/nextstrain/auspice/blob/master/.eslintrc.yaml) - not saying nextstrain.org has to do the same, but if there's no objectively better way then it'd be nice to have consistency.

genehack:
Okay, there's a bit of yak shaving potential here...

Installing `@typescript-eslint/eslint-plugin` fails due to ...some sort of tangled conflict between peer deps and `eslint-config-next` (which is a regular dependency, not a devDep, for some reason. We're also installing `eslint-plugin-jest` but don't seem to be loading/using that anywhere. Finally, we've got a lot of things that are fairly out of date and probably contributing to the kinda high number of audit warnings (currently: `44 vulnerabilities (1 low, 20 moderate, 18 high, 5 critical)`):

| Package                | Installed | Current |
|------------------------|-----------|---------|
| `eslint`               | 8.23.1    | 9.4.0   |
| `@babel/eslint-parser` | 7.19.1    | 7.24.7  |
| `eslint-config-next`   | 14.2.3    | 14.2.3  | 
| `eslint-plugin-jest`   | 27.0.4    | 28.6.0  |

Due to [`eslint-config-next` lacking ESLint v9 support](https://github.com/vercel/next.js/issues/64409), we need to stay with ESLint v8, but we could bump to v8.57.0. 

I propose to do the following: 

1. remove `eslint-plugin-jest`
2. Update `eslint` to v8.57.0
3. install `@typescript-eslint/eslint-plugin` and `@typescript-eslint/parser` (I _think_ this will work once the jest plugin is removed, I think it is actually the source of the failing peer dep, because it wants an older version of the typescript parser than `eslint-config-next` allows for)
4. update `static-site/.eslintrc.yaml to load the TS parser, the TS plugin, and add the `@typescript-eslint/recommended` ruleset to the `extends` list
[...]

PR #906, opened 2024-06-07
Lint Typescript files too [#905]

genehack:

Extend linting of the `static-site` tree to also look at `*.ts` and `*.tsx` files. Includes updates to existing files to correct or silence errors based on established usage.

N.b., I elected to add individual `eslint-disable <rule>` comments rather than globally disabling in the config, as that felt like it would probably be less controversial during review. Happy to convert any of those changes to global rule changes instead; each type of rule is in a distinct commit to facilitate these sorts of changes. 

## Related issue(s)

#905 

## Checklist

- [ ] Checks pass


package-lock.json
tsibley, line 1: The diff of this lock file change here is a mess.

It changes from format version 2 → 3 and includes a lot of unnecessary/harmful changes unrelated to installing typescript-eslint, including removal of cross-platform deps like `@next/swc-linux-x64-gnu`, which leads to this warning [during build](https://github.com/nextstrain/nextstrain.org/actions/runs/9422275879/job/25958182405?pr=906#step:3:609):

```
          Downloading swc package @next/swc-linux-x64-gnu...
 ⚠ Found lockfile missing swc dependencies, run next locally to automatically patch
```

which is what ultimately triggered the error resolved by #907.

tsibley: It also includes diff hunks like this:

```diff
@@ -8869,7 +9723,6 @@                                                                                                                                                                          
         "url": "https://github.com/sponsors/isaacs"
       }
     },
-<<<<<<< HEAD                                                                                                                                                                                  
     "node_modules/@typescript-eslint/utils": {
       "version": "5.62.0",
       "resolved": "https://registry.npmjs.org/@typescript-eslint/utils/-/utils-5.62.0.tgz",
@@ -8993,15 +9846,6 @@                                                                                                                                                                         
       "integrity": "sha512-zuVdFrMJiuCDQUMCzQaD6KL28MjnqqN8XnAqiEq9PNm/hCPTSGfrXCOfwj1ow4LFb/tNymJPwsNbVePc1xFqrQ==",
       "dev": true
     },
-||||||| parent of 25246597 (`npm up eslint` [#905])                                                                                                                                           
-=======                                                                                                                                                                                       
-    "node_modules/@ungap/structured-clone": {                                                                                                                                                 
-      "version": "1.2.0",                                                                                                                                                                     
-      "resolved": "https://registry.npmjs.org/@ungap/structured-clone/-/structured-clone-1.2.0.tgz",                                                                                          
-      "integrity": "sha512-zuVdFrMJiuCDQUMCzQaD6KL28MjnqqN8XnAqiEq9PNm/hCPTSGfrXCOfwj1ow4LFb/tNymJPwsNbVePc1xFqrQ==",                                                                         
-      "dev": true                                                                                                                                                                             
-    },                                                                                                                                                                                        
->>>>>>> 25246597 (`npm up eslint` [#905])                                                                                                                                                     
     "node_modules/abab": {   
       "version": "2.0.6",
       "resolved": "https://registry.npmjs.org/abab/-/abab-2.0.6.tgz",
```

so some prior commit included merge conflict markers in the committed `package-lock.json`.

genehack: Sorry, that was borne out of trying to undo the `eslint-plugin-jest` removal in a way that told a good story.

tsibley: No worries. Lock file changes are tedium taken to the nth degree. It's super tempting to ignore them, but unfortunately lock files are also central and critical. A bad mix!

genehack: Aside: this is one reason I like having a `pre-commit` (or `lint-staged`, etc.) running locally: you can have a check for whether a commit includes conflict resolution marks and avoid looking like a doofus.

package.json
```diff
@@ -91,19 +90,24 @@
...
     "@babel/plugin-proposal-class-properties": "^7.18.6",
+    "@eslint/js": "^9.4.0",
     "@jest/globals": "^29.0.2",
+    "@types/eslint__js": "^8.42.3",
     "@types/react": "^18.2.73",
     "babel-plugin-lodash": "^3.2.11",
     "benchmark": "^2.1.4",
     "eslint": "^8.23.0",
-    "eslint-plugin-jest": "^27.0.1",
+    "eslint-config-next": "^14.1.4",
```
jameshadfield, line 100: What are dev-dependencies in the context of `nextstrain.org`? Building the site without these dev dependencies installed will cause next.js to download a bunch of them itself (potentially with different pins), which just seems worse, and then fail to build because of missing dependencies, which is categorically worse.

Maybe some could be classed as developer-only (e.g. `nodemon`) but anything that's necessary for CI to pass or the site to run seems like it should be a proper dependency.

jameshadfield (commented):
Just skimmed through the code but it looks ok

For what it's worth I'm not too worried about getting the perfect set of TS rules at this stage - others can amend these over time as they see fit.

package.json
genehack: The [node package.json docs](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#devdependencies) have this to say about `devDependencies`:

> If someone is planning on downloading and using your module in their program, then they probably don't want or need to download and build the external test or documentation framework that you use.

The [dependencies section](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#devdependencies) also says: 

> Please do not put test harnesses or transpilers in your dependencies object. See devDependencies, below.

Linting config feels (to me) very akin to "test harnesses or transpilers" — it's only something that's going to be needed if you're working on the code. 

As far as CI, etc — `npm ci` installs devDeps (as long as the lock file that's present was built with a normal `npm install` and not `npm install --production). Additionally the CI GitHub workflow does some extra dance steps to make sure devDeps are installed. 

I don't care about this _super_ strongly, but it doesn't make a whole lot of sense to have several linting related packages (`eslint`, `eslint-plugin-jest`, `typescript-eslint`, etc) in `devDeps` and one (`eslint-config-next`) as a regular dep, does it?

jameshadfield: If it were up to me I'd put everything in dependencies and call it a day. But I don't want us to get sidetracked on the minutiae so... let's consider this resolved.

genehack: I would probably be inclined to +1 a PR that did that :grin: …but I wanted to land the linting changes, so I went ahead and merged.

victorlin (commented):
✅

Issue #895, opened 2024-06-05
CI failures when npm installing deps from GitHub 


jameshadfield:
Recent / ongoing build failures are observed during CI for [this PR](https://github.com/nextstrain/nextstrain.org/pull/891) (full logs [failure 1](https://github.com/nextstrain/nextstrain.org/actions/runs/9377730007/job/25819748851), [failure 2](https://github.com/nextstrain/nextstrain.org/actions/runs/9391578974/job/25864032722)) when installing Auspice's dependencies:

```console
npm error code 1
npm error git dep preparation failed
```

however CI is passing for master branch. These errors are not reproducible for me when building the PRs branch in node20, but I haven't tested building them locally using the heroku buildpack container. The Heroku review app succeeds and is working as (I think) we don't yet in-house the building & deployment of these, which differs from how we deploy canary/production. I believe the salient difference (and hinted at in the error) is that we install auspice [from a GitHub SHA](https://github.com/nextstrain/nextstrain.org/pull/891/files) for these automated testing PRs. @tsibley do you know more about what's happening here and if it is a known limitation of using build-packs the way we are? Hopefully it's as simple as adding some GitHub credentials / config to the build container.

tsibley:
Hmm. Do you figure the error with Puppeteer is unrelated/a red herring?

```
npm error code 1
npm error git dep preparation failed
npm error command /build/app/.heroku/node/bin/node /build/app/.heroku/node/lib/node_modules/npm/bin/npm-cli.js install --force --cache=/tmp/npmcache.OQOXF --prefer-offline=false --prefer-online=false --offline=false --no-progress --no-save --no-audit --include=dev --include=peer --include=optional --no-package-lock-only --no-dry-run
npm error npm error code 1
npm error npm error path /tmp/npmcache.OQOXF/_cacache/tmp/git-cloneb8Kpvu/node_modules/puppeteer
npm error npm error command failed
npm error npm error command sh -c node install.js
npm error npm error ERROR: Failed to set up Chrome r114.0.5735.90! Set "PUPPETEER_SKIP_DOWNLOAD" env variable to skip download.
npm error npm error [Error: EACCES: permission denied, mkdir '/.cache'] {
npm error npm error   errno: -13,
npm error npm error   code: 'EACCES',
npm error npm error   syscall: 'mkdir',
npm error npm error   path: '/.cache'
npm error npm error }
```

tsibley:
[...]

PR #907, opened 2024-06-07
scripts/heroku-build: Set HOME=/tmp for build

tsibley:
Some things (Puppeteer during `npm` install, in some cases; `next build` when SWC deps are missing) want to store files in $HOME/.cache/ and that doesn't work due to permissions when HOME=/ (the fallback value provided by bash because it can't lookup a user for the uid we use).

I believe these caches are needed at build-time only and can be safely discarded afterwards.  If that turns out to not be the case, then we can switch to setting HOME=/build/app instead to include them in the slug.

Resolves: <https://github.com/nextstrain/nextstrain.org/issues/895>

## Checklist

- [x] Checks pass

### DIFF
--- a/scripts/heroku-build
+++ b/scripts/heroku-build
@@ -63,6 +63,7 @@ docker run \
   --volume "$(realpath build/pack)":/build/pack:ro \
   --volume "$(realpath build/cache)":/build/cache:rw \
   --volume "$(realpath build/env)":/build/env:ro \
+  --env HOME=/tmp \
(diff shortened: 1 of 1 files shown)


genehack (approved):

jameshadfield:
Following up - the [dependabot-triggered PR](https://github.com/nextstrain/nextstrain.org/pull/908) from this morning [successfully built & deployed](https://github.com/nextstrain/nextstrain.org/actions/runs/9440605160) 🎉

PR #914, opened 2024-06-12
Swap splash tiles to use viz screenshots

trvrb:
[_preview_](https://nextstrain-s-trvrb-sema-uhwlnj.herokuapp.com/)

This commit swaps the tiles used on the splash page from the previous doctored electron micrographs to be simple screenshots of the pages / pathogens that are linked to. This has two advantages:
1. It's super simple and no longer requires sourcing a copyright-free image and doctoring it in LunaPic
2. It's a more up front, more transparent window into what you get when you click on a tile

I used tree vs map depending on spatial / epi patterns. For pathogens like WNV and Ebola, the map is meaningful and interesting, while for things like H3N2, it's less relevant. I generally tried to stick close to the content that's directly linked and not update coloring, etc..., but I made a couple of exceptions for visual interest.

## Other considerations

- Replace images on `/pathogens`
- Remove unused images
- Remove credit from footer


jameshadfield:
Nice! I was a big fan of the styled cards in the early days but they didn't scale well in multiple senses of the word. This should give us a lot more scope to implement [new functionality](https://github.com/nextstrain/nextstrain.org/issues/849#issuecomment-2161839537) in future.

static-site/static/splash_images/h3n2_tree.png
jameshadfield, line 1: A lot of these images are big file sizes for what they are, and it means for the splash page we'll end up loading multiple Mb of image data. We can either convert them in this PR (`sips`, `imagemagick` etc) or add the appropriate image optimisation layers to NextJS. Or both.

jameshadfield:
Parent PR (LHS) vs this PR (RHS). Colour matching between tiles and rainbow Nextstrain title is nice.

image

static-site/static/splash_images/h3n2_tree.png
trvrb: I meant to be cutting all of these down to 250x250px (the same as the previous tiles). I think I missed a couple however.

victorlin:
Rebasing this onto latest `master`

victorlin:
I added a few more things to the PR description to consider if pursuing this direction.

victorlin:
Rebasing to resolve merge conflicts

victorlin:
Bundling these changes into #922 (aeffbe28333a75282857ded46c26a30fa6287274)

victorlin:
Actually there's enough going on in the screenshot swap that I'm going to put those changes back into this PR. This should make it easier to review each batch of related commits.

static-site/static/splash_images/h3n2_tree.png
victorlin: Downsized all screenshots. JPEG or WebP might be even better in terms of file size but I'll leave it as-is for now.

static-site/README.md
```diff
@@ -73,16 +73,11 @@ We may wish to rename this directory to avoid any doubt.
...
-    * "Edit" -> "Scale Image" to 250 x 250px
-    * "File" -> "Save Image" & use PNG
-3. Save the file in `./static/splash_images`
-4. Edit `./src/components/splash/showcase.yaml` to include the card & title.
-5. Edit `./src/components/Footer/index.jsx` to provide credit for the photo.
+1. Navigate to the Auspice view and take a square screenshot of a panel that is representative of the data.
+2. If using a tree, make sure branch labels are turned off.
+3. Downsize the image to 250px by 250px.
+4. Save the file in `./static/splash_images`
+5. Edit `./src/components/splash/showcase.yaml` to include the card & title.
```
joverlee521, line 80: Seeing these instructions made me realize that we don't have instructions for updating the cards on the `/pathogens` page (that I know of). I think makes sense to add here since they share the same pool of images.

_nitpick_ : Maybe renaming `/static/splash_images` to `/static/pathogen_images` to reflect the shared use of images.

joverlee521 (approved):

static-site/README.md
victorlin: Added as a task in https://github.com/nextstrain/nextstrain.org/issues/934

PR #922, opened 2024-06-14
Redesign showcase on splash

victorlin:
> [!NOTE]
> I've created this PR as draft to have an interactive preview of my design mockups. Code is not ready for review but you can get a general sense of direction from the scattered `FIXME` comments.

_preview to come_

Adding more information to the showcase ("featured analyses") section.

## Related issue(s)

- #849

## Checklist

- [ ] Settle on design
- [ ] Clean up code
- [ ] Checks pass
- [ ] Merge base PR #914


trvrb:
I really like where this is headed @victorlin (and it seems like most everyone agrees with this). I just made some minor tweaks. I'd be happy for this to go live except the issues with /pathogens, etc... and code clean up.

I think it's your call for whether to:
1. Just merge this PR into PR #914 (branch `trvrb/semantic-tiles`) as this PR is set up in GitHub to do, and then clean up the PR there
2. Close PR #914 and swap this PR to merge into `master`

victorlin:
@trvrb thanks for those tweaks, visually it's looking very polished. One more thing I want to try is border change on hover to mimic behavior of buttons on the same page.

Code is quite messy so I'll clean it up with the aim to keep things visually identical.

Since there are additional screenshot updates in this PR, I'll close #914 and consolidate all the screenshot changes into this PR as a single commit.

trvrb:
Sounds good. Thanks @victorlin!

static-site/src/components/splash/index.jsx
```diff
@@ -215,57 +217,163 @@ const UrlShowcaseTile = ({ card }) => {
...
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const gitHubLogo = require(`../../../static/logos/github-mark.png`).default.src;
+
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const nextstrainLogo = require(`../../../static/logos/nextstrain-logo-tiny.png`).default.src;
+
+  let maintainers, image;
+
+  if (url.startsWith('/community')) {
+    maintainers = 'community members on GitHub';
```
jameshadfield, line 249: [suggestion]

Change this text to "{orgName} via GitHub"

victorlin: Good suggestion, added in 694ff93b02a94d1434e990433e67cd6c3a82a7b3

static-site/src/components/splash/index.jsx
```diff
@@ -246,13 +246,16 @@ function CardSourceIcon({ url }) {
   let maintainers, image;
 
   if (url.startsWith('/community')) {
-    maintainers = 'community members on GitHub';
+    const owner = url.split('/')[2];
```
jameshadfield, line 249: For narratives it's `[3]`, [e.g.](https://next.nextstrain.org/community/narratives/blab/ebola-narrative-ms/2019-09-13-sit-rep-ENGLISH)

victorlin: Thanks, fixed in 2a2e9f202be8cc9a7a4bc6f5fcd25e454f3330e4.

This made me think of something else: 32d2c7f869ed7708cfcdd5a694bfca01fe11368d

joverlee521 (approved):
Thanks for all of the work you've put into the redesign @victorlin 🌈

static-site/src/components/splash/index.jsx
jameshadfield: > This made me think of something else

Nice improvement!

static-site/src/components/splash/index.jsx
```diff
@@ -211,61 +213,179 @@ export default Splash;
...
         <a href={card.url}>
-          <CardTitle $squashed>
+          <CardTitle>
             {card.name}
           </CardTitle>
-          <CardImgWrapper filename={card.img}/>
+          <CardImgContainer>
+            <CardImgWrapper filename={card.img}/>
+            <InfoIcons>
+              <CardSourceIcon url={card.url} isNarrative />
```
jameshadfield, line 236: The prop here needs to be changed to `isNarrative={isNarrative}` - the current syntax is a [default true prop](https://legacy.reactjs.org/docs/jsx-in-depth.html#props-default-to-true), which is why it passes linting, but it's not behaving like it does when declaring objects. This has the flow on effect (within `CardSourceIcon`) of returning the wrong GitHub org name because everything's a narrative!

static-site/src/components/splash/index.jsx
```diff
@@ -211,61 +213,179 @@ export default Splash;
...
+function CardSourceIcon({ url, isNarrative }) {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const gitHubLogo = require(`../../../static/logos/github-mark.png`).default.src;
+
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const nextstrainLogo = require(`../../../static/logos/nextstrain-logo-tiny.png`).default.src;
+
+  let maintainers, image;
+
+  if (url.startsWith('/community')) {
```
jameshadfield, line 261: Implicit in this logic is that the urls (as defined in the YAML) will always have a leading `/`. But the links continue to work if we forget this, and we never assert it anywhere. As far as I know we can't test for this at build time, due to the async nature of `import`s (although I feel like we _should_ be able to somehow 🤷 ) but it'd be easy to add a (jest) test for this.

All of the URLs currently do have leading slashes, so this can be split out into another issue.

static-site/src/components/splash/index.jsx
```diff
@@ -211,61 +213,179 @@ export default Splash;
...
+          {card.description}
+        </CardDescription>
       </CardInner>
     </CardOuter>
   )
 }
 
-const cardWidthHeight = 160; // pixels
+const cardWidth = 220; // pixels
+const cardHeight = 285; // pixels
```
jameshadfield, line 250: This height works when the description fits on two lines, but (e.g.) I wanted to expand the description for the cattle-flu card but the text becomes clipped (1st screenshot)

A long title which ends up as >1 line makes things look even worse (2nd screenshot)

None of the current entries have this problem, so we can defer work on this to a separate issue.

image

image

jameshadfield (changes requested):
Really nice work @victorlin! Only one little bug which needs fixing (`isNarrative` prop), the rest of my comments can be split out to future work / discussion

static-site/src/components/splash/index.jsx
trvrb: In writing these titles I made sure to stick to one line. Otherwise we have an alignment issue across cards. I think it's fair to demand these be short enough to fit on one line.

static-site/src/components/splash/index.jsx
```diff
@@ -211,61 +213,179 @@ export default Splash;
...
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const faUsers = require(`../../../static/logos/fa-users-solid.svg`).default.src;
+  const [groupLogo, setGroupLogo] = useState(faUsers);
+
+  useEffect(() => {
+    async function getGroupLogo(group) {
+      const response = await fetch(`/charon/getSourceInfo?prefix=/groups/${group}`);
+      const data = await response.json();
+      if (data.avatar) {
+        setGroupLogo(data.avatar);
```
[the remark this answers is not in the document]
victorlin: I was thinking this too. I picked `fa-users-solid.svg` as a placeholder without too much thought into it. I imagine we'll only see this in review apps though and would feature resources that have a a group logo available.

We could do something similar for community and fetch the GitHub org/account's avatar. I'll consider it in https://github.com/nextstrain/nextstrain.org/issues/934

trvrb (approved):
I really like where this ended up @victorlin. 😀

static-site/src/components/splash/index.jsx
```diff
@@ -211,61 +213,179 @@ export default Splash;
...
+}
+
+function GroupImage({ group }) {
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
+  const faUsers = require(`../../../static/logos/fa-users-solid.svg`).default.src;
+  const [groupLogo, setGroupLogo] = useState(faUsers);
+
+  useEffect(() => {
+    async function getGroupLogo(group) {
+      const response = await fetch(`/charon/getSourceInfo?prefix=/groups/${group}`);
```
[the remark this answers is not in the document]
joverlee521: > Hardcoding group logos in the YAML feels off to me. 

+1. I think the groups logo should be dynamic since we don't have control over them.

static-site/src/components/splash/index.jsx
victorlin: This came back to bite in https://github.com/nextstrain/nextstrain.org/pull/993#discussion_r1729422975. Removing the note and properly handling the absence of `/` in #996.

PR #928, opened 2024-06-21
Update group cards further

genehack:

* Links group name to group page; updates tooltip on group name

## Related issue(s)

#870

## Checklist

- [ ] Checks pass

### DIFF
--- a/static-site/src/components/ListResources/ResourceGroup.tsx
+++ b/static-site/src/components/ListResources/ResourceGroup.tsx
@@ -15,9 +15,11 @@ interface ResourceGroupHeaderProps {
   isCollapsed: boolean
   resourcesToShowWhenCollapsed: number
   quickLinks: QuickLink[]
+  tooltipText: string
 }
 
-const ResourceGroupHeader = ({group, isMobile, setCollapsed, collapsible, isCollapsed, resourcesToShowWhenCollapsed, quickLinks}: ResourceGroupHeaderProps) => {
+const ResourceGroupHeader = ({group, isMobile, setCollapsed, collapsible, isCollapsed, resourcesToShowWhenCollapsed, quickLinks, tooltipText}: ResourceGroupHeaderProps) => {
+  if (!tooltipText) tooltipText="Click to load the default (and most recent) analysis";
   const setModalResource = useContext(SetModalResourceContext);
   if (!setModalResource) throw new Error("Context not provided!")
 
@@ -34,7 +36,7 @@ const ResourceGroupHeader = ({group, isMobile, setCollapsed, collapsible, isColl
 
         <HeaderRow>
           {group.groupUrl ? (
-            <TooltipWrapper description={`Click to load the default (and most recent) analysis for ${group.groupDisplayName || group.groupName}`}>
+            <TooltipWrapper description={`${tooltipText} for ${group.groupDisplayName || group.groupName}`}>
               <GroupLink style={{ fontSize: '2rem', fontWeight: '500'}} href={group.groupUrl} target="_blank" rel="noreferrer">
                 {group.groupDisplayName || group.groupName}
               </GroupLink>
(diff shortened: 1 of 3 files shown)


genehack:
Pushing this for preliminary feedback on this first bit; more to come.

static-site/src/pages/groups.jsx
```diff
@@ -115,6 +115,7 @@ class GroupsPage extends React.Component {
 
         <ListResources
           resourceType="dataset"
+          defaultGroupLinks={true}
```
jameshadfield, line 118: [crystal ball gazing -- not a request for changes]

Implicit in the original design is that card titles will become links by simply adding the card name to the current URL. It works well for the groups page (e.g. `groups` + `blab`), but I don't think it'll work for an individual groups page.¹ We'll probably just disable the links for those pages.

Also, for what it's worth, the `={true}` can be dropped, if you prefer. 

¹ Depends on how card titles are chosen of course, but the blab group will probably have a "229e" card and [/groups/blab/229e](https://next.nextstrain.org/groups/blab/229e) is 404.

static-site/src/components/ListResources/ResourceGroup.tsx
```diff
@@ -15,9 +15,11 @@ interface ResourceGroupHeaderProps {
   isCollapsed: boolean
   resourcesToShowWhenCollapsed: number
   quickLinks: QuickLink[]
+  tooltipText: string
```
jameshadfield, line 18: There are quite a few tooltips within the resources UI, so I'd consider a prop name such as `defaultGroupLinksTooltip`.

static-site/src/pages/groups.jsx
```diff
@@ -117,7 +117,10 @@ class GroupsPage extends React.Component {
           resourceType="dataset"
           defaultGroupLinks={true}
           versioned={false}
-          resourceListingCallback={resourceListingCallback}/>
+          resourceListingCallback={resourceListingCallback}
+          // FIXME I do not love this text string; suggestions welcome
+          tooltipText="Click to load the group page"
```
jameshadfield, line 122: [suggestion]

"Click to view this group in isolation, including additional information specific to that group"

Won't look nice if you then tack on `for ${group.groupDisplayName || group.groupName}` but we could change the prop to be a function accepting the group name.

static-site/src/pages/groups.jsx
joverlee521: > Depends on how card titles are chosen of course, but the blab group will probably have a "229e" card and [/groups/blab/229e](https://next.nextstrain.org/groups/blab/229e) is 404.

I would think on the individual groups page, the titles would not include `groups/<group_name>`. The `resourceListingCallback` would just return `pathPrefix: "groups/<group_name>"`? (Maybe I'm completely off here, still catching up on all the ListResources work)

static-site/src/pages/groups.jsx
joverlee521: The [Nextstrain docs for Groups](https://docs.nextstrain.org/en/latest/learn/groups/index.html#pages) call it a "splash page", so we can follow that pattern here:

"Click to view the individual group splash page"

static-site/src/components/ListResources/ResourceGroup.tsx
victorlin: This is not the entire text and gets added to later. I would call it something like `groupTooltipPrefix`.

static-site/src/components/ListResources/ResourceGroup.tsx
```diff
@@ -15,9 +15,11 @@ interface ResourceGroupHeaderProps {
...
 }
 
-const ResourceGroupHeader = ({group, isMobile, setCollapsed, collapsible, isCollapsed, resourcesToShowWhenCollapsed, quickLinks}: ResourceGroupHeaderProps) => {
+const ResourceGroupHeader = ({group, isMobile, setCollapsed, collapsible, isCollapsed, resourcesToShowWhenCollapsed, quickLinks, tooltipText}: ResourceGroupHeaderProps) => {
+  if (!tooltipText) tooltipText="Click to load the default (and most recent) analysis";
   const setModalResource = useContext(SetModalResourceContext);
```
victorlin, line 23: Suggestion: don't set a default here. Make sure every usage of `ListResources` defines the right text based on the context.

static-site/src/pages/groups.jsx
victorlin: Or something less wordy:

> Go to splash page for blab

> Visit splash page for blab

Issue #883, opened 2024-05-30
Add mumps and EV-D68 to /pathogens showcase


victorlin:
Proposed in https://github.com/nextstrain/nextstrain.org/pull/878#discussion_r1619548498

This will go well with changes from #878.

PR #930, opened 2024-06-21
Showcase refactor and additions

victorlin:
_[preview](https://nextstrain-s-victorlin--7dxrsz.herokuapp.com/)_

Small fixes - see commit messages.

## Related issue(s)

- Prepping for #922
- Resolves https://github.com/nextstrain/nextstrain.org/pull/878#discussion_r1617752924
- Closes https://github.com/nextstrain/nextstrain.org/issues/883

## Checklist

- [x] Checks pass

### DIFF
--- a/static-site/src/components/ListResources/index.tsx
+++ b/static-site/src/components/ListResources/index.tsx
@@ -11,7 +11,7 @@ import { ResourceGroup } from './ResourceGroup';
 import { ErrorContainer } from "../../pages/404";
 import { TooltipWrapper } from "./IndividualResource";
 import {ResourceModal, SetModalResourceContext} from "./Modal";
-import { CardImgWrapper, CardInner, CardOuter, CardTitle, Showcase } from "../Showcase";
+import { Showcase } from "../Showcase";
 import { FilterCard, FilterOption, Group, QuickLink, Resource, ResourceListingInfo } from './types';
 import { HugeSpacer } from "../../layouts/generalComponents";
 
@@ -83,7 +83,7 @@ function ListResources({
         </Byline>
 
         <SetSelectedFilterOptions.Provider value={setSelectedFilterOptions}>
-          <Showcase cards={showcaseCards} CardComponent={FilterShowcaseTile} />
+          <Showcase cards={showcaseCards} cardWidth={cardWidthHeight} cardHeight={cardWidthHeight} CardComponent={FilterShowcaseTile} />
         </SetSelectedFilterOptions.Provider>
         </>
       )}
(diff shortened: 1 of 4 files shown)


static-site/src/components/Showcase/index.tsx
```diff
@@ -112,65 +110,13 @@ const PreviewOverlay = styled.div`
 
 const CardsContainer = styled.div`
   /* background-color: #ffeab0; */
-  display: grid;
-  grid-template-columns: repeat(auto-fit, minmax(${cardWidthHeight}px, max-content));
-  grid-gap: 1%;
+  display: inline-flex;
+  flex-wrap: wrap;
+  gap: 10px;
```
jameshadfield, line 115: I would use a smaller (or no) row-gap to better match the previous (grid) layout where we could mostly read the titles of the collapsed row. An alternative solution would be to keep the gap as-is and increase the space allocated to the UI when collapsed

image

jameshadfield (commented):
Scanned through the code / used the review app and this looks good to me, barring the one in-line comment. (I [preferred left-aligned dangling rows](https://github.com/nextstrain/nextstrain.org/pull/878#discussion_r1617899607), contrary to the majority.)

static-site/src/components/Showcase/index.tsx
victorlin: Thanks for pointing this out. I realized this was a side effect of using a relative value for `gap`. Fixed in dcd8a92829f110d5be32280cb857c2d27ec6d1b6.

This also prompted me to add some other CSS grid fixes to this PR: 883998a33a2a0f43e6d3eac4b82976ccb913c017, f8388ea0f5bda3ca08a6c3e9a21030d56ccd29f2

static-site/src/components/Showcase/index.tsx
victorlin, line 1: Re: https://github.com/nextstrain/nextstrain.org/pull/930#pullrequestreview-2135023379

> (I [preferred left-aligned dangling rows](https://github.com/nextstrain/nextstrain.org/pull/878#discussion_r1617899607), contrary to the majority.)

I also prefer left-aligned dangling rows (which is what [prompted](https://github.com/nextstrain/nextstrain.org/pull/878#discussion_r1617935653) using CSS grid here). That makes it a tie, and since left-aligned is the current behavior on nextstrain.org, I've updated this PR to keep it. The other reason I had changed to flexbox here was to avoid passing the width as a prop since it's required for CSS grid. This PR now keeps CSS grid and passes the width as a prop, which isn't terrible since the height is already necessary for different reasons.

static-site/src/components/ListResources/index.tsx
```diff
@@ -310,3 +310,58 @@ const useShowcaseCards = (cards?: FilterCard[], groups?: Group[]) => {
   }, [cards, groups]);
   return restrictedCards;
 }
+
+const cardWidthHeight = 160; // pixels
+
+const CardOuter = styled.div`
```
joverlee521, line 316: _non-blocking_

I'm not 100% convinced by this change since I'd be worried about the splash and ListResources card styling diverging. Would it make sense to pull the card styling out as a separate shared component? (Maybe confusing with the [existing Cards component](https://github.com/nextstrain/nextstrain.org/tree/master/static-site/src/components/Cards)...)

joverlee521 (approved):

static-site/src/components/ListResources/index.tsx
joverlee521: Oh actually, there's a lot of overlap in the styling with the [existing Cards styles](https://github.com/nextstrain/nextstrain.org/blob/05aba2bc93ef855de7b0c932b977d98c11b40408/static-site/src/components/Cards/styles.tsx), would it make sense to just have a single centralized Card component?

victorlin: I think it's fine to have separate copies, at least while we figure out what design works best for each page. In my experience so far, the idea of a centralized Card design has been limiting and it's easiest to have different components (hence the parameterization of `CardComponent` in 1414985c92488423035a1470534c22a8ef479595 and follow-up to that in this PR as 9beac0fa2c324b326587063e0f2e3ef48d9ac653).

> I'd be worried about the splash and ListResources card styling diverging

I've already made that divergence happen in #922. I think we've all agreed that the new splash page card design looks good on that page. This design doesn't make sense for the ListResources showcase pages (`/pathogens`, `/staging`). While some parts of the design may work well, there's still some obvious differences (e.g. no need for the "maintainer" icon since all those are maintained by the Nextstrain team).

I haven't touched the existing Cards component (`static-site/src/components/Cards`) but after these PRs it's probably time to re-scope that to its only usage which is on the `/groups` page. It also doesn't make sense to apply the new splash page card design (linking to datasets/narratives) to the groups cards (linking to group pages).

joverlee521: Ah, I haven't gotten to that PR yet, but this separation makes sense in that context 👍

jameshadfield: > I haven't touched the existing Cards component (static-site/src/components/Cards) but after these PRs it's probably time to re-scope that to its only usage which is on the /groups page. It also doesn't make sense to apply the new splash page card design (linking to datasets/narratives) to the groups cards (linking to group pages).

The cards UI currently in use on `/groups` (the one using blank colours rather than images) isn't something we need to preserve. That is to say, we can re-organise the source files to make it clear it's only used in one place, but we shouldn't invest time developing it without considering alternatives.

victorlin: > we can re-organise the source files to make it clear it's only used in one place

Added as a task in https://github.com/nextstrain/nextstrain.org/issues/934

static-site/src/components/Showcase/index.tsx
trvrb: I also prefer left-aligned dangling rows.

trvrb (approved):

PR #939, opened 2024-07-01
Update routes for seasonal cov

genehack:

Adds route and manifest_core entry for `seasonal-cov` datasets.

## Related issue(s)

https://github.com/nextstrain/seasonal-cov/issues/13

## Checklist

- [ ] Checks pass

### DIFF
--- a/data/manifest_core.json
+++ b/data/manifest_core.json
@@ -666,6 +666,15 @@
           "default": "a"
         }
       },
+      "seasonal-cov": {
+        "resolution": {
+          "229e": "",
+          "hku1": "",
+          "nl63": "",
+          "oc43": "",
+          "default": "229e"
+        }
+      },
       "tb": {
         "resolution": {
           "global": "",
--- a/src/routing/core.js
+++ b/src/routing/core.js
@@ -43,6 +43,7 @@ const coreBuildPaths = [
   "/ncov",
   "/nextclade",
   "/rsv",
+  "/seasonal-cov",
   "/seasonal-flu",
   "/tb",
   "/WNV",


data/manifest_core.json
```diff
@@ -666,6 +666,15 @@
           "default": "a"
         }
       },
+      "seasonal-cov": {
+        "resolution": {
```
joverlee521, line 670: _nit-pick_

I don't think the `resolution` key gets surfaced anywhere on the website (yet?) but I think more appropriate to label these as `virus` or `type`?

joverlee521 (approved):
Testing app is redirecting `/seasonal-cov` to `/seasonal-cov/229e` as expected and the Auspice dataset dropdown is appropriately populated with the 4 different viruses 🎉

data/manifest_core.json
jameshadfield: > I don't think the resolution key gets surfaced anywhere on the website (yet?)

As far as I remember it's only used during when working out what options to expose as potential second trees (part of the `charon/getAvailable` API):

https://github.com/nextstrain/nextstrain.org/blob/7a8e92ffc6448ee1be96186dbdef279c7961d088/src/endpoints/charon/parseManifest.js#L78-L80

(Oh how that needs to be improved one day, but not here!)

jameshadfield (approved):
Nice one! Consider adding a splash card / pathogens-showcase card?

data/manifest_core.json
genehack: Is there a list of what are acceptable keys at that level of the data structure, and a key as to what those things mean? Reviewing the whole file, I see @joverlee521 `s point about `"resolution"` not generally being used to distinguish between strains, but also: I don't see `"virus"` used at all, and the only place `"type"` is used is for RSV, where it is common to talk about viral subtypes (at least, that's what a quick web search indicates).

`"virus"` (or `"species"`, I guess?) **do** feel more correct. Feedback welcome.

genehack:
> Nice one! Consider adding a splash card / pathogens-showcase card?

Tracking in #940.

data/manifest_core.json
joverlee521: Yeah, the actual value is ignored and just indicates a separate level for the Auspice dropdown. I was mostly flagging for internal reading of the manifest file. 

If we do decide to surface these values in the UI in the future, we'll probably have to do a whole overhaul of the manifest anyways...

jameshadfield: I think the technically correct word here is "strain". Let's not let this decision hold up this PR!

> the actual value here is mostly ignored

Correct - ignored unless it's "segment".

> If we do decide to surface these values in the UI in the future

I think that was on the cards when Nextstrain was only core datasets, but now that we have such a diverse range of external datasets I don't think we'll build UIs which use manually curated information like this.

joverlee521: > I think the technically correct word here is "strain". 

Oh boy. I'm always confused by "strain" vs our use of strain names in Nextstrain 🙈 

> Let's not let this decision hold up this PR!

Yeah, I didn't mean to hold up this PR. Fine to merge as-is.

genehack: > > I think the technically correct word here is "strain".
> 
> Oh boy. I'm always confused by "strain" vs our use of strain names in Nextstrain 🙈

I don't think "strain" is appropriate here because these are all distinct viral _species_, unless we're using "strain" differently than I think is typical in biology? 

> > Let's not let this decision hold up this PR!
> 
> Yeah, I didn't mean to hold up this PR. Fine to merge as-is.

Merging, still curious about what "strain" means in this context :wink:

jameshadfield: > I'm always confused by "strain"

To be fair, the entire field is too.

I think any of "subtype" "species" "strain" "cov-strain" is better than "resolution", but it's a minor thing.

PR #941, opened 2024-07-02
Replace seasonal flu with Zika in API examples

huddlej:

Replace seasonal flu with Zika in API examples based on [Slack discussion](https://bedfordlab.slack.com/archives/C03KWDET9/p1719943529502889?thread_ts=1719612970.204079&cid=C03KWDET9).

[Preview of changes to docs](https://nextstrain--941.org.readthedocs.build/projects/nextstrain-dot-org/en/941/api-restful.html#synopsis).

## Checklist

- [x] Checks pass

### DIFF
--- a/docs/api-restful.rst
+++ b/docs/api-restful.rst
@@ -33,16 +33,16 @@ Synopsis
 .. code-block:: bash
 
     # Download a main dataset JSON file…
-    curl https://nextstrain.org/seasonal-flu/h3n2/ha/2y                         \
+    curl https://nextstrain.org/zika                         \
         --header 'Accept: application/vnd.nextstrain.dataset.main+json'         \
         --compressed                                                            \
-            > seasonal-flu_h3n2_ha_2y.json
+            > zika.json
 
     # …and re-upload it another location.
-    curl https://nextstrain.org/groups/blab/flu/latest                          \
+    curl https://nextstrain.org/groups/blab/zika/latest                          \
         --header 'Authorization: Bearer eyJraWQiOiJyM3…'                        \
         --header 'Content-Type: application/vnd.nextstrain.dataset.main+json'   \
-        --upload-file seasonal-flu_h3n2_ha_2y.json
+        --upload-file zika.json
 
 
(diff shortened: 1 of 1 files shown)


trvrb (approved):
Looks great!

docs/api-restful.rst
```diff
@@ -33,16 +33,16 @@ Synopsis
 .. code-block:: bash
 
     # Download a main dataset JSON file…
-    curl https://nextstrain.org/seasonal-flu/h3n2/ha/2y                         \
+    curl https://nextstrain.org/zika                         \
```
victorlin, line 36: Whitespace to align `\` across multiple lines is a bit off now. I'll fix in another PR

docs/api-restful.rst
```diff
@@ -33,16 +33,16 @@ Synopsis
...
         --header 'Accept: application/vnd.nextstrain.dataset.main+json'         \
         --compressed                                                            \
-            > seasonal-flu_h3n2_ha_2y.json
+            > zika.json
```
victorlin, line 39: I also contemplated replacing with another example that uses the `/` to `_` pattern, but realized that's not a feature relevant to this API and more for something like [`nextstrain remote download`](https://docs.nextstrain.org/projects/cli/en/8.5.0/commands/remote/download/) where that's clearly documented already.

docs/api-restful.rst
victorlin: https://github.com/nextstrain/nextstrain.org/pull/943

docs/api-restful.rst
tsibley: Hmm. I used seasonal-flu exactly for this reason: to highlight how multi-part pathogen URLs work and to show by example the local filesystem naming convention (even if you can ignore it, you're likely better off not). I'd suggest switching to mpox or ncov/open or something similar where we can maintain the multi-part example.

victorlin: Done: https://github.com/nextstrain/nextstrain.org/pull/946

Issue #947, opened 2024-07-08
Groups page improvements


victorlin:
## Potential tasks

- [ ] Make cards container expandable
- [ ] Remove cards in favor of resource listing
- [ ] Use group logo/title in card

victorlin:
> Remove cards in favor of resource listing

Rather than removing it, I think it can be improved. With an expandable cards container, `/groups` would look visually similar to `/pathogens`. Keeping the cards maintains that similarity, though the behavior on click is slightly different.

An idea to integrate the two parts of the page:

1. Make the click behavior of `/groups` act as a filter similar to the behavior on `/pathogens`.
2. Make the resource listing titles clickable (to preserve links to group pages). This is currently in a PR: https://github.com/nextstrain/nextstrain.org/pull/928/commits/f137aea34b358a23c997e39aa732edaaaec4cd6b
3. Make the resource listing show groups without resources (to preserve links to all groups).

joverlee521:
> Make the resource listing show groups without resources (to preserve links to all groups).

Ah right, I hadn't considered groups without datasets! We should definitely preserve links to all groups even if they do not have published resources. 

I think the expandable cards container makes sense for the /pathogens page because we plan to have curated cards. Will the expandable display still make sense if we display ~all groups (we're already at 30+ public groups) or would it feel like clutter?

joverlee521:
> Use group logo/title in card

Related to https://github.com/nextstrain/nextstrain.org/issues/870 which also has a task to display group logos in the resource listing.

victorlin:
> Will the expandable display still make sense if we display ~all groups (we're already at 30+ public groups) or would it feel like clutter?

I'll make a PR so we can get a better sense of it interactively.

PR #949, opened 2024-07-09
Make group tiles expandable

victorlin:
[_preview_](https://nextstrain-s-victorlin--rgwhw6.herokuapp.com/groups)

ExpandableTiles is already used in two other places where tiles are used. Use it here too for consistency.

## Related issue(s)

https://github.com/nextstrain/nextstrain.org/issues/947

## Checklist

- [ ] Checks pass
- [ ] [Preview](https://nextstrain-s-victorlin--rgwhw6.herokuapp.com/groups) looks good
- [ ] Merge base PR #950

### DIFF
--- a/static-site/src/components/Groups/Tiles/index.tsx
+++ b/static-site/src/components/Groups/Tiles/index.tsx
@@ -1,22 +1,22 @@
 import React, { useContext } from "react";
 import * as Styles from "./styles";
-import { H1 } from "../../splash/styles";
-import { MediumSpacer } from "../../../layouts/generalComponents";
+import { width, height } from "./styles";
 import Padlock from "./padlock";
 import { theme } from "../../../layouts/theme";
 import { UserContext } from "../../../layouts/userDataWrapper";
 import { GroupTile } from "./types";
 import { Group } from "../types";
+import { ExpandableTiles } from "../../ExpandableTiles";
 
 
-export const GroupTiles = ({squashed}: { squashed: boolean }) => {
+export const GroupTiles = () => {
   const { visibleGroups } = useContext(UserContext);
   return (
-    <Tiles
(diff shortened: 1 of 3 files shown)


jameshadfield:
Looks nice (didn't test with private groups). If we want to keep the current design (or at least make minimal changes), making the cards a more similar size to those on /pathogens would be nice. Longer term I think it's still an open question about what UI we should display above the fold on the /groups page.

victorlin:
> making the cards a more similar size to those on /pathogens would be nice

Done: f8122d1b6b5b63d4f4e007a9f87e6ba28c1094a3

jameshadfield (approved):

joverlee521 (approved):
Definitely much cleaner to have the expandable tiles! 

A couple of non-blocking notes

1. The [Groups docs](https://docs.nextstrain.org/en/latest/learn/groups/index.html#names) state the group name can have 3-50 characters. We should probably ensure 50 character group names will still be displayed "nicely" in the smaller tiles.

2. To me, it feels like we are preferentially displaying the groups in the first row. Others may or may not feel the same. Might seem less biased if we note that groups are listed in the order they were added? (The order comes from the order in the groups.json, which is roughly the order in which groups are added.)

victorlin:
Good points!

1. I just tested on the longest group name at 29 characters. It's slightly cut off:

    image

    I've [force-pushed](https://github.com/nextstrain/nextstrain.org/compare/f8122d1b6b5b63d4f4e007a9f87e6ba28c1094a3..20e88d780ef737f21993c23a895a5557d3aa1894) to reduce the left padding so it now looks like this:

    image

    Going forwards, we can check PR previews and update the styles if necessary.

2. I'll make a separate issue for this

PR #951, opened 2024-07-10
Reorganize files related to splash

victorlin:
_[preview](https://nextstrain-s-victorlin--lzgmyt.herokuapp.com/)_

Lots of small changes. See commit messages.

## Checklist

- [x] Checks pass
- [x] These pages still load fine: [`/`](https://nextstrain-s-victorlin--lzgmyt.herokuapp.com/), [`/groups/blab`](https://nextstrain-s-victorlin--lzgmyt.herokuapp.com/groups/blab)


static-site/src/components/nav-bar/index.tsx
```diff
@@ -19,6 +18,21 @@ const NavContainer = styled.div`
   left: 0px;
   z-index: 1001;
   transition: left .3s ease-out;
+
+  // TODO: Put this component in a container which sets these styles for all pages.
```
jameshadfield, line 22: [for consideration] You could use a global CSS class ([here](https://github.com/nextstrain/nextstrain.org/blob/d823926cab0180625cdeceeb48d4ba087dd6ae43/static-site/pages/_app.tsx#L5-L7) are the global CSS imports we currently use) or a CSS module which both pages import.

victorlin: I haven't actually tried so this could get complicated, but I would rather follow through here and create a common page-level container component to be used across all pages so that there's not so many different ones with varying styles and navbar + footer logos added separately.

Or maybe the split is intentional – if that's the case, this TODO doesn't really make sense and I'll consider using the global CSS class here.

victorlin: Moved this change to #958

joverlee521 (approved):
Reorganization makes sense to me 👍

PR #952, opened 2024-07-10
Rearrange and update splash contents

victorlin:
[_preview_](https://nextstrain-s-victorlin--xbfmig.herokuapp.com)

Follow-up to #901 which added a section without much thinking on how it fits with the rest of the page. Design and contents directly reflects [mockup](https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1719520299456109) by @trvrb.

## Related issue(s)

- #849 

## Checklist

- [x] Decide on section links and contents
- [x] Checks pass
- [x] Merge base PR #951
- [x] Merge PR #958 and drop 6358a8c7


trvrb:
Awesome work here @victorlin! One quick I just noticed is that the responsive behavior at phone width is not ideal. Here's what I just got:

Screenshot 2024-07-25 at 1 05 02 PM

I'd recommend:
1. Dropping the circular logo when hitting phone width. The "Nextstrain" logotype should fit in this case.
2. I'd _think_ I'd drop to 1 entry per row when hitting phone width. The two columns feel squeezed when it's this narrow.

Exact breakpoint for both of the above would have to be experimented with.

trvrb:
I also just made a few small tweaks to the language in the links (in part to make everything a noun rather than mix of nouns and verbs).

victorlin:
Thanks @trvrb! I'll play around with it more at narrow widths and push some fixes for that.

👍 link to installation docs should be more meaningful to users. I'll change the icon from GitHub logo to something that represents "tooling".

victorlin:
> Exact breakpoint for both of the above would have to be experimented with.

I felt myself reaching for the more precise breakpoints from Bootstrap v4 so I created a [PR](https://github.com/nextstrain/nextstrain.org/pull/958) to switch to that and used it here. Responsive design now looks like:

https://github.com/user-attachments/assets/9ee10e79-cf8e-4071-9a14-2cb7764a0efa

static-site/src/components/splash/index.tsx
```diff
@@ -1,135 +1,127 @@
...
+          link="https://clades.nextstrain.org"
         />
         <Section
-          id="narratives"
-          title="Narratives"
-          abstract="Narratives are a method of data-driven storytelling. They allow authoring of content which is displayed alongside a view into the data."
-          buttonText="Find out more"
-          buttonLink="https://docs.nextstrain.org/en/latest/guides/communicate/narratives-intro.html"
+          id="groups"
+          // eslint-disable-next-line @typescript-eslint/no-var-requires
```
genehack, line 103: i wonder if it would be easier/better to disable this rule for all `*.tsx` files in `static-site/.eslintrc.yaml`?

genehack (approved):

static-site/src/components/splash/index.tsx
victorlin: Good idea, opened an issue: #959

static-site/src/components/nav-bar/index.tsx
```diff
@@ -119,8 +119,10 @@ const NavBar = ({ minified }: {
 
   return (
     <NavContainer>
-      <Logo />
-      <LogoType />
+      {Router.pathname != "/" && <>
+        <Logo />
+        <LogoType />
+      </>}
```
joverlee521, line 125: _non-blocking_ 

Since there's discussion in this PR regarding the responsiveness on mobile, I wanted to note that the "LOGIN" button is lost in mobile view because of the presence of the logo (looking at the [/pathogens page](https://nextstrain-s-victorlin--xbfmig.herokuapp.com/pathogens)). 

Screenshot 2024-07-29 at 1 13 16 PM

A nice future improvement would be to collapse the nav links behind button at certain breakpoints, but for now maybe just hide the logo based on breakpoint?

joverlee521 (approved):

static-site/src/components/nav-bar/index.tsx
victorlin: That's already the behavior without this PR, so I've created a separate issue: #962

joverlee521: > That's already the behavior without this PR

Ah, guess I just haven't looked at the mobile view in a while!

trvrb:
I really like what you ended up with here @victorlin in terms of responsive breakpoints.

PR #955, opened 2024-07-17
manifest: update rsv resolutions

rneher:
Add different time resolution to the RSV builds to produce the hierarchy (defaults in italics):

- a
  - *genome*
    - all-time
    - *6y*
    - 3y
  - F
    - all-time
    - *6y*
    - 3y
  - G 
    - all-time
    - *6y*
    - 3y
- b
  - *genome*
    - all-time
    - *6y*
    - 3y
  - F
    - all-time
    - *6y*
    - 3y
  - G 
    - all-time
    - *6y*
    - 3y

### DIFF
--- a/data/manifest_core.json
+++ b/data/manifest_core.json
@@ -649,17 +649,59 @@
         "type": {
           "a": {
             "region":{
-              "genome":"",
-              "G":"",
-              "F":"",
+              "genome": {
+                "resolution":{
+                  "all-time":"",
+                  "6y":"",
+                  "3y":"",
+                  "default":"6y"
+                }
+              },
+              "G": {
+                "resolution":{
+                  "all-time":"",
+                  "6y":"",
+                  "3y":"",
+                  "default":"6y"
+                }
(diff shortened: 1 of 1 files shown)


jameshadfield (approved):
Manifest looks good 👍 

cc @joverlee521 - I don't think these kinds of changes necessitate an update to the resource index version, but I'm not 100% sure...

rneher:
it all works on the preview instance. no rush getting this live. but the rsv-pipeline will push builds to the new location from now on.

joverlee521:
> cc @joverlee521 - I don't think these kinds of changes necessitate an update to the resource index version, but I'm not 100% sure...

This particular change is not breaking so it doesn't necessitate an update to the version. However, this change does mean that the previous `@version` URLs will no longer work.

https://next.nextstrain.org/rsv/a/F@2023-04-18 redirects to https://next.nextstrain.org/rsv/a/F/6y@2023-04-18 which does not exist.

PR #960, opened 2024-07-29
Allow require() outside of import statements

### LINKED ISSUE #959: Disable `@typescript-eslint/no-var-requires` across the project
_proposed by @genehack in https://github.com/nextstrain/nextstrain.org/pull/952#discussion_r1695622111_

Currently, the rule [@typescript-eslint/no-var-requires](https://typescript-eslint.io/rules/no-var-requires/) is disabled per-line (88ec4bcea271e7289d4bc6e6be6e26810cdf71de).

However, the pattern of using `require("path/to/image").default.src` seems to be project-wide rather than individual one-offs. In this case, it makes more sense to disable the rule across the project in [static-site/.eslintrc.yaml](https://github.com/nextstrain/nextstrain.org/blob/916ea06647334d9d2f1ac336a78e944f816b7072/static-site/.eslintrc.yaml).

victorlin:

The pattern of using require("path/to/image").default.src is project-wide rather than individual one-offs. In this case, it makes more sense to disable the rule across the project in files which may import images.

## Related issue(s)

Closes #959

## Checklist

- [x] Checks pass

### DIFF
--- a/static-site/.eslintrc.yaml
+++ b/static-site/.eslintrc.yaml
@@ -31,6 +31,13 @@ rules:
   '@next/next/no-img-element': off # Remove this if we use next.js optimisations for <img>
   '@next/next/no-html-link-for-pages': off
 
+overrides:
+  # require() is used for images that are not under static-site/public
+  - files:
+      - '**.[jt]sx'
+    rules:
+      '@typescript-eslint/no-var-requires': off
+
 parserOptions:
   sourceType: module
   ecmaVersion: 2022
--- a/static-site/src/components/Groups/Tiles/index.tsx
+++ b/static-site/src/components/Groups/Tiles/index.tsx
@@ -50,7 +50,6 @@ const Tile = ({ tile }: { tile: GroupTile }) => {
               <Padlock/>
             </Styles.BottomRightLabel>
           ) : null}
-          {/* eslint-disable-next-line @typescript-eslint/no-var-requires */}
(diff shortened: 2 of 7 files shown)


static-site/.eslintrc.yaml
```diff
@@ -30,6 +30,7 @@ rules:
   react/prop-types: off # Remove this override once all props have been typed using PropTypes or TypeScript.
   '@next/next/no-img-element': off # Remove this if we use next.js optimisations for <img>
   '@next/next/no-html-link-for-pages': off
+  '@typescript-eslint/no-var-requires': off # require() is used for images that are not under static-site/public
```
genehack, line 33: doing something like this (note: did not check syntax carefully...) will make this specific to JSX/TSX files, which feels like a better scoping.


```suggestion
  overrides:
    # require() is used for images that are not under static-site/public
    - files:
      - "**.[jt]sx"
      rules:
        "@typescript-eslint/no-var-requires": off
```

genehack (approved):

static-site/.eslintrc.yaml
victorlin: Thanks, good suggestion! Added in [force-push](https://github.com/nextstrain/nextstrain.org/compare/d6baa4f8957c59cbbf4786d71ce4298968054983..0e566de918aae41b399800dda8e343047621d174).

Your syntax worked! I just changed indentation and quoting for consistency.

PR #956, opened 2024-07-23
Bump RESOURCE_INDEX version

joverlee521:
## Description of proposed changes

Follow up to https://github.com/nextstrain/nextstrain.org/pull/955

Bumping the RESOURCE_INDEX version should make the transition smoother when we promote the changes to prod.

## Checklist

- [ ] Checks pass
- [ ] Check if changes affect the [resource index JSON revision](https://docs.nextstrain.org/projects/nextstrain-dot-org/en/latest/resource-collection.html#resource-index-revisions)

PR #961, opened 2024-07-29
Fix typo in h5n1-cattle-outbreak segment 'pa'

victorlin:
Follow-up to https://github.com/nextstrain/nextstrain.org/pull/957#discussion_r1695748905

## Checklist

- [x] Checks pass


joverlee521 (approved):
Good catch!

victorlin:
> Check if changes affect the [resource index JSON revision](https://docs.nextstrain.org/projects/nextstrain-dot-org/en/latest/resource-collection.html#resource-index-revisions)

`data/manifest_core.json` is changed, but next.nextstrain.org already has the correct link to `pa` builds. I think this is because #956 got merged after the new segment builds were already available in S3. So no need to bump with this PR.

Issue #934, opened 2024-06-27
Small tweaks to be done after splash page redesign


victorlin:
- [ ] Update `/static-site/README.md` screenshot instructions to account for use on `/pathogens` page
- [ ] `/static/splash_images` → `/static/pathogen_images` to reflect the shared use of images
- [ ] Re-scope the Cards component to reflect sole usage on groups page

victorlin:
All tasks have been completed except one:

> Consider consolidating designs of Card components (splash, `/pathogens`, `/groups`)

I'm going to close this without checking that off since it was partially addressed by #949 and there aren't any clear improvements to be made.

PR #965, opened 2024-07-30
Show GitHub account avatar on community tiles

victorlin:
[_preview_](https://nextstrain-s-victorlin--c1t8eq.herokuapp.com)

This is more consistent with the Nextstrain Groups tiles showing the group's personalized logo.

## Related issue(s)

- #934

## Checklist

- [x] [Preview](https://nextstrain-s-victorlin--c1t8eq.herokuapp.com) looks good
- [x] Checks pass


genehack:
Doesn't seem to work as intended for organizations? Compare the Chikungunya tile and the small generic icon there with what's on [the Vienna RNA org](https://github.com/viennarna). Whereas the Yersina pestis build does show the same icon as [ktmeaton's Github](https://github.com/ktmeaton). 

I would also consider relocating the mini-icon overlay from the top left of the tile to near the bottom left, to the left of and top-aligned with the title/caption text. I think it will be more noticeable and having the attribution next to the title feels appropriate. (Suggestion only, no strong opinion really.)

victorlin:
> Doesn't seem to work as intended for organizations? Compare the Chikungunya tile and the small generic icon there with what's on [the Vienna RNA org](https://github.com/viennarna). Whereas the Yersina pestis build does show the same icon as [ktmeaton's Github](https://github.com/ktmeaton).

The ViennaRNA tile is for a Nextstrain Groups resource which is not affected by the changes here. It won't work on PR review apps due to AWS permissions, but you can see it on https://next.nextstrain.org

> I would also consider relocating the mini-icon overlay from the top left of the tile to near the bottom left, to the left of and top-aligned with the title/caption text. I think it will be more noticeable and having the attribution next to the title feels appropriate. (Suggestion only, no strong opinion really.)

Could you provide a mockup? I'm confused on what "top-aligned with the title/caption text" means, and what I'm visualizing seems to be opposite of "having the attribution next to the title".

joverlee521:
> I would also consider relocating the mini-icon overlay from the top left of the tile to near the bottom left, to the left of and top-aligned with the title/caption text. I think it will be more noticeable and having the attribution next to the title feels appropriate. (Suggestion only, no strong opinion really.)

The mini-icons were placed over the main image to avoid issues with title widths ([Slack context](https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1718412436850759?thread_ts=1718403195.109039&cid=C7SDVPBLZ))

trvrb (approved):

trvrb:
I definitely like the personalized avatars (for GitHub and for Groups).

joverlee521 (approved):

tsibley (approved):

genehack:
> Could you provide a mockup? I'm confused on what "top-aligned with the title/caption text" means, and what I'm visualizing seems to be opposite of "having the attribution next to the title".

This is an extremely rough mockup; pretend I did the extra work to remove the logo at top left:

Screenshot 2024-07-30 at 12 11 26

May not be worth the extra hassle of dealing with the spacing around the title/caption being center-aligned but I think this makes the logo more apparent; some of the titles because of how colorful they are; the icon gets visually lost.

victorlin:
Thanks! I prefer the current placement, but happy to consider your alternative if others also express preference for it.

Note that one of the requirements of the featured analyses is that [descriptions take up at most 2 lines](https://github.com/nextstrain/nextstrain.org/blob/ad0478b85ec6a2e05288dbcf3730602450537dff/static-site/content/featured-analyses.yaml#L4-L5). The alternative placement would reduce the amount of space available for description text which would require rewording some existing tiles such as:

image

Or we could make the tiles bigger, but I think any bigger might be too big.

PR #971, opened 2024-07-31
Sort group tiles alphabetically

### LINKED ISSUE #953: Reflect on order of group tiles
## Context

The group tiles are displayed like this:

image

@joverlee521 noted in https://github.com/nextstrain/nextstrain.org/pull/949#pullrequestreview-2170101869 that it may seem like we are preferentially displaying the groups in the first row when it's really just a direct reflection of order in [env/production/groups.json](https://github.com/nextstrain/nextstrain.org/blob/20e88d780ef737f21993c23a895a5557d3aa1894/env/production/groups.json#L271) which is roughly the order in which groups are added.

## Potential solutions

1. Add a note saying that the order is mostly arbitrary.
2. Use a less arbitrary ordering. Alphabetical (easy) or most recently updated (maybe not so trivial).

victorlin:
_[preview](https://nextstrain-s-victorlin--3ajzg9.herokuapp.com/groups)_

There's no good reason to keep the order from the groups.json file and I can't imagine any users finding that order useful. Alphabetical could make it easier to find and see related groups (e.g. public/private pairs that share a prefix).

## Related issue(s)

Closes #953

## Checklist

- [x] [Preview](https://nextstrain-s-victorlin--3ajzg9.herokuapp.com/groups) looks good
- [x] Checks pass

### DIFF
--- a/static-site/src/components/Groups/Tiles/index.tsx
+++ b/static-site/src/components/Groups/Tiles/index.tsx
@@ -22,20 +22,22 @@ export const GroupTiles = () => {
 
 
 function createGroupTiles(groups: Group[], colors = [...theme.titleColors]): GroupTile[] {
-  return groups.map((group) => {
-    const groupColor = colors[0]!;
-    colors.push(colors.shift()!);
+  return groups
+    .sort((a, b) => a.name.localeCompare(b.name))
+    .map((group) => {
+      const groupColor = colors[0]!;
+      colors.push(colors.shift()!);
 
-    const tile: GroupTile = {
-      img: "empty.png",
-      url: `/groups/${group.name}`,
-      name: group.name,
-      color: groupColor,
-      private: group.private
-    };
+      const tile: GroupTile = {
+        img: "empty.png",
+        url: `/groups/${group.name}`,
+        name: group.name,
+        color: groupColor,
+        private: group.private
+      };
 
-    return tile;
-  });
+      return tile;
+    });
 }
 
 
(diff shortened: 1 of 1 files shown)


static-site/src/components/Groups/Tiles/index.tsx
```diff
@@ -21,20 +21,24 @@ export const GroupTiles = () => {
...
-    name: group.name,
-    color: groupColor,
-    private: group.private
-  };
-
-  return tile;
-});
+function createGroupTiles(groups: Group[], colors = [...theme.titleColors]): GroupTile[] {
+  return groups
+    .sort((a, b) => a.name.localeCompare(b.name))
```
joverlee521, line 26: Briefly wondered about using [numeric sorting](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare#numeric_sorting), but probably not worth it since options arguments are not supported in all browsers yet.

joverlee521 (approved):

static-site/src/components/Groups/Tiles/index.tsx
victorlin: I also considered [faster alternatives](https://stackoverflow.com/q/14677060) but decided to keep as-is because:

1. `(a<b?-1:(a>b?1:0))` is case-sensitive which we don't want
2. The number of groups is small enough that I don't think the  `localeCompare` slowness matters much

Issue #804, opened 2024-03-19
Groups: Add UI for listing members in Group


joverlee521:
### Context
The API endpoints for listing group members were added in https://github.com/nextstrain/nextstrain.org/pull/581 but we have not surfaced it through the group's splash page. I think this UI has been discussed before but I couldn't find any existing issue for it. 

This was prompted by @trvrb's comment on [Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1710863523001559):

> Just like on GitHub you can see who has view / edit access to a repo you have access to. Currently you can see which groups you belong to as a user, but you can't see which users belong to groups. However, you'd only want this to be visible to people you belong to the group. I don't think immediate priority, but would be nice.

tsibley:
This is part of the larger umbrella of UI for Groups. https://github.com/nextstrain/private/issues/57

PR #973, opened 2024-08-06
Add page for listing Group members 

joverlee521:

The new page is available at `groups/<groupName>/settings/members`. I chose to use the same URL as the API so that we don't have to take up another namespace (e.g. `groups/<groupName>/members`). The main Group page will only link out to the members page if the current user is authorized to list Group members.

The members list only displays the most privileged role for each member to keep the display simple. It assumes the group roles API returns the roles in the order from least to most privileged.

The plan is to make future changes to this page to allow Group owners to edit members here as well. I'm making that a separate change to keep this PR simple. 

## Related issue(s)

Related to #804 

## Checklist

- [x] Checks pass

### DIFF
--- a/static-site/pages/groups/[...groupName].jsx
+++ b/static-site/pages/groups/[...groupName].jsx
@@ -1,12 +1,14 @@
 import dynamic from 'next/dynamic'
 import { useRouter } from 'next/router'
 const IndividualGroupPage = dynamic(() => import("../../src/sections/individual-group-page"), {ssr: false})
+const GroupMembersPage = dynamic(() => import("../../src/sections/group-members-page.tsx"), {ssr: false})
 const GroupSettingsPage = dynamic(() => import("../../src/sections/group-settings-page"), {ssr: false})
 
 
 /**
  * NextJS dynamic routing doesn't allow us to set up routes where we have (e.g.)
  * groups/:groupName/settings -> <GroupSettingsPage>
+ * groups/:groupName/settings/members -> <GroupMembersPage>
  * groups/:groupName[/*] -> <IndividualGroupPage>
  * so we use some JS routing here to serve the right component
  */
@@ -20,6 +22,10 @@ const Index = () => {
 
   if (resourcePath === 'settings') {
     return <GroupSettingsPage groupName={groupName}/>
+  }
+
+  if (resourcePath === 'settings/members') {
+    return <GroupMembersPage groupName={groupName}/>
   }
 
   return (
--- a/static-site/pages/groups/[...groupName].jsx
+++ b/static-site/pages/groups/[...groupName].jsx
@@ -14,19 +14,17 @@ const Index = () => {
   const {query} = useRouter();
   if (!query.groupName) return null; // wait until query ready
 
-
-
-
   /* nextJs dynamic routing will set the part parts _beyond_ "groups/" as query.groupName */
-  if (query.groupName.length===2 && query.groupName[1]==='settings') {
+  const groupName = query.groupName[0]
+  const resourcePath = query.groupName.slice(1).join("/")
+
+  if (resourcePath === 'settings') {
     // param location TODO!!!
-    return <GroupSettingsPage groupName={query.groupName[0]}/>
+    return <GroupSettingsPage groupName={groupName}/>
   }
 
-  const resourcePath = query.groupName.slice(1);
-
   return (
-    <IndividualGroupPage groupName={query.groupName[0]} resourcePath={resourcePath.length ? resourcePath : undefined}/>
+    <IndividualGroupPage groupName={groupName} resourcePath={resourcePath.length ? resourcePath : undefined}/>
   );
 
 }
--- a/static-site/pages/groups/[...groupName].jsx
+++ b/static-site/pages/groups/[...groupName].jsx
@@ -20,18 +20,17 @@ const Index = () => {
   const groupName = query.groupName[0]
   const resourcePath = query.groupName.slice(1).join("/")
 
-  if (resourcePath === 'settings') {
-    return <GroupSettingsPage groupName={groupName}/>
+  switch(resourcePath) {
+    case 'settings':
+      return <GroupSettingsPage groupName={groupName}/>
+    case 'settings/members':
+      return <GroupMembersPage groupName={groupName}/>
+    default:
+      return <IndividualGroupPage
+                groupName={groupName}
+                resourcePath={resourcePath.length ? resourcePath : undefined}
+             />
   }
-
-  if (resourcePath === 'settings/members') {
-    return <GroupMembersPage groupName={groupName}/>
-  }
-
-  return (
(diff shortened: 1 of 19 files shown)


static-site/src/sections/group-members-page.tsx
```diff
@@ -0,0 +1,139 @@
...
+    <CenteredContainer>
+      <div className="row">
+        <div className="col-6">
+          <splashStyles.CenteredFocusParagraph>
+            <strong>Username</strong>
+          </splashStyles.CenteredFocusParagraph>
+        </div>
+        <div className="col-6">
+          <splashStyles.CenteredFocusParagraph>
+            <strong>Roles</strong>
```
victorlin, line 101: Some related suggestions:

1. Make the header singular. Only the most privileged role is shown per person.
2. Make the values singular. This reads better: "trs is an owner"
3. Capitalize the values. This looks more polished.

image

static-site/src/sections/group-members-page.tsx
```diff
@@ -0,0 +1,139 @@
...
+          </splashStyles.CenteredFocusParagraph>
+        </div>
+      </div>
+
+    {sortedMembers.map((member) =>
+      <div className="row" key={member.username}>
+        <div className="col-6">
+          <splashStyles.CenteredFocusParagraph>
+            {member.username}
```
victorlin, line 110: Users can have odd usernames that make it hard to tell who it actually is. It would be cool if we could also list the "Name" attribute here since we have it on many users. This is what it looks like with e0e7b03f140fdda4be8bbc4a16a6e322172df6bf (I didn't think too much about the design, just wanted to see if adding the info was technically feasible):

image

static-site/src/sections/group-members-page.tsx
```diff
@@ -0,0 +1,139 @@
...
+      <BigSpacer/>
+
+      {roles && members
+        ? <MembersTable roles={roles as string[]} members={members as GroupMember[]} />
+        : <splashStyles.H4>Fetching group members...</splashStyles.H4>}
+    </GenericPage>
+  )
+};
+
+const MembersTable = ({ roles, members }: { roles: string[], members: GroupMember[]}) => {
```
victorlin, line 84: Current spacing makes it hard to visually link username to the role which is far to the right, especially for a long list of members. Consider adding horizontal lines between rows or borders on the entire table to add visual guides across the whitespace.

victorlin (approved):
The page loads fine in local testing, so consider everything non-blocking.

My suggestions are largely influenced by GitHub's design on https://github.com/orgs/nextstrain/people

static-site/src/sections/group-members-page.tsx
joverlee521: > Users can have odd usernames that make it hard to tell who it actually is. It would be cool if we could also list the "Name" attribute here since we have it on many users.

Good point! I am just always hesitant to display any personal information because of data privacy issues.

static-site/src/sections/group-members-page.tsx
joverlee521: All good suggestions! Updated in https://github.com/nextstrain/nextstrain.org/pull/973/commits/a2c9dc4094f9c4c0823407073b9615f93cc846de

static-site/src/sections/group-members-page.tsx
joverlee521: Thanks for the push here, I add more styling in https://github.com/nextstrain/nextstrain.org/pull/973/commits/fe814eef7e44f162515a692572c482fc159db4f8

Screenshot 2024-08-09 at 4 30 18 PM

static-site/src/sections/group-members-page.tsx
joverlee521: Thinking on this more, I'm going to punt on this for now. We can add more detailed user info later if we determine that it's necessary.

static-site/pages/groups/[...groupName].jsx
```diff
@@ -1,34 +1,37 @@
...
 
-  const resourcePath = query.groupName.slice(1);
+  if (resourcePath === 'settings/members') {
+    return <GroupMembersPage groupName={groupName}/>
+  }
 
   return (
-    <IndividualGroupPage groupName={query.groupName[0]} resourcePath={resourcePath.length ? resourcePath : undefined}/>
+    <IndividualGroupPage groupName={groupName} resourcePath={resourcePath.length ? resourcePath : undefined}/>
   );
```
Comment (genehack) [inline static-site/pages/groups/[...groupName].jsx:33]: I wonder if it's worth structuring this as either `if/else if/else` or a `switch` on `resourcePath`, to make it more clear that there are three mutually exclusive exits here.

tsibley:
_not a full review, but passing comment_

> The members list only displays the most privileged role for each member to keep the display simple. It assumes the group roles API returns the roles in the order from least to most privileged.

This feels fragile. It makes the display (and eventual management UI) at odds with the realities of our data model, which leaves a space for bugs to lurk. I think it's particularly important to represent the data model accurately when dealing with permissions/ACLs/security/etc.

static-site/pages/groups/[...groupName].jsx
Comment (tsibley) [reply static-site/pages/groups/[...groupName].jsx:33]: +1 for `switch`

Comment (joverlee521) [reply static-site/pages/groups/[...groupName].jsx]: Good suggestion! Updated to `switch` in 195aa4f24c735222f8815bfaa2809eb26b2e5fd7

joverlee521:
> _not a full review, but passing comment_
> 
> > The members list only displays the most privileged role for each member to keep the display simple. It assumes the group roles API returns the roles in the order from least to most privileged.
> 
> This feels fragile. It makes the display (and eventual management UI) at odds with the realities of our data model, which leaves a space for bugs to lurk. I think it's particularly important to represent the data model accurately when dealing with permissions/ACLs/security/etc.

Thanks for flagging @tsibley, I agree this can be fragile if we ever decide to update the Group roles. Is the suggestion to update the UI to display multiple roles or to update the data model to only allow users to be a part of one role? 

I'm personally leaning towards updating the data model to a single role per user per group (unless there was a specific reason to allow a member to have multiple roles). This seems to be inline with the current [provision script](https://github.com/nextstrain/nextstrain.org/blob/c3b3b0382392aba2525c33ad5ea317daf0644392/scripts/provision-group.js#L219-L226) which removes the user from other roles when they get assigned a new role.

tsibley:
Update the UI to display multiple roles. Updating the data model is not trivial/possible, as users can inherently be in multiple Cognito groups and we can't control that. I never saw that as a problem because RBAC is inherently multi-role and we could easily have reasons to use more than the basic three we started with.

joverlee521:
> Update the UI to display multiple roles. 

Updated in https://github.com/nextstrain/nextstrain.org/pull/973/commits/027910d01adba0a43e6433327ed5d358ffd809d1: 

Screenshot 2024-08-14 at 1 25 36 PM

If there's no other feedback, I'll merge this tomorrow.

Issue #975, opened 2024-08-09
Rebuild index job failed


victorlin:
First failed on yesterday's [scheduled run](https://github.com/nextstrain/nextstrain.org/actions/runs/10313465649):

> inventory for core - manifest key: nextstrain-data/config-v1/2024-08-08T01-00Z/manifest.json
> There was an error while fetching the S3 inventory for core. This is fatal.
>  The manifest file for the S3 inventory for bucket nextstrain-data includes more than one inventory file. This situation was not encountered during development, but this is presumably caused by the inventory size exceeding some threshold and being chunked into multiple files. Please check this is indeed the case and, if so, amend the code to parse and join each file.

I noticed when pushing ceb01dc3740f89ca9f6b47db63f069485b0969fd.

## Progress

- [x] #976
- [x] #979
- [ ] #981

victorlin:
I need the canary build to complete in order to complete the work associated with the commit, so I'll investigate.

victorlin:
Comparing the manifest used for the [previous commit on `master`](https://github.com/nextstrain/nextstrain.org/commit/ceb01dc3740f89ca9f6b47db63f069485b0969fd~) to the current one:

```sh
diff \
    <(aws s3 cp s3://nextstrain-inventories/nextstrain-data/config-v1/2024-07-30T01-00Z/manifest.json -) \
    <(aws s3 cp s3://nextstrain-inventories/nextstrain-data/config-v1/2024-08-08T01-00Z/manifest.json -)
```

Indeed there are two inventory files now:

```diff
5c5
<   "creationTimestamp" : "1722301200000",
---
>   "creationTimestamp" : "1723078800000",
9,11c9,15
<     "key" : "nextstrain-data/config-v1/data/ace7ffe4-a9de-4cae-a17e-af5102e61a82.csv.gz",
<     "size" : 25209320,
<     "MD5checksum" : "86f6b86c69c1080abc25dcce78c2f4ed"
---
>     "key" : "nextstrain-data/config-v1/data/027be65e-df06-4962-8233-bd87cf606c08.csv.gz",
>     "size" : 24492873,
>     "MD5checksum" : "159639032cb63deac7abeba189b2af51"
>   }, {
>     "key" : "nextstrain-data/config-v1/data/48f7d241-e89b-4765-8d04-8a446371d0a0.csv.gz",
>     "size" : 945430,
>     "MD5checksum" : "4b9c6e8cd352124ed608515aa2575f8b"
```

[...]

PR #977, opened 2024-08-09
Add rabies

kimandrews:

Add rabies to `data/manifest_core.json` and `src/routing/core.js` so tree can be viewed at `nextstrain.org/rabies`

## Checklist

- [ ] Checks pass

### DIFF
--- a/data/manifest_core.json
+++ b/data/manifest_core.json
@@ -653,6 +653,7 @@
           "default": "sars-cov-2"
         }
       },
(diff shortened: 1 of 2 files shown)


joverlee521 (approved):
LGTM. 
 
The rabies dataset will be missing from the /pathogens listing until https://github.com/nextstrain/nextstrain.org/issues/975 is resolved, but that shouldn't block this PR.

Issue #980, opened 2024-08-10
./scripts/get-resource-index-ref-matrix not producing the correct matrix


joverlee521:
Noticed in https://github.com/nextstrain/nextstrain.org/pull/979#issuecomment-2278891739

I ran the index-resources GH Action workflow on the `victorlin/fix-resource-indexer` branch, expecting it to upload as the new `v3.json`, but it [uploaded as `v2.json` instead](https://github.com/nextstrain/nextstrain.org/actions/runs/10327086205/job/28591692144#step:7:30). 

Looks like something went wrong in `./scripts/get-resource-index-ref-matrix` because the [produced matrix points to the `v2.json`](https://github.com/nextstrain/nextstrain.org/actions/runs/10327086205/job/28591681062#step:4:10).

joverlee521:
I am able to replicate behavior locally:

```

export IS_WORKFLOW_CALL=false
export HEROKU_TOKEN=bogus
export GITHUB_EVENT_NAME='workflow_dispatch'
export GITHUB_REF='victorlin/fix-resource-indexer'
export GITHUB_SHA='fc5eac219950cb7de8ce639a1c37f535a9521ba9'

./scripts/get-resource-index-ref-matrix
[{"ref":"fc5eac219950cb7de8ce639a1c37f535a9521ba9","resource_index":"s3://nextstrain-inventories/resources/v2.json.gz"}]
```

The GITHUB_SHA and the `ref` in the matrix match, so something is going on with getting the resource index.

joverlee521:
OH! It's because the `/env/testing/config.json` was not updated as part of fc5eac219950cb7de8ce639a1c37f535a9521ba9. 
Without explicitly setting `NODE_ENV` or `CONFIG_FILE` envvars, the app defaults to using `/env/testing/config.json` which still pointed to `v2.json`.

joverlee521:
Bumped the resource index version in `/env/testing/config.json` (https://github.com/nextstrain/nextstrain.org/pull/979/commits/da7f8eb110faaf56765a79a0378eea20157f2f3b) and the manually triggered workflow does [produce the expected matrix](https://github.com/nextstrain/nextstrain.org/actions/runs/10327521575/job/28592851906#step:4:10).

joverlee521:
[...]

PR #983, opened 2024-08-13
Always use production config.json in index-resources GH Action

joverlee521:

Ensures that the index-resource workflow always uses the production config as our Heroku apps default to NODE_ENV=production.¹

¹ <https://docs.nextstrain.org/projects/nextstrain-dot-org/en/latest/infrastructure.html#environment-or-config-variables>

## Related issue(s)

Resolves https://github.com/nextstrain/nextstrain.org/issues/980

## Checklist

- [ ] Checks pass

### DIFF
--- a/docs/resource-collection.rst
+++ b/docs/resource-collection.rst
@@ -22,7 +22,8 @@ The nextstrain.org testing & production configs currently set this to
 ``s3://nextstrain-inventories/resources/v<revision_number>.json.gz``.
 
 If you make any updates that changes the structure or the contents of the resource
-index JSON, then bump the ``<revision_number>`` within the configs (``env/production/config.json``)
+index JSON, then bump the ``<revision_number>`` within the configs
+(``env/testing/config.json`` and ``env/production/config.json``)
 so that any uploads to S3 does not disrupt the production server.
 
 These updates include changes to:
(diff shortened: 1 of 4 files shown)


docs/resource-collection.rst
```diff
@@ -22,7 +22,8 @@ The nextstrain.org testing & production configs currently set this to
 ``s3://nextstrain-inventories/resources/v<revision_number>.json.gz``.
 
 If you make any updates that changes the structure or the contents of the resource
-index JSON, then bump the ``<revision_number>`` within the configs (``env/production/config.json``)
+index JSON, then bump the ``<revision_number>`` within the configs
+(``env/testing/config.json`` and ``env/production/config.json``)
```
jameshadfield, line 26: Is there a situation where the `<revision_number>` should differ in these files?

joverlee521: Hmm, `env/testing/config.json` is only used for local testing right now so I don't think so? 
(Maybe if we ever set up a completely separate testing instance on AWS there would be reason for them to differ?)

jameshadfield: Ok - in this case [I continue to think](https://github.com/nextstrain/nextstrain.org/issues/980#issuecomment-2282869865) that we should add a check in CI that they are the same. Otherwise it's too easy to overlook the docs, get things working locally, and then overwrite the production version.

joverlee521: Good suggestion! I added in https://github.com/nextstrain/nextstrain.org/pull/983/commits/9d45dd059d4712d997c9a57229294999ae023565

victorlin (approved):

PR #987, opened 2024-08-19
[provision group] Check emails / usernames against existing users

jameshadfield:
The provision-users command requires us to provide both a username and email for each user we want to add. A common use case is being given a (long) list of emails and wanting to check if a user already exists for each email, which is exactly what this script does.

- [ ] Checks pass

### DIFF
--- a/scripts/provision-group-check-users.js
+++ b/scripts/provision-group-check-users.js
@@ -1,0 +1,130 @@
+#!/usr/bin/env node
+import { ArgumentParser } from 'argparse';
+import {
+    CognitoIdentityProviderClient,
+    ListUsersCommand,
+  } from '@aws-sdk/client-cognito-identity-provider';
+import fs from 'fs';
+import yaml from 'js-yaml';
+import process from 'process';
+import { COGNITO_USER_POOL_ID } from '../src/config.js';
+import { reportUnhandledRejectionsAtExit } from '../src/utils/scripts.js';
+
+const REGION = COGNITO_USER_POOL_ID.split("_")[0];
+const cognito = new CognitoIdentityProviderClient({ region: REGION });
+
+function parseArgs() {
+    const argparser = new ArgumentParser({
+      description: `
+        A helper script to check provided emails against existing cognito users.
+        If a username is provided that is also checked against existing users.
+        Messages are printed to STDOUT.
+        Set 'CONFIG_FILE=env/production/config.json' to use the production cognito
+        user pool (default: pool defined in env/testing/config.json)
+      `,
+    });  
+    argparser.addArgument("--members", {
+      dest: "membersFile",
+      metavar: "<file.yaml>",
+      required: true,
+      help: `
+        A YAML file intended for 'provision-group.js'. Each entry must contain a 'email'
+        key and optionally a 'username' key.
+      `,
+    });
+  
+    return argparser.parseArgs();
+  }
+
+
+async function main({membersFile}) {
+  const members = readMembersFile(membersFile)
+  const {usersByEmail, usersByUsername} = await cognitoUsers()
+  queryUsernames(members, usersByEmail, usersByUsername)
+}
+
+
+function readMembersFile(file) {
+  const members = yaml.load(fs.readFileSync(file));
+
+  const validationErrors = !Array.isArray(members)
+    ? ["Not an array"]
+    : members.filter(m => !m.email)
+      .map(m => `Email missing for member ${JSON.stringify(m)}`)
+    ;
+
+  if (validationErrors.length) {
+    const msg = validationErrors.map((err, i) => ` ${i+1}. ${err}`).join("\n");
+    const s = validationErrors.length === 1 ? "" : "s";
+    throw new Error(`Members file contains ${validationErrors.length} error${s}:\n${msg}`);
+  }
+
+  return members;
+}
+
+function getEmailFromUser(user) {
+  return user.Attributes.filter(({Name}) => Name==='email')[0].Value;
+}
+
+async function cognitoUsers() {
+  // https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/client/cognito-identity-provider/command/ListUsersCommand/
+  const params = {
+    AttributesToGet: ['email'], // all users must have this else cmd fails
+    UserPoolId: COGNITO_USER_POOL_ID,
+    // Limit: 60, // default: 60
+  }
+  let {Users, PaginationToken} = await cognito.send(new ListUsersCommand(params));
+  while (PaginationToken) {
+    const data = await cognito.send(new ListUsersCommand({...params, PaginationToken}));
+    Users = [...Users, ...data.Users]
+    PaginationToken = data.PaginationToken
+  }
+  // there may be duplicate emails (different users)
+  const [usersByEmail, usersByUsername] = [{}, {}]
+  for (const user of Users) {
+    const email = getEmailFromUser(user);
+    Object.hasOwn(usersByEmail, email) ? usersByEmail[email].push(user) : (usersByEmail[email]=[user])
+    usersByUsername[user.Username] = user; // Username is unique (within a user pool)
+  }
+
+  return {usersByEmail, usersByUsername};
+}
+
+
+function queryUsernames(members, usersByEmail, usersByUsername) {
+  for (const member of members) {
+    if (Object.hasOwn(usersByEmail, member.email)) {
+      const existingMsg = `Email ${member.email} exists with username(s) ${usersByEmail[member.email].map((u) => `'${u.Username}'`).join(', ')}`
+      if (member.username) {
+        const usernameAssociatedWithEmail = usersByEmail[member.email].filter((u) => u.Username===member.username).length>0
+        if (usernameAssociatedWithEmail) {
+          console.log(`${existingMsg} ALL GOOD.`)
+        } else if (Object.hasOwn(usersByUsername, member.username)) {
+          console.log(`${existingMsg} but the username '${member.username}' is already associated with '${getEmailFromUser(usersByUsername[member.username])}'`)
+        } else {
+          console.log(`${existingMsg} and you are asking for the new username '${member.username}' to be created. ALL GOOD.`)
+        }
+      } else {
+        console.log(existingMsg)
+      }
+    } else {  // new email (not associated with any cognito user)
+      const existingMsg = `Email ${member.email} doesn't yet exist`
+      if (member.username) {
+        if (Object.hasOwn(usersByUsername, member.username)) {
+          console.log(`${existingMsg} but the username '${member.username}' is already associated with '${getEmailFromUser(usersByUsername[member.username])}'`)
+        } else {
+          console.log(`${existingMsg} and neither does the username '${member.username}. ALL GOOD.'`)
+        }
+      } else {
+        console.log(`${existingMsg} (and you haven't specified a username)`)
+      }
+    }
+  }
+}
+
+reportUnhandledRejectionsAtExit();
+main(parseArgs())
+  .catch((error) => {
+    process.exitCode = 1;
+    console.error(error)
+  });


victorlin:
This seems useful! Instead of a separate script, how about adding it as a validation step within the `provision-group` script? Example: 434dff708dd82649d04dd76aea45ecb661ab29ae

```
$ ./scripts/provision-group test --members new-members.yaml

PROGRAMMING ERROR: Caught unhandled rejection of 1 promise:
  Promise {
    <rejected> Error: Email another@example.com doesn't yet exist but the username 'victorlin-test' is already associated with 'existing@example.com'
        at queryUsernames (file:///Users/vlin/Dropbox/repos/nextstrain/nextstrain.org/scripts/provision-group.js:487:17)
        at readMembersFile (file:///Users/vlin/Dropbox/repos/nextstrain/nextstrain.org/scripts/provision-group.js:123:3)
        at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
        at async main (file:///Users/vlin/Dropbox/repos/nextstrain/nextstrain.org/scripts/provision-group.js:69:7)
  }
```

jameshadfield:
> Instead of a separate script, how about adding it as a validation step within the provision-group script? 

There's a few reasons I chose not to go down this road. My script needs to list all the cognito users (as you can't query by email, AFAIK), which is unnecessary overhead for most use cases of `provision-group`. Secondly, I allow the username to be left out, whereas `provision-group` enforces it (could be fixed). Thirdly, I wanted to use the script with confidence no changes would be made on the cognito side; I sketched out adding `--dry-run` to `provision-group` (which I think would be a very good thing) but ultimately found it easier to make a separate script. Finally I deemed the bar for addition to `provision-group` to be higher than a script like this (ensuring all exceptions are handled etc), as their purposes and importance are quite different. If you want to take this direction on then by all means please do - there are a few emails with multiple usernames in the production pool, which I don't think is ideal.

scripts/provision-group-check-users.js
victorlin, line 1: > there are a few emails with multiple usernames in the production pool, which I don't think is ideal.

This seems intentional. Skimming through the code, everything is keyed on username (also apparent on https://nextstrain.org/login). Many services allow multiple accounts for a single email. This can be useful for things such as one person having separate accounts for different levels of access.

victorlin (commented):
Thanks for the explanation. I only checked against the testing user pool so I didn't realize the overhead on the production user pool. I understand your decision better now and feel less strongly about adding to `provision-group`.

Thread for tangential discussion below.

scripts/provision-group-check-users.js
jameshadfield: Yeah - for sure cognito is keyed off username, and by extension nextstrain is as well.

> This can be useful for things such as one person having separate accounts for different levels of access

I guess so? Outside of testing purposes I don't quite see the point of one person maintaining different access levels for the same group, but I'm sure there are use cases.

In the few examples I've seen (and thus anecdotal) I think we've created a username/email combo for one group, then for a subsequent group been given a list of emails and used the `provision-groups` script to add a new username for the same email. From the users perspective that's not ideal (need to login with different usernames to see different groups), but for the users in question I believe their usage is very low at the moment.

genehack (approved):

tsibley:
> My script needs to list all the cognito users (as you can't query by email, AFAIK),

You can absolutely search by email, e.g.
https://docs.aws.amazon.com/cognito/latest/developerguide/how-to-manage-user-accounts.html#manage-user-accounts-searching-user-attributes

scripts/provision-group-check-users.js
```diff
@@ -0,0 +1,130 @@
...
+      } else {
+        console.log(existingMsg)
+      }
+    } else {  // new email (not associated with any cognito user)
+      const existingMsg = `Email ${member.email} doesn't yet exist`
+      if (member.username) {
+        if (Object.hasOwn(usersByUsername, member.username)) {
+          console.log(`${existingMsg} but the username '${member.username}' is already associated with '${getEmailFromUser(usersByUsername[member.username])}'`)
+        } else {
+          console.log(`${existingMsg} and neither does the username '${member.username}. ALL GOOD.'`)
```
victorlin, line 116:
```suggestion
          console.log(`${existingMsg} and neither does the username '${member.username}'. ALL GOOD.`)
```

victorlin (approved):
I just used this for 88c309241909a80d2ebdda657676a826d5e1e480 – very useful! I'm good for this to be merged pending one small change below.

It would be nice to run this automatically as part of validation in `provision-group`. But given the long-term plan to allow self-service, maybe no need to invest much additional effort here.

scripts/provision-group-check-users.js
```diff
@@ -0,0 +1,130 @@
+#!/usr/bin/env node
```
victorlin, line 1: Shebang indicates that this should be executable but it isn't:

```console
$ ./scripts/provision-group-check-users.js members.yaml       
zsh: permission denied: ./scripts/provision-group-check-users.js
```

Fix:

```sh
chmod a+x scripts/provision-group-check-users.js
```

scripts/provision-group-check-users.js
victorlin: [Example use case](https://support.nextstrain.org/agent/nextstrain/nextstrain/tickets/details/668036000007930037) for multiple usernames under the same email.

I don't doubt that some of the existing many-to-one username/email instances are not ideal, so maybe it'd be good to print a warning such as

```
WARNING: Email <email> exists with username(s) <name> and you are asking for the new username <name> to be created. Confirm that this is intentional.
```

In the future when user creation is self-service, it would be good to add an extra confirmation step to this scenario.

victorlin:
Bump: It's been helpful to run this while handling Groups requests. Currently I switch to the PR branch just to run it.

victorlin:
Bump: I'm now doing this

```sh
git checkout james/groups-helper-script scripts/provision-group-check-users.js
chmod a+x scripts/provision-group-check-users.js
AWS_PROFILE=victorlin-admin-read-only ./scripts/provision-group-check-users.js --members members.yaml
```

scripts/provision-group-check-users.js
jameshadfield: Fixed - thanks!

jameshadfield:
I moved this forward today with three major improvements:

We now check the roles we're attempting to add against an existing users roles, which helps correctly identify no-op situations and (more importantly) avoids the previous issue where a prospective change in user role wouldn't be considered and so the script would say "no changes needed here" (or similar)

The verbosity is also increased. I find this really helpful when I'm only working on this part of the ecosystem from time to time.

Finally we now check emails against other emails in a case-insensitive way which closes #1258 

P.S. If you run this against the testing pool you'll see I've added a "duplicate" email with different casing. I'll remove this when I merge this PR.

scripts/provision-group-check-users.js
```diff
@@ -0,0 +1,319 @@
...
+ * way we have it) is case-sensitive, but emails are generally case-insensitive.
+ * Ultimately we want to avoid situations where we provision multiple emails
+ * which differ only in their casing.
+ * @param {object} emailDb
+ * @param {string} email
+ * @returns {object} exact: boolean (whether the email is an exact match),
+ *  candidates: string[] cognito URLs which are identical to the provided email
+ *  except for potential different casing
+ */
+function searchEmails(emailDb, email) {
```
victorlin: If the warning that's given as a byproduct of creating `emailDb` is not needed (https://github.com/nextstrain/nextstrain.org/pull/987#discussion_r2520458024), the code could be simplified to remove the entire `emailDb` intermediate data structure and instead do case-insensitive searches on-the-fly. Example: a1a599f8c5ad8cf87258e086ef39d94405d73545

scripts/provision-group-check-users.js
```diff
@@ -0,0 +1,319 @@
...
+function summariseEmails(usersByEmail) {
+  const db = {}
+  for (const email of Object.keys(usersByEmail)) {
+    const key = email.toUpperCase();
+    Object.hasOwn(db, key) ? db[key].push(email) : (db[key]=[email]);
+  }
+  // eslint-disable-next-line no-unused-vars
+  for (const [_key, emails] of Object.entries(db)) {
+    if (emails.length>1) {
+      console.log(`WARNING: Multiple emails with different casings present in pool: ${emails.join(', ')}`)
```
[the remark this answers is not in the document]
victorlin: I checked – we don't have any duplicates in the production user pool at the moment. With the new warning when adding them, duplicates that make it into the user pool would likely be intentional, leading to noisy warnings whenever provisioning members for other groups. While I don't see the benefit, I'm fine with keeping it if you would find it beneficial.

The other motivation, though not as important, is simpler code as sketched out in https://github.com/nextstrain/nextstrain.org/pull/987#discussion_r2520540285.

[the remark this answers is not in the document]
jameshadfield: I went a different direction and this ended up being dropped, but

> duplicates that make it into the user pool would likely be intentional

I don't think we should ever have duplicates-which-differ-by-case-only, either for email or username. I can see the value in multiple users with the same email (e.g. as we do this in our testing pool), and so we don't loudly warn against this anymore.

scripts/provision-group-check-users.js
```diff
@@ -0,0 +1,319 @@
...
+  }
+  return db;
+}
+
+/**
+ * Match a provided case-sensitive *email* against the *emailDb* which is
+ * indexed by all-caps emails. This is necessary because cognito (at least the
+ * way we have it) is case-sensitive, but emails are generally case-insensitive.
+ * Ultimately we want to avoid situations where we provision multiple emails
+ * which differ only in their casing.
```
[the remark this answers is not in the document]
jameshadfield: Rewrote the script more centered around usernames now. I've tested it reasonably extensively and am happy with the situation. The outputs somewhat verbose, but given how often we actually add users I think this is helpful rather than annoying. @victorlin can you take another look?

scripts/provision-group-check-users.js
```diff
@@ -0,0 +1,324 @@
...
+    argparser.addArgument("groupName", {metavar: "<name>", help: "Name of the Nextstrain Group"});
+    argparser.addArgument("--members", {
+      dest: "membersFile",
+      metavar: "<file.yaml>",
+      required: true,
+      help: `
+        A YAML file describing the members to add to the Group.
+        The file must be an array of objects (i.e.  dict/map/hash).
+        Each object must contain an "email" key which is case-sensitive.
+        The "username" key is optional for this script, but is required when you actually create provision the group membership.
```
victorlin: Why allow "username" to be optional in this script? I think it should be required to match the other script, and also simplify the code by avoiding the need to handle its absence.

scripts/provision-group-check-users.js
victorlin: (non-blocking) Another follow-up suggested simplification: 9332153280e419dfe7c512d8843e984eaf8ca656

scripts/provision-group-check-users.js
```diff
@@ -0,0 +1,324 @@
...
+
+
+/**
+ * Given a proposed username/email pairing, cross-reference against the existing
+ * cognito pool. We primarily search by username, as that's the index of the pool.
+ *
+ */
+async function crossReferenceUsernameAndEmail(group, member, usersByEmail, usersByUsername) {
+  console.log()
+  console.log(`Membership request for user "${member.username}" (${member.email}). Proposed role: ${member.role}`)
```
victorlin: Generally on board with the level of verbosity here. Suggested rewording from:

```
Membership request for user "testUser2" (test@example.com). Proposed role: owners
        No matching usernames in the pool (case insensitive) but the email did match.
        There is one email which matches exactly. Associated users:
                testUser (existing group role: owners)
        ACTION: Use a username from above and cross-reference with existing roles.
```

to:

```
Username: 'testUser2'; email: 'test@example.com'; role: 'test/owners'
        ✔️ Username is unique
        ⚠ Email matches 1 other user:
                Username: 'testUser'; email: 'test@example.com'; role: 'test/owners'
        ↳ ACTION: Use a username from above and cross-reference with existing roles.
```

I don't think it's necessary to have separate code paths to distinguish between exact and not-exact email matches. The suggested rewording reflects this.

scripts/provision-group-check-users.js
```diff
@@ -0,0 +1,324 @@
...
+  if (userNameMatch && userNameMatch.candidates.length>1) {
+    console.log(`\tThere are multiple matching usernames which differ by case:`)
+    for (const cognitoUsername of userNameMatch.candidates) {
+      const exactMsg = cognitoUsername===member.username ? '(exact match)' : '(case differs)';
+      const associatedEmail = getEmailFromUser(usersByUsername[cognitoUsername]);
+      const emailMsg = cognitoUsername===member.email ?
+        '(exact match)' :
+        cognitoUsername.toUpperCase()===member.email.toUpperCase() ? 
+          '(case differs)' :
+          '(not a match with the provided email)';
```
victorlin: Bug: username compared against email

```suggestion
      const emailMsg = associatedEmail===member.email ?
        '(exact match)' :
        associatedEmail.toUpperCase()===member.email.toUpperCase() ? 
          '(case differs)' :
          '(not a match with the provided email)';
```

PR #988, opened 2024-08-21
group-members-page: Replace `toSorted()` with shallow copy + `sort()`

joverlee521:

I didn't realize that `toSorted` is a relatively new method that is not supported in older browser versions.¹

Thanks to @tsibley for using an old version of FireFox and catching this.

¹ <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted#browser_compatibility>

## Related issue(s)

Follow up to https://github.com/nextstrain/nextstrain.org/pull/973

## Checklist

- [x] Checks pass

### DIFF
--- a/static-site/src/sections/group-members-page.tsx
+++ b/static-site/src/sections/group-members-page.tsx
@@ -100,7 +100,7 @@ const MembersTableContainer = styled.div`
 `;
 
 const MembersTable = ({ members }: { members: GroupMember[]}) => {
-  const sortedMembers = members.toSorted((a, b) => a.username.localeCompare(b.username));
+  const sortedMembers = [...members].sort((a, b) => a.username.localeCompare(b.username));
   function prettifyRoles(memberRoles: string[]) {
     // Prettify the role names by making them singular and capitalized
     return memberRoles.map((roleName) => startCase(roleName.replace(/s$/, ''))).join(", ");


static-site/src/sections/group-members-page.tsx
```diff
@@ -100,7 +100,8 @@ const MembersTableContainer = styled.div`
 `;
 
 const MembersTable = ({ members }: { members: GroupMember[]}) => {
-  const sortedMembers = members.toSorted((a, b) => a.username.localeCompare(b.username));
+  const sortedMembers = [...members];
+  sortedMembers.sort((a, b) => a.username.localeCompare(b.username));
```
tsibley, line 104: _non-blocking, purely style_

```suggestion
  const sortedMembers = [...members].sort((a, b) => a.username.localeCompare(b.username));
```

tsibley (approved):
WFM locally. Thanks for saving me the proper bug report/fix myself! :-)

(I really need to upgrade the base OS on this computer so I can get more than security updates to Firefox…)

static-site/src/sections/group-members-page.tsx
joverlee521: Ah, right Javascript's `sort()` returns the reference to the array! Mixing up with Python's `sort()` which returns `None`

joverlee521: Updated in [force-push](https://github.com/nextstrain/nextstrain.org/compare/572934557853e33b81d28dab5557ef783f13982c..c7d9a0b2aa9c226f7b5096a26b96eb7334125a4e)

PR #989, opened 2024-08-21
Add IBRB's mpox clade-I public build

jameshadfield:
The underlying dataset has been generated by EKL at INRB (DRC) and we've been given permission to advertise this public build. Adding as the first tile as this was declared a PHEIC a ~week ago [1]

[1] <https://www.who.int/news/item/14-08-2024-who-director-general-declares-mpox-outbreak-a-public-health-emergency-of-international-concern>

- [ ] Checks pass


static-site/content/featured-analyses.yaml
```diff
@@ -21,7 +26,7 @@
   url: /avian-flu/h5n1/ha/2y
 
 - name: Mpox
-  description: Evolution and spread mpox virus
+  description: Evolution and spread of mpox virus (clade-IIb)
```
miparedes, line 29: slightly pedantic but this description is a bit misleading because the dropdown options in this build also include clade I as well as all-clade views

Not sure what the best solution would be. Maybe "Mpox DRC Clade-I" as the name for the new one?

miparedes:
I looked at the deployment and tried to click on the new build and it said that `The dataset "nextstrain.org/groups/inrb-mpox/clade-I" doesn't exist.` not sure if that's just a deployment issue or not but wanted to highlight that.

static-site/content/featured-analyses.yaml
jameshadfield: Good call - have dropped the " (clade-IIb)" and changed the new tile's title

jameshadfield:
> not sure if that's just a deployment issue or not but wanted to highlight that.

Thanks for highlighting! It's an unfortunate limitation of how we run the testing apps (they only have permissions to interact with the blab, test & test-private groups). Once I merge and it's running on next.nextstrain.org it'll work as intended.

PR #990, opened 2024-08-22
Add RSS feed [#932]

### LINKED ISSUE #932: The blog should have an RSS feed
Having an RSS feed is the best way to surface content for periodically updated things like a weblog — yes, posting on social media helps, but an RSS feed is standard. 

Blocks #1001

genehack:

Add RSS generation script and hook it into build [#932]

Note: Atom and JSON feeds will be added to this once any review of the overall metadata is complete.

* Add function to `blogPosts.js` to read blog post and generate feed with the 10 most recent posts into the `public` directory
* Add new top-level script that calls the new `generateBlogFeeds` function
* Add package.json:scripts target to run this script
* Hook that package.json:scripts target into build.sh
* Ignore the new build-time generated file

## Related issue(s)

Closes #932 

## Checklist

- [x] Checks pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,6 +100,7 @@
     "eslint": "^8.23.0",
     "eslint-config-next": "^14.1.4",
     "eslint-plugin-jest": "^28.6.0",
+    "feed": "^4.2.2",
     "http-proxy": "^1.18.1",
     "http-proxy-middleware": "^1.3.1",
     "jest": "^27.5.1",
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
   },
   "scripts": {
     "build": "./build.sh",
+    "build:feeds": "node scripts/generate-blog-feeds.js",
     "lint": "npm run lint:server",
     "lint:server": "DEBUG=eslint:cli-engine npx eslint --ext .js,.jsx .",
     "lint:static-site": "cd static-site && DEBUG=eslint:cli-engine npx eslint --ext .js,.jsx,.ts,.tsx .",
--- a/static-site/data/BaseConfig.js
+++ b/static-site/data/BaseConfig.js
@@ -1,0 +1,32 @@
+// This file contains basic configuration information that is used in
+// non-React contexts, such that putting it into `SiteConfig.jsx`
+// would not work. Everything that is exported from this file will
+// be re-exported from `SiteConfig.jsx`.
+
+export const siteTitle = "Nextstrain";
+
+// Alternative site title for SEO.
+export const siteTitleAlt = "Real-time tracking of pathogen evolution";
+
+// Logo used for SEO
+export const siteLogo = "/nextstrain-logo-small.png";
+
+export const siteUrl = "https://nextstrain.org";
+
+// Website description used for RSS feeds/meta description tag.
+export const siteDescription = "Real-time tracking of pathogen evolution";
+
+export const groupsApp = false;
+
+export const groupsTitle = "Scalable Sharing with Nextstrain Groups";
+
+export const blogUrl = `${siteUrl}/blog`;
+
+export const blogDescription = "Updates from the Nextstrain core team";
+
+export const blogFeeds = {
+  atom: `atom.xml`,
+  json: `feed.json`,
+  rss2: `rss2.xml`,
+}
+export const blogFeedUrls = Object.fromEntries( Object.entries(blogFeeds).map(([k,v]) => [k, `${blogUrl}/${v}`]));
--- a/static-site/src/components/layout.jsx
+++ b/static-site/src/components/layout.jsx
@@ -1,7 +1,7 @@
 import React from "react";
 import Helmet from "react-helmet";
 import styled, {ThemeProvider} from "styled-components";
-import { siteTitle, siteDescription } from "../../data/SiteConfig";
+import { blogFeedUrls, groupsApp, siteTitle, siteDescription } from "../../data/SiteConfig";
 import {theme} from '../layouts/theme';
 
 /**
@@ -22,7 +22,23 @@ export default class MainLayout extends React.Component {
         <Helmet>
           <title>{`${siteTitle}`}</title>
           <meta name="description" content={siteDescription} />
-          <link rel="me" href="https://mstdn.science/@nextstrain" />
+          {
+            // react-helmet doesn't support React fragments, so we have to do this
+            // in a maximally silly way; see https://github.com/nfl/react-helmet/issues/342
+            // for details
+          }
+          {!groupsApp &&
+            <link rel="me" href="https://mstdn.science/@nextstrain" />
+          }
+          {!groupsApp &&
+            <link href={`${blogFeedUrls.atom}`} rel="alternate" title="Atom feed for nextstrain.org/blog" type="application/atom+xml" />
+          }
+          {!groupsApp &&
+            <link href={`${blogFeedUrls.json}`} rel="alternate" title="JSON feed for nextstrain.org/blog" type="application/json" />
+          }
+          {!groupsApp &&
+            <link href={`${blogFeedUrls.rss2}`} rel="alternate" title="RSS2 feed for nextstrain.org/blog" type="application/rss+xml" />
+          }
         </Helmet>
         <ThemeProvider theme={theme}>
           <GlobalStyles>
--- a/static-site/src/util/blogFeeds.js
+++ b/static-site/src/util/blogFeeds.js
@@ -1,0 +1,75 @@
+import fs from 'fs';
+import path from 'path';
+import { fileURLToPath } from 'url';
+import { Feed } from 'feed';
+
+import { getBlogPosts} from "./blogPosts.js";
+import { parseMarkdown } from "./parseMarkdown.js";
+import { blogDescription, blogFeeds, blogFeedUrls, blogUrl, siteLogo, siteTitle, siteUrl  } from '../../data/BaseConfig.js';
+
+const NUMBER_OF_POSTS_IN_FEED = 10;
+
+/**
+ * Writes RSS2, Atom, and JSON feeds into the `public/blog` directory.
+ */
+export async function generateBlogFeeds() {
+  const feed = new Feed({
+    title: `${siteTitle} Blog`,
+    description: blogDescription,
+    id: blogUrl,
+    link: blogUrl,
+    image: `${siteUrl}${siteLogo}`,
+    favicon: `${siteUrl}/favicon.ico`,
+    copyright: "Copyright Trevor Bedford and Richard Neher.",
+    updated: new Date(),
+    feedLinks: blogFeedUrls,
+    author: {
+      name: "The Nextstrain Team",
+      email: "hello@nextstrain.org",
+      link: siteUrl,
+    },
+  });
+
+  const posts = getBlogPosts().slice(0, NUMBER_OF_POSTS_IN_FEED);
+
+  for (const post of posts) {
+    if (post) { // getBlogPosts() _might_ return `false` list members
+      try {
+        const content = parseMarkdown(post.mdstring);
+        const url = `${blogUrl}/${post.blogUrlName}`;
+
+        feed.addItem({
+          author: [ { name: post.author } ],
+          content: content,
+          date: new Date(post.date),
+          id: url,
+          link: url,
+          title: post.title,
+        });
+      } catch (error) {
+        console.error(`Skipping post entitled "${post.title}" due to Markdown parsing error:\n${error}`);
+      }
+    }
+  }
+
+  const __dirname = path.dirname(fileURLToPath(import.meta.url));
+  const feedDir = `${__dirname}/../../public/blog`;
+  const feeds = [
+    {
+      file: `${feedDir}/${blogFeeds.atom}`,
+      data: feed.atom1,
+    },
+    {
+      file: `${feedDir}/${blogFeeds.json}`,
+      data: feed.json1,
+    },
+    {
+      file: `${feedDir}/${blogFeeds.rss2}`,
+      data: feed.rss2,
+    },
+  ];
+
+  for (const feed of feeds) {
+    fs.writeFileSync(feed.file, feed.data());
+  }
+}
--- a/static-site/src/util/blogPosts.js
+++ b/static-site/src/util/blogPosts.js
@@ -2,7 +2,9 @@ import fs from 'fs';
 import path from 'path';
 import { fileURLToPath } from 'url';
 import matter from 'gray-matter';
-import { startCase } from "lodash"
+import lodash from "lodash";
+
+const { startCase } = lodash;
 
 /**
  * Scans the ./static-site/content/blog directory for .md files
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,8 @@ static-site/.next
 
 # TypeScript
 static-site/tsconfig.tsbuildinfo
+
+# Blog feeds
+static-site/public/blog/atom.xml
+static-site/public/blog/feed.json
+static-site/public/blog/rss2.xml
--- a/build.sh
+++ b/build.sh
@@ -33,6 +33,7 @@ main() {
 
 build-static() {
     echo "Building the static Next.JS app (pages defined in static-site/pages, assets written to static-site/.next)"
+    npm run build:feeds
     ./node_modules/.bin/next build static-site
 }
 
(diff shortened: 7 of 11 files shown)


genehack:
This is my first time mucking with the build/infra side of nextstrain.org, so I'm marking this as draft and tagging @jameshadfield and @victorlin for any feedback.

There are a couple additional (fairly trivial) changes needed to get Atom and JSON feeds working, which I will probably land tomorrow, but I wanted to get this pushed so folks could give me initial feedback. There are a couple of FIXMEs noted, and I'm open to any and all feedback about what's being included in the feed.

static-site/src/util/blogPosts.js
```diff
@@ -2,7 +2,9 @@ import fs from 'fs';
 import path from 'path';
 import { fileURLToPath } from 'url';
 import matter from 'gray-matter';
-import { startCase } from "lodash"
+import lodash from 'lodash';
+
+const { startCase } = lodash;
```
jameshadfield, line 8: `import { startCase } from "lodash"` often allows tree-shaking (cf. `import lodash from 'lodash';`). I can't remember the details of what nextjs is doing here / what we've asked it to do. Might be worth [analysing the bundle](https://nextjs.org/docs/pages/building-your-application/optimizing/package-bundling#analyzing-javascript-bundles) here to check.

jameshadfield (commented):
`npm run build:feeds` ran without trouble for me locally. I don't have experience with RSS spec, so no comments from me on that. How are feeds normally discovered? I.e. should we not add an icon on the blog somewhere linking to the (newly added) `/feed.xml` URL? Finally my presumption of the URL would be `/blog/feed`.

static-site/src/util/blogPosts.js
genehack: > `import { startCase } from "lodash"` often allows tree-shaking (cf. `import lodash from 'lodash';`). I can't remember the details of what nextjs is doing here / what we've asked it to do. 

When this file contains `import { startCase } from 'lodash';`, running the feed generation script produces this error:

```javascript
> npm run build:feeds                                                                                                                 v20.9.0

> build:feeds
> node scripts/generate-blog-feeds.js

file:///Users/genehack/nextstrain/nextstrain.org/static-site/src/util/blogPosts.js:9
import { startCase } from 'lodash';
         ^^^^^^^^^
SyntaxError: Named export 'startCase' not found. The requested module 'lodash' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'lodash';
const { startCase } = pkg;

    at ModuleJob._instantiate (node:internal/modules/esm/module_job:131:21)
    at async ModuleJob.run (node:internal/modules/esm/module_job:213:5)
    at async ModuleLoader.import (node:internal/modules/esm/loader:316:24)
    at async loadESM (node:internal/process/esm_loader:34:7)
    at async handleMainPromise (node:internal/modules/run_main:66:12)

Node.js v20.9.0
```

(Trying to evaluate the `blogPosts.js` file directly, in the `static-site` directory, generates the exact same error.)

I'm not sure why this error doesn't happen when the file is included in the 2 JSX files that consume it. 

> Might be worth [analysing the bundle](https://nextjs.org/docs/pages/building-your-application/optimizing/package-bundling#analyzing-javascript-bundles) here to check.

The only place this file gets included in the site is in 2 JSX files that are used to produce static site files — I don't think the `lodash` Javascript would need to end up in the bundle? But maybe I'm confused about how the "static" file generation works in the production build.

static-site/src/util/blogPosts.js
```diff
@@ -36,6 +39,59 @@ export function getBlogPosts() {
...
+
+  // FIXME is there a config file where this should go?
+  const FEED_POSTS = 10;
+
+  const posts = getBlogPosts().slice(0, FEED_POSTS);
+
+  if (posts && posts.length > 0) {
+    for (const post of posts) {
+      if (post) { // make the type checker happier
+        const url = `${siteURL}/${post.blogUrlName}`;
```
victorlin, line 75: Did you see linting errors? I removed the conditionals and `npm run lint:static-site` still runs fine.

```js
for (const post of posts) {
const url = `${siteURL}/${post.blogUrlName}`;

feed.addItem({
    title: post.title,
    id: url,
    link: url,
    author: post.author,
    date: new Date(post.date),
});
}
```

genehack:
> `npm run build:feeds` ran without trouble for me locally. I don't have experience with RSS spec, so no comments from me on that.

W3C has a [feed validator](https://validator.w3.org/feed/#validate_by_input), that you can copypasta a file into. Once I add the Atom and JSON feed versions, I will validate each of them. (A [JSON Feed validator](https://validator.jsonfeed.org/) also exists.)

> How are feeds normally discovered? I.e. should we not add an icon on the blog somewhere linking to the (newly added) `/feed.xml` URL?

We could add an icon (maybe next to the `BLOG` title in the sidebar?), but generally, feed autodiscovery works via including something like this in the `<head>` of (at least) the index page -- but I would advocate for adding it to `https://nexstrain.org` as well as all URLs that start with `https://nextstrain.org/blog`:

```html
<link href="https://genehack.blog/feed.rss2" rel="alternate" title="RSS2 feed for genehack.blog" type="application/rss+xml">
<link href="https://genehack.blog/atom.xml" rel="alternate" title="Atom feed for genehack.blog  type="application/atom+xml">
<link href="https://genehack.blog/feed.json" rel="alternate" title="JSON feed for genehack.blog" type="application/json">"
```

(Examples from my personal blog just because I had them handy.)

> Finally my presumption of the URL would be `/blog/feed`.

Containing it under the `/blog` stem is fine, but note there will be multiple feed versions, so we'll have `/blog/feed/rss2.xml`, `/blog/feed/atom.xml`, etc.

static-site/src/util/blogPosts.js
genehack: I didn't see _linting_ errors, but my editor typechecks javascript code too. 

Reviewing it in the cold light of morning, the `if (posts && posts.length > 0)` guard isn't needed (and I'll remove it). 

The interior `if (post)` guard _is_ needed, because [elements in the returned list _might_ be `false`](https://github.com/nextstrain/nextstrain.org/blob/master/static-site/src/util/blogPosts.js#L27) instead of an object, which would cause the element accesses in `feed.addItem()` to fail.

genehack:
Okay, I have added Atom and JSON feeds, linked them into the `<head>` of the template, and updated various other bits.

The one thing remaining is to add a top-level "description" to the feed. This _could_ be "Real-time tracking of pathogen evolution", as we use as the top-level description, but I'm not sure that's appropriate for the blog? 

Suggestions? (/cc @trvrb for visibility on the description question.)

genehack:
> W3C has a [feed validator](https://validator.w3.org/feed/#validate_by_input), that you can copypasta a file into. Once I add the Atom and JSON feed versions, I will validate each of them. 

The Atom and RSS feeds validate.

> (A [JSON Feed validator](https://validator.jsonfeed.org/) also exists.)

The JSON feed needs a little work. 

I think I'm also going to add the full text content of the posts to the feeds as well, unless somebody thinks that's too much.

static-site/src/util/blogPosts.js
jameshadfield: > The only place this file gets included in the site is in 2 JSX files that are used to produce static site files — I don't think the lodash Javascript would need to end up in the bundle? But maybe I'm confused about how the "static" file generation works in the production build.

In this case it's not ending up in the bundle - next.js is (correctly) not including it even tho the files which import it themselves export frontend JSX.

> Might be worth [analysing the bundle](https://nextjs.org/docs/pages/building-your-application/optimizing/package-bundling#analyzing-javascript-bundles) here to check.

I did this to double check and it identified some other issues. Will make an issue.

jameshadfield: > I did this to double check and it identified some other issues. Will make an issue.

Cancel that, was looking at stat sizes not parsed/compressed sizes.

genehack:
> The JSON feed needs a little work.

The JSON feed validates now. 

> I think I'm also going to add the full text content of the posts to the feeds as well, unless somebody thinks that's too much.

And the content chunks have been added. 

This is ready to go, I think...

static-site/src/components/layout.jsx
```diff
@@ -23,6 +23,9 @@ export default class MainLayout extends React.Component {
           <title>{`${siteTitle}`}</title>
           <meta name="description" content={siteDescription} />
           <link rel="me" href="https://mstdn.science/@nextstrain" />
+          <link href={`${siteUrl}/blog/atom.xml`} rel="alternate" title="Atom feed for nextstrain.org/blog" type="application/atom+xml" />
+          <link href={`${siteUrl}/blog/feed.json`} rel="alternate" title="JSON feed for nextstrain.org/blog" type="application/json" />
+          <link href={`${siteUrl}/blog/rss2.xml`} rel="alternate" title="RSS2 feed for nextstrain.org/blog" type="application/rss+xml" />
```
victorlin, line 28: _non-blocking_

All these (including mastodon link) should be conditioned on SiteConfig's `groupsApp` being `false`, but no need to block this PR on that.

static-site/src/util/blogPosts.js
```diff
@@ -36,6 +42,76 @@ export function getBlogPosts() {
...
 
+/**
+ * Writes RSS2, Atom, and JSON feeds into the `public/blog` directory.
+ */
+export async function generateBlogFeeds() {
+  const feedDir = `${__dirname}/../../public/blog`;
+  const siteURL = "https://nextstrain.org/blog";
+
+  const feed = new Feed({
+    title: "Nextstrain.org Blog",
```
victorlin, line 53: _non-blocking_

```suggestion
    title: "Nextstrain Blog",
```

This is the blog of Nextstrain that happens to be on nextstrain.org, not a blog specifically about nextstrain.org

static-site/src/util/blogPosts.js
```diff
@@ -36,6 +42,76 @@ export function getBlogPosts() {
...
+    description: "FIXME",
```
victorlin, line 54:
```suggestion
    description: "Updates from the Nextstrain core team",
```

Doesn't have to be this, just something other than `FIXME` 😄

static-site/src/util/blogPosts.js
```diff
@@ -36,6 +42,76 @@ export function getBlogPosts() {
...
+    id: siteURL,
+    link: siteURL,
+    image: "https://nextstrain.org/_next/static/media/nextstrain-logo-small.dcf6bdd2.png",
```
victorlin, line 57:
```suggestion
    image: "https://nextstrain.org/nextstrain-logo-small.png",
```

Use `static-site/public/nextstrain-logo-small.png` which is served at the top level using the filename as-is, not `static-site/static/logos/nextstrain-logo-small.png` which is bundled by webpack with a hash in the filename that may change over time.

static-site/src/util/blogPosts.js
```diff
@@ -36,6 +42,76 @@ export function getBlogPosts() {
   return blogPosts;
 }
 
+/**
+ * Writes RSS2, Atom, and JSON feeds into the `public/blog` directory.
+ */
+export async function generateBlogFeeds() {
+  const feedDir = `${__dirname}/../../public/blog`;
+  const siteURL = "https://nextstrain.org/blog";
```
victorlin, line 50:
```suggestion
  const blogURL = "https://nextstrain.org/blog";
```

I was confused why these paths seemed to look different:

https://github.com/nextstrain/nextstrain.org/blob/5709c3ce11f60cf66fa4ae3604643018e70e2e00/static-site/src/util/blogPosts.js#L62

https://github.com/nextstrain/nextstrain.org/blob/5709c3ce11f60cf66fa4ae3604643018e70e2e00/static-site/src/components/layout.jsx#L26

then I realized that `siteURL` and `siteUrl` are not the same. Renaming `siteURL` (or even better, [putting the Atom/JSON/RSS2 paths in shared variables](https://github.com/nextstrain/nextstrain.org/pull/990#discussion_r1731861727)) would reduce confusion.

static-site/src/components/layout.jsx
```diff
@@ -23,6 +23,9 @@ export default class MainLayout extends React.Component {
...
         </Helmet>
```
victorlin, line 29: _non-blocking_

These href values could be more tighly coupled with where they are defined:

https://github.com/nextstrain/nextstrain.org/blob/5709c3ce11f60cf66fa4ae3604643018e70e2e00/static-site/src/util/blogPosts.js#L61-L65

They could be top-level constants in `blogPosts.js` and imported here.

package.json
```diff
@@ -47,6 +48,7 @@
     "express-naked-redirect": "^0.1.2",
     "express-session": "^1.16.2",
     "express-static-gzip": "^0.2.2",
+    "feed": "^4.2.2",
```
victorlin, line 51: _non-blocking_

I wondered if `feed` should be under `devDependencies` because the feed is only updated during build time, but realized that's not possible with the way this has been implemented. If `scripts/generate-blog-feeds.js` were the only thing that used `blogPosts.js` then it should be a dev dependency. However, Next.js uses `getBlogPosts()` (in the same file) during runtime.

If this is reorganized so that `generateBlogFeeds()` is not in the same file as `getBlogPosts()`, `feed` could be made a dev dependency. I'll let you decide if that's worthwhile.

victorlin (changes requested):

static-site/src/components/layout.jsx
genehack: done.

static-site/src/util/blogPosts.js
genehack: Kinda ended up doing both of these things...

static-site/src/util/blogPosts.js
genehack: I've been waiting for somebody to suggest something! :grin:

genehack:
Okay, @victorlin, refactored so that `feed` can be a devDep, and adjusted various constants to be imported from SiteConfig (which turned into maybe a larger than expected refactoring due to the need to access things from a non-JSX file...)

I _think_ this is overall cleaner now, but also maybe a little more convoluted? Interested to hear what you think.

static-site/data/BaseConfig.js
```diff
@@ -0,0 +1,31 @@
...
+
+export const siteUrl = "https://nextstrain.org";
+
+// Website description used for RSS feeds/meta description tag.
+export const siteDescription = "Real-time tracking of pathogen evolution";
+
+export const groupsApp = false;
+
+export const groupsTitle = "Scalable Sharing with Nextstrain Groups";
+
```
victorlin, line 22: I believe CDC's internal "groups app instance" has customized these in SiteConfig.jsx. This would be a breaking change for them, which should be fine as long as we give them a heads up on the `#nextstrain-cdc` Slack channel.

static-site/src/util/blogFeeds.js
```diff
@@ -0,0 +1,76 @@
...
+    },
+    {
+      file: `${feedDir}/feed.json`,
+      data: feed.json1,
+    },
+    {
+      file: `${feedDir}/rss2.xml`,
+      data: feed.rss2,
+    },
+  ];
```
victorlin, line 71: Get the `file` values from BaseConfig?

https://github.com/nextstrain/nextstrain.org/blob/43a922fb76960f818e28cb64a0b0139c27ec81ff/static-site/data/BaseConfig.js#L27-L31

victorlin (commented):
Quickly skimmed before heading off to lunch

static-site/data/BaseConfig.js
genehack: Tracked in #1001

static-site/src/util/blogFeeds.js
genehack: refactored.

genehack:
> Quickly skimmed before heading off to lunch

* stripped comment
* created [tracking issue](#1001) for CDC notification about config rearrangement
* refactored blog feed data structure to enable SSoT design
* updated groupsApp.md docs to reflect SiteConfig/BaseConfig split

static-site/data/BaseConfig.js
victorlin, line 1: Re: https://github.com/nextstrain/nextstrain.org/pull/990#issuecomment-2313359755

> I _think_ this is overall cleaner now, but also maybe a little more convoluted? Interested to hear what you think.

Separating dev vs. runtime deps seems more important than keeping all config in SiteConfig.jsx, so this seems fine to me.

victorlin (approved):
Code looks good. I'm not familiar enough with feeds so I don't know how to properly test the Atom/JSON/RSS files, but I trust your judgement there!

genehack:
> I'm not familiar enough with feeds so I don't know how to properly test the Atom/JSON/RSS files, but I trust your judgement there!

They were validating in the preview build in the W3C and JSON feed validators. I'll doublecheck in the morning before I merge.

PR #993, opened 2024-08-23
Revise splash page tiles

trvrb:
1. Slightly revise text
2. Add tile for newly launched Lassa build
3. Remove tile for avian influenza (this feels currently extraneous with the cattle outbreak tile)

PR #996, opened 2024-08-23
Auto-prepend `/` to tile URLs

victorlin:

Previously, the code assumed that all URLs started with `/` and was noted in featured-analyses.yaml. However, there was nothing to check this. A `/` is perfectly valid for everything except for the source check, which silently assumes everything that doesn't start with `/` is maintained by Nextstrain. This is a bad assumption.

The source check could either (1) be updated to work with the error or (2) throw a runtime error. I chose to go with (1) to make things easier when updating featured analyses.

## Related issue(s)

Follow-up to https://github.com/nextstrain/nextstrain.org/pull/993#discussion_r1729422975

## Checklist

- [x] Checks pass
- [x] Manually reverting 7a93cbd1c8bb1f71c86824f70572f7e42fcf9e3e still shows the correct source icon

### DIFF
--- a/static-site/content/featured-analyses.yaml
+++ b/static-site/content/featured-analyses.yaml
@@ -3,7 +3,6 @@
 #   rendered on the splash page.
 # - When adding or modifying descriptions, ensure it takes up at most two lines
 #   when rendered on the splash page.
-# - Ensure URLs start with `/`.
 
 - name: Mpox in the DRC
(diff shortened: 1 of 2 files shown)


genehack (approved):

jameshadfield (approved):
Nice work! (we discussed this back in [the original PR](https://github.com/nextstrain/nextstrain.org/pull/922#discussion_r1656384318))

PR #1006, opened 2024-08-28
Remove PreviewOverlay when expanded

### LINKED ISSUE #1010: Last row of groups cannot be easily clicked
### Current Behavior

Clicking the lower part of a tile from the last row of groups will collapse the section.

### Expected behavior

Clicking anywhere in the a tile should navigate to the group page.

### How to reproduce

https://github.com/user-attachments/assets/40a1fe86-ec7c-4c1c-89f6-8b5048f839bf

### Possible solution

- https://github.com/nextstrain/nextstrain.org/pull/1006

victorlin:
In all current usages of ExpandableTiles, it is possible for and sometimes even expected for users to interact with tiles by clicking. Having an invisible overlay on top of the last row of tiles with a separate onClick behavior can be confusing for users expecting another onClick behavior.

When the tiles are expanded, the only benefit to having PreviewOverlay available is for the onClick behavior of hiding the tiles. This can already be accomplished by the ArrowButton which also covers full width and has a height of 1em.

This removes the opacity transition, which in my eyes does not make a noticeable difference. The visually distinct part of the transition action is on the height change to show more rows of tiles, not the opacity here.

## Related issues

Closes #1010

## Checklist

- [x] Checks pass
- [x] Bug cannot be reproduced in review app

### DIFF
--- a/static-site/src/components/ExpandableTiles/index.tsx
+++ b/static-site/src/components/ExpandableTiles/index.tsx
@@ -46,7 +46,7 @@ export const ExpandableTiles = <AnyTile extends Tile>({tiles, tileWidth, tileHei
             return <TileComponent tile={el} key={el.name} />
           })}
         </TilesContainer>
-        <PreviewOverlay onClick={toggleExpand} className={!isExpandable || isExpanded ? "hidden" : "visible"} />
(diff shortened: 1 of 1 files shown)


victorlin:
Merging this because it worked well in my own testing.

Issue #940, opened 2024-07-02
Add splash tile / pathogen showcase entry for `seasonal-cov`


genehack:
### Context
[Suggested by @jameshadfield](https://github.com/nextstrain/nextstrain.org/pull/939#pullrequestreview-2152492485) during review of #939 

Creating a new issue for this because I'm not sure if the intent of the showcase is for there to be a tile for _everything_ or just for "notable" datasets. If the latter, I'm not sure `seasonal-cov` qualifies?

To be clear, I think we're talking about a tile in the showcase on https://nextstrain.org/pathogens, not the top-level splash page. 

Opinions sought!

joverlee521:
Should we showcase any pathogen that is being automatically updated? I'd hope any pathogen that we think is important enough automate is also "notable"

victorlin:
I believe the original list added in f57bf94a257127f0bc1d9ea8fdc6d4af7fff3b02 was short because the showcase was not expandable and thus implicitly capped at 9 entries at max width of `.container`. Now that it's expandable, I would propose adding at the very least any automatically updated pathogen, if not all available pathogens and repurposing it as more like a "table of contents".

victorlin:
For splash page tiles defined by [featured-analyses.yaml](https://github.com/nextstrain/nextstrain.org/blob/-/static-site/content/featured-analyses.yaml), my understanding is that @trvrb is maintaining this so I would run any new entries through him to determine if/where to place the tile relative to existing tiles.

victorlin:
> Now that it's expandable, I would propose adding at the very least any automatically updated pathogen, if not all available pathogens and repurposing it as more like a "table of contents".

If adding every available pathogen, I would also propose alphabetical sorting to align with how we are now sorting groups. The current sorting is something along the lines of [some "popular" items] + [other items based on time of addition] which is  pretty arbitrary.

trvrb:
If the showcase gets too long, it no longer serves the purpose to highlight new / notable pages. I wouldn't have every automatically updated pathogen (this is what /pathogens is for). I'd also be rotating out previous highlights pretty judiciously. That said, I think the current splash page could handle up 18 cards maximum (3 rows of 6). So we don't have to rotate out quite yet. The top 4-6 should have lots of thought as they're above the fold (as many users won't click the carat).

In this particular case, I think rotating in seasonal CoV near the bottom would be fine, but it would rotate out sooner than most.

[...]

PR #1022, opened 2024-09-12
Add rabies card to pathogens page

kimandrews:

Add rabies card to pathogens page

## Checklist

- [ ] Checks pass


genehack (approved):

jameshadfield (commented):
This adds it as a tile on the [/pathogens](https://nextstrain-s-add-rabies-imcdjj.herokuapp.com/pathogens) page, but we should also add it to the main splash page (which is [this YAML file](https://github.com/nextstrain/nextstrain.org/blob/master/static-site/content/featured-analyses.yaml)).

victorlin:
> we should also add it to the main splash page (which is [this YAML file](https://github.com/nextstrain/nextstrain.org/blob/master/static-site/content/featured-analyses.yaml)).

https://github.com/nextstrain/nextstrain.org/issues/940#issuecomment-2349631750

PR #1015, opened 2024-09-05
Add doc page on Redis

victorlin:
_[preview](https://nextstrain--1015.org.readthedocs.build/projects/nextstrain-dot-org/en/1015/redis.html)_

## Description of proposed changes

Docs used for [6.2 → 7.2 upgrade](https://github.com/nextstrain/private/issues/121). Reflects lessons learned from the 2 upgrades that have been done so far.

## Checklist

- [x] Checks pass
- [x] Convert to RST
- [x] Doc page preview looks good
- [x] Do the 6.2 → 7.2 upgrade and update PR contents with any new findings
- [x] Check if changes affect the [resource index JSON revision](https://docs.nextstrain.org/projects/nextstrain-dot-org/en/latest/resource-collection.html#resource-index-revisions)

PR #1031, opened 2024-09-26
Update Redis docs

victorlin:
Follow-up to #1015

## Checklist

- [x] Checks pass

### DIFF
--- a/docs/redis.rst
+++ b/docs/redis.rst
@@ -145,6 +145,19 @@ described in Heroku's own documentation.
           heroku addons:detach REDIS -a "$app"
       done
 
+   .. warning::
+
+      This step causes the site to enter the degraded state described earlier.
+      If the need arises, you can roll back to the old instance:
+
+      .. code-block:: bash
+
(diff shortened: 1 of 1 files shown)


docs/redis.rst
```diff
@@ -218,7 +218,11 @@ described in Heroku's own documentation.
           heroku addons:attach --as REDIS "$new_instance" -a "$app"
       done
 
-10. Test that your login session is now "remembered" again.
+10. Test that the new instance works:
+
+   1. Load the website and check that your login session is now "remembered" again.
+   2. Check that you can successfully log out and log back in.
+   3. Check that you can remove/add a member from a group.
```
victorlin, line 238: I didn't do this yesterday but it seems like something that should be tested, though setup is complicated – I think it requires a testing grouop and two accounts (one admin to make the change, one user that's logged in for testing).

victorlin:
Merging given +1 on https://github.com/nextstrain/nextstrain.org/pull/1015#discussion_r1777551144

Issue #134, opened 2020-03-14
Add Open Graph and Twitter Card meta tags for rich embeds/previews


tsibley:
[Open Graph](https://ogp.me) is a microformat that specifies `<meta>` tags which can be added to the `<head>` of pages in order to provided structured data about the page. Apps like Facebook, Twitter, and Slack will look for these tags and use them to provide richer embeds/previews when someone links to the page.

[Twitter Cards](https://developer.twitter.com/en/docs/tweets/optimize-with-cards/overview/abouts-cards) build upon Open Graph to provide some extra Twitter-specific information. These may or may not be necessary in our case.

By setting this up for nextstrain.org pages, links in social apps will display a better preview. Doing so will require changes to both Gatsby and the Nextstrain extensions to Auspice.

One challenge will be picking an appropriate "cover" image, since nearly all of the fancy visuals of Nextstrain are dynamic and not static. A first-pass option might be to use the pathogen cards on the front page of Nextstrain. Later, if desired, we could provide an endpoint which generates screencaps of the tree or map or something (probably with significant caching).

victorlin:
It seems like these tags for the nextstrain.org static site have already been implemented in [SEO.js](https://github.com/nextstrain/nextstrain.org/blob/938aabcab71993c1ebdd938c92227147d9fe92f6/static-site/src/components/SEO/SEO.jsx). See this [Open Graph Check of https://nextstrain.org](https://opengraphcheck.com/result.php?url=https%3A%2F%2Fnextstrain.org) - the tags are there. It's worth noting that this seems to be outdated and the `og:image` points to a non-existent file.

The problem is when linking to a page for datasets or narratives. Examples:

- [Open Graph Check for https://nextstrain.org/ncov](https://opengraphcheck.com/result.php?url=https%3A%2F%2Fnextstrain.org%2Fncov)
- [Open Graph Check for https://nextstrain.org/narratives/ncov/sit-rep/2020-03-20](https://opengraphcheck.com/result.php?url=https%3A%2F%2Fnextstrain.org%2Fnarratives%2Fncov%2Fsit-rep%2F2020-03-20)

Since these are generated by auspice (the HTML base seems to be [auspice/index.html](https://github.com/nextstrain/auspice/blob/master/index.html)), I believe this issue should be moved over to that repository, unless you all would rather track it here.

psthomas:
[...]

PR #1032, opened 2024-09-27
Convert 'contact', 'team', and 'whoami' to app router; partially convert '404' [#134]

genehack:

This converts the global/shared layout and navigation bits, the `/contact`, `/team`, and `/whoami` pages (including login state display in that page and the global nav), and the general outline of the `404/not-found` page over to using the App Router. The `404` page has been changed to provide a link to the root page, rather than loading the (as-yet-unported) `<Splash>` component. 

This also required converting all `styled-components` usage in those parts over to either regular CSS, or to CSS Modules, depending on the scope of the usage. In general, I have tried to keep things scoped either very tightly (i.e., in specific CSS Modules imported into a single component) or put them in a global CSS file that is imported in the root layout. 

I have elected to migrate components to `/static-site/components` as they are converted over for use with App Router; that felt like the cleanest option. I also elected to put page content directly into the `app/*/pages.tsx` files, rather than propagating the "page content is actually under `static-site/src/pages/*.jsx`" pattern, reasoning that this reduces the number of moving parts and makes it more clear where the content is relative to the the URL it is displayed at. 

To be explicit: when this conversion work is done, I expect the contents of `/static-site` to be:

```
README.md
app/
components/
content/        ; maybe unify 
data/           ; these two?
next-env.d.ts
next.config.mjs
public/
static/         ; maybe unify with data?
tsconfig.json
types.d.ts
vendored/
```

## Related issue(s)

#1052 
#134 

## Checklist

- [x] Checks pass


tsibley:
Comparison via flicker:

before-after

genehack:
Is that level of whitespace variation worth trying to nail down, IyourO?

tsibley:
Maybe not? Though if I were doing the work I'd probably still try. It also might have more impact on mobile viewports? Would be good to check. The logo position change seems like it'll have more impact when navigating between a page like the contact page and an Auspice page.

genehack:
> Maybe not? Though if I were doing the work I'd probably still try. 

Oh believe me, there has been some previous effort in this direction, I assure you. 

On first blush, I think I would push further effort getting the header alignment perfect to after adding the login/user tracking bit, as I suspect some of the delta is due to that. 

> It also might have more impact on mobile viewports? Would be good to check. The logo position change seems like it'll have more impact when navigating between a page like the contact page and an Auspice page.

Yeah, fair; I haven't done a lot of checking around mobile viewport sizes.

package.json
```diff
@@ -108,6 +108,7 @@
     "jest-extended": "^1.1.0",
     "nodemon": "^2.0.22",
     "request": "^2.88.2",
+    "sass": "^1.79.3",
```
jameshadfield, line 111: > converting all styled-components usage in those parts over to either regular CSS/SASS, or to CSS Modules, depending on the scope of the usage.

Could you write a little about the guidelines you used to decide which system to use, and also why SASS was needed?

static-site/app/styles/browserCompatability.css
```diff
@@ -0,0 +1,446 @@
+/* HTML5 display definitions
```
jameshadfield, line 1: How hard would it be to structure the commits such that we can track these "new" files?

This stylesheet is a duplicate (with some commented out sections removed¹). Can we not simply make this a file move and have the `/pages/*` files use it from its new location? bootstrap.css is a straight copy and globals.css looks like a conversion from `theme.js` (and maybe others), so I can see here the desire to maintain things in parallel until the entire `/pages` section is removed.

This applies to other files in this PR as well, e.g. `static-site/components/people/teamMembers.ts` is largely a copy of `static-site/src/components/People/teamMembers.js`

¹ `git diff master:./static-site/src/styles/browserCompatability.css HEAD:static-site/app/styles/browserCompatability.css `

genehack:
BTW, what browser is this, and do you happen to remember the rough viewport size?

package.json
genehack: Generally, if a style or set of styles was specific to an individual component, I moved it to a CSS Module sidecar file; if there was styling that was shared across multiple components, or that was affecting HTML elements that were part of the basic layout, I put it into global.css. 

I'm also, as much as possible, removing rules that aren't having any effect (due to overriding rules higher in the cascade), to reduce the amount of surface that needs to be considered when understanding how the styling works. 

With regard to SASS, as we discussed yesterday, I reached for that because I didn't realize that [CSS Nesting](https://caniuse.com/?search=css+nesting) was now a mainstream browser feature. I'll take the SASS stuff back out and just do everything as plain 'ol CSS.

static-site/app/styles/browserCompatability.css
genehack: > How hard would it be to structure the commits such that we can track these "new" files?

I think it would be trickier than it's worth, because at the end of all this, the original/source files will be removed, along with all the Pages Router bits. 

Duplicating them now, at the beginning, and scoping them only to the new App Router section, means that I don't have to worry about some change having a temporary bad side-effect on Pages Router pages, the way I would need to if I changed those pages to reference the new copies.

genehack:
> BTW, what browser is this, and do you happen to remember the rough viewport size?

@tsibley this was a question for you, about that comparison GIF — I'm having trouble reproducing some of the differences visible there with Firefox v131 on macOS.

static-site/app/styles/browserCompatability.css
jameshadfield: If the conversion for the entire site will be done in one PR, squashing all commits would be an option. We haven't done this much in the team, but we also haven't had a migration of this scope before!

package.json
genehack: > With regard to SASS, as we discussed yesterday, I reached for that because I didn't realize that [CSS Nesting](https://caniuse.com/?search=css+nesting) was now a mainstream browser feature. I'll take the SASS stuff back out and just do everything as plain 'ol CSS.

One issue that I've run into with switching to plain CSS in CSS Modules, is that CSS Modules requires selectors to be "pure", and doesn't support the nesting selector, so you can't do something like: 

```
.componentClassName {
  margin: 0 auto;

  & h1 {
    font-weight: 500;
  }

  & a { 
    color: black;
    text-decoration: none;
  }
}
```

you best you can do would be something like this: 

```
.componentClassName {
  margin: 0 auto;
}
.componentClassName > h1 { 
  font-weight: 500;
}
.componentClassName > a { 
  color: black;
  text-decoration: none;
}
```

which is …fairly annoying by comparison.

tsibley:
> BTW, what browser is this, and do you happen to remember the rough viewport size?

Firefox 113.0.2 on Linux (specifically `113.0.2+build1-0ubuntu0.18.04.1`) with a viewport that's as wide as you see in the full size image (minus the red border).

package.json
tsibley: > One issue that I've run into with switching to plain CSS in CSS Modules, is that CSS Modules requires selectors to be "pure", and doesn't support the nesting selector…

I'm confused by your example. I believe you could write it without the nesting selector as:

```css
.componentClassName {
  margin: 0 auto;

  h1 {
    font-weight: 500;
  }

  a { 
    color: black;
    text-decoration: none;
  }
}
```

and have the same effect. An older version of the spec required `&` in this case, but that's no longer the case. I don't know why CSS Modules would support one and not the other at this point.

Stepping back though, do we think CSS nesting is widely supported enough yet? It was only supported across all browsers as of December 2023. The usage and date relative views on the [Can I use… page](https://caniuse.com/css-nesting) for it show a small but maybe significant number of older browser usage lacking support. (Our actual visitor usage may vary from Can I use…'s numbers.)

genehack: > I'm confused by your example. I believe you could write it without the nesting selector as:
> 
> ```css
> .componentClassName {
>   margin: 0 auto;
[snip]

This code:

```css
.teamPage {
  h1 {
    color: var(--darkGrey);
  }
}
```

produces this output:

```
Syntax error: Selector "h1" is not pure (pure selectors must contain at least one local class or id)

  1 | .teamPage {
> 2 |   h1 {
    |   ^
  3 |     color: var(--darkGrey);
```

> Stepping back though, do we think CSS nesting is widely supported enough yet? It was only supported across all browsers as of December 2023. The usage and date relative views on the [Can I use… page](https://caniuse.com/css-nesting) for it show a small but maybe significant number of older browser usage lacking support. (Our actual visitor usage may vary from Can I use…'s numbers.)

I wondered about this too but didn't go beyond wondering, as I wasn't 100% sure what our browser support policy was.

The other potentially significant thing, which I've already run into once, is that SASS/SCSS supports [`@extend`](https://sass-lang.com/documentation/at-rules/extend/) and I don't think there's any equivalent for that in standard CSS. There are a number of `styled-elements` rules that are defined as:

```js
const NewRule = styled(OtherRule)`selector: rule`;
```
Porting those to a future new state is _way_ easier with `@extend`.

tsibley: > This code […] produces this output […]

Well that's dumb! I misread your prior statement as «CSS Modules require selectors to be "pure" **and so** don't support the nesting selector (`&`)» and so I thought "Ah, but the nesting selector isn't required for that example." But the correct reading of your statement is actually, «CSS Modules require selectors to be "pure" **and also** don't support the nesting selector (`&`)». Thanks for the clarifying example.

genehack: > Well that's dumb!

I was less than impressed to discover it -- it apparently has to be that way in order to prevent "impure" selectors escaping the module scope and affecting the global document, from which I infer that the underlying implementation is, well, bad. 

While I'm bitching, it's also pretty underwhelming that using the CSS Modules still requires you to set a `className` prop to something imported from the CSS file, instead of just applying the CSS module to the entire hunk (and only the hunk) of JSX coming out of the component.

tsibley: > it apparently has to be that way

"has", sure. Ugh. If this works

```css
.componentClassName {
}
.componentClassName h1 { 
}
```

but this doesn't

```css
.componentClassName {
  h1 {
  }
}
```

it's extra dumb because they're equivalent. Difference is purely syntactical, so the issue is with the logic behind CSS Modules "pure" mode, not with anything fundamental. (To be clear, not complaining about you/your words, but about this dumb thing.)



> …instead of just applying the CSS module to the entire hunk (and only the hunk) of JSX coming out of the component.

I think you get that by doing:

```js
import "./foo.css";
```

in the `app/…/page.tsx`. That is, not using CSS Modules at all.

static-site/app/contact/page.tsx
victorlin, line 1: This new file is (1) a rename of `src/pages/contact.jsx` + (2) necessary TypeScript/App Router changes + (3) additional touch-ups. It's hard to tell from 1f5099ad56edee40bf0e2516cdc480ee10763783 – the additional touch-ups are throwing off the git rename detection. I think splitting those out into a separate commit would make it easier to review. Example: 36d5bf4802e484ffd9ff39a2db85302cf94de150 (1, 2) + aa8c7535d9863f4fc27ff3027931b8adc0b33324 (3)

package.json
genehack: > > …instead of just applying the CSS module to the entire hunk (and only the hunk) of JSX coming out of the component.
> 
> I think you get that by doing:
> 
> ```js
> import "./foo.css";
> ```
> 
> in the `app/…/page.tsx`. That is, not using CSS Modules at all.

I don't think that's quite what I want either -- as I understand it, doing `import "foo.css"` adds stuff to the stylesheet for the entire document; it's not scoped to just the component. So a top level `h1` rule in `foo.css` would apply to an `<h1>` that was in the document but didn't arise from component doing the import.

tsibley: Ahh, you're totally right.

static-site/app/styles/bootstrap.css
```diff
@@ -0,0 +1,3872 @@
...
+  .container, .container-sm, .container-md, .container-lg {
+    max-width: 960px;
+  }
+}
+
+@media (min-width: 1200px) {
+  .container, .container-sm, .container-md, .container-lg, .container-xl {
+    max-width: 1140px;
+  }
+}
```
joverlee521, line 53: I noticed some additional whitespace on a wider screen and traced it back to the `container` class. The App router layout is respecting the bootstrap.css max-width of 1140px while the Pages router styled component `Container` has an additional @media rule 

https://github.com/nextstrain/nextstrain.org/blob/7617c384f3aa8fb3083c46a5c29f9f90db8c2558/static-site/src/components/splash/styles.tsx#L20-L22

joverlee521 (commented):
I really like the direction this is heading, thank you for simplifying the styling ❤️

static-site/components/footer/index.tsx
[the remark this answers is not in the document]
genehack: > Is the plan to put all individual components under `components/` and only put pages under `app/`? Or should page specific components be colocated with the page.tsx under `app/`?

The plan is anything that's used on more than one page goes under `components/`; single-page components will be in `app/$page/` — not tightly wedded to that but it seems like the most intuitive organization in terms of keeping things as close as possible to where they get used.

static-site/app/styles/bootstrap.css
genehack: I'll have to dig and see where I got that 1500px rule from and how it's working on the current prod site...

joverlee521: Ah, sorry I wasn't clear. I meant the 1500px rule is included in the current prod site but is not here.

genehack: @joverlee521 Actually, [it is there](https://github.com/nextstrain/nextstrain.org/pull/1032/files#diff-414250c6938b686d9b4ef1135e2c3038098a52c8fb5d223792e2894c830a6226R49), unless I'm still misunderstanding you. 

I do see a layout difference on a larger screen, but it's only in the width of the divs starting with the team avatar icons -- from what I see, the layout from the top of the page down to and including "The core Nextstrain team is" matches exactly; stuff after that is clamped to 1140 ...

but yeah that's because the `@media(min-width: 1550px)` rule isn't cascading down from the `html` selector... 🤔

genehack: ugh, nvm, stupid bootstrap.

static-site/components/people/avatars.tsx
```diff
@@ -0,0 +1,42 @@
+import { ReactElement, ReactNode } from "react";
+
+import styles from "./avatars.module.scss";
```
victorlin, line 3: FYI since we just discussed build failures on a call, it's [failing](https://github.com/nextstrain/nextstrain.org/actions/runs/11391880676/job/31696648454) here because of this line

```
./components/people/avatars.tsx
Module not found: Can't resolve './avatars.module.scss'
```

victorlin: (My [build failures](https://github.com/nextstrain/nextstrain.org/pull/1025#issuecomment-2420582251) are unrelated)

genehack:
If folks want to take another look at this, I've addressed the feedback about the basic layout (I think! 🤞 ) and also finished the `/team` page port. 

Heroku seems to be having some deployment issues at the moment; when a review app finally builds, I'll link a preview. 

Next stage: dealing with the user context.

static-site/components/people/avatars.tsx
genehack: thanks, that got cleaned up in the most recent pushes.

static-site/app/styles/browserCompatability.css
genehack: I don't think squashing everything into one monster commit is going to improve the understandability here; I do think keeping all changes related to one "step" of the process is the best balance between the not-always-aligned goals of "make sure things work at each step", "make it possible to see the changes", and "don't waste massive amounts of time jumping through git hoops".

static-site/app/layout.tsx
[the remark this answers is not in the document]
genehack: Having just done the 404 page — I needed to adjust the root layout a _tiny_ bit, but it wasn't horrible, and the final result is *more* consistent (e.g., the logo doesn't change for no reason on the 404 page, and the nav bar links are the same size as usual). 

I'm going to try to keep it this way as I move into the `whoami` page, which is the next thing to convert.

genehack:
> If folks want to take another look at this, I've addressed the feedback about the basic layout (I think! 🤞 ) and also finished the `/team` page port.
> 
> Heroku seems to be having some deployment issues at the moment; when a review app finally builds, I'll link a preview.
> 
> ~Next stage: dealing with the user context.~

Ported the 404 page over to App Router, which required porting a stubbed-out version of the `<Splash>` component as well. That should, I hope, resolve the test failures.

**Next stage:** dealing with the user context.

static-site/app/layout.tsx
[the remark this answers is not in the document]
genehack: Compare [the Pages Router 404](https://nextstrain-s-convert-bl-98fca6.herokuapp.com/404) with [the App Router 404](https://nextstrain-s-convert-bl-98fca6.herokuapp.com/foo) to see what I'm talking about.

static-site/app/whoami/styles.module.css
```diff
@@ -0,0 +1,23 @@
...
+}
+
+.subText {
+  font-size: 18px;
+  line-height: var(--niceLineHeight);
+}
+
+.userGroupList {
+  display: table;
+  margin: 0 auto !important;
```
genehack, line 22: note to self: see if you can remove the `!important`

static-site/components/footer/site-map.module.css
```diff
@@ -0,0 +1,46 @@
+.siteMap {
+  /* Center contents */
+  margin: 0 auto !important;
+
+  /* Ensure that, on wide screens, 4 sections renders to roughly align
+  with the width of IconParagraph (rendered below this element) */
```
genehack, line 6: note to self: check if `IconParagraph` is still the correct label here.

static-site/components/logos/index.tsx
```diff
@@ -0,0 +1,105 @@
...
+}
+
+function Logo({
+  href,
+  imgSrc,
+  width,
+}: {
+  href: string;
+  imgSrc: string;
+  width: number;
```
genehack, line 86: note to self: can probably re-use the `Logo` type here

static-site/components/nav/nav-logo.tsx
```diff
@@ -0,0 +1,43 @@
...
+  <a href="/" className="logo">
+    <Image src={logo} alt="Nextstrain (logo)" width="40" />
+  </a>
+);
+
+const LogoType = (): React.ReactElement => {
+  const rainbowTitle = siteTitle
+    .split("")
+    .map((letter, i) => (
+      <span key={i} style={{ color: "var(--titleColor" + i + ")" }}>{letter}</span>
```
genehack, line 31: note to self: can you use variable interpolation here and make this string a little cleaner?

static-site/app/contact/page.tsx
victorlin, line 1: I noticed a new loading spinner at the bottom right. This is the output associated with it:

```
 ○ Compiling /contact ...
 ✓ Compiled /contact in 1869ms (1224 modules)
```

I assume it's dev only and pages are pre-compiled in an actual build?

https://github.com/user-attachments/assets/06725641-e5fa-45d9-8fcc-e08ba5f9065d

static-site/app/styles/globals.css
```diff
@@ -0,0 +1,141 @@
...
+  --titleColor0: #4377cd; /* N */
+  --titleColor1: #5097ba; /* E */
+  --titleColor2: #63ac9a; /* X */
+  --titleColor3: #7cb879; /* T */
+  --titleColor4: #9abe5c; /* S */
+  --titleColor5: #b9bc4a; /* T */
+  --titleColor6: #d4b13f; /* R */
+  --titleColor7: #e49938; /* A */
+  --titleColor8: #e67030; /* I */
+  --titleColor9: #de3c26; /* N */
```
victorlin, line 14: Why not put these in `static-site/components/nav/styles.css`?

static-site/app/styles/globals.css
```diff
@@ -0,0 +1,141 @@
...
+  /* We use nextjs' next/font to manage our primary font (Lato). `var(--lato)` is defined in app/layout.tsx */
+  font-family: var(--lato), "Helvetica Neue", sans-serif;
+  font-weight: 300;
+}
+
+html {
+  font-size: 62.5%;
+  margin: auto;
+  -webkit-overflow-scrolling: touch;
+}
```
victorlin, line 40: Maybe worth noting that these should be synced with the styles used by pages router?

https://github.com/nextstrain/nextstrain.org/blob/544b59e4a95f19ea71504cb9579fc0ecbda22330/static-site/src/styles/globals.css#L4-L15

static-site/app/contact/page.tsx
victorlin: I split these up again for ease of review: 3e7fe67d8a2fc0ffbdc85f60063347332200206f...5f8ec8d5ac3c46c2cf0bc2a6f8aaa60212a78150

victorlin (commented):
Reviewed all the contents of converted pages. Will continue reviewing styles next week.

static-site/components/nav/user-or-login-link.tsx
```diff
@@ -0,0 +1,19 @@
+"use client";
```
joverlee521, line 1: Is this a client component only to be able to use the `UserContext`? Would this be an opportunity to just completely remove the `UserContext` and `UserDataWrapper` component and switch to getting this data server side?

joverlee521 (commented):
I didn't look through the styling changes too closely but just compared the updated pages to prod. I like that the `/whoami` page's layout matches the rest of the site now 👍 

I think the move away from styled-components should make it easier to remove `UserContext`/`UserDataWrapper` completely and just fetch the current user data server side.

static-site/components/nav/user-or-login-link.tsx
joverlee521: @tsibley did this in his previous demo of migrating to Next.js: 

1. https://github.com/nextstrain/nextstrain.org/commit/c98760c sets up the `getCurrentUser` function on the server side 
2. https://github.com/nextstrain/nextstrain.org/commit/164579fe9653bc5d5f9684efc1ed918cd4665b91 uses the server function in the nav bar
3. https://github.com/nextstrain/nextstrain.org/commit/65a62ee06cda7f27ad651971b4c7d7bdef5a1e28 uses the server function for the whoami page

static-site/app/contact/page.tsx
genehack: Given that the new page is basically a complete re-write, I don't think the ROI on maintaining this granularity in the commits is worth the pain.

static-site/app/not-found.tsx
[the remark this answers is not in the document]
genehack: I think it will be cleaner to just let them both get cleaned up when they're not needed anymore. 

In my local testing, I have noticed that I sometimes still end up getting served `/404` from the Pages router, possibly due to internal redirects. I think leaving that alone until we're ready to completely remove the Pages Router is easier than re-arranging code in places that will ultimately go away.

static-site/app/contact/page.tsx
genehack: AFAIK everything is pre-compiled as part of the build process, yes.

static-site/app/styles/globals.css
genehack: My long term plan is to remove the duplicated version of this list that's currently in `static-site/src/components/splash/title.tsx` and have that component get these variables from the root.

static-site/app/styles/globals.css
genehack: The differences between the old version and the new one are basically removing rules that were redundant.

static-site/app/contact/page.tsx
victorlin: That's fair, I don't mind keeping the commits as-is. I mostly did it for myself (and anyone other reviewers looking for the same).

static-site/components/footer/site-map.module.css
```diff
@@ -0,0 +1,46 @@
...
+/* Wrapper is used to center the actual Section while allowing it to be left-aligned. */
+.sectionWrapper {
+  /* Center section contents when all sections are displayed in a single row */
+  @media (min-width: 768px) {
+    text-align: center;
+  }
+
+  /* Otherwise, when there are multiple rows, left-align so columns are aligned */
+  text-align: left;
+}
```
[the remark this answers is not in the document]
genehack: good catch, will fix

static-site/components/nav/user-or-login-link.tsx
genehack: Yes, this is a client component so that it can get access to the context. 

I elected to port the context layers over because it felt like a smaller change, like it was more idiomatic Next.js style, and because the scale of changing the contents of the `request` object felt like it was expanding the already large scope here. 

I think if we want to try to get away from using React context for this, that should be a follow-up change, not part of this conversion.

joverlee521: That's fair, this has been a long thread to pull on!

static-site/app/whoami/styles.module.css
genehack: you can!

static-site/components/footer/site-map.module.css
```diff
@@ -0,0 +1,46 @@
+.siteMap {
+  /* Center contents */
+  margin: 0 auto !important;
```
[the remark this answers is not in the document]
genehack: This one needs to stay, because otherwise bootstrap stomps on the left/right margin.

static-site/components/footer/site-map.module.css
genehack: it is.

static-site/components/nav/nav-logo.tsx
genehack: yes, you can.

static-site/components/focus-paragraph/styles.module.css
```diff
@@ -0,0 +1,22 @@
+/* this is the basic style for all FocusParagraph */
+.focus {
+  font-size: var(--niceFontSize);
+  font-weight: 300;
```
victorlin, line 4:
```suggestion
```

Remove this, and let it inherit from https://github.com/nextstrain/nextstrain.org/blob/a331d36fd8ea14a51a7f4b2fac0926bbc2d41899/static-site/app/styles/globals.css#L28-L34

static-site/components/nav/styles.css
```diff
@@ -0,0 +1,30 @@
...
+  display: flex;
+  height: 50px;
+  justify-content: space-between;
+  left: 0px;
+  margin: auto;
+  overflow: hidden;
+
+  a {
+    color: var(--darkGrey);
+    font-size: 16px;
```
victorlin, line 12: Suggestions:

1. Set font-size on `nav` so it's applicable to non-links as well
2. Use `var(--niceFontSize)`

static-site/components/focus-paragraph/styles.module.css
```diff
@@ -0,0 +1,22 @@
...
+}
+
+/* this is specifically *added* to the base style for FocusParagraphCentered */
+.centered {
+  margin: 20px 0 0;
+  text-align: center;
+}
+.centered > strong {
+  font-weight: 500;
+}
```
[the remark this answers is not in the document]
genehack: So, I had intended to loop back and fix the font-weight situation once the App Router conversion was complete. 

From experimenting in Firefox, the `font-weight: 500` _does_ have a visual impact, and it is different (lighter) than setting `font-weight: 700` in the same place — I was specifically experimenting with the hyperlinks in the footer, and 700 is (IMO) too heavy; 500 looks good.

This _specific_ rule is relevant to the as-yet-unported [static-site/src/sections/group-members-page.tsx](https://github.com/nextstrain/nextstrain.org/blob/master/static-site/src/sections/group-members-page.tsx#L113-L122).

static-site/components/nav/nav-logo.tsx
```diff
@@ -0,0 +1,43 @@
...
+    .split("")
+    .map((letter, i) => (
+      <span key={i} style={{ color: `var(--titleColor${i})` }}>{letter}</span>
+    ));
+  const SubTitle = (): React.ReactElement => (
+    <div className="wordmark-subtitle">Groups Server</div>
+  );
+
+  return (
+    <a href="/" className="wordmark">
```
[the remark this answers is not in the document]
genehack: Looks like an oversight; I can look at converting it over. 

I will say, this will make the resulting CSS uglier, because the `nav` selector won't be allowed in a CSS Module, because it is not "pure".

static-site/components/focus-paragraph/styles.module.css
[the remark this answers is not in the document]
victorlin: Can you check if there is a visual difference between 500 and 400? I would be surprised if there is. 400 should be used since that is a valid weight.

genehack:
(Not sure why I can't `quote reply` this one comment, but...)

> My biggest concern is the lack of clarity around where the duplicates are.

I guess I don't see this as a huge issue, because when all this work is done, all the duplicates will be gone, as all the "old" components will be removed/replaced — more specifically, `static-site/pages` and `static-site/src` will no longer be present. 

If the work is introducing multiple duplicates where there is currently only a single copy, I would consider that a bug/mis-implementation — but I don't think that's the case. 

The only way to do this without having some transient duplication would be to convert the entire `static-site` directory at once, and even then, you'd have to decide between the equally unpalatable options of "have some commits where the site is broken" and "have some commits that are far too large and change multiple things at once due to shared component usage". Compared to those two options, I think having some duplication present in the tree in some commits is the least bad option.

static-site/components/focus-paragraph/styles.module.css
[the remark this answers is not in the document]
genehack: Yeah, on deeper investigation, it's using 400 when we set 500; I thought it was interpolating but it isn't.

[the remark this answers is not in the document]
victorlin: Ok, I didn't notice the other changes that use `font-weight: 500`. I'd suggest replacing those with `font-weight: 400`.

static-site/components/nav/nav-logo.tsx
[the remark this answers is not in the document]
victorlin: Couldn't the `nav` selector be replaced with `.nav` and used with `<nav className={styles.nav}>`?

static-site/app/styles/browserCompatability.css
victorlin: Re: https://github.com/nextstrain/nextstrain.org/pull/1032#issuecomment-2463428609

> Not sure why I can't `quote reply` this one comment, but...

Oh I replied to an existing thread so you have to go back to the thread to add a comment. GitHub UI doesn't do a great job of showing this... copying back to the thread.

> > My biggest concern is the lack of clarity around where the duplicates are.
> 
> I guess I don't see this as a huge issue, because when all this work is done, all the duplicates will be gone, as all the "old" components will be removed/replaced — more specifically, `static-site/pages` and `static-site/src` will no longer be present.
> 
> If the work is introducing multiple duplicates where there is currently only a single copy, I would consider that a bug/mis-implementation — but I don't think that's the case.
> 
> The only way to do this without having some transient duplication would be to convert the entire `static-site` directory at once, and even then, you'd have to decide between the equally unpalatable options of "have some commits where the site is broken" and "have some commits that are far too large and change multiple things at once due to shared component usage". Compared to those two options, I think having some duplication present in the tree in some commits is the least bad option.

Sorry if I wasn't clear – I'm not suggesting to remove the duplication. It's obviously unavoidable with the iterative approach! I'm suggesting to improve the clarity. I don't see anything obvious that maps copy A to copy B unless the filenames are similar, which is not always the case.

victorlin:
> (Not sure why I can't `quote reply` this one comment, but...)

Oh I replied to an existing thread so you have to go back to the thread to add a comment. GitHub UI doesn't do a great job of showing this... copying back to the thread: https://github.com/nextstrain/nextstrain.org/pull/1032#discussion_r1833537040

static-site/components/focus-paragraph/styles.module.css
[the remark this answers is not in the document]
genehack: ok, will change all the new CSS to use `font-weight: 400` instead of `font-weight: 500`

static-site/components/nav/nav-logo.tsx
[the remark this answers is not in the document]
genehack: > Couldn't the `nav` selector be replaced with `.nav` and used with `<nav className={styles.nav}>`?

I mean, yeah, that's what I'm planning on doing, but that ☝️ is IMO uglier than `<nav>`, and the CSS will change from the clean and nested stuff that's there now to having to prepend `.nav` to all other selectors. 

I'm not going to change it at this point but I have mild regrets about going the CSS Modules route; the bad parts > the good, IMO.

static-site/components/footer/site-map.module.css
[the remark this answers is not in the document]
genehack: @joverlee521 can you explain a little bit more about what you're seeing and how it differs from what nextstrain.org is doing now? 

When I flipped the order of these two rules, the footer links shifted -- and I see that at multiple viewport sizes. 

With the rules in the original order (with the `@media` query first), I get the same layout as I see in production.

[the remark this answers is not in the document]
genehack: (Mac Firefox, in case this is a browser thing.)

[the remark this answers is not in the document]
joverlee521: I'm on mac/Firefox 132.0.1. 

With the original order of `@media` query first, I see the footer links shift:

https://github.com/user-attachments/assets/85c674b8-014a-491a-ab00-cd2e1ae2dc95

With new order of `@media` query second, there is no shift:

https://github.com/user-attachments/assets/cc186c78-7668-4654-8532-eadebdd53ac3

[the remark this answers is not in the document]
genehack: that's really weird; that is the opposite of what i'm seeing.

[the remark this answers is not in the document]
joverlee521: Huh, if it helps, the currently deployed testing app does _not_ have a flicker for me.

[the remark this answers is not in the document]
genehack: I was turned around, all good.

PR #1034, opened 2024-10-01
/pathogens: Add tiles for all pathogens

victorlin:
_[preview](https://nextstrain-s-victorlin--8r2s7q.herokuapp.com/pathogens)_

## Description of proposed changes

See linked issue and commit messages

## Related issue(s)

Closes #1033

## Checklist

- [x] [Add Nextclade tile?](https://github.com/nextstrain/nextstrain.org/issues/1033#issuecomment-2384512123)
- [x] Sort tiles alphabetically in code instead of YAML
- [x] Checks pass
- [x] Check if changes affect the [resource index JSON revision](https://docs.nextstrain.org/projects/nextstrain-dot-org/en/latest/resource-collection.html#resource-index-revisions)

PR #1035, opened 2024-10-01
Merge display name and tile lists

victorlin:

`coreGroupDisplayNames` and `coreTiles` had some overlap. This commit merges the two lists into a single a list of pathogens. Separate lists for display names and tiles are still used. The difference is that they are derived from the merged list.

This removes some implementation details from the YAML file and should make content updates more straightforward.

## Related issue(s)

N/A

## Checklist

- [ ] Checks pass
- [ ] Merge base PR #1034


static-site/src/components/ListResources/index.tsx
```diff
@@ -73,6 +72,12 @@ function ListResources({
     )
   }
 
+  const tiles = pathogens.map((p) => ({
+    name: p.name,
+    img: p.img,
+    filters: [p.id]
+  } as FilterTile));
```
genehack, line 79: why is this cast needed?

static-site/src/components/ListResources/index.tsx
```diff
@@ -293,14 +296,14 @@ const Tile = ({ tile }: { tile: FilterTile }) => {
...
+ * for which the filters are valid given the resources known to the resource
+ * listing UI
  */
-const useTiles = (tiles?: FilterTile[], groups?: Group[]) => {
-  const [restrictedTiles, setRestrictedTiles] = useState<FilterTile[]>([]);
+const usePathogens = (pathogens: Pathogen[], groups?: Group[]) => {
+  const [restrictedTiles, setRestrictedTiles] = useState<Pathogen[]>([]);
   useEffect(() => {
-    if (!tiles || !groups) return;
+    if (!pathogens || !groups) return;
```
genehack, line 306: The way this is written, `!pathogens` is always going to be false -- you've got `pathogens: Pathogen[]` in the function signature, so at a minimum, there will be an empty array there. 

I'm not sure if the correct fix here is to change the function signature to be `pathogens?: Pathogen[]` (which would more closely match the old version) or change this conditional to be `if (!pathogen.length || !groups)`

genehack (approved):

static-site/src/components/ListResources/index.tsx
victorlin: Good catch – I inadvertently removed the `?`. Added back in [force-push](https://github.com/nextstrain/nextstrain.org/compare/f9a617103fb9f2a22ab0277f94fd1b3649e646df..5c46de7e9269ca8a64bbf3863de9e8b196a91f0a). I don't think the `?` is necessary given existing usage which guarantees a value, but that was already the case before this PR.

static-site/src/components/ListResources/index.tsx
victorlin: I've fixed my intentions in [force-push](https://github.com/nextstrain/nextstrain.org/compare/f9a617103fb9f2a22ab0277f94fd1b3649e646df..5c46de7e9269ca8a64bbf3863de9e8b196a91f0a). Hopefully this makes more sense: any issues with the map output (e.g. missing `filters`) will show a type error here instead of further down.

genehack (approved):

victorlin:
cc @trvrb: you recently modified this list in #1095 and may be interested in the user-facing text changes.

PR #1036, opened 2024-10-02
Finish adding types to ListResources

victorlin:

With this PR, all files under [static-site/src/components/ListResources](https://github.com/nextstrain/nextstrain.org/blob/5d85bd9d18b175c0d7a304960eccbce53e411f82/static-site/src/components/ListResources/) (with the exception of vendored `dodge.js`) are TypeScript files.

## Related issue(s)

- #871
- Follow-up to #874
- Prompted by thinking around https://github.com/nextstrain/nextstrain.org/pull/1035

## Checklist

- [x] Checks pass

### DIFF
--- a/static-site/src/components/ListResources/useSortAndFilter.ts
+++ b/static-site/src/components/ListResources/useSortAndFilter.ts
@@ -1,30 +1,42 @@
 import { useMemo } from 'react';
+import { FilterOption, Group, Resource, SortMethod } from './types';
 
 
-export const useSortAndFilter = (sortMethod, selectedFilterOptions, originalData, setState) => {
+export const useSortAndFilter = (
+    sortMethod: SortMethod,
+    selectedFilterOptions: readonly FilterOption[],
+    setState: React.Dispatch<React.SetStateAction<Group[]>>,
+    originalData?: Group[],
+  ) => {
   useMemo(() => {
     if (!originalData) return;
     /* Following console log is really useful for development */
     // console.log(`useSortAndFilter() sortMethod "${sortMethod}" ` + (selectedFilterOptions.length ? `filtering to ${selectedFilterOptions.map((el) => el.value).join(", ")}` : '(no filtering)'))
 
-    let _sortGroups, _sortResources;
+    let _sortGroups: (groupA: Group, groupB: Group) => 1 | -1 | 0,
+        _sortResources: (a: Resource, b: Resource) => 1 | -1 | 0;
     switch (sortMethod) {
       case "lastUpdated":
-        _sortGroups = (groupA, groupB) => _newestFirstSort(groupA.lastUpdated, groupB.lastUpdated);
-        _sortResources = (a, b) => a.lastUpdated === b.lastUpdated ?
-          _lexicographicSort(a.name, b.name) : // resources updated on the same day sort alphabetically
-          _newestFirstSort(a.lastUpdated, b.lastUpdated); // newest updated resources first
+        _sortGroups = (groupA: Group, groupB: Group) => _newestFirstSort(groupA.lastUpdated, groupB.lastUpdated);
+        _sortResources = (a: Resource, b: Resource) => {
+          if (!a.lastUpdated || !b.lastUpdated || a.lastUpdated === b.lastUpdated) {
+            // resources updated on the same day or without a last updated date
+            // sort alphabetically
+            return _lexicographicSort(a.name, b.name)
+          }
+          else {
+            return _newestFirstSort(a.lastUpdated, b.lastUpdated);
+          }
+        }
         break;
       case "alphabetical":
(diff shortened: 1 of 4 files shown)


static-site/src/components/ListResources/useSortAndFilter.ts
```diff
@@ -49,13 +58,13 @@ export const useSortAndFilter = (sortMethod, selectedFilterOptions, originalData
 /* helper function to sort alphabetically. If provided with YYYY-MM-DD strings
  * this is the same as a chronological sort (oldest to newest)
  */
-function _lexicographicSort(a,b) {
+function _lexicographicSort(a: string, b: string) {
```
genehack, line 61:
```suggestion
function _lexicographicSort(a: string, b: string): 1 | -1 | 0 {
```

static-site/src/components/ListResources/useSortAndFilter.ts
```diff
@@ -49,13 +58,13 @@ export const useSortAndFilter = (sortMethod, selectedFilterOptions, originalData
...
   return a<b ? -1 : a>b ? 1 : 0;
 }
 
 /* If provided with YYYY-MM-DD strings this sorts chronologically with newest
  * first (this is just reverse lexicographic sort)
  */
-function _newestFirstSort(a,b) {
+function _newestFirstSort(a: string, b: string) {
```
genehack, line 68:
```suggestion
function _newestFirstSort(a: string, b: string) : 1 | -1 | 0 {
```

static-site/src/components/ListResources/useSortAndFilter.ts
```diff
@@ -1,30 +1,39 @@
 import { useMemo } from 'react';
+import { FilterOption, Group, Resource } from './types';
 
 
-export const useSortAndFilter = (sortMethod, selectedFilterOptions, originalData, setState) => {
+export const useSortAndFilter = (sortMethod: string, selectedFilterOptions: readonly FilterOption[], setState: React.Dispatch<React.SetStateAction<Group[]>>, originalData?: Group[]) => {
```
genehack, line 5:
```
export const useSortAndFilter = (sortMethod: string, selectedFilterOptions: readonly FilterOption[], setState: React.Dispatch<React.SetStateAction<Group[]>>, originalData?: Group[]) => {
```

Since `sortMethod` here is being used in a `switch`, it might be worth typing it explicitly as `"lastUpdated"|"alphabetical"`, and then you can remove the `default` case.

genehack (approved):

static-site/src/components/ListResources/useSortAndFilter.ts
victorlin: Updated in [force-push](https://github.com/nextstrain/nextstrain.org/compare/5d85bd9d18b175c0d7a304960eccbce53e411f82..1ecff3bd305af2b9942c6d8d9b1283e6ff80a8a2).

static-site/src/components/ListResources/useSortAndFilter.ts
victorlin: Updated in [force-push](https://github.com/nextstrain/nextstrain.org/compare/5d85bd9d18b175c0d7a304960eccbce53e411f82..1ecff3bd305af2b9942c6d8d9b1283e6ff80a8a2).

static-site/src/components/ListResources/useSortAndFilter.ts
victorlin: Updated in [force-push](https://github.com/nextstrain/nextstrain.org/compare/5d85bd9d18b175c0d7a304960eccbce53e411f82..1ecff3bd305af2b9942c6d8d9b1283e6ff80a8a2).

What do you think about using a type alias `SortMethod = "lastUpdated" | "alphabetical"`? Example: 3284beb526722968cdac41dadbbe0b091b82bd57

In the end, at least without more involved code changes, it requires a cast `event.currentTarget.value as SortMethod`, but I think it's a good improvement.

genehack (approved):

static-site/src/components/ListResources/useSortAndFilter.ts
genehack: yeah, i like making the type alias; it makes it more explicit that the two different places it is used are actually the same thing. 

agreed on the cast as a necessary evil in this case.

victorlin: Added the type alias to 6d3bf4f9cdc3a9a7d6c830068b34b0894895ed3a in [force-push](https://github.com/nextstrain/nextstrain.org/compare/1ecff3bd305af2b9942c6d8d9b1283e6ff80a8a2..fb2fb81020eaca1644dadc9d3041d6120c178a7d).

Issue #1052, opened 2024-10-25
Convert `static-site` from Pages Router to App Router


genehack:
This is a prerequisite for #134, but has expanded beyond that to the point where I think a distinct isuse will be helpful. This should be considered a meta-issue, to facilitate building on top of the initial conversion work to parallelize work to finalize the conversion.

This is a list of pages that need to be converted:

- [ ] index page
    - [ ] components/SEO
    - [ ] components/splash
- [x] 404/not-found
    - [ ] components/splash
- [ ] /blog
- [ ] /community
    - [ ] components/Datasets/dataset-select
    - [ ] components/errorMessages
- [x] /contact
- [ ] /groups
    - [ ] components/Datasets/dataset-select
    - [ ] components/errorMessages
    - [ ] components/Groups/Tiles
    - [ ] components/ListResources
- [ ] /ncov
    - [ ] components/Datasets/*
    - [ ] components/errorMessages
- [ ] /pathogens
    - [ ] components/ListResources
- [ ] /sars-cov-2
    - [ ] components/Datasets/*
    - [ ] components/errorMessages
- [ ] /sars-cov-2/forecasts
- [ ] /staging
    - [ ] components/errorMessages
    - [ ] components/ListResources
- [x] /team
- [x] /whoami

If you'd like to build on top of #1032 and convert one of these pages, please create a new issue and then link that issue to the appropriate page in the above list.

In many cases, converting one of these will also require porting a component away from using Styled Components; some of these components may be used across more than one top-level page (e.g., <ListResources> is used by `/groups`, `/pathogens`, and `/staging`). Before you start working on converting a page, please make sure that somebody isn't already working on a page that would require both of you working on the same underlying component! To make it a little easier to see where these overlaps are, I have listed some of the components used in particular pages as sub-items of the page.

PR #1059, opened 2024-10-31
Port `/blog` to App router; add blog preview cards [#134]

### LINKED ISSUE #134: Add Open Graph and Twitter Card meta tags for rich embeds/previews

genehack:

This ports the `/blog` and `/blog/[id]` pages to App Router. It also adds metadata to blog posts, including OpenGraph tags, to support preview cards. Finally, it updates the `marked` Markdown parser to the current release, in order to get type support, and adds the `@types/sanitize-html` dev dependency to provide type support for `sanitize-html`. 

~Note that this code is on top of, and includes, the changes from #1032; when that is merged, I will rebase this on top of it.~ This is rebased and ready to review. 

## Related issue(s)

- resolves #134 
- #1052 

## Checklist

- [x] Checks pass


static-site/app/not-found.tsx
```diff
@@ -0,0 +1,19 @@
...
+export default function FourOhFour(): React.ReactElement {
+  return (
+    <>
+      <div className={styles.errorContainer}>
+        Oops - that page doesn’t exist! (404).
+        <br />
+        Here’s the splash page instead…
+      </div>
+
+      <Splash />
```
jameshadfield, line 16: Todo - drop this, as it's a placeholder (literally!) while the splash page is still part of the pages router.

static-site/app/whoami/page.tsx
```diff
@@ -0,0 +1,54 @@
+"use client";
```
jameshadfield, line 1: I'm trying to understand nextjs behaviour here. This code must be run client-side due to the use of react context (ultimately because of an API call to get the users state). However using a production nextjs build in a browser with JS disabled we have:
* `/contact` is rendered server side and works fine (as expected, `contact/page.tsx` doesn't have `"use client"`)
* `/` (splash page) renders nothing. Expected, as it's the pages router and we don't use SSR/SSG there. (See https://github.com/nextstrain/nextstrain.org/pull/824, although I'd priorise moving to the app router instead.)
* This page renders fine ("You are not logged in.") despite indicating it will only run client-side. How does this happen?

jameshadfield (commented):
A couple of comments while trying out the PR. I didn't do a thorough comparison of pages as I think others have done so.

static-site/app/not-found.tsx
genehack: this is actually already fixed in the [prior PR](https://github.com/nextstrain/nextstrain.org/pull/1032); once that is merged, this PR will get rebased on top of it and will pick up [that change](https://github.com/nextstrain/nextstrain.org/pull/1032/commits/4f649994c05ace441e03e42f4fe88c5b0a367c8f).

static-site/app/whoami/page.tsx
genehack: >     * This page renders fine ("You are not logged in.") despite indicating it will only run client-side. How does this happen?

What happens if you log in, though?

jameshadfield: I should have read [the docs](https://nextjs.org/docs/app/building-your-application/rendering/client-components). 

> To optimize the initial page load, Next.js will use React's APIs to render a static HTML preview on the server for both Client and Server Components. This means, when the user first visits your application, they will see the content of the page immediately, without having to wait for the client to download, parse, and execute the Client Component JavaScript bundle.

> What happens if you log in, though?

As expected, it still thinks i'm logged out (until hydration & the JS code runs)

joverlee521 (commented):
I only did a quick review in the test app, I didn't dig into specifics. Really love the new layout of the blog page! 

I only flagged a weird behavior in loading images when redirecting from the main `/blog` page to the most recent blog post.

static-site/app/blog/page.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+  // _technically_ getBlogPosts() could return an empty array and then
+  // mostRecentPost would be undefined -- to make the type checker
+  // happy, if for some reason mostRecentPost is undefined, we will
+  // detect that and redirect to the 404 page
+  const redirectTo = mostRecentPost
+    ? `/blog/${mostRecentPost.blogUrlName}`
+    : `/404`;
+
+  redirect(redirectTo);
```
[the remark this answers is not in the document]
genehack: yeah, i see that. 

seems to be because the image URLs are relative to `/blog` (i.e., `img/foo.png`, not `/blog/img/foo.png`), maybe? The images are showing as 404 and the HTTP request was `GET /img/oropouche_host_view.png`. 

oddly, it works fine on localhost... I will dig into the `redirect()` API and see if I need to do something different, but an initial fix might be tweaking the image URLs in the Oropouche post.

static-site/app/blog/utils.ts
```diff
@@ -0,0 +1,72 @@
...
+  date: string;
+  mdstring: string;
+  sidebarTitle: string;
+  title: string;
+}
+
+// Scans the ./static-site/content/blog directory for .md files and
+// returns a chronologically-sorted array of posts, each with some
+// basic metadata and the raw (unsanitized) markdown contents.
+export function getBlogPosts(): BlogPost[] {
```
victorlin, line 20: There are two functions that have duplicated JS and TS versions:

- getBlogPosts
    - static-site/src/util/blogPosts.js
    - static-site/app/blog/utils.ts
- parseMarkdown
    - static-site/app/blog/parseMarkdown.ts
    - static-site/src/util/parseMarkdown.js

I understand that the JS versions are still used by `scripts/generate-blog-feeds.js`. Is there a plan to deduplicate these at some point?

static-site/app/blog/page.tsx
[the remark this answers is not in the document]
genehack: So, yeah, this is because of how the `redirect()` API works, apparently. 

I think I'm going to take this as the push from the universe to move ahead with converting the front page to be the five most recent blog posts, which will require some additional refactoring in this PR.

package.json
```diff
@@ -63,7 +64,7 @@
     "luxon": "^3.4.4",
     "make-fetch-happen": "^11.1.1",
     "mapbox-gl": "^3.2.0",
-    "marked": "^0.7.0",
+    "marked": "^14.1.3",
```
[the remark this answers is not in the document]
genehack: The primary reason was I needed to update it off 0.7 to get typing support, and so I figured I might as well update it to "current". 

FWIW, 0.7 is "only" five years old — the package [versions fairly aggressively](https://www.npmjs.com/package/marked?activeTab=versions), for whatever reason.

static-site/app/blog/utils.ts
genehack: I could convert `scripts/generate-blog-feeds` to TS — which would allow the old `getBlogPosts()` to be removed — but it felt out of scope of the issue.

The old `parseMarkdown()` is also used in `static-site/src/components/sourceInfoHeading.jsx`, so that needs both community and groups (at least) to be converted over before it can be removed.

victorlin: Thanks for the explanation. I [still think](https://github.com/nextstrain/nextstrain.org/pull/1032#discussion_r1833206146) it'd be useful to add a small comment at the "deprecated" JS file indicating that it's been ported over to TS. Otherwise I will keep asking for clarification about these duplicates in PR reviews.

package.json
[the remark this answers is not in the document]
victorlin: Makes sense!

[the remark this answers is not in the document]
genehack: Since I need to rework other stuff in this PR, I have gone back and updated the commit message to include this information.

static-site/app/blog/utils.ts
genehack: Added deprecation notices to these two files.

static-site/app/blog/page.tsx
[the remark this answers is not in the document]
genehack: In the end, because of how Next.JS generates URLs without trailing slashes, I had to update the blog image URLs; there was no other way to resolve this (not without other knock-on damage that would have been worse). 

So the "multiple pages on the front page" thing will get kicked down the road a bit, and this PR is once again ready to be reviewed.

Note that I also added a section to the static-site README with some minimal info about blog post creation.

jameshadfield (approved):

PR #1062, opened 2024-11-11
Remove font-size scaling

victorlin:

The purpose of the scaling is unclear. It was added in "parse /groups/:groupName markdown;" (347ee033) which is the commit that added the markdown rendering of group descriptions, so maybe it was a stylistic adjustment relative to whatever else was on the page at that time.

A similar 94% font-size scaling is used in the custom Auspice splash page (auspice-client/customisations/splash.js) as well as other parts of the Auspice code base. The purposes of those are also unclear.

The main motivation for removing it here is that there was an [internal bug report](https://bedfordlab.slack.com/archives/C0K3GS3J8/p1731343211138099) of a group's description rendering with what looks like partially scaled fonts. The issue self-resolved without any code changes and we were unable to understand what caused it. My suspicion is that this font-size scaling is a contributing factor. It can't be the only factor because it has been around for many years now and this is the first report of such an issue. Maybe a recent browser update handles scaled fonts poorly.

Another motivation is that, after seeing the unscaled version, the scaled version has noticeably smaller text for what seems to be no apparent reason. I think removing the scaling improves readability on this page, but of course that is subjective.

## Related issues

- Maybe fixes [text rendering issue](https://bedfordlab.slack.com/archives/C0K3GS3J8/p1731343211138099)?

## Checklist

- [x] Checks pass

### DIFF
--- a/static-site/src/components/sourceInfoHeading.jsx
+++ b/static-site/src/components/sourceInfoHeading.jsx
@@ -24,9 +24,6 @@ const OverviewContainer = styled.div`
     width: 200px;
     padding: 0;
   }
-  a, p {
-    font-size: 94%;
-  }
 `;
 
 export const AvatarImg = styled.img`


genehack (approved):

static-site/src/components/sourceInfoHeading.jsx
```diff
@@ -24,9 +24,6 @@ const OverviewContainer = styled.div`
     width: 200px;
     padding: 0;
   }
-  a, p {
-    font-size: 94%;
-  }
```
tsibley, line 29: Any idea _why_ this was here in the first place?

(I mean, seems like we should remove it, but also, want to consider Chesterton's Fence and all.)

victorlin: I can't tell. It's from 347ee0336606e4546d6e92bd4df97d4710dca891 which was the commit that added the markdown rendering of group descriptions, so maybe it was a stylistic adjustment relative to whatever else was on the page at that time.

I can update the commit message to clarify this uncertainty (or certainty if someone else has a better explanation).

tsibley: Nod, thanks for looking. I'd say it's worth adding that to the commit message, as well as a reason _for_ the change in the commit message (i.e. font rendering issue reported by Emma, which you did link the PR). But not a blocker.

jameshadfield: 94% font-size is elsewhere in the code too (and was there in 347ee03 also):

```jsx
/* lifted from auspice's default splash page */
function formatDataset(requestPath, dispatch, changePage) {
  return (
    <li key={requestPath}>
      <div
        style={{... fontSize: "94%"}}
```

Following the above comment back to Auspice, 94% font-size is used in 5 places. You could trace them back to see if there was a commit message explaining _why_, but I suspect there wasn't.

Given it's relatively widespread and long-term use, I don't understand why this could be responsible for any recent changes (unless we think it's been buggy for years but we only noticed recently?). Moreover, [this comment in slack](https://bedfordlab.slack.com/archives/C0K3GS3J8/p1731356299825229?thread_ts=1731343211.138099&cid=C0K3GS3J8) indicates the jagged-text went away when viewing next.nextstrain.org (with the app-router changes) and thus seems unrelated to the changes in this PR.

victorlin: I've updated the PR description and commit message 943fa3ea7ba93cb20435b8b6c8baed8d296079f9 with a summary. The case for merging this is definitely weak and I'm fine with closing it out if it doesn't seem useful to others.

genehack: I'm +1 for merging it. As Victor notes in the commit message, it makes the overall text sizing more consistent, and the current scaling down doesn't appear to convey any semantic information. Additionally, I see a benefit (small, but potentially cumulative over time) from reducing the amount of CSS surface area in play — it's currently extremely hard to reason about CSS changes, and the only way I see to make it easier to try to reduce the amount that's in play.

jameshadfield: To be clear, I'm fine with this being merged. I'm extremely skeptical that this was the underlying reason for the original bug report on slack, as [I explained above](https://github.com/nextstrain/nextstrain.org/pull/1062#discussion_r1838756758), but that doesn't prevent us from making this change.

PR #1071, opened 2024-11-13
Reformat TypeScript files

victorlin:

Reformatting for consistency with files in Auspice codebase

## Related issue(s)

- https://github.com/nextstrain/auspice/issues/1890

## Checklist

- [x] Checks pass


genehack (approved):
be nice if this also added in return types where they're missing, but i also understand that's not a "pure" formatting change, so it's not a blocker.

PR #1073, opened 2024-11-14
Improve type narrowing and error handling

### LINKED ISSUE #892: List Resources: improve null checks on dates
## 1

from https://github.com/nextstrain/nextstrain.org/pull/874#discussion_r1626765546

https://github.com/nextstrain/nextstrain.org/blob/aae874c06cd6d17616710c501ad73104a7c5e051/static-site/src/components/ListResources/Modal.tsx#L134-L140

@ivan-aksamentov's recommendation:

> Why is this bang needed? the preceding line ensures that there is at lease one element in d.

Compiler is not smart enough to figure this out from this kind of the conditional. Connecting `.length` to `.at()` is tough for a type system.

What it can easily figure out is null checks. So a cleaner way would be to check and throw after non-fallible array access:

```ts
const t0 = d.at(0)?.getTime()
if(isNil(t0)) { // careful to not exclude the legit value `0`
    throw ...
}
// `t0` is guaranteed to be `number` in this branch
```

I would go further. Finding first and last value is a common enough "algorithm" that I would introduce a utility function:

```ts
export function firstLastOrThrow<T>(a: T[]): [T, T] {
    // TODO: use .at() or lodash head() and tail() along with null checks 
}
```

This would make the code very pretty and safe, with all dirty details hidden:

```ts
const [t0, t1] = firstLastOrThrow(dates) // guaranteed to return a pair of numbers or to  throw
```

## 2

from https://github.com/nextstrain/nextstrain.org/pull/874#discussion_r1626767197

https://github.com/nextstrain/nextstrain.org/blob/aae874c06cd6d17616710c501ad73104a7c5e051/static-site/src/components/ListResources/Modal.tsx#L148-L156

https://github.com/nextstrain/nextstrain.org/blob/aae874c06cd6d17616710c501ad73104a7c5e051/static-site/src/components/ListResources/Modal.tsx#L180-L182

@ivan-aksamentov's recommendation:

The problem with comments like this:
```ts
// presence of dates on resource has already been checked so this assertion is safe
```
is that 3 months from now a fresh intern comes and inserts new code between the check and the assertion (not knowing that this assertion exists - it's hard to find even when looking for it specifically) and then :boom:. Though, to be fair, it did not happen with js previously (due to lack of interns?)

Again, a small wrapper would make it safe and clean:

```ts
export function at<T>(a: T[], index: number): T {}
```

If you can find a fallback value, then something like this might work:

```ts
flatData[0]?.date ?? new Date()
``` 

This cannot fail and requires no hacks.

Continuing with the wrapper (though not directly applicable here sadly):

```ts
export function at<T>(a: T[], index: number, fallback?: T): T {}
```

Another option, although more involved, is to write a custom array type which always returns value or throws and never returns undefined. There might be libraries implementing non-empty arrays.

victorlin:

I mainly went through 54be4d60c17e62d236ad95d10aff3895ca0a4553 and looked for things that can be improved. See commit messages for details.

## Related issue(s)

- Closes #892

## Checklist

- [x] Checks pass

### DIFF
--- a/static-site/src/components/ListResources/IndividualResource.tsx
+++ b/static-site/src/components/ListResources/IndividualResource.tsx
@@ -3,7 +3,7 @@ import React, {useState, useRef, useEffect, useContext} from 'react';
 import styled from 'styled-components';
 import { MdHistory } from "react-icons/md";
 import { SetModalResourceContext } from './Modal';
-import { ResourceDisplayName, Resource } from './types';
+import { ResourceDisplayName, Resource, DisplayNamedResource } from './types';
 import { IconType } from 'react-icons';
 import { InternalError } from './errors';
 
@@ -20,10 +20,8 @@ export const LINK_HOVER_COLOR = '#31586c'
 const [resourceFontSize, namePxPerChar, summaryPxPerChar] = [16, 10, 9];
 const iconWidth = 20; // not including text
 const gapSize = 10;
-export const getMaxResourceWidth = (displayResources: Resource[]) => {
+export const getMaxResourceWidth = (displayResources: DisplayNamedResource[]) => {
   return displayResources.reduce((w, r) => {
-    if (!r.displayName) return w
-
     /* add the pixels for the display name */
     let _w = r.displayName.default.length * namePxPerChar;
     if (r.nVersions && r.updateCadence) {
--- a/static-site/src/components/ListResources/IndividualResource.tsx
+++ b/static-site/src/components/ListResources/IndividualResource.tsx
@@ -133,10 +133,11 @@ export const IndividualResource = ({
   const ref = useRef<HTMLDivElement>(null);
   const [topOfColumn, setTopOfColumn] = useState(false);
   useEffect(() => {
-    // don't do anything if the ref is undefined or the parent is not a div (IndividualResourceContainer)
-    if (!ref.current
-     || !ref.current.parentNode
-     || ref.current.parentNode.nodeName != 'DIV') return;
+    if (ref.current === null ||
+        ref.current.parentNode === null ||
+        ref.current.parentNode.nodeName != 'DIV') {
+      throw new InternalError("ref must be defined and the parent must be a div (IndividualResourceContainer).");
+     }
 
     /* The column CSS is great but doesn't allow us to know if an element is at
     the top of its column, so we resort to JS */
--- a/static-site/src/components/ListResources/IndividualResource.tsx
+++ b/static-site/src/components/ListResources/IndividualResource.tsx
@@ -5,6 +5,7 @@ import { MdHistory } from "react-icons/md";
 import { SetModalResourceContext } from './Modal';
 import { ResourceDisplayName, Resource } from './types';
 import { IconType } from 'react-icons';
+import { InternalError } from './errors';
 
 export const LINK_COLOR = '#5097BA'
 export const LINK_HOVER_COLOR = '#31586c'
@@ -129,7 +130,7 @@ export const IndividualResource = ({
   isMobile: boolean
 }) => {
   const setModalResource = useContext(SetModalResourceContext);
-  if (!setModalResource) throw new Error("Context not provided!")
+  if (!setModalResource) throw new InternalError("Context not provided!")
 
   const ref = useRef<HTMLDivElement>(null);
   const [topOfColumn, setTopOfColumn] = useState(false);
--- a/static-site/src/components/ListResources/Modal.tsx
+++ b/static-site/src/components/ListResources/Modal.tsx
@@ -146,14 +146,15 @@ function _snapshotSummary(dates: string[]) {
 }
 
 function _draw(ref, resource: VersionedResource) {
-  // do nothing if resource has no dates
-  if (!resource.dates) return
-
   /* Note that _page_ resizes by themselves will not result in this function
   rerunning, which isn't great, but for a modal I think it's perfectly
   acceptable */  
   const sortedDateStrings = [...resource.dates].sort();
   const flatData = sortedDateStrings.map((version) => ({version, 'date': new Date(version)}));
+
+  if (flatData[0] === undefined) {
+    throw new InternalError("Resource does not have any dates.");
+  }
 
   const width = ref.clientWidth;
   const graphIndent = 20;
@@ -178,8 +179,7 @@ function _draw(ref, resource: VersionedResource) {
 
   /* Create the x-scale and draw the x-axis */
   const x = d3.scaleTime()
-    // presence of dates on resource has already been checked so this assertion is safe
-    .domain([flatData[0]!.date, new Date()]) // the domain extends to the present day
+    .domain([flatData[0].date, new Date()]) // the domain extends to the present day
     .range([graphIndent, width-graphIndent])
   svg.append('g')
     .attr("transform", `translate(0, ${heights.height - heights.marginBelowAxis})`)
--- a/static-site/src/components/ListResources/Modal.tsx
+++ b/static-site/src/components/ListResources/Modal.tsx
@@ -4,7 +4,7 @@ import styled from 'styled-components';
 import * as d3 from "d3";
 import { MdClose } from "react-icons/md";
 import { dodge } from "./dodge";
-import { Resource } from './types';
+import { Resource, VersionedResource } from './types';
 import { InternalError } from './errors';
 
 export const SetModalResourceContext = createContext<React.Dispatch<React.SetStateAction<Resource | undefined>> | null>(null);
@@ -17,7 +17,7 @@ export const ResourceModal = ({
   resource,
   dismissModal,
 }: {
-  resource?: Resource
+  resource: VersionedResource
   dismissModal: () => void
 }) => {  
   const [ref, setRef] = useState(null); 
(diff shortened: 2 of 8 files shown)


static-site/src/components/ListResources/types.ts
```diff
@@ -31,6 +49,36 @@ export interface Resource {
...
+      resource.nVersions !== undefined &&
+      resource.updateCadence !== undefined) {
+    return {
+      ...resource,
+      lastUpdated: resource.lastUpdated,
+      firstUpdated: resource.firstUpdated,
+      dates: resource.dates,
+      nVersions: resource.nVersions,
+      updateCadence: resource.updateCadence
+    }
```
genehack, line 77: I don't understand what this code is doing — you just checked that all those properties exist on `resource`, so they will be there in the spread version, and then you're overwriting them with ...the same values? 

How is this different from just doing `return resource;`?

static-site/src/components/ListResources/IndividualResource.tsx
```diff
@@ -122,15 +121,16 @@ interface IndividualResourceProps {
...
   const ref = useRef<HTMLDivElement>(null);
   const [topOfColumn, setTopOfColumn] = useState(false);
   useEffect(() => {
-    // don't do anything if the ref is undefined or the parent is not a div (IndividualResourceContainer)
-    if (!ref.current
-     || !ref.current.parentNode
-     || ref.current.parentNode.nodeName != 'DIV') return;
+    if (ref.current === null ||
+        ref.current.parentNode === null ||
+        ref.current.parentNode.nodeName != 'DIV') {
```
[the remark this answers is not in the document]
victorlin: This should be enforced by ESLint rule `eqeqeq`. I've started the journey in #1076

static-site/src/components/ListResources/Modal.tsx
```diff
@@ -133,28 +131,30 @@ const Title = styled.div`
...
   else duration=`${Math.round(days/365)} years`;
-  return {duration, first: d[0], last:d.at(-1)};
+  return {duration, first: d1, last: d2};
 }
 
-function _draw(ref, resource: Resource) {
-  // do nothing if resource has no dates
-  if (!resource.dates) return
-
+function _draw(ref, resource: VersionedResource) {
```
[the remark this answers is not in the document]
victorlin: > ref is typeless

Not ideal (it was easiest to not bother when typing in 432577b5efc54173c76f8483c67342f70a28a07f), but beyond the scope of this PR to address. See https://github.com/nextstrain/nextstrain.org/blob/25341aece1ebe3a7c62c2cadb49403b45b242aa9/static-site/tsconfig.json#L32

> return type is not annotated

Also not ideal (I didn't know this was good practice in 432577b5efc54173c76f8483c67342f70a28a07f) and beyond the scope of this PR to address. I think it should be enforced with ESLint (see https://github.com/nextstrain/auspice/issues/1890).

static-site/src/components/ListResources/errors.tsx
```diff
@@ -0,0 +1,54 @@
...
+  static getDerivedStateFromError(error: Error): State {
+    return {
+      hasError: true,
+      errorMessage: error instanceof InternalError ? error.message : "Unknown error (thrown value was not an InternalError)",
+    };
+  }
+
+  override componentDidCatch(error: Error, info: ErrorInfo) {
+    console.error(error);
+    console.error(info);
```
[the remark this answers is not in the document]
victorlin: Logging the object without conversion to string allows for easier inspection in the console logs. Also, this was copied from [Auspice](https://github.com/nextstrain/auspice/blob/6e047d378bd6c4f7b7affa2500d41838ae8dc723/src/components/errorBoundaries/panelErrorBoundary.tsx#L41-L45).

static-site/src/components/ListResources/errors.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  override render() {
+    if (this.state.hasError) {
+      return (
+        <ErrorContainer>
+          {"Something isn't working!"}
+          <br/>
+          {`Error: ${this.state.errorMessage}`}
+          <br/>
+          {'Please '}<a href="/contact" style={{fontWeight: 300}}>get in touch</a>{" if this keeps happening"}
```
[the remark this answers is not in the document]
victorlin: I agree! Though this was copied from elsewhere and I would like to consider it beyond the scope of the PR to change:

https://github.com/nextstrain/nextstrain.org/blob/25341aece1ebe3a7c62c2cadb49403b45b242aa9/static-site/src/components/ListResources/index.tsx#L61-L65

static-site/src/components/ListResources/types.ts
victorlin: `return resource` gives a type error:

```
error TS2322: Type 'Resource' is not assignable to type 'VersionedResource'.
  Types of property 'lastUpdated' are incompatible.
    Type 'string | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.
```

Another alternative is `return resource as VersionedResource` but that loses the tight coupling between the condition and return value.

jameshadfield: This was the main TS concept I dove into (see comments in [this issue](https://github.com/nextstrain/auspice/issues/1889)) and as best as I can tell changing the type of an object (i.e. the same reference) goes against the grain of TS. Most suggestions involved using something like Victor's done to change the type. (You can cast - it's not like TS forbids it - but it seems an antipattern.)

static-site/src/components/ListResources/IndividualResource.tsx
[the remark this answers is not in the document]
jameshadfield: [Added recently to Auspice](https://github.com/nextstrain/auspice/pull/1872)

static-site/src/components/ListResources/types.ts
victorlin: My impression is also that we should avoid casting (and similarly non-null assertions) if possible, even if the proper way is more verbose like in this example. I'll explore removing them and enforcing that in a separate PR.

jameshadfield: > return resource gives a type error:

What about `return {...resource}` - is TS smart enough to match the non-undefined checks and therefore allow the newly created object to be of `VersionedResource` type?

victorlin: No, this is the error:

```
Type '{ name: string; displayName?: ResourceDisplayName; groupName: string; nameParts: string[]; sortingName: string; url: string; lastUpdated?: string; firstUpdated?: string; dates?: string[]; nVersions?: number; updateCadence?: UpdateCadence; }' is not assignable to type 'VersionedResource'.
  Types of property 'lastUpdated' are incompatible.
    Type 'string | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.
```

The first object shape is just a reflection of `resource: Resource` from the function signature. `resource: Resource | VersionedResource` doesn't work either.

victorlin: Separate PR for stricter types: #1077

genehack: > The first object shape is just a reflection of `resource: Resource` from the function signature. `resource: Resource | VersionedResource` doesn't work either.

Maybe only marginally more attractive, but I wonder if turning this into something like a wrapped type guard would work? Something like 

```typescript
function foo(bar: Resource): VersionedResource {
  const _guard = (x:Resource): x is VersionedResource => 
    return x.lastUpdaded && the other checks here;

  if (_guard(bar)) { return bar }
  else { throw }
}
```

static-site/src/components/ListResources/errors.tsx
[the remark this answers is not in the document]
genehack: 🥺

static-site/src/components/ListResources/types.ts
victorlin: Hmm, I find that less attractive because `x is VersionedResource` is effectively a type assertion. A type error won't be raised if you accidentally forget to check one of the the properties in the return statement. So it's about the same as `return resource as VersionedResource`.

static-site/src/components/ListResources/errors.tsx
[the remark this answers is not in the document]
victorlin: #1078

static-site/src/components/ListResources/types.ts
genehack: > Hmm, I find that less attractive because `x is VersionedResource` is effectively a type assertion.

...a type assertion supported by exactly the same logic that underlies the spread + override `return` in your code.

> A type error won't be raised if you accidentally forget to check one of the the properties in the return statement.

I mean, yes, if you write a bug in the code, the code is going to be buggy. :grin:

> So it's about the same as `return resource as VersionedResource`.

I disagree; there's zero validation associated with a cast on the return.

victorlin: > if you write a bug in the code, the code is going to be buggy

Yes, and we should leverage TypeScript to catch the bug at compile time. I haven't articulated my point very well. This example bug + screenshots might help.

Scenario: a property `newVersionedProperty` is added. Just like `lastUpdated`, `firstUpdated`, etc., it is optional on `Resource` and required on `VersionedResource`. The bug is that we forget to add it to this function.

typing

1. This is the code in the PR. There is TypeScript error: `Types of property 'newVersionedProperty' are incompatible`. Useful!
2. This is `return resource as VersionedResource`. No TypeScript error.
3. This is the guard `resource is VersionedResource`. No TypeScript error.

PR #1076, opened 2024-11-14
Lint everything

victorlin:

Apply linting rules more uniformly and (TODO) address violations.

## Related issue(s)

Prompted by https://github.com/nextstrain/nextstrain.org/pull/1073#discussion_r1842659569. I added `eqeqeq: error` to `.eslintrc.yaml`, realized it didn't apply to `static-site`, and went down the rabbit hole.

## Checklist

- [ ] Figure out what to do with violations
- [ ] Checks pass


.eslintrc.yaml
victorlin, line 1: Task: address `prefer-const` violations

There is 1. I will fix it.

```
static-site/src/components/ListResources/dodge.js
  50:13  error  'y' is never reassigned. Use 'const' instead  prefer-const
```

victorlin, line 1: Task: address `no-use-before-define` violations

There are 110. Example:

```
static-site/src/templates/displayMarkdown.jsx
  52:10  error  'MobileToggleIconContainer' was used before it was defined  no-use-before-define
  53:12  error  'MobileToggleIcon' was used before it was defined           no-use-before-define
  57:10  error  'GreyOverlay' was used before it was defined                no-use-before-define
  72:10  error  'SidebarBodyFlexContainer' was used before it was defined   no-use-before-define
  73:12  error  'SidebarContainer' was used before it was defined           no-use-before-define
  82:12  error  'ContentContainer' was used before it was defined           no-use-before-define
  84:16  error  'PostDate' was used before it was defined                   no-use-before-define
  85:16  error  'PostTitle' was used before it was defined                  no-use-before-define
  86:16  error  'PostAuthor' was used before it was defined                 no-use-before-define
```

This was added per https://github.com/nextstrain/auspice/pull/1665#discussion_r1184486701. It's categorized as a "Code quality rule" in our ESLint config file, but seems to be more of a stylistic thing. The easiest thing might be to drop the rule, at least for `static-site`.

victorlin, line 1: Task: address `operator-linebreak` violations

There are 16. Example:

```
static-site/src/components/sourceInfoHeading.jsx
   96:31  error  '&&' should be placed at the beginning of the line  operator-linebreak
  102:28  error  '&&' should be placed at the beginning of the line  operator-linebreak
```

This was added per https://github.com/nextstrain/auspice/pull/1665#discussion_r1184489708. It's a code style rule, I'm ambivalent to dropping or addressing.

victorlin, line 1: Task: address `quote-props` violations

There are 7. Example:

```
static-site/src/components/People/teamMembers.js
    2:3  error  Unnecessarily quoted property 'founders' found  quote-props
   16:3  error  Unnecessarily quoted property 'core' found      quote-props
   85:3  error  Unnecessarily quoted property 'alumni' found    quote-props
  152:3  error  Unnecessarily quoted property 'board' found     quote-props
```

This was added per https://github.com/nextstrain/auspice/pull/1665#discussion_r1184491797. It's a code style rule, I'm ambivalent to dropping or addressing.

genehack:
I would say that maybe the time for this is _after_ the App Router conversion is finished, rather than in the middle of it?

victorlin:
@genehack sure, I've updated the PR description and will plan to leave this in draft until that's done.

.eslintrc.yaml
jameshadfield: We should use it but allow function hoisting. (see [this thread](https://github.com/nextstrain/auspice/pull/1450#discussion_r1152377156). Most of those errors are from a (formerly? currently?) recommended react style along the lines of:
```jsx
export const Component = (props) => {
   return (<Container>jsx</Container>)
}
const Container = styled.div``
```

which looks like is shouldn't work (it works, I think, because `Component` isn't called by that file, rather it's imported elsewhere so that `Container` is in scope by the time it runs).

jameshadfield: We shouldn't be enforcing this in my opinion.

victorlin: Agreed. Let's drop the rule entirely from nextstrain.org and auspice?

victorlin: https://github.com/nextstrain/nextstrain.org/pull/1117 + https://github.com/nextstrain/auspice/pull/1941

victorlin: I've dropped the commit above and replaced it with 2 commits related to `no-use-before-define`: 676b6ce6fdff9729c16f8ce2f6edf608ae0520c0, c4aa1b55d57d384cda469285f3b9246bf8364578

package.json
```diff
@@ -14,8 +14,8 @@
   "scripts": {
     "build": "./build.sh",
     "build:feeds": "node scripts/generate-blog-feeds.js",
-    "lint": "npm run lint:server && npm run lint:static-site",
-    "lint:server": "DEBUG=eslint:cli-engine npx eslint --max-warnings=0 .",
+    "lint": "DEBUG=eslint:cli-engine npx eslint --max-warnings=0 .",
+    "lint:server": "DEBUG=eslint:cli-engine npx eslint --max-warnings=0 --ignore-pattern static-site .",
     "lint:static-site": "cd static-site && DEBUG=eslint:cli-engine npx eslint --max-warnings=0 .",
```
jameshadfield: From cc763094726d7ea6501f8087da026d880fddd3e1

> For unknown reasons, the invocations 'eslint static-site' and 'cd
static-site && eslint .' behaved differently when next.js rules were
involved, with the former reporting a subset of the errors/warnings
picked up by the latter. The next.js config uses a lot of path
manipulation¹ so I presume that's the root cause, but the simple
solution is to change directory before invoking the rule.

I just tested and didn't see any differences as things stand (both with the commands in the PR and those from that commit message), but I think any rules in the base config won't be applied to the static site with the `lint:static-site` invocation. For this reason, as well as general simplicity, I'd think to drop the `lint:server` and `lint:static-site` commands completely.

.eslintrc.yaml
```diff
@@ -19,10 +19,28 @@ ignorePatterns:
   - "!/src"
   - "!/test"
```
jameshadfield: Similar to the comment for the other config, thoughts on inverting this? It requires excluding a few more folders, but it reveals some relevant code in `auspice-client` which wasn't being linted (and has errors)

```
ignorePatterns:
  - "node-modules"
  - "/devData"
  - "dist"
  - "/docs/build"
```

static-site/.eslintrc.yaml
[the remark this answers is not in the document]
victorlin: I'd prefer to keep the inclusion-based targets. We intentionally switched to this in #1158.

The goal is to only lint our own source code, and I think an inclusion-based target aligns better with that goal. This is also how the [TSConfig targets](https://github.com/nextstrain/nextstrain.org/blob/78fd627742e2355a2e569025c7ac90e14bc949d0/static-site/tsconfig.json#L64-L69) work.

While this means there is potential for new files to be missed, unless there is another major reshuffling, the include patterns should cover all new files.

Because it's auto-generated, I don't think we should lint `static-site/next-env.d.ts`. It passes right now, but if it fails for whatever reason in the future, there's nothing we can do about it.

package.json
jameshadfield: Thanks for digging!

PR #1089, opened 2024-12-11
blog post: Auspice 2024 summary

victorlin:
_[preview](https://nextstrain-s-victorlin--45xpgm.herokuapp.com/blog/2024-12-16-auspice-2024)_

## Description of proposed changes

A well-deserved blog post for Auspice features!

## Related issue(s)

- Closes #1049

## Checklist

- [x] Check if changes affect the [resource index JSON revision](https://docs.nextstrain.org/projects/nextstrain-dot-org/en/latest/resource-collection.html#resource-index-revisions)
- [x] Fix header anchors ([ref](https://github.com/nextstrain/nextstrain.org/pull/1089#discussion_r1879139706))
- [x] Squash all commits
- [x] Update date in filename and front matter
- [x] Update count of released versions
- [ ] Checks pass

PR #1091, opened 2024-12-12
blog: Add anchor links to headings

victorlin:
_[preview](https://nextstrain-s-victorlin--45xpgm.herokuapp.com/blog/2024-12-10-auspice-2024)_

This PR adds [GitHub-style anchor links](https://github.com/markedjs/marked/blob/10020d018d4448b1d2d4612baf650c4e3a5949bf/docs/USING_PRO.md#the-renderer--renderer) to heading elements. I also added a '#' on hover. Styles added in globals.css. Not ideal since it is also added to non-blog pages, but this was the easiest thing to do in absence of a blog-specific static CSS f

## Related issue(s)

N/A, thought of this while working on #1089

## Checklist

- [x] Works on [#1089](https://github.com/nextstrain/nextstrain.org/pull/1089) ([preview](https://nextstrain-s-victorlin--45xpgm.herokuapp.com/blog/2024-12-10-auspice-2024))
- [x] Checks pass

### DIFF
--- a/static-site/app/blog/parseMarkdown.ts
+++ b/static-site/app/blog/parseMarkdown.ts
@@ -61,8 +61,11 @@ allowedTags.push("pattern", "polygon", "polyline", "radialGradient", "rect", "se
 allowedTags.push("text", "textPath", "title", "tref", "tspan", "use", "view", "vkern");
 
 const allowedAttributes = ['href', 'src', 'width', 'height', 'alt'];
+/* "style" is not safe for untrusted code.¹ Styles should be defined in globals.css.
+ * ¹ https://stackoverflow.com/a/4547037/4410590
+ */
+
 // We add the following Attributes for SVG via https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute
-// Certain values have been excluded here, e.g. "style"
 allowedAttributes.push("accent-height", "accumulate", "additive", "alignment-baseline", "allowReorder", "alphabetic", "amplitude", "arabic-form", "ascent", "attributeName", "attributeType", "autoReverse", "azimuth");
 allowedAttributes.push("baseFrequency", "baseline-shift", "baseProfile", "bbox", "begin", "bias", "by");
 allowedAttributes.push("calcMode", "cap-height", "class", "clip", "clipPathUnits", "clip-path", "clip-rule", "color", "color-interpolation", "color-interpolation-filters", "color-profile", "color-rendering", "cursor", "cx", "cy");
--- a/static-site/app/blog/parseMarkdown.ts
+++ b/static-site/app/blog/parseMarkdown.ts
@@ -7,11 +7,11 @@ export default async function parseMarkdown(mdString: string): Promise<string> {
   const rawDescription = await marked.parse(mdString);
 
   const sanitizerConfig: IOptions = {
-    allowedTags, // see below
-    allowedAttributes: { "*": allowedAttributes }, // see below
+    allowedTags,
+    allowedAttributes: { "*": allowedAttributes },
     nonTextTags: ["style", "script", "textarea", "option"],
     transformTags: {
-      a: transformA, // see below
+      a: transformA,
     },
   };
 
--- a/static-site/app/blog/parseMarkdown.ts
+++ b/static-site/app/blog/parseMarkdown.ts
@@ -31,7 +31,7 @@ function transformA(tagName: string, attribs: Attributes): Tag {
   };
 
   const href = attribs.href;
-  if (href) {
+  if (href && !href.startsWith("#")) {
     const baseUrl = new URL(siteUrl);
     try { // sometimes the `href` isn't a valid URL…
       const linkUrl = new URL(href);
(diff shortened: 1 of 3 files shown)


static-site/app/blog/parseMarkdown.ts
```diff
@@ -62,7 +78,7 @@ allowedTags.push("text", "textPath", "title", "tref", "tspan", "use", "view", "v
 
 const allowedAttributes = ['href', 'src', 'width', 'height', 'alt'];
 // We add the following Attributes for SVG via https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute
-// Certain values have been excluded here, e.g. "style"
+allowedAttributes.push("style");
```
joverlee521, line 81: I remember we excluded `style` for a specific reason, I wish we had added as a comment here...
Tracked this down in https://github.com/nextstrain/auspice/pull/834#discussion_r357755220

static-site/app/styles/globals.css
```diff
@@ -139,3 +139,14 @@ p.wideParagraph {
...
+
+/* For blog posts */
+
+.anchor {
+   opacity: 0;
+}
+
+.anchor:hover {
+  opacity: 100;
+}
```
joverlee521, line 152: > this was the easiest thing to do in absence of a blog-specific static CSS file.

Can this be added to the [app/blog/[id]/styles.modules.css](https://github.com/nextstrain/nextstrain.org/blob/master/static-site/app/blog/%5Bid%5D/styles.module.css) as a blog-specific styling?

joverlee521 (commented):
Flagging that right now with the split between the Page and App routers, we also have [static-site/src/util/parseMarkdown.js](https://github.com/nextstrain/nextstrain.org/blob/5ca6efb84c0b0b156dd7abad94aabb60b4afb199/static-site/src/util/parseMarkdown.js). I think we want to keep the two in-sync?

static-site/app/blog/parseMarkdown.ts
victorlin: Thanks, I made a mental note to track this down but forgot. Keeping it here with reasoning in comment:

https://github.com/nextstrain/nextstrain.org/blob/ac19562745599f8150795ae7e13c37b99a54ada9/static-site/app/blog/parseMarkdown.ts#L81-L85

static-site/app/styles/globals.css
victorlin: No, CSS modules are not "static"; the classes get dynamically modified to something like

```css
.styles_anchor__1sHC6 {
   opacity: 0;
}

.styles_anchor__1sHC6:hover {
  opacity: 100;
}
```

and `className={styles.hover}` can't be used since it's not a React/Next.js component.

victorlin:
> Flagging that right now with the split between the Page and App routers, we also have [static-site/src/util/parseMarkdown.js](https://github.com/nextstrain/nextstrain.org/blob/5ca6efb84c0b0b156dd7abad94aabb60b4afb199/static-site/src/util/parseMarkdown.js). I think we want to keep the two in-sync?

Hmm, I think it's reasonable to have different Markdown parsing based on each of the 3 usages:

- Rendering blog posts (using app router version)
- Rendering the blog feeds (using pages router version)
- Rendering community/group descriptions (using pages router version)

Anchor links are only relevant for blog posts and I don't see the other two benefitting from this functionality. I can rewrite `parseMarkdown` to only add anchor links with an option (e.g. `parseMarkdown({mdString, addHeadingAnchors: true})` so that this function is still unchanged for when we convert the other two to use app router. I can also revert allowing `"style"` tags and put all the styles in globals.css.

static-site/app/styles/globals.css
joverlee521: Ah gotcha...I'm guessing it'd be possible to pass in `styles.hover` as a param to `parseMarkdown` but maybe that's just overcomplicating things.

joverlee521:
> I can rewrite parseMarkdown to only add anchor links with an option (e.g. parseMarkdown({mdString, addHeadingAnchors: true}) so that this function is still unchanged for when we convert the other two to use app router. I can also revert allowing "style" tags and put all the styles in globals.css.

+1 for both of these

static-site/app/styles/globals.css
victorlin: I've reorganized the commits and wrote more about this in c562b67f

joverlee521: > I'm guessing it'd be possible to pass in styles.hover as a param to parseMarkdown but maybe that's just overcomplicating things.

Not asking for a change, I was curious what this would look like: https://github.com/nextstrain/nextstrain.org/commit/a20a9aed7d5c7947663ee6608bd170c72653fbae.

joverlee521 (approved):

static-site/app/styles/globals.css
victorlin: Oh nice! I'll open a separate PR to consider that.

victorlin: #1092

PR #1092, opened 2024-12-12
Use page-specific CSS Module for anchor styling

victorlin:

This allows for better scoping of the style definitions.

## Related issue(s)

Follow-up to https://github.com/nextstrain/nextstrain.org/pull/1091#discussion_r1881202912

## Checklist

- [x] [Preview](https://nextstrain-s-victorlin--6dustq.herokuapp.com/blog/2024-09-03-augur-v25-features) works
- [x] Checks pass

### DIFF
--- a/static-site/app/blog/parseMarkdown.ts
+++ b/static-site/app/blog/parseMarkdown.ts
@@ -6,13 +6,20 @@ import { siteUrl } from "../../data/BaseConfig";
 export default async function parseMarkdown({
   mdString,
   addHeadingAnchors = false,
+  headingAnchorClass = undefined,
 }: {
   mdString: string
 
   /** Should `<a>` tags be added to headings? */
   addHeadingAnchors?: boolean
+
+  /** Class name for heading anchors (from page-specific CSS Module) */
+  headingAnchorClass?: string
 }): Promise<string> {
   if (addHeadingAnchors) {
+    if (headingAnchorClass === undefined) {
+      throw Error("parseMarkdown: headingAnchorClass must be defined when addHeadingAnchors is true.")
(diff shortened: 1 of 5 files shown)


static-site/app/blog/parseMarkdown.ts
```diff
@@ -6,11 +6,15 @@ import { siteUrl } from "../../data/BaseConfig";
...
   addHeadingAnchors = false,
+  headingAnchorClass = "",
 }: {
   mdString: string
 
   /** Should `<a>` tags be added to headings? */
   addHeadingAnchors?: boolean
+
+  /** Class name for heading anchors (from page-specific CSS Module) */
+  headingAnchorClass?: string
```
genehack, line 17: does it ever make sense to call `parseMarkdown({ mdstring, addHeadingAnchors: true })`, i.e., without providing a `headingAnchorClass`? Since that's only going to be used when `addHeadingAnchors` is `true`, I wonder if there's a better default than `""`, so that if somebody forgets it, we don't end up with `class=""` in the output?

This is probably nitpicky.

victorlin: Valid case. How about error?

https://github.com/nextstrain/nextstrain.org/blob/cf053361645509cd488ec421fc27ea7840313881/static-site/app/blog/parseMarkdown.ts#L20-L22

It shows a client-side error

image

and server-side error

```
[verbose]       next.js handling request for "/blog/2024-12-10-auspice-1password-issue"
Error parsing markdown: Error: parseMarkdown: headingAnchorClass must be defined when addHeadingAnchors is true.
```

PR #1096, opened 2024-12-17
manifest: add hmpv

rneher:
This adds the HMPV build to the manifest.


genehack (approved):

victorlin (approved):
Also consider adding a tile to https://nextstrain.org/pathogens by modifying [static-site/content/resource-listing.yaml:coreTiles](https://github.com/nextstrain/nextstrain.org/blob/bc19faf7ce16e01af8aeac1764851f7ef84dc015/static-site/content/resource-listing.yaml#L55-L131)

jameshadfield (changes requested):
We need to add hmpv here:

https://github.com/nextstrain/nextstrain.org/blob/d7e7a7349852fe990f1b1d88d9288b7805ef9b82/src/routing/core.js#L25-L28

Without this accessing URLs directly doesn't work (e.g. [/hmpv](https://nextstrain-s-add-hmpv-vt7cndy3.herokuapp.com/hmpv)) because the nextststrain.org routing doesn't recognse it as an Auspice dataset and thus 404s. 

Going to [/zika](https://nextstrain-s-add-hmpv-vt7cndy3.herokuapp.com/zika) and then using the dropdown _does_ work, because we're already in the Auspice SPA and thus avoid the nextstrain.org routing.

genehack:
> We need to add hmpv here:
[snip]

@rneher As I was doing similar work for yellow fever, I went ahead and updated this branch as well:

* add HMPV to core routing list per @jameshadfield 
* update manifest version number in production and testing configs
* add HMPV tree tile image
* add HMPV to /pathogens and to "featured analyses" on front page

static-site/content/featured-analyses.yaml
```diff
@@ -78,3 +78,8 @@
   description: Rabies virus evolution and spread
   img: rabies_tree.png
   url: /rabies
+
+- name: Human Metapneumovirus
```
joverlee521, line 82: Maybe abbreviate name to 'HMPV', the current display cuts off the description

Screenshot 2024-12-18 at 10 57 38 AM

genehack: good call, updated and force pushed.

jameshadfield (commented):
I presume [/pathogens](https://nextstrain-s-add-hmpv-vt7cndy3.herokuapp.com/pathogens) isn't working because the resource index got bumped to v5 which hasn't yet been built? I don't understand the versioning there so will leave to @joverlee521 to review that.

Splash page & routing are working.

@rneher / @isabeljoia - some comments from looking through the data for the first time (and knowing essentially nothing about hmpv):
* The root of b/G has negative time, I think due to the outlier `KEN/252/Dec-14`
* The B1 clade calls aren't capturing all the strains they should b/G & b/F, at least cf. the genome
* It'd be nice to use consistent colours across genome/G/F, e.g. for clade, to help with tanglegrams and general familiarity as we change between builds
* Looks like `KC562241` is clade A1 in a/F but A2b1 in a/G

genehack:
> I presume [/pathogens](https://nextstrain-s-add-hmpv-vt7cndy3.herokuapp.com/pathogens) isn't working because the resource index got bumped to v5 which hasn't yet been built? I don't understand the versioning there so will leave to @joverlee521 to review that.

Wondered about that too. 

> @rneher / @isabeljoia - some comments from looking through the data for the first time (and knowing essentially nothing about hmpv):

Maybe those are more appropriate as [HMPV repo](https://github.com/nextstrain/hmpv/) issue(s)?

env/production/config.json
```diff
@@ -110,5 +110,5 @@
   "OIDC_GROUPS_CLAIM": "cognito:groups",
   "SESSION_COOKIE_DOMAIN": "nextstrain.org",
   "GROUPS_DATA_FILE": "groups.json",
-  "RESOURCE_INDEX": "s3://nextstrain-inventories/resources/v4.json.gz"
+  "RESOURCE_INDEX": "s3://nextstrain-inventories/resources/v5.json.gz"
```
joverlee521, line 113: > I presume [/pathogens](https://nextstrain-s-add-hmpv-vt7cndy3.herokuapp.com/pathogens) isn't working because the resource index got bumped to v5 which hasn't yet been built? I don't understand the versioning there so will leave to @joverlee521 to review that.

Yup, this is not working in the test app because v5.json.gz is not generated yet. Technically this bump is not necessary because it's adding a new pathogen and doesn't change any of the existing data structure.

jameshadfield: > Technically this bump is not necessary because it's adding a new pathogen and doesn't change any of the existing data structure.

Right - let's drop this from this PR then. I just generated a new (core) index from `master` and can confirm the hmpv datasets are there.

genehack: > Right - let's drop this from this PR then.

Will do that as part of rebasing on top of the yellow fever changes.

genehack: okay, rebased on top of master with the yellow fever changes, dropped the manifest version bump commit, force-pushed, ready to rock.

joverlee521: > Ah, so -- just for my edification -- why was it needed when I added yellow fever? That was also new and didn't change anything existing.

Ah sorry, I must have been confused for yellow-fever 🤦‍♀️ . Since the old dataset existed, I thought it was updating the existing default.

static-site/content/resource-listing.yaml
```diff
@@ -133,3 +133,7 @@ coreTiles:
     img: yellow_fever_tree.png
     filters:
       - yellow-fever
+  - name: Human Metapneumovirus
```
joverlee521, line 136: Another candidate for abbreviating to 'HMPV'

Screenshot 2024-12-19 at 10 13 50 AM

Everything else looks good to me 👍

genehack: force-pushed that change; will merge after tests go green and deployed version looks good.

Issue #962, opened 2024-07-29
LOGIN button missing from mobile


victorlin:
_from @joverlee521 in https://github.com/nextstrain/nextstrain.org/pull/952#discussion_r1695826595_

The "LOGIN" button is lost in mobile view because of the presence of the logo (looking at the [/pathogens page](https://nextstrain.org/pathogens)).

![Screenshot 2024-07-29 at 1 13 16 PM](https://github.com/user-attachments/assets/356fea52-a2f0-493d-8314-dd212fc2cb8d)

A nice future improvement would be to collapse the nav links behind button at certain breakpoints, but for now maybe just hide the logo based on breakpoint?

PetWolowitz:
I found this issue on https://ovio.org/projects and would love to contribute!

victorlin:
@PetWolowitz thank you for your interest, I forgot to add the label but this is an appropriate https://github.com/nextstrain/nextstrain.org/labels/good%20first%20issue. The repo's readme has instructions on how to get things running locally – that should be a good place to get started. Let us know if you have any other questions.

bspangler21:
@victorlin I'm new to open source contributing and hope I'm not being disrespectful by jumping in here. I think I have a proposed MVP implementation here: [https://github.com/bspangler21/nextstrain.org/tree/issue-962](https://github.com/bspangler21/nextstrain.org/tree/issue-962)

Would it be okay to submit a pull request?

victorlin:
Hi @bspangler21, thank you for asking!

Right now we're in the process of a [code migration](https://github.com/nextstrain/nextstrain.org/issues/1052) so the nav bar is actually defined in two places:

- [static-site/src/components/nav-bar/index.tsx](https://github.com/nextstrain/nextstrain.org/blob/40be39bae236785e5fd477c4877ed1090c53da88/static-site/src/components/nav-bar/index.tsx) (which you've edited)
- [static-site/components/nav/index.tsx](https://github.com/nextstrain/nextstrain.org/blob/40be39bae236785e5fd477c4877ed1090c53da88/static-site/components/nav/index.tsx) (the migrated version)

I will defer to @genehack on this, but I think a PR is great if you are willing to make the changes to the migrated version of the nav bar as well. Notably, we've switched from styled-components to CSS Modules during the migration, but hopefully the same fix is still possible.

bspangler21:
@victorlin great, thank you for the guidance! I updated the migrated version as well and opened PR #1105

[...]

PR #1105, opened 2025-01-14
Issue 962 - Login button missing from mobile

bspangler21:
Update the legacy and migrated versions of the nav bar to hide the site title when on mobile, thus allowing the "Login" button to appear.

#962 
Tangentially related to #1052

### DIFF
--- a/static-site/components/nav/nav-logo.tsx
+++ b/static-site/components/nav/nav-logo.tsx
@@ -41,9 +41,11 @@ const LogoType = (): React.ReactElement => {
   );
 
   return (
-    <a href="/" className={styles.wordmark}>
-      {rainbowTitle}
-      {groupsApp && <SubTitle />}
-    </a>
+    <div className={styles.rainbowTitle}>
+      <a href="/" className={styles.wordmark}>
+        {rainbowTitle}
+        {groupsApp && <SubTitle />}
+      </a>
+    </div>
   );
 };
--- a/static-site/components/nav/nav-logo.tsx
+++ b/static-site/components/nav/nav-logo.tsx
@@ -27,11 +27,15 @@ const Logo = (): React.ReactElement => (
 );
 
 const LogoType = (): React.ReactElement => {
-  const rainbowTitle = siteTitle
-    .split("")
-    .map((letter, i) => (
-      <span key={i} style={{ color: `var(--titleColor${i})` }}>{letter}</span>
-    ));
+  const rainbowTitle = siteTitle.split("").map((letter, i) => (
+		<span
+			key={i}
+			style={{ color: `var(--titleColor${i})` }}
+			className={styles.rainbowTitle}
+		>
+			{letter}
+		</span>
+  ));
   const SubTitle = (): React.ReactElement => (
     <div>Groups Server</div>
   );
--- a/static-site/components/nav/nav-logo.tsx
+++ b/static-site/components/nav/nav-logo.tsx
@@ -27,15 +27,11 @@ const Logo = (): React.ReactElement => (
 );
 
 const LogoType = (): React.ReactElement => {
-  const rainbowTitle = siteTitle.split("").map((letter, i) => (
-		<span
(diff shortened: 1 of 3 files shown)


static-site/src/components/nav-bar/index.tsx
```diff
@@ -98,23 +98,28 @@ const NavBar = ({ minified }: {
...
-          {groupsApp &&
-            <SubTitle>Groups Server</SubTitle>}
-        </a>
-    );
+		const title = "Nextstrain";
+		const rainbowTitle = title.split("").map((letter, i) => (
+			<NavLogoCharacter key={i} $colorIndex={i}>
+				{letter}
+			</NavLogoCharacter>
+		));
```
genehack, line 106: please revert this change, it isn't necessary for the fix and adds unnecessary whitespace.

static-site/src/components/nav-bar/index.tsx
```diff
@@ -98,23 +98,28 @@ const NavBar = ({ minified }: {
...
+			color: black;
+		`;
+		const ResponsiveTitle = styled.span`
+			@media (max-width: 768px) {
+				display: none;
+			}
+		`;
+		return minified ? (
+			<div />
+		) : (
```
genehack: this isn't needed.

static-site/src/components/nav-bar/index.tsx
```diff
@@ -98,23 +98,30 @@ const NavBar = ({ minified }: {
...
+			<a href="/">
+				<ResponsiveTitle>{rainbowTitle}</ResponsiveTitle>
+				{groupsApp && <SubTitle>Groups Server</SubTitle>}
```
genehack, line 122: If we're going to hide the title, we want to hide all of it: 

```suggestion
				<ResponsiveTitle>
				  {rainbowTitle}
  				{groupsApp && <SubTitle>Groups Server</SubTitle>}
  			</ResponsiveTitle>
```

note: this may require changing L112 to be `styled.div` instead of the current `styled.span`.

static-site/components/nav/nav-logo.tsx
```diff
@@ -27,11 +27,15 @@ const Logo = (): React.ReactElement => (
 );
 
 const LogoType = (): React.ReactElement => {
-  const rainbowTitle = siteTitle
-    .split("")
-    .map((letter, i) => (
-      <span key={i} style={{ color: `var(--titleColor${i})` }}>{letter}</span>
-    ));
+  const rainbowTitle = siteTitle.split("").map((letter, i) => (
```
genehack, line 30: Please don't combine the previously split lines:

```suggestion
  const rainbowTitle = siteTitle
    .split("")
    .map((letter, i) => (
```

static-site/components/nav/nav-logo.tsx
```diff
@@ -27,11 +27,15 @@ const Logo = (): React.ReactElement => (
...
+		<span
+			key={i}
+			style={{ color: `var(--titleColor${i})` }}
+			className={styles.rainbowTitle}
```
genehack, line 34: This is not the right place for this, as we also want the subtitle hidden.

Instead of this, please handle this by either wrapping the returned value (around L44) in a `<div>` that applies this class, or just apply the class to the `<a>` that's wrapping both the title and the subtitle.

static-site/components/nav/styles.module.css
```diff
@@ -28,3 +28,9 @@
   font-size: 20px;
   font-weight: 400;
 }
+
+@media (max-width: 768px) {
+  .rainbowTitle {
+    display: none;
+  }
+}
```
genehack, line 36: Please put this at the top of the file — it's a more generic query than the `.nav` selector, so it should be above those styles.

genehack (changes requested):
Thanks for the contribution, @bspangler21 — I left some requests for changes, mainly some re-organization and removing some unneeded changes.

PR #1109, opened 2025-01-15
Fix login button on mobile [#962]

genehack:

This builds upon #1105 to add in some final (nitpicky!) tweaks to what @bspangler21 already did.

@bspangler21: Rather than make you go through another round of tweaking, I just went ahead and applied a couple additional changes on top of what you've done to better align it with the project standards. All of your commits are still present; the commits I added do the following:

* use `responsiveTitle` as the name for the CSS-ish bits in both the old and the new component
* add a linebreak in the old component so that the JSX layout between the old and new components is more obviously parallel
* restored a couple of terminal newlines to files where they were (accidentally, I assume) deleted

Thanks again for your contribution to Nextstrain.org! Let us know if you'd like to do more work on the project!


bspangler21:
@genehack Thank you for your feedback! It's been a while since I've had other eyes on my coding and I learned a lot from it (namely: disable Prettier when contributing!).

I'd love to help on more items.

genehack:
> (namely: disable Prettier when contributing!).=

Nextstrain projects (broadly but not exclusively) tend to stay away from automated linting/formatting approaches, for better or worse. (I've been working here for ~9 months and it's been an adjustment for me, relative to what I've been used to :grin:)

> I'd love to help on more items.

If there's anything in the issue list that looks interesting that is unassigned, feel free to grab it — probably a good idea to let us know what you're thinking about before you do a ton of work on it, just to make sure we're all moving in the same direction.

Issue #1093, opened 2024-12-13
Invalid blog URLs are unhandled 500 errors


victorlin:
Going to an invalid URL under `/blog` such as https://nextstrain.org/blog/invalid shows Next.js's **500: Internal Server Error** page.

<img width="403" alt="image" src="https://github.com/user-attachments/assets/9dfc42a6-9367-4fd7-ab32-627c49aa4c5f" />

## Possible solutions

Catch the error and:

1. Redirect to `/blog`
2. Show an error page similar to the one that appears when there is a Markdown parsing error:

    <img width="932" alt="image" src="https://github.com/user-attachments/assets/b85962f4-338f-44ea-95cc-e0abf38af365" />

genehack:
@victorlin this is probably my fault — I can get it unless you want it?

victorlin:
> this is probably my fault

No worries, I don't remember if we had anything in place before your blog changes. It's possible that it was already broken.

> I can get it unless you want it?

All yours!

victorlin:
Are you planning to (1) redirect or (2) show a custom error page? I've a slight preference for (2)

genehack:
> Are you planning to (1) redirect or (2) show a custom error page? I've a slight preference for (2)

I was going to make it properly 404, ala https://next.nextstrain.org/contact/foo

genehack:
Can't reproduce this locally (even using `USE_PREBUILT_STATIC_SITE=1`), but still see it happening on the live site.

jameshadfield:
Here's the log (from nextstrain-canary) when attempting /blog/invalid:

```console
$ heroku logs --app nextstrain-canary --tail
2025-02-04T01:46:12.841773+00:00 heroku[router]: at=info method=GET path="/blog/invalid" host=next.nextstrain.org request_id=aebb6c5b-9e20-42dd-ba40-b0f3bf4a4988 fwd="125.236.205.234" dyno=web.1 connect=0ms service=36ms status=500 bytes=1999 protocol=https
2025-02-04T01:46:12.809271+00:00 app[web.1]: [verbose]  next.js handling request for "/blog/invalid"
2025-02-04T01:46:12.816296+00:00 app[web.1]: Error: ENOENT: no such file or directory, scandir '/build/app/static-site/content/blog'
2025-02-04T01:46:12.816298+00:00 app[web.1]: at Object.readdirSync (node:fs:1506:26)
[...]

Issue #879, opened 2024-05-28
Enable `noImplicitAny` in TSConfig


victorlin:
#877 allows incremental adoption of TypeScript in this codebase by disabling `noImplicitAny`:

https://github.com/nextstrain/nextstrain.org/blob/68264db2acd9204cbccec531420eef6935b9c2c0/static-site/tsconfig.json#L32

With increased TypeScript usage, strictness should be preferred by ensuring every TypeScript file is typed exhaustively.

PR #1115, opened 2025-02-03
use notFound() instead of redirect("/404") [#1093]

genehack:
## Checklist

- [ ] Checks pass


victorlin (approved):

static-site/app/blog/utils.ts
```diff
@@ -22,6 +22,28 @@ export function getBlogPosts(): BlogPost[] {
...
+  try {
+    fs.statSync(postsDirectory);
+  }
+  catch (err: unknown) {
+    if (
+      typeof err === "object" &&
+      err !== null &&
+        Object.prototype.hasOwnProperty.call(err, "code") &&
+      err["code"] === "ENOENT"
+    ) {
```
victorlin, line 40: _(non-blocking)_

Simpler and should work equally well:

```suggestion
    if (
      err instanceof Error &&
      'code' in err &&
      err["code"] === "ENOENT"
    ) {
```

genehack: yeah, thanks, applied that change, that's better than the pretzel mine was twisted into by the time the linter was happy.

static-site/app/blog/utils.ts
```diff
@@ -22,6 +22,27 @@ export function getBlogPosts(): BlogPost[] {
 
   const postsDirectory = path.join(__dirname, "..", "..", "content", "blog");
 
+  // if this code is running in production, it won't have access to
+  // the postsDirectory, because it doesn't exist there — so check to
```
joverlee521, line 26: I'm totally missing something here, but why doesn't it have access to the `postsDirectory`? 
How is it generating the list of blog posts for the "Blog Archives" sidebar?

genehack: I agree it's kinda confusing. 

We build all the known blog post pages ahead of time, [thanks to this special `generateStaticParams` method](https://github.com/nextstrain/nextstrain.org/blob/master/static-site/app/blog/%5Bid%5D/page.tsx#L22-L28) — so any blog post that was in `postsDirectory` at build time already exists as a static HTML file in the build, and will get returned when that path is accessed. 

If you try to access a blog page that _doesn't_ exist, the `app/blog/[id]/page.tsx` endpoint is still hit, it calls `getBlogPosts()`, and (prior to this PR) that would blow up trying to access `postsDirectory`, which isn't there. 

This PR makes `getBlogPosts()` return an empty list when it can't read `postsDirectory`, and then [this later code in `BlogPost`](https://github.com/nextstrain/nextstrain.org/blob/master/static-site/app/blog/%5Bid%5D/page.tsx#L97-L100) can do the right thing and bounce to `notFound()`.

joverlee521: Ahh, thank you! I was missing the `generateStaticParams` piece. Okay that makes more sense now.

static-site/app/blog/utils.ts
victorlin: That would work with `err: any`, but I think it's good practice to use `unknown` instead, even if it requires a couple more type guards.

static-site/app/blog/utils.ts
jameshadfield: I'm still missing something, why does this not happen when running locally? I.e. what are the salient differences between the heroku install and a local (git repo) install such that the former doesn't have the `postsDirectory` but the latter does?

static-site/app/blog/utils.ts
tsibley: Ah, I see. Thanks!

More reading:
- https://www.typescriptlang.org/docs/handbook/2/functions.html#unknown
- https://devblogs.microsoft.com/typescript/announcing-typescript-3-0-rc-2/#the-unknown-type

tsibley: Ok, another question, if you'll humor me: why doesn't TS know that Node's `Error` class has a [`code` property](https://nodejs.org/api/errors.html#errorcode)? i.e. why is `"code" in err` necessary?

jameshadfield: Yeah, i'd like to know this too. Even if you subclass `Error` and `delete this.code` you can still access `err.code` directly, just like you can do `({}).whatever`.

genehack: > Ok, another question, if you'll humor me: why doesn't TS know that Node's `Error` class has a [`code` property](https://nodejs.org/api/errors.html#errorcode)? i.e. why is `"code" in err` necessary?

After a bit of local testing, I think it probably isn't -- this code seems to be fine:

```
    if (
      err instanceof Error &&
      err["code"] === "ENOENT"
    ) {
```

…and this version is throwing a typing error in my editor but seems to lint okay? :shrug:

```
    if (
      err instanceof Error &&
      err.code === "ENOENT"
    ) {
```

static-site/app/blog/utils.ts
genehack: This is the error that gets thrown on the server side that results in the 500 error:

```
ENOENT: no such file or directory, scandir '/build/app/static-site/content/blog'
```

That was the result of this code:

```typescript
  const __dirname = path.dirname(fileURLToPath(import.meta.url));

  const postsDirectory = path.join(__dirname, "..", "..", "content", "blog");

  const markdownFiles = fs
    .readdirSync(postsDirectory)
    .filter((fileName) => fileName.endsWith(".md"));
```

…which in Heroku, is running from this path:

```
/app/static-site/.next/server/app/blog/[id]/page.js
```

I don't know enough about how Next.js/Webpack/Heroku/whatever other build magic happens, or how all those pieces interact, to say _why_ it's not finding `postsDirectory`, but based on the error message, it very clearly wasn't. 

I did download the build artifact, and if you expand out the embedded slug, the `*.md` files are in there; perhaps the problem is the leading `/` on the `/build/app/static-site/content/blog` path?

static-site/app/blog/utils.ts
victorlin: > why doesn't TS know that Node's `Error` class has a [`code` property](https://nodejs.org/api/errors.html#errorcode)?

TypeScript limitations are showing here 😅

The [built-in type definition](https://github.com/microsoft/TypeScript/blob/71716a2868c87248af5020e33a84a2178d41a2d6/src/lib/es2022.error.d.ts#L7-L9) for `Error` doesn't contain the `code` property because it is reflective of [JavaScript's Error class](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error#instance_properties), not Node.js's Error class. The latter is [defined by `@types/node`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ab4715a8aead1731ced56619128c31e57bcc682d/types/node/v16/globals.d.ts#L167), however we can't use `err instanceof NodeJS.ErrnoException` [because it's not available at run time](https://github.com/threema-ch/threema-desktop/blob/3e1f4ea5a04ee59d745fb20a1a96e36051a51c05/src/common/node/utils.ts#L10-L13).

victorlin: > [`err["code"]`] seems to be fine

This is allowed because we [bypass `noImplicitAny`](https://github.com/nextstrain/nextstrain.org/blob/7d4d6a8c46a60b549a09201de38d5fcdf447bea4/static-site/tsconfig.json#L32). +1 for [removing the bypass](https://github.com/nextstrain/nextstrain.org/issues/879).

> [`err.code`] is throwing a typing error in my editor but seems to lint okay

It errors when running `npm run type-check`: `error TS2339: Property 'code' does not exist on type 'Error'.`

victorlin: > Even if you subclass `Error` and `delete this.code` you can still access `err.code` directly, just like you can do `({}).whatever`.

You can do many things in JavaScript that are arguably prone to error and TypeScript helps reduce the potential for error. I think for this particular case, the guardrails set up by TypeScript are not very useful, as checking `err.code === "ENOENT"` already handles the case where it is `undefined`, so something like this is sufficient:

```ts
  catch (err: any) {
    if (
      err?.code === "ENOENT"
    )
```

but it's nice to get into the habit of using `unknown` when possible.

tsibley: Thanks for helping me learn a bit more about the nuances TypeScript! 

My high-level takeaway is that I still dislike (though also still find value in) type checking that is bolted onto languages rather than an intrinsic part of them.

static-site/app/blog/utils.ts
tsibley: > perhaps the problem is the leading `/` on the `/build/app/static-site/content/blog` path?

Do you mean the leading `/build`? I think that's the issue unless I'm misunderstanding something else. It seems a build-phase path is being baked into the built artifact, and that's a problem because the path is only valid during the build and not when run in release later. The `/build` prefix is almost certainly coming from our `scripts/heroku-build`:

https://github.com/nextstrain/nextstrain.org/blob/7d4d6a8c46a60b549a09201de38d5fcdf447bea4/scripts/heroku-build#L58-L74

In the snippet @genehack quoted, `__dirname` has to be the thing starting with `/build` and I guess that has to be because `import.meta.url` does too. Is that because Next.js is fixing that value during its own build? I guess it must be, since Next.js is a big [source filter](https://metasocial.com/@trs/113077209770668069). Gah.

tsibley: I can reproduce the issue locally like this:

```console
~/nextstrain/.org $ ./build.sh static
~/nextstrain/.org $ cd ..
~/nextstrain $ mv .org .org-moved
~/nextstrain $ cd .org-moved
~/nextstrain/.org-moved $ USE_PREBUILT_STATIC_SITE=1 node server.js --verbose
```

and then visiting `http://localhost:4000/blog/invalid`. It hits `ENOENT` trying to access the old `.org` path instead of `.org-moved`.

joverlee521: Ah, I don't think Next.js is doing anything magical here. [Next.js is building within `./scripts/heroku-build`](https://github.com/nextstrain/nextstrain.org/actions/runs/13142250189/job/36671979699#step:3:696) so the `/build/app` prefix is correct in that environment?

tsibley: The Next.js build is hardcoding/embedding a fully-qualified build-time path into its build artifacts such that they're not portable at run-time. This is like putting `input: "/home/tom/data/sequences.fasta"` in a Snakemake rule that others are expected to run.

PR #1123, opened 2025-02-26
Add A/H5N1 D1.1 build to featured analyses

jameshadfield:
@lmoncla I'll leave this here for review - it adds D1.1 as a tile on the front page - not entirely convinced it's necessary right now but my hunch is that it soon will be.

image


static-site/content/featured-analyses.yaml
```diff
@@ -15,10 +15,15 @@
   url: /groups/inrb-mpox/clade-I
 
 - name: H5N1 cattle outbreak
-  description: Analysis of the ongoing influenza A/H5N1 cattle outbreak
+  description: Analysis of the ongoing B3.13 outbreak
   img: h5n1_cattle_outbreak_tree.png
   url: /avian-flu/h5n1-cattle-outbreak/genome
 
+- name: H5N1 D1.1 outbreaks
```
victorlin, line 22: Suggestion: add "avian" somewhere in the title (general public doesn't know that H5N1 implies avian)

lmoncla:
Thanks @jameshadfield! I agree. D1.1 is the predominant strain in North America right now, and I think it is good to have this as a tile, since its essentially replacing other prior lineages.

trvrb:
I'd add to the list of tiles, but I don't think that D1.1 genome build is impactful enough to live above the fold. Somehow we lost the general "Avian influenza" tile. This should get added back and link to nextstrain.org/avian-influenza. This should live after "SARS-CoV-2 forecasts". After this "Avian influenza" tile should be the D1.1 tile.

I don't like "H5N1 D1.1 outbreaks". This is continuous circulation among birds for the most part and an outbreak in cattle nested within it. I'd rename to something closer to "Avian influenza H5N1 D1.1" (whatever fits).

victorlin:
@trvrb "Avian influenza" was removed in a075d29000eccaa0385a02ae2155f10828e2c3f3 😄

I've translated your comment into PR changes. Feel free to make additional changes. Do you think the "H5N1 cattle outbreak" tile placement should stay as-is?

static-site/content/featured-analyses.yaml
victorlin: Renamed to "Avian influenza (D1.1)".

trvrb:
Thanks @victorlin! I just made some additional changes and will merge.

PR #1127, opened 2025-03-11
Add global, wa, lineage-1A, and lineage-2 trees to WNV display

j23414:

PR that comes after merging the following WNV PRs

* https://github.com/nextstrain/WNV/pull/70
* https://github.com/nextstrain/WNV/pull/61

## Checklist

- [ ] Checks pass

### DIFF
--- a/data/manifest_core.json
+++ b/data/manifest_core.json
@@ -778,8 +778,12 @@
       },
       "WNV": {
         "resolution": {
+          "global": "",
+          "wa": "",
+          "lineage-1A": "",
+          "lineage-2": "",
           "NA": "",
-          "default": "NA"
+          "default": "global"
         }
       },
       "yellow-fever": {
--- a/test/date_descriptor.test.js
+++ b/test/date_descriptor.test.js
@@ -302,20 +302,20 @@ describe("Valid narratives", () => {
 })
 
 const redirects = [
-  ["WNV", "WNV/NA"],
+  ["WNV", "WNV/global"],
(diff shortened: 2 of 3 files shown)


data/manifest_core.json
```diff
@@ -778,8 +778,12 @@
       },
       "WNV": {
         "resolution": {
+          "global": "",
+          "wa": "",
+          "lineage-1A": "",
+          "lineage-2": "",
           "NA": "",
-          "default": "NA"
+          "default": "global"
```
victorlin, line 786: Re: failing checks, since the default changed, this test on the default needs to be updated:

https://github.com/nextstrain/nextstrain.org/blob/2998b8067be993d6ebc7f6f5731c06629ada95e4/test/date_descriptor.test.js#L312

j23414: Looking at "git blame", @jameshadfield, do we still need to hardcode a redirect from "WNV" to "WNV/NA"? I'm trying to get the context for this date_descriptor check to figure out what to change here.

victorlin: No problem! I think it's fine to rely on PR checks to know when a test should be changed. In this case it required scrolling up more in the [run output](https://github.com/nextstrain/nextstrain.org/actions/runs/13799033821/job/38673874498) to see these lines:

```
FAIL test/inventory_parsing.test.js
  ● core URL redirects (via our manifest) are correctly obtained

    expect(received).toEqual(expected) // deep equality

    Expected: "WNV/NA"
    Received: "WNV/global"
```

genehack (approved):

PR #1131, opened 2025-03-17
Convert staging to app router [#1067]

### LINKED ISSUE #1067: Convert `/staging/[[...staging]].jsx` to App Router
This is part of the [Convert static-site from Pages Router to App Router](https://github.com/nextstrain/nextstrain.org/issues/1052) work. See there for additional context.

This page uses essentially the same set of components as the `/pathogens` page. Once #1066 is completed, this should be a small amount of additional work, moving this page into `/static-site/app/`, pointing it to the new versions of components added during the `/pathogens` conversion, and updating the dynamic routing code. (We should be able to render all versions of this server-side, I believe?)

genehack:

Converts `/staging` over to the App Router style; see individual commit comments for details. 

[Preview](https://nextstrain-s-convert-st-d0asv2.herokuapp.com/staging) — note that the [error message](https://nextstrain-s-convert-st-d0asv2.herokuapp.com/staging/foo) for a bad dataset looks slightly different, due to how the base App Router layout is done. This feels like an acceptable modification. 

## Related issue(s)

Closes #1067 

## Checklist

- [x] Checks pass


static-site/components/list-resources/index.tsx
```diff
@@ -29,13 +29,15 @@ import useSortAndFilter from "./use-sort-and-filter";
 import useDataFetch from "./use-data-fetch";
 
 import { pathogenResourceListingCallback } from "../../app/pathogens/callback";
+import { stagingResourceListingCallback} from "../../app/staging/[[...staging]]/callback";
```
victorlin, line 32: Is `staging/[[...staging]]` necessary and not `pathogens/[[...pathogens]]` because there is `nextstrain.org/staging/*` and not `nextstrain.org/pathogens/*`? I was initially confused because this was not the case before even with the same routes. Might be worth noting somewhere.

static-site/components/list-resources/index.tsx
```diff
@@ -142,14 +147,27 @@ function ListResourcesContent({
...
+  switch(resourceName) {
+    case "pathogens":
+      resourceListingCallback = pathogenResourceListingCallback;
+      break;
+    case "staging":
+      resourceListingCallback = stagingResourceListingCallback;
+      break;
+    default:
+      throw new InternalError(`${resourceName} is not a valid resource name`);
+  }
```
victorlin, line 164: Suggestion: pass callback function directly (ac4e552995d4b28dcd6ed7858626cff4b6e24993)

static-site/components/list-resources/index.tsx
genehack: Memorializing conversations I had with both @tsibley and @victorlin about this:

As far as I can tell, the reason we have `/staging/[[...staging]]/` (and `/sars-cov-2/[[...sars-cov-2]]` and `/ncov/[[...ncov]]`) is to enable the display of the `<ErrorMessage>` component when somebody visits, e.g., https://nextstrain.org/staging/foo (sidebar: the page rendering is slightly broken for this in prod; in this PR, the full staging page actually loads under the error banner message). 

There are some Express-level routes for specific URLs that are intercepted and displayed before the Next.js-level routing comes into play, so if the Next.js router is seeing a value for the "splat" part of the URL, we can be sure that the requested dataset does not exist, and displaying the error message (and offering up the full list of resources) is an appropriate thing to do.

Note: this isn't currently documented as far as I can tell; I am inferring that this is for the error message display only because we don't do anything with the `[[...staging]]` bits, other than trigger the error message if it is set to _anything_, and use that anything in the error message that is displayed. 

- [ ] I will add some commentary to `/staging/[[...staging]]/page.tsx` that spells this out a little bit more explicitly, and propagate that commentary to additional places where we have this behavior as they get ported over to App Router.

Currently, when this error message is displayed, we are still returning a 200 status code, which is arguably incorrect; it should be a 404, even though we are displaying the static page. 

- [ ] I will update this PR so that, on the conditional branch where the `<ErrorMessage>` component is displayed, the HTTP response status will be 404, not 200, and I will propagate that behavior to the other places where we have this behavior as they get ported to App Router.

Also currently (as @victorlin notes!), we do _not_ have this behavior for the `/pathogens/` page -- loading, for example, https://nextstrain.org/pathogens/foo, returns the regular site 404 page (and a 404 status code). 

- [ ] I am going to create a new sub-task in this overall "port to App Router" issue for converting `/pathogens/page.tsx` to `/pathogens/[[...pathogens]]/page.tsx, add in the conditional `<ErrorMessage>` display when `[[...pathogens]]` is set, and retain the 404 status code. (I.e., make the pathogens page act like staging and the others, rather than maintaining the current inconsistent behavior.)

I think that covers all the aspects of this that I have discussed with Tom and Victor, but if I left anything out, I invite them both to comment here.

static-site/components/list-resources/index.tsx
genehack: Again, memorializing a decision made in a conversation: I will move the callback usage in the direction Victor suggests here; I had previously tried this approach and failed, not understanding that the callback file needs to be a React Client Component.

static-site/components/list-resources/index.tsx
genehack: This ended up being kinda more hinky than I expected, but I got it done. 

[Commentary on how it works](https://github.com/nextstrain/nextstrain.org/blob/convert-staging-to-app-router-1067/static-site/app/staging/%5B%5B...staging%5D%5D/page.tsx#L16).

genehack: Also, I have created #1133 for the third item above.

genehack:
Okay, @victorlin, I think I addressed all the issues you raised; particularly interested in any feedback on [the comment about how /staging versus /staging/foo works now](https://github.com/nextstrain/nextstrain.org/blob/convert-staging-to-app-router-1067/static-site/app/staging/%5B%5B...staging%5D%5D/page.tsx#L16).

I will plan to merge this early next week, sans additional feedback.

static-site/components/list-resources/index.tsx
tsibley: Thank you!

There's some reasoning/commentary in the original commit message which introduced these pages: 669c95302f57ec6b8280a7f63829b61b93539992

static-site/app/staging/[[...staging]]/validateStagingUrl.tsx
```diff
@@ -0,0 +1,23 @@
...
+  * redirected at the Express router level, before the Next.js router
+  * is engaged, so if we are trying to render a `/staging/<something>`
+  * URL, it _is_ an error
+ */
+export default function ValidateStagingUrl(): null {
+  const params = useParams();
+
+  if (params && params["staging"]) {
+    notFound();
+  }
```
Comment (victorlin) [inline static-site/app/staging/[[...staging]]/validateStagingUrl.tsx:20]: I wonder if another name would be more informative here. Maybe `staging/[[...slug]]/…` to match [Next.js docs](https://nextjs.org/docs/pages/building-your-application/routing/dynamic-routes#optional-catch-all-segments)? Same would go for `pathogens/[[...slug]]`. `blog/[id]` is meaningful as-is.


```suggestion
  if (params && params["slug"]) {
    notFound();
  }
```

Not the fault of this PR since the old implementation [uses [[...staging]]](https://github.com/nextstrain/nextstrain.org/blob/db66c00f9ca610978ec4df6e46570b9858a33611/static-site/pages/staging/%5B%5B...staging%5D%5D.jsx)

static-site/app/staging/[[...staging]]/not-found.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+
+import StagingPageContent from "./content";
+import ErrorBanner from "./error-banner";
+import { metadata } from "./page";
+
+/**
+ * A React Server component that renders the usual `/staging` page
+ * content, with an error banner up-top explaining that the requested
+ * dataset doesn't actuall exist.
```
Comment (victorlin) [inline static-site/app/staging/[[...staging]]/not-found.tsx:10]:
```suggestion
 * dataset doesn't actually exist.
```

victorlin (approved):
Didn't test anything but the code looks good. Only comments are a non-blocking discussion for naming and a small typo correction.

static-site/app/staging/[[...staging]]/validateStagingUrl.tsx
Comment (genehack) [reply static-site/app/staging/[[...staging]]/validateStagingUrl.tsx:20]: Opened #1137 for this.

PR #1135, opened 2025-03-21
Migrate WA tree from WNV list to NW-PaGe

j23414:

Part of migration of WA WNV trees to NW-PaGe

## Checklist

- [x] Checks pass
- [ ] ~~Optional step: Add NW-PaGe to manifest~~ [Edit: 2025-03-24] Noop, migrate to Nextstrain group so different system

### DIFF
--- a/data/manifest_core.json
+++ b/data/manifest_core.json
@@ -779,7 +779,6 @@
       "WNV": {
         "resolution": {
           "all-lineages": "",
-          "wa": "",
           "lineage-1A": "",
           "lineage-2": "",
           "NA": "",


victorlin:
> Optional step: Add NW-PaGe to manifest

I assume migration to NW-PaGe is to either a Nextstrain group or community dataset. If that's the case, it doesn't make sense to add to `data/manifest_core.json` which is only for core datasets.

j23414:
Thanks for the catch! Agree, the migration afaik is to a Nextstrain group, not [core](https://github.com/nextstrain/nextstrain.org/blob/master/data/manifest_core.json), or [community](https://github.com/nextstrain/nextstrain.org/blob/master/static-site/content/community-datasets.yaml). 

I'll update this PR as ready for review

data/manifest_core.json
victorlin, line 1: _(non-blocking)_

> Check if changes affect the [resource index JSON revision](https://docs.nextstrain.org/projects/nextstrain-dot-org/en/latest/resource-collection.html#resource-index-revisions)

It looks like yes, changes to `data/manifest_core.json` do affect the index, so the revision number in configs should be bumped according to docs.

However, this was not done for several other recent changes to this file (8a1e19d7b972c95d7684a27655189c64a287f691, 8ece4d382b12801aea83a7a55d47a2690848ce59, 41855e5ecf607f8d25783c0d20157bcef1a0683d) yet both https://next.nextstrain.org/pathogens and https://nextstrain.org/pathogens seem to reflect those changes, so I'm not sure an update is actually necessary.

data/manifest_core.json
```diff
@@ -779,7 +779,6 @@
       "WNV": {
         "resolution": {
           "all-lineages": "",
-          "wa": "",
```
victorlin, line 782: This renders the URL https://nextstrain.org/WNV/wa invalid. Is the new NW-PaGe dataset ready? It should be possible to encode a redirect in [src/redirects.js](https://github.com/nextstrain/nextstrain.org/blob/db66c00f9ca610978ec4df6e46570b9858a33611/src/redirects.js) from `/WNV/wa` to `/groups/…` if desired. Note that it would be a redirect for the direct URL, not an entry in the dataset selector which is based on manifest.

victorlin: Wait, actually it will still work, but we don't want that as it'll become out of date if it's not automatically updated by the workflow: https://nextstrain-s-drop-wnv-w-o8gyt6.herokuapp.com/WNV/wa

To properly remove `/WNV/wa`, it needs to be removed from S3.

j23414: Since we just migrated last week and I know some people are busy this week, I'd estimate that the NW-PaGe dataset might not be ready until April 4th or later.

Good to know about the redirects! I can ping NW-PaGe next week for an update/if they would like the redirect option. In the meantime, I don't think we need to maintain a valid URL https://nextstrain.org/WNV/wa as it seems like NW-PaGe would like to manage the WA view themselves

j23414: > To properly remove /WNV/wa, it needs to be removed from S3.

I'm happy to `nextstrain remote rm` from S3

victorlin: The redirect would be more for ourselves so we don't break our own URL. I see that `/WNV/wa` was only recently added in ff1e6d25638da49c9a23e1e1468448b90dd20535 so maybe it's fine to break. Another option is to redirect to an existing core dataset, i.e. `/WNV/wa → /WNV/all-lineages`. I don't have any opinions here, just want to list out some options.

victorlin (approved):
I assume this needs coordination with the workflow side of things, i.e. https://github.com/nextstrain/WNV/issues/81, so I trust your judgement on when to merge this.

data/manifest_core.json
j23414: That's fair! Thanks for listing the other options, and this thread does help clarify my thinking on this and similar PRs 😄 No worries!

I lean toward not supporting `/WNV/wa -> /WNV/all-lineages` redirects as `/WNV/wa` was only recently added, and was fated to be dropped anyway. And as a user, I'd be confused on why "WNV/wa (state)" was sending me to `WNV/all-lineages (global)`

victorlin: > I'm happy to `nextstrain remote rm` from S3

👍

> I probably need to remove some old non-updating "global".

You could also redirect `/WNV/global` to `/WNV/all-lineages` (in addition to removing the global JSON from S3).

PR #1140, opened 2025-03-27
Link to BA.2-derived viruses for 6m ncov builds

trvrb:
Circulating SARS-CoV-2 viruses are increasing evolved from the root of the tree. The default tree view should focus more on circulating viruses. A simple way to do this to have the default zoom be the most recent clade prior to the common ancestor of circulating viruses. At the moment this is clade 21L (lineage BA.2).

In the future we may be able to set these defaults as part of the Auspice JSON, but at the moment this isn't possible. So instead, this commit updates links to 6m builds to include the URL param that zooms to clade 21L.

### DIFF
--- a/static-site/content/featured-analyses.yaml
+++ b/static-site/content/featured-analyses.yaml
@@ -7,7 +7,7 @@
 - name: SARS-CoV-2
   description: Ongoing evolution and spread of SARS-CoV-2
   img: ncov_freq.png
-  url: /ncov/gisaid/global/6m
+  url: /ncov/gisaid/global/6m?label=clade:21L%20%28BA.2%29
 
 - name: Mpox in the DRC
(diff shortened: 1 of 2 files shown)


genehack (approved):

jameshadfield (changes requested):
The 21L branch label has disappeared from the latest global tree, and thus you'll get an error message

> The label 'clade' value '21L (BA.2)' was not found in the tree - no zoom performed

(Previous builds had it, e.g. [2025-03-20](https://next.nextstrain.org/ncov/gisaid/global/6m@2025-03-20?label=clade:21L%20%28BA.2%29).) 

I think the 21L mutations (21M + `?gt=nuc.21618T,22775A`) are correctly found, but that same branch is subsequntly (re-)labelled as 22B.

PR #1152, opened 2025-04-22
Add /workshop redirect link

victorlin:

Workshop attendees will be typing a URL into their browser as the first step. 'https://nextstrain.org/workshop' is easier to type than 'https://github.com/nextstrain/nextstrain-walkthrough', plus it looks nicer.

## Checklist

- [x] Checks pass
- [x] Redirect works on review app: https://nextstrain-s-victorlin--73x4gz.herokuapp.com/workshop

### DIFF
--- a/src/redirects.js
+++ b/src/redirects.js
@@ -122,7 +122,8 @@ const setup = (app) => {
     "/help/coronavirus/FAQ": `${mainReadTheDocs}/learn/pathogens/coronavirus/FAQ.html`,
     "/help/coronavirus/SARS-CoV-2": `${mainReadTheDocs}/learn/pathogens/coronavirus/SARS-CoV-2.html`,
     "/help/coronavirus/Technical-FAQ": `${mainReadTheDocs}/learn/pathogens/coronavirus/Technical-FAQ.html`,
-    "/help/coronavirus/human-CoV": `${mainReadTheDocs}/learn/pathogens/coronavirus/human-CoV.html`
+    "/help/coronavirus/human-CoV": `${mainReadTheDocs}/learn/pathogens/coronavirus/human-CoV.html`,
+    "/workshop": "https://github.com/nextstrain/nextstrain-walkthrough"
   };
 
   for (const [from, to] of Object.entries(docsRedirects)) {


tsibley:
@victorlin What's the longer-term plan here? Is https://nextstrain.org/workshop going to get re-pointed in the future? ([I hope not](https://www.w3.org/Provider/Style/URI).) Is https://github.com/nextstrain/nextstrain-walkthrough going to be made timeless and generic instead of specific to the GSEMOR workshop next week?

tsibley:
I think it's unfortunate if a URL given to a workshop attendee stops working at some future point after they've gone home, or even more confusingly, starts pointing to something similar but _different_ than what they saw (e.g. when we update `nextstrain-walkthrough` for a new workshop audience down the road).

But I get the desire to have an easy single entrypoint URL for workshops. My suggestion is to use a short moniker for each workshop, e.g. `nextstrain.org/workshop/gsemore2025` or similar. I would also make a branch for each workshop where workshop-specific changes are made and have the redirect go specifically to that branch rather than whatever's on `main`. This approach makes both the URL and content stable for attendees to refer back to.

> I realize most of this was done without any explicit approval.

And that's totally fine! :-)

victorlin:
Thanks, agree with all of that. Since I've already shared `nextstrain.org/workshop`, I see two options:

1. Update `nextstrain.org/workshop` to a static reference of what was shown that day (currently tagged as [nextstrain-walkthrough@uf-epi-2025](https://github.com/nextstrain/nextstrain-walkthrough/releases/tag/uf-epi-2025)). Future workshops could start using the `nextstrain.org/workshop/<event>`.
2. Make `nextstrain.org/workshop` a web page that lists out all workshops. It would be an extra step for last week's attendees, but less prone to error that could happen if someone at a hypothetical `nextstrain.org/workshop/uf-epi-2026` only types `nextstrain.org/workshop` or makes a typo and ends up at `nextstrain.org/workshop/uf-epi-2025` which may be misleading with slightly different contents.

genehack:
>     2. Make `nextstrain.org/workshop` a web page that lists out all workshops. 

+1 for 2.

tsibley:
+1 for 2.

PR #1155, opened 2025-04-25
Pathogen resolution endpoints

tsibley:
See commit messages.

## Checklist

- [x] Checks pass

### DIFF
--- a/src/endpoints/pathogenRepos.js
+++ b/src/endpoints/pathogenRepos.js
@@ -1,0 +1,158 @@
+import { fetch, paginatedFetch } from "../fetch.js";
+import { BadRequest, InternalServerError, NotFound } from '../httpErrors.js';
+import { contentTypesProvided } from "../negotiate.js";
+import { uri } from '../templateLiterals.js';
+import { slurp } from '../utils/iterators.js';
+
+
+const headers = {
+  Authorization: process.env.GITHUB_TOKEN
+    ? `token ${process.env.GITHUB_TOKEN}`
+    : "",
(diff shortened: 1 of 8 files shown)


genehack (approved):

src/endpoints/pathogenRepos.js
```diff
@@ -0,0 +1,158 @@
+import { fetch, paginatedFetch } from "../fetch.js";
+import { BadRequest, InternalServerError, NotFound } from '../httpErrors.js';
+import { contentTypesProvided } from "../negotiate.js";
+import { uri } from '../templateLiterals.js';
+import { slurp } from '../utils/iterators.js';
+
+
+const headers = {
+  "Authorization": process.env.GITHUB_TOKEN
```
joverlee521, line 9: Confirmed for my own sanity that the `GITHUB_TOKEN` we use has no additional permissions, so that this should not expose any private repos.

joverlee521 (approved):
Thanks for updating the docs! Changes LGTM (except whatever is making the linter unhappy 🙃)

src/endpoints/pathogenRepos.js
tsibley: Yep, it's the existing token we use for other unprivileged GitHub requests. Thanks for checking!

tsibley:
I have [appeased the linter](https://github.com/nextstrain/nextstrain.org/compare/33180b9b3952cd5512304bd2d77d0e9f30572f61..13dcbe3542b6b4c772def7833fde17514951a8b5).

PR #1158, opened 2025-04-28
Update linting targets

victorlin:

Align ESLint and TSConfig linting targets to be inclusion-based.

## Related issue(s)

[from Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1745612337373529)

## Checklist

- [x] Checks pass


static-site/tsconfig.json
```diff
@@ -61,7 +61,6 @@ be set and will rewrite this file if they aren't set.²
     ]
   },
   "include": [
-    ".next/types/**/*.ts",
```
tsibley, line 64: I guess this is unnecessary to include Next.js' internal type defs?

static-site/.eslintrc.yaml
```diff
@@ -24,7 +24,13 @@ settings:
...
 
 ignorePatterns:
-  - public/
+  - "/*"
+  - "!/app"
+  - "!/components"
+  - "!/pages"
+  - "!/src"
+  - "!/types.d.ts"
+  - "!/vendored"
```
tsibley, line 33: Does this intentionally omit things like `data/BaseConfig.js` and `data/SiteConfig.jsx`? I'd think we'd want to keep linting those.

.eslintrc.yaml
```diff
@@ -11,10 +11,12 @@ globals:
...
-  - auspice-client/
-  - docs/
-  - node_modules/
-  - static-site/
+  - "/*"
+  - "!/resourceIndexer"
+  - "!/scripts"
+  - "!/server.js"
+  - "!/src"
+  - "!/test"
```
tsibley, line 19: Do we want to keep linting `jest.config.js`?

package.json
```diff
@@ -15,8 +15,8 @@
     "build": "./build.sh",
     "build:feeds": "node scripts/generate-blog-feeds.js",
     "lint": "npm run lint:server",
-    "lint:server": "DEBUG=eslint:cli-engine npx eslint --max-warnings=0 --ext .js,.jsx .",
-    "lint:static-site": "cd static-site && DEBUG=eslint:cli-engine npx eslint --max-warnings=0 --ext .js,.jsx,.ts,.tsx .",
+    "lint:server": "DEBUG=eslint:cli-engine npx eslint --max-warnings=0 .",
+    "lint:static-site": "cd static-site && DEBUG=eslint:cli-engine npx eslint --max-warnings=0 .",
```
tsibley, line 19: Tangential, but can `--max-warnings=0` move into the eslint configs?

tsibley (approved):

.eslintrc.yaml
victorlin: I don't see why not. Added back in force-push:

https://github.com/nextstrain/nextstrain.org/blob/6e1ec7f8d094214e3b3de66e8359f7d40f9d4282/.eslintrc.yaml#L15

static-site/.eslintrc.yaml
victorlin: Good point, done: 1526d25a974f9133b50e1035d666a49b853ba2bf

package.json
victorlin: [Not possible](https://github.com/eslint/eslint/issues/16804#issuecomment-1460887876) 😕

static-site/tsconfig.json
victorlin: I'm not sure what it's used for but it does seem like Next.js wants it there and running `next lint` will add it back. I've added it back in 78d8b9a6accbed2502241071c3ddd674ce72b6b7.

Maybe we should be running `next lint` instead of ESLint on static-site?

jameshadfield: > running `next lint` will add it back

next.js has a habit of silently modifying configs, e.g. cc763094726d7ea6501f8087da026d880fddd3e1

PR #1161, opened 2025-04-29
Record Plausible analytics events for endpoints intended for Nextstrain CLI

tsibley:
These will help us in the future gain insight into Nextstrain CLI version usage and update patterns as well as pathogen repository setup and update patterns.

## Checklist

- [x] Checks pass

### DIFF
--- a/src/analytics.js
+++ b/src/analytics.js
@@ -1,0 +1,85 @@
+/**
+ * Analytics for server-side events.
+ *
+ * @module analytics
+ */
+import { default as UserAgentBag } from "user-agent-bag";
+import { PLAUSIBLE_ANALYTICS_DOMAIN, PLAUSIBLE_ANALYTICS_ENDPOINT } from "./config.js";
+import { fetch } from "./fetch.js";
+
+
+/**
+ * Generate middleware to record an analytics event via a request to
+ * Plausible's API.
+ *
+ * Intended for use in an Express routing chain.  The generated middleware
+ * function does not block on the analytics request before returning.
+ *
+ * If the User-Agent of the incoming request starts with "Nextstrain-CLI/",
+ * then it is parsed for several bits of information and those are
+ * automatically attached to the analytics event as custom properties.
+ *
+ * @function recordEvent
+ * @param {object} data
+ * @param {string} data.name - Event name.  Defaults to "pageview" (Plausible's default).
+ * @param {object} data.props - Custom properties.  Defaults to {}.
+ * @param {boolean} data.interactive - Interactive session or not?  Defaults to
+ *   false (opposite of Plausible's default), with the reasoning that most events
+ *   we're likely to record server-side aren't for interactive browser-based
+ *   sessions.
+ * @returns {expressMiddlewareAsync}
+ */
+export const recordEvent = ({name = "pageview", props = {}, interactive = false} = {}) => async (req, res, next) => {
+  if (PLAUSIBLE_ANALYTICS_DOMAIN) {
+    /* We intentionally do not "await fetch()" as we do not want to block
+     * request processing on analytics; we'd rather ignore failures.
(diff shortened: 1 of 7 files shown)


tsibley:
This appears to work with actual submission to Plausible, BTW.

With Nextstrain CLI built from https://github.com/nextstrain/cli/pull/434, I ran:

```
envd mitmproxy NEXTSTRAIN_DOT_ORG=https://nextstrain-s-trs-cli-an-7zooyl.herokuapp.com/ -- nextstrain check-setup docker
```

and I saw the appropriate request/response in mitmproxy and [I see the event](https://plausible.io/nextstrain.org?f=is,hostname,nextstrain-s-trs-cli-an-7zooyl.herokuapp.com&period=7d&keybindHint=W) in Plausible when I filter to the review app's hostname.

genehack (approved):

src/analytics.js
```diff
@@ -0,0 +1,85 @@
...
+ */
+export const recordEvent = ({name = "pageview", props = {}, interactive = false} = {}) => async (req, res, next) => {
+  if (PLAUSIBLE_ANALYTICS_DOMAIN) {
+    /* We intentionally do not "await fetch()" as we do not want to block
+     * request processing on analytics; we'd rather ignore failures.
+     *    -trs, 28 April 2025
+     */
+    fetch(
+      // <https://plausible.io/docs/events-api>
+      PLAUSIBLE_ANALYTICS_ENDPOINT, {
```
joverlee521, line 40: I'm surprised this doesn't require some king of API key, but maybe that's normal for tracking?

joverlee521 (approved):

src/analytics.js
```diff
@@ -0,0 +1,85 @@
...
+};
+
+
+const userAgentToPlausibleKeys = new Map([
+  ["Nextstrain-CLI",  "nextstrain-cli/version"],
+  ["Python",          "nextstrain-cli/python"],
+  ["installer",       "nextstrain-cli/installer"],
+  ["platform",        "nextstrain-cli/platform"],
+  ["tty",             "nextstrain-cli/tty"],
+]);
```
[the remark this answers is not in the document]
tsibley: Yep!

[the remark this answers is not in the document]
tsibley: I was going to leave them for ad-hoc analytics viewing until we settle on what's the most useful view (if we do).

src/analytics.js
tsibley: It's pretty normal for analytics that are focused on the client-side (like Plausible), since there is no way to have any actual secrets without server-side involvement. Analytics spam is totally a thing that happens as a result, though, and as such providers like Plausible have to do a bit of spam/abuse prevention work.

src/analytics.js
```diff
@@ -0,0 +1,85 @@
...
+ *   sessions.
+ * @returns {expressMiddlewareAsync}
+ */
+export const recordEvent = ({name = "pageview", props = {}, interactive = false} = {}) => async (req, res, next) => {
+  if (PLAUSIBLE_ANALYTICS_DOMAIN) {
+    /* We intentionally do not "await fetch()" as we do not want to block
+     * request processing on analytics; we'd rather ignore failures.
+     *    -trs, 28 April 2025
+     */
+    fetch(
```
tsibley, line 38: I'm a little hesitant to use the interpreter's event loop/task queue here as our own offline work queue, as it seems like badness awaits if we start clogging it up with requests that take a long time to complete (e.g. if network cxn to Plausible is very slow to connect or transfer data). And it's not a queue we can manage on our own or have much insight into by default. But it seems an ok place to start, probably, esp. in the absence of some other obvious work queue in our stack.

tsibley:
https://plausible.io/nextstrain.org?f=contains,page,/cli,/pathogen-repos

PR #1163, opened 2025-04-30
Make use of Plausible Analytics configurable [#1162]

### LINKED ISSUE #1162: Analytics should be configurable, not hardcoded
### Current Behavior

[Slack context](https://bedfordlab.slack.com/archives/C04RRD7DEQN/p1745985905354629)

Quoting Tom; TL,DR is there are two places where analytics via Plausible is hard-coded in the source when it should be configurable: 

> Unfortunately the last two are source code modifications instead of config file modifications. We should update that code to a) look to the server config file for the analytics domain instead of hardcoding it and b) omit analytics entirely if the config key is unset. 
> 
> [static-site/pages/_app.tsx](https://github.com/nextstrain/nextstrain.org/blob/eb9c4a54f9d35839cb0e3208324fdba95ebce353/static-site/pages/_app.tsx#L32)
> [static-site/components/plausible-analytics/index.tsx](https://github.com/nextstrain/nextstrain.org/blob/eb9c4a54f9d35839cb0e3208324fdba95ebce353/static-site/components/plausible-analytics/index.tsx#L12)

### Expected behavior

Change the source as described above.

genehack:

Adds config values to BaseConfig for script URL and `data-domain` prop. Setting the config values to `undefined` will disable inclusion of the analytics script completely.

## Related issue(s)

Closes #1162 

## Checklist

- [ ] Checks pass

### DIFF
--- a/static-site/components/plausible-analytics/index.tsx
+++ b/static-site/components/plausible-analytics/index.tsx
@@ -1,17 +1,31 @@
 import Script from "next/script";
 import React from "react";
 
+// import from BaseConfig to avoid a conflict with SiteConfig, due to
+// it importing an older component that's still using styled.
+//
+// TODO convert over to SiteConfig once this is no longer true.
+import {
+  plausibleAnalyticsDataDomain,
+  plausibleAnalyticsScript,
+} from "../../data/BaseConfig";
+
 /** React Server Component to add analytics javascript */
 export default function PlausibleAnalytics(): React.ReactElement {
-  /* See <https://plausible.io/docs/plausible-script>. Analytics are disabled on
-     localhost, but use `script.local.js` if you want to track localhost for
-     testing purposes */
-  return (
-    <Script
-      src="https://plausible.io/js/script.js"
-      data-domain="nextstrain.org"
-      async
-      defer
-    />
-  );
+  /* See <https://plausible.io/docs/plausible-script>. Analytics are
+     disabled on localhost; see comments in BaseConfig if you want to
+     change configuration, enable analytics for localhost, or disable
+     them completely. */
+  if (plausibleAnalyticsDataDomain && plausibleAnalyticsScript) {
+    return (
+      <Script
+        src={plausibleAnalyticsScript}
+        data-domain={plausibleAnalyticsDataDomain}
+        async
+        defer
+      />
+    );
+  } else {
+    return <></>;
+  }
 }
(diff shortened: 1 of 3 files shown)


static-site/data/BaseConfig.js
```diff
@@ -30,3 +30,13 @@ export const blogFeeds = {
...
 export const blogFeedUrls = Object.fromEntries( Object.entries(blogFeeds).map(([k,v]) => [k, `${blogUrl}/${v}`]));
+
+/*
+  See <https://plausible.io/docs/plausible-script>. Analytics are
+  disabled on localhost, but use `script.local.js` if you want to
+  track localhost for testing purposes. If you want to disable all
+  analytics, set both of these constants to `undefined`.
+*/
+export const AnalyticsSource = "https://plausible.io/js/script.js";
+export const AnalyticsDomain = "nextstrain.org";
```
tsibley, line 41: I'd suggest

```suggestion
export const analyticsSource = "https://plausible.io/js/script.js";
export const analyticsDomain = "nextstrain.org";
```

to match the rest of the vars in this file.

And maybe also make the names specific to Plausible since they're tightly coupled to it?


```suggestion
export const plausibleAnalyticsScript = "https://plausible.io/js/script.js";
export const plausibleAnalyticsDomain = "nextstrain.org";
```

static-site/components/plausible-analytics/index.tsx
```diff
@@ -1,17 +1,24 @@
 import Script from "next/script";
 import React from "react";
 
+import { AnalyticsDomain, AnalyticsSource } from "../../data/BaseConfig";
```
tsibley, line 4: Since this is a React/JSX file, should this be importing from `../../data/SiteConfig` instead? I'm not sure of the intent for the two files going forward from 82d1cf3097d836ad3e7b7855ea267d72fdff892f…

tsibley (approved):

static-site/components/plausible-analytics/index.tsx
genehack: Pragmatically, the reason why this file is importing from `BaseConfig` (and partially the reason `BaseConfig` exists) is `SiteConfig` imports (the old, pre-App Router) `Footer` component, which imports `styled`, which won't work in App Router-style React Server Components.

static-site/data/BaseConfig.js
genehack: Fixed in force-push.

PR #1170, opened 2025-05-27
Remove unused <Sidebar> component

tsibley:
Last use was removed in "Convert blog to App Router [#134]" (fcfd1411).

## Checklist

- [x] Checks pass


tsibley:
@genehack Tagging you for review since you've been neck deep in the Next.js transition here and will have the best sense of things.

PR #1205, opened 2025-08-15
Convert index to app router [#1148]

### LINKED ISSUE #1148: Port index page to App Router
Components used in this page, which will also need to be ported:

- [x] `components/SEO`
    - requires porting to [Metadata API](https://nextjs.org/docs/app/api-reference/functions/generate-metadata)
- [x] `components/splash`
    - has a fair number of moving parts at least some of which (e.g., `ExpandableTiles`) overlap with other components (e.g., ListResources) — might be easiest to leave this for the end

genehack:

Converts the index page and the `<Splash>` component to App Router/CSS Modules style.

Note that there's one more change that will be made to the "convert /groups" PR that this is based on; that should get made before EOD tomorrow, after which I will rebase this PR too. None of those changes will affect any of the content in this PR. 

## Related issue(s)

Closes #1148 

## Checklist

- [ ] Checks pass


static-site/components/splash/index.tsx
```diff
@@ -0,0 +1,457 @@
...
+    <div
+      className={styles.tileOuter}
+      style={{ height: `${tileHeight}px`, width: `${tileWidth}px` }}
+    >
+      <div className={styles.tileInner}>
+        <a href={tile.url}>
+          <div className={styles.tileName}>{tile.name}</div>
+
+          <div className={styles.tileImgContainer}>
+            <img className={styles.tileImg} src={tileImg} alt={""} />;
```
victorlin, line 364: Remove stray `;` that is being rendered:

```suggestion
            <img className={styles.tileImg} src={tileImg} alt={""} />
```

static-site/app/styles/globals.css
```diff
@@ -62,10 +62,6 @@ html {
   }
 }
 
-main {
-  padding: 0 25px;
-}
```
victorlin, line 67: Re: f9d2ebd9ec64fdc197219a928c9db5cc50bc1231

> The padding: 0 25px style that was applied to the `<main>` tag in global.css is not needed (or wanted!) on the index page, only on …every other page on the site.

Is this padding needed or wanted on the other pages? I took it out and can't tell a difference from a few spot checks. It would be cleaner to just remove the padding entirely and not need to add it back with the pageContent class applied everywhere, which seems fragile.

static-site/src/pages/404.jsx
victorlin, line 1: In #1032 the 404 page was noted as "partially converted" because it links to `/` rather than serving the Splash component. I'm guessing the plan is now to keep it as a link, which seems fine to me.

victorlin (changes requested):

static-site/app/layout.tsx
```diff
@@ -39,6 +40,31 @@ export const metadata: Metadata = {
...
+};
+
+/**
+ * data used to generate JSON-LD schema script in the <head>
+ */
+const jsonLd = {
+  "@context": "http://schema.org",
+  "@type": "WebSite",
+  url: "https://nextstrain.org/",
+  name: "Nextstrain",
```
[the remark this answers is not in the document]
genehack: refactored.

static-site/app/styles/globals.css
genehack: dropped the commit.

Issue #1224, opened 2025-09-09
index-resources: `build-ref-matrix` not returning the expected matrix


joverlee521:
With the deployment of https://github.com/nextstrain/nextstrain.org/pull/828, there are two different versions of the resources index on prod (v5) vs canary (v6). 

### Current behavior

Both the [scheduled](https://github.com/nextstrain/nextstrain.org/actions/runs/17571447775/job/49908268378#step:4:11) and my [manually triggered](https://github.com/nextstrain/nextstrain.org/actions/runs/17595974453/job/49988191462#step:4:11) workflow runs created an unexpected matrix that used the **prod** ref (6983efeb) with the **canary** version (v6):

```json
[{"ref":"6983efebfbbebfce23d54c048678909d07f5b70e","resource_index":"s3://nextstrain-inventories/resources/v6.json.gz"}]
```

### Expected behavior

I would expect the build-ref-matrix output in the index-resources workflow to show this with two refs pointing to the two versions:

```json
[{"ref":"6983efebfbbebfce23d54c048678909d07f5b70e","resource_index":"s3://nextstrain-inventories/resources/v5.json.gz"},{"ref":"c686e0c204698009a169108f3ae199ed87bdb7c1","resource_index":"s3://nextstrain-inventories/resources/v6.json.gz"}]
```

This is the output I get when I run `./scripts/get-resource-index-ref-matrix` locally, so the bug is specific to the GH Action environment.

jameshadfield:
For those following along, from 6983efe (current heroku production, and therefore https://nextstrain.org) we have:

https://github.com/nextstrain/nextstrain.org/blob/6983efebfbbebfce23d54c048678909d07f5b70e/env/production/config.json#L113

Whereas current master and heroku canary (next.nextstrain.org) is at c686e0c2:

https://github.com/nextstrain/nextstrain.org/blob/c686e0c204698009a169108f3ae199ed87bdb7c1/env/production/config.json#L113

PR #1225, opened 2025-09-10
Fix scripts/get-resource-index-ref-matrix

joverlee521:

Change directory into the temp "$repo" instead of using the `--prefix` flag to  install the tmp app for running `scripts/get-resource-index.js`.

This commit includes a bunch of extra logs to stderr that I found useful for  debugging the issue, so keeping here for future issue.

## Related issue(s)

Resolves https://github.com/nextstrain/nextstrain.org/issues/1224

## Checklist

- [ ] Checks pass

### DIFF
--- a/scripts/get-resource-index-ref-matrix
+++ b/scripts/get-resource-index-ref-matrix
@@ -17,6 +17,7 @@ set -euo pipefail
 
 : "${PROD_APP_NAME:=nextstrain-server}"
 : "${CANARY_APP_NAME:=nextstrain-canary}"
+: "${RESOURCE_INDEX:=}"
 
 main () {
     if [[ "$IS_WORKFLOW_CALL" == true || \
@@ -35,6 +36,8 @@ main () {
 }
 
 build_prod_and_canary_ref_matrix() {
+    >&2 echo "Creating ref matrix from prod (${PROD_APP_NAME}) and canary (${CANARY_APP_NAME}) apps"
+
     local prod_commit prod_resource_index
     local canary_commit canary_resource_index
     local jq_array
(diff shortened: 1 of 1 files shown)


scripts/get-resource-index-ref-matrix
```diff
@@ -100,7 +109,15 @@ get_resource_index_at_commit() {
     # script. We can remove it once all the servers have been updated.
     #   -Jover, 01 May 2024
     if [[ -f "$repo/scripts/get-resource-index.js" ]]; then
```
jameshadfield, line 111: Thoughts on dropping this conditional while we're here? Would make debugging simpler and us more confident in the code which was executed

scripts/get-resource-index-ref-matrix
```diff
@@ -100,7 +109,15 @@ get_resource_index_at_commit() {
...
-        npm ci --silent --prefix "$repo"
+        >&2 echo "Running $repo/scripts/get-resource-index.js"
+        >&2 echo "RESOURCE_INDEX in $repo/env/production/config.json: $(jq -r '.RESOURCE_INDEX' $repo/env/production/config.json)"
+
+        if [[ -n "${RESOURCE_INDEX}" ]]; then
+            >&2 echo "RESOURCE_INDEX envvar: $RESOURCE_INDEX"
+        fi
+
+        cd "$repo"
+        npm ci --silent
```
jameshadfield, line 120: Why do we need to run `npm ci` at all? All that script needs is `node`...

Did you figure out _why_ the old code (`npm ci --silent --prefix "$repo"`) resulted in the bug?

joverlee521:
Merging without review to run this on the master branch in GH Actions, happy to take any post merge reviews.

joverlee521:
Ran the [index resources workflow](https://github.com/nextstrain/nextstrain.org/actions/runs/17599633518) and the matrix correctly linked prod/canary refs to the resource versions

```shell
ref-matrix=[{"ref":"6983efebfbbebfce23d54c048678909d07f5b70e","resource_index":"s3://nextstrain-inventories/resources/v5.json.gz"},{"ref":"c686e0c204698009a169108f3ae199ed87bdb7c1","resource_index":"s3://nextstrain-inventories/resources/v6.json.gz"}]
```

scripts/get-resource-index-ref-matrix
jameshadfield: Or, alternatively, just use the `jq` code which seems much simpler to reason with to my eyes. I'm presuming there's some reason you (pl) wanted to use a custom script for this?

Issue #1228, opened 2025-09-16
pathogens/files: Versioned URL still downloads latest version


joverlee521:
While investigating https://github.com/nextstrain/rsv/issues/99, I tried to use https://nextstrain.org/pathogens/files?filter=rsv to download previous versions of metadata for comparisons. However, no matter which past version I clicked on, I ended up downloading the latest version of the file. 

This seems to be an issue with downloading through the data.nextstrain.org CloudFront URL since I can download the correct version with the   nextstrain-data.s3.amazonaws.com S3 URL:

```
$ curl -sI https://data.nextstrain.org/files/workflows/rsv/a/metadata.tsv.gz?versionId=YVj6GR8MGATh3LSjJyCEqG3JcGlJ89Aa
HTTP/2 200 
content-type: application/gzip
content-length: 1254218
date: Tue, 16 Sep 2025 23:06:54 GMT
x-amz-meta-recordcount: 33510
x-amz-meta-sha256sum: 53fa60387f2faac2376e0ca628ef007f4124a9a0ffc44115640db834c83bbbdc
last-modified: Tue, 16 Sep 2025 16:08:04 GMT
x-amz-version-id: l2DqWnQ78zyb8RCgSrXz8VAS9f8r6IgC
etag: "eaa2cdaf1f5593fe1da3399b70cca801"
server: AmazonS3
x-cache: Hit from cloudfront
via: 1.1 2f743a0ca9cc9da3bb8d81eb8b012916.cloudfront.net (CloudFront)
x-amz-cf-pop: MCI50-P1
x-amz-cf-id: nYvc7pTywwa7n4-Tu-Kkv4Y-49jvJw_E1KptGiHKhmU8c45dNmAe9Q==
age: 818

$ curl -sI https://nextstrain-data.s3.amazonaws.com/files/workflows/rsv/a/metadata.tsv.gz?versionId=YVj6GR8MGATh3LSjJyCEqG3JcGlJ89Aa
HTTP/1.1 200 OK
x-amz-id-2: poNqNU4jIFOhUPynOGVI8gKaivkoxlgckKd8xh6P2bWAmvmpxTDRVgo1uRz5AOjRMEqiA9ipPaA=
x-amz-request-id: QSVH9H0GDJ4BPBYR
Date: Tue, 16 Sep 2025 23:20:50 GMT
Last-Modified: Tue, 05 Aug 2025 16:08:27 GMT
[...]

PR #1230, opened 2025-09-17
pathogens/files: Use S3 URL for intermediate files

joverlee521:

Based on feedback from @jameshadfield on [Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1758068502650279?thread_ts=1758067622.694609&cid=C01LCTT7JNN).

Resolves https://github.com/nextstrain/nextstrain.org/issues/1228

## Checklist

- [x] Checks pass

### DIFF
--- a/src/resourceIndex.js
+++ b/src/resourceIndex.js
@@ -130,7 +130,7 @@ async function updateResourceVersions() {
 
 /**
  * ListResources creates the response data for resource listing queries
- * 
+ *
  * Note: 'word' as used here relates to the resource name/ID in the sense that the
  * ID is made up of slash-separated words.
  */
@@ -285,6 +285,8 @@ function _onlyLatestIntermediates([name, data]) {
       if (Object.hasOwn(files, filenameOrDate)) continue;
       const link = new URL(indexedUrlOrDate);
       link.searchParams.delete('versionId')
+      // Use the CloudFront URLs for the unversioned link
+      link.host = "data.nextstrain.org"
       files[filenameOrDate] = [link.toString(), date]
     }
   }


env/production/config.json
```diff
@@ -110,6 +110,6 @@
   "OIDC_GROUPS_CLAIM": "cognito:groups",
   "SESSION_COOKIE_DOMAIN": "nextstrain.org",
   "GROUPS_DATA_FILE": "groups.json",
-  "RESOURCE_INDEX": "s3://nextstrain-inventories/resources/v6.json.gz",
+  "RESOURCE_INDEX": "s3://nextstrain-inventories/resources/v7.json.gz",
```
jameshadfield, line 113: I don't think this needs an update, but also fine to update it...

resourceIndexer/coreStagingS3.js
```diff
@@ -308,13 +308,9 @@ function validIntermediate(id, date, objects) {
...
         .map((s3object) => {
           const filename = s3object.key.split('/').pop();
-          let url = s3object.versionId ?
+          const url = s3object.versionId ?
             `${s3object.baseUrl}?versionId=${encodeURIComponent(s3object.versionId)}` :
             s3object.baseUrl;
-          /* we prefer to serve objects on the nextstrain-data bucket via the cloudfront-ed
-          data.nextstrain.org domain rather than nextstrain-data.s3.amazonaws.com. This
-          continues to work with versionId queries */
-          url = url.replace('https://nextstrain-data.s3.amazonaws.com', 'https://data.nextstrain.org');
```
jameshadfield, line 317: [discussion]

The server has 2 APIs to provide data to /pathogen/files, one which lists the latest versions of every known file (across all pathogens), and one which provides the version history for a single pathogen (only called when a modal opens). For the former would it be better to use data.nextstrain.org (cloudfronted) URLs? If so we can add a similar `replace()` to the server code.

jameshadfield (approved):
👍 for this.

env/production/config.json
joverlee521: Yeah, not totally necessary, but I didn't want this change propagating to prod before we promote.

resourceIndexer/coreStagingS3.js
jameshadfield: Probably in this function if we were to do so

https://github.com/nextstrain/nextstrain.org/blob/d8f3c5691e3befd87a674e899964689b439b4ad0/src/resourceIndex.js#L269-L277

joverlee521: Yeah, I think we would want CloudFront for the unversioned URLs. Thanks for the pointer.

jameshadfield (approved):
Thanks for picking this up so quickly. Tested locally & things working (now!) as expected. I'm going to merge to get this onto canary straight away.

Issue #1252, opened 2025-11-06
Listing previous versions fails for some datasets


tsibley:
Navigate to https://nextstrain.org/pathogens?filter=ncov (for example).

Click on one of the many "turn back the clock" links with an n= in several hundreds, e.g. ncov/gisaid/africa/1m's "every few days (n=455)". Note that it opens the previous versions modal as expected. Close the modal.

Click on one of the same "turn back the clock" links with a higher n=, e.g. ncov/gisaid/africa/2m's n=616 or …/6m's n=1119. Note that it crashes the list resources component with the error:

> Something isn't working! Error: \`Unknown error (thrown value was not an InternalError)\`

Reproducing this locally, I see the underlying error is:

> TypeError: can't access property Symbol.iterator, items is undefined

from inside d3 and implicates our code at:

https://github.com/nextstrain/nextstrain.org/blob/a1f44b894073e04e35ccad22a4cec9ecf653b73b/static-site/components/list-resources/modal_draw.js#L125

This points to `beeswarmData` being undefined, which makes sense as possible since it starts as undefined and may not ever be defined before the iterative sizing algorithm gives up:

https://github.com/nextstrain/nextstrain.org/blob/a1f44b894073e04e35ccad22a4cec9ecf653b73b/static-site/components/list-resources/modal_draw.js#L97-L119

jameshadfield:
I can't reproduce this via:
1. https://nextstrain.org/pathogens?filter=africa&filter=gisaid&filter=ncov (or the same page in local dev mode) (Firefox 144, Chrome 141)
2. Load the /1m (n=455) modal
3. Load 6m (n=1119) modal

No crashes or errors were observed in the console and I don't think (!) any dots are missing. I don't doubt there is such an error lurking here, so if anyone else can reproduce it please let me know!

PR #1253, opened 2025-11-06
static-site[list-resources]: Fix crash in modal_draw() when bottoming out during radius calculations

tsibley:
Fixes the algorithm which calculates an ideal radius for the points on the beeswarm plot by ensuring that beeswarmData, beeswarmHeight, and radius are all set to their last computed values when bottoming out as well as when topping out.  Simplifies state tracking instead of doubling variables unnecessarily.

Previously when bottoming out on downward iterations, the algorithm would leave beeswarmData unset, which caused errors like this from inside d3:

    TypeError: can't access property Symbol.iterator, items is undefined

This was triggered when many data points were close together causing beeswarm heights to soar.

Bottoming out will mean that some points in the plot are cut off, i.e. above the top of the chart.  This seems preferable to a crash (or unclickably tiny points).  Future improvements could dynamically scale the chart height in such cases, or alternatively the chart width to spread out points horizontally and make the longer timeline scrollable.

Resolves: <https://github.com/nextstrain/nextstrain.org/issues/1252>
## Checklist

- [x] Checks pass

### DIFF
--- a/static-site/components/list-resources/modal_draw.js
+++ b/static-site/components/list-resources/modal_draw.js
@@ -92,7 +92,6 @@ export default function modal_draw(ref, resource, lightGrey) {
     heights.marginBelowAxis;
   let radius = 12;
   const padding = 1;
-  let nextRadius = radius;
   let iterCount = 0;
   let beeswarmData;
   let beeswarmHeight = 0;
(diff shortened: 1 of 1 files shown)


static-site/components/list-resources/modal_draw.js
```diff
@@ -92,29 +92,25 @@ export default function modal_draw(ref, resource, lightGrey) {
...
     if (nextSpareHeight <= spareHeight && nextSpareHeight > 0) {
-      beeswarmData = nextBeeswarmData;
-      beeswarmHeight = nextBeeswarmHeight;
       spareHeight = nextSpareHeight;
-      radius = nextRadius;
-      nextRadius += radiusJump;
+      radius += radiusJump;
     } else {
-      nextRadius -= radiusJump;
+      radius -= radiusJump;
```
victorlin: This fixes the bug, but the radius seems off:

Image

PR #1261, opened 2025-11-13
Dependency updates

jameshadfield:
I started down this road then moved onto more interesting things, but thought we might as well push these improvements through. The next (bigger) updates would be to express & nextjs but they're a lot more involved.


victorlin (approved):

victorlin:
@jameshadfield do you mind if I rebase this PR and merge?

PR #1264, opened 2025-11-19
/groups: Use custom group logos

victorlin:

Preview: https://nextstrain-dev.herokuapp.com/groups

logos

For groups that have custom logos, using those logos in place of the generic users icon gives a more personalized look to the page. The generic icon is still used for groups without custom logos.

This is done with a new server-side function getGroupAvatars() to return logos for all visible groups.

## Related issue(s)

Follow-up to #1263

## Checklist

- [x] [Explore server component rewrite](https://github.com/nextstrain/nextstrain.org/pull/1264#issuecomment-3555286523)
- [x] Checks pass

### DIFF
--- a/src/endpoints/charon/getAvailable.js
+++ b/src/endpoints/charon/getAvailable.js
@@ -47,9 +47,23 @@ async function collectAllAvailableGroups(user) {
   const source = new metaSources.Groups(user);
   const datasets = await source.availableDatasets();
   const narratives = await source.availableNarratives();
+  const avatars = await source.getGroupAvatars();
+  /* TODO: Consider using a single call to GroupSource.getInfo() to return data
+  for all 3 calls above. I think this means a structure like below, but the
+  current flat structure should also work fine.
+  {
+    [groupName]: {
+      datasets: …
+      narratives: …
+      avatar: …
+    }
+  }
+  */
+
   return {
     datasets: datasets.map((request) => ({request})),
-    narratives: narratives.map((request) => ({request}))
+    narratives: narratives.map((request) => ({request})),
+    avatars
   };
 }
 
--- a/src/metaSources.js
+++ b/src/metaSources.js
@@ -82,6 +82,26 @@ const Groups = (() => {
         .map((e) => e[1])
         .flat();
     }
+    async getGroupAvatars() {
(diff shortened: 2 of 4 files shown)


static-site/app/groups/[[...groups]]/group-logo.tsx
```diff
@@ -1,9 +1,32 @@
...
+export default function GroupLogo({ name }: {
+  /** Group name */
+  name: string
+}): React.ReactElement {
+  const [imgSrc, setImgSrc] = useState(usersIcon.src);
+  const [imgAlt, setImgAlt] = useState("default group logo");
+
+  useEffect(() => {
+    async function fetchImgSrc(): Promise<void> {
+      const response = await fetch(`/charon/getSourceInfo?prefix=/groups/${name}`);
```
Comment (victorlin) [inline static-site/app/groups/[[...groups]]/group-logo.tsx]: _from https://github.com/nextstrain/nextstrain.org/pull/1263#discussion_r2539311362_

This is 44 calls to the server when loading nextstrain.org/groups – not great! There should be better ways to do this such as [putting the request in a server component](https://github.com/nextstrain/nextstrain.org/pull/1263#discussion_r2536028542), but that requires extra work to convert parent components to also be server components.

joverlee521:
> Comments in the codebase remind me of long-ago prototypes where both tom & me explored making the server-rendering code aware of the request object in order to do this; it wasn't easy.

Previous [Slack thread](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1724449585236639?thread_ts=1724263606.541899&cid=C01LCTT7JNN) that explored server-side user contexts. General summary was use of the old NextJS "Pages" router made this difficult. I was able to prototype getting the [user/group contexts in the App router](https://github.com/nextstrain/nextstrain.org/blob/f648cdaaa02b3c9afdca83c2ef4c2f25584504c8/static-site/app/groups/%5BgroupName%5D/settings/members/page.tsx#L10-L13).

jameshadfield:
> While it's nice to have custom logos on the page, I think the costs outweigh the benefits for any simple solution, no matter client or server side. 

Agreed - and a good example of how server-side wouldn't have helped. I think there's many paths we can take to improve the situation, such as fetching the `overview.md` each time we renew the groups listings & storing a small version of the image on the server, but I agree it's not worth prioritizing for now.

PR #1267, opened 2025-11-21
Use new dataset browser on individual group pages

victorlin:

Two refactoring commits followed by a commit that uses ListResources on IndividualGroupPage.

Preview: https://nextstrain-s-victorlin--v9gx4f.herokuapp.com/groups/blab

datasets

## Related issue(s)

Part of #870

## Checklist

- [x] Checks pass

### DIFF
--- a/static-site/app/groups/[[...groups]]/individual-group-page.tsx
+++ b/static-site/app/groups/[[...groups]]/individual-group-page.tsx
@@ -10,6 +10,8 @@ import DatasetSelect from "../../../components/dataset-select";
 import { DatasetType } from "../../../components/dataset-select/types";
 import ErrorMessage from "../../../components/error-message";
 import { FlexGridRight } from "../../../components/flex-grid";
+import ListResources from "../../../components/list-resources";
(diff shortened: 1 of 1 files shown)


static-site/app/groups/[[...groups]]/individual-group-page.tsx
```diff
@@ -247,6 +227,46 @@ export default function IndividualGroupPage({
   }
 }
 
+// It is unfortunate that this method repeats the request that we're already
+// making above in the `useEffect()` hook. Ideally we could make that request
+// once, and re-use the response.
+//   - jsja, 08 Aug 2025
+//   - copied/modified by victor, 21 Nov 2025
```
Comment (joverlee521) [inline static-site/app/groups/[[...groups]]/individual-group-page.tsx]: I'm confused by this comment. I'd think we could keep the `datasets` state and pass that into `_resourcesCallback` to be used to create the `Group` object (something like https://github.com/nextstrain/nextstrain.org/commit/69073ccf1c612aed444d2436bdb4ffdbeb871935). Am I missing something here?

joverlee521 (approved):

Issue #1201, opened 2025-08-13
Refactor `/groups` routing/file organization


genehack:
FWIW, the client side router for all `/groups` URLs will make it harder to convert child components to server components in the future. (Migrating to server side components was where I stalled out on adding the groups management stuff previously...)

I don't think it was possible to set up nested dynamic routes in the Page router, but it should be possible with the new App router. Instead of using the catch all segment `[[...groups]]`, we could switch to the simple dynamic route `[group]` with the file structure

```
static-site/app/groups/
├── [group]
│   ├── page.tsx
│   └── settings
│       ├── members
│       │   ├── page.module.css
│       │   └── page.tsx
|       ├── page.module.css
│       └── page.tsx
├── group-tiles.module.css
├── group-tiles.tsx
├── page.tsx
└── utils.ts
```

_Originally posted by @joverlee521 in https://github.com/nextstrain/nextstrain.org/pull/1199#discussion_r2271589524_

joverlee521:
> Groups URLs such as https://nextstrain.org/groups/blab/foo are 404 pages (in spirit), as they are a result of [our error handling routes](https://github.com/nextstrain/nextstrain.org/blob/1efc22ec62431c8c54b11e4df70e4d4a86daa4d0/src/routing/errors.js#L116-L127).

Ah, since the `[group]` route is not a catch-all, this results in the [top level FourOhFour page](https://github.com/nextstrain/nextstrain.org/blob/1efc22ec62431c8c54b11e4df70e4d4a86daa4d0/static-site/app/not-found.tsx#L7). We could add a catch-all `/[group]/[...dataset]` route to catch this and display the group page with the error banner. (I have it working locally, but can't push up at the moment due to Git outage).

joverlee521:
> (I also didn't think that [[...foo]] routes had to be client side, but I don't claim to be a nextjs expert.)

You're totally right. I think it currently needs `use client` because the router component calls [useParams](https://github.com/nextstrain/nextstrain.org/blob/1efc22ec62431c8c54b11e4df70e4d4a86daa4d0/static-site/app/groups/%5B%5B...groups%5D%5D/router.tsx#L50) and uses state. This could be converted to take the params as a functional param and not store state to be a server component.

jameshadfield:
[...]

PR #1268, opened 2025-11-22
/groups: Resource listing improvements

victorlin:

group

Grab bag of improvements to resource listing on the groups page. Functional changes:

- No more duplicate fetch
- Added link to individual group page
- Removed generic users icon
- Removed group name prefix from list
- Changed links to open in current window (this also affects other pages using ListResources)

## Notable review threads

- [Why did the mis-typing of `available: AvailableGroups[]` work?](https://github.com/nextstrain/nextstrain.org/pull/1268#discussion_r2554327609) The only usage was `available["narratives"]` which was inferred as string-indexing to an implicit type `any`.
- [Enforce co-occurrence of `groupUrl` and `groupUrlTooltip` in TypeScript?](https://github.com/nextstrain/nextstrain.org/pull/1268#discussion_r2554337569) yes
- [Re-use parent component fetch without changing ListResources implementation?](https://github.com/nextstrain/nextstrain.org/pull/1268#discussion_r2554356557) yes
- [Cache the ListResources function with `useCallback`?](https://github.com/nextstrain/nextstrain.org/pull/1268#discussion_r2558082050) Consider after #1201

## Checklist

- [x] Checks pass

### DIFF
--- a/static-site/app/groups/[[...groups]]/group-listing-page.tsx
+++ b/static-site/app/groups/[[...groups]]/group-listing-page.tsx
@@ -61,7 +61,7 @@ export default function GroupListingPage(): React.ReactElement {
           "/charon/getAvailable?prefix=/groups",
         );
         setDatasets(available.datasets);
-        setNarratives(_cleanUpAvailable(available["narratives"]));
+        setNarratives(_cleanUpAvailable(available.narratives));
         setDataLoaded(true);
       } catch (err) {
         console.error(


static-site/app/groups/[[...groups]]/group-listing-page.tsx
```diff
@@ -55,7 +55,7 @@ export default function GroupListingPage(): React.ReactElement {
   useEffect((): void => {
     async function fetchData(): Promise<void> {
       try {
-        const available = await fetchAndParseJSON<AvailableGroups[]>(
+        const available = await fetchAndParseJSON<AvailableGroups>(
```
Comment (jameshadfield) [inline static-site/app/groups/[[...groups]]/group-listing-page.tsx]: [question] How can something appease the type checker as an array of X and also just `X`?

jameshadfield (changes requested):
Nice grab-bag of improvements! Reading through these (and being at arms length from the ResourceListing since I did the original implementation) really made me want the file reorganisation being [discussed in #1201](https://github.com/nextstrain/nextstrain.org/issues/1201#issuecomment-3554638790).

static-site/components/list-resources/resource-group.tsx
```diff
@@ -148,9 +148,9 @@ function ResourceGroupHeader({
 
       <div className={styles.flexColumnContainer}>
         <div className={styles.headerRow}>
-          {group.groupUrl ? (
+          {group.groupUrl && group.groupUrlTooltip ? (
```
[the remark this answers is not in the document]
victorlin: Good point, I'll look into this. Thinking of something like:

```ts
export type Group = {
  // ...
} & GroupUrlProperties;

type GroupUrlProperties =
  | { groupUrl: string; groupUrlTooltip: string }
  | { groupUrl: undefined; groupUrlTooltip: undefined };
```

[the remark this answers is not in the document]
jameshadfield: Yeah - something like that is what I was imagining 👍

static-site/app/groups/[[...groups]]/group-listing-page.tsx
Comment (jameshadfield) [reply static-site/app/groups/[[...groups]]/group-listing-page.tsx]: Gah - what! TS has so many of these gotchas. They really me wonder about the limitations of TS that I am not fully aware of

Anyway, nice work identifying the original mis-typing!

Comment (victorlin) [reply static-site/app/groups/[[...groups]]/group-listing-page.tsx]: I dug a little more. Explanation: string-indexing an array (`AvailableGroups[]["narratives"]` is valid JS which returns an implicit type `any`. This is currently allowed by our TypeScript config which disables `noImplicitAny`. Enabling that (#879) shows an error with the original mis-typing.

Comment (victorlin) [reply static-site/app/groups/[[...groups]]/group-listing-page.tsx]: Switched to `available.narratives` in c6f9ab9508cc37d4a8a1c67dbb08438e568a5d0f

static-site/components/list-resources/resource-group.tsx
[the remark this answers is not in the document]
victorlin: `AllOrNothing` allows defining an object without `groupUrl`, but it doesn't allow accessing `groupUrl` explicitly such as in the JSX this thread is anchored to.

I've addressed this with `& GroupUrlProperties` in a3488040648e1558c66172053ec92f2fe7508c98.

static-site/app/groups/[[...groups]]/group-listing-page.tsx
```diff
@@ -72,6 +75,41 @@ export default function GroupListingPage(): React.ReactElement {
     fetchData();
   }, []);
 
+  // NOTE: "group" has two meanings here - a nextstrain group and a group of
+  // resources for listing. Luckily for us the "group name" is the same for both
+  async function resourcesCallback(): Promise<Group[]> {
```
Comment (joverlee521) [inline static-site/app/groups/[[...groups]]/group-listing-page.tsx]: _non-blocking discussion only_

> Note that a subtle difference is that the callbacks are now redefined on
> every change of parent component state, instead of being defined once at
> the top level.

I was going to suggest `useCallback` to cache the function so that it only gets updated with the `dataset` state gets updated. However, this doesn't matter much right now because all state is only set on the initial render of the component...This brings me to the tangential thought that maybe these shouldn't be state. Similar to https://github.com/nextstrain/nextstrain.org/issues/1201, we could lean more into the [NextJS routing files](https://nextjs.org/docs/app/getting-started/project-structure#routing-files) and use [loading](https://nextjs.org/docs/app/api-reference/file-conventions/loading) + [error](https://nextjs.org/docs/app/api-reference/file-conventions/error) files instead of component states.

PR #1270, opened 2025-11-25
Restore "no datasets" message

victorlin:

This was inadvertently removed in "Use new dataset browser on individual group pages" (#1267). Without this check, the ListResources component will display a forever-spinning Nextstrain logo for empty results. Such a scenario has always been possible with URL query filters that return no results, but it became an obvious bug on individual group pages without any datasets.

## Related issue(s)

Follow-up to #1267

## Checklist

- [x] Checks pass

### DIFF
--- a/static-site/components/list-resources/index.tsx
+++ b/static-site/components/list-resources/index.tsx
@@ -170,7 +170,7 @@ function ListResourcesContent({
 
   const [sortMethod, changeSortMethod] = useState<SortMethod>("alphabetical");
 
-  const [resourceGroups, setResourceGroups] = useState<Group[]>([]);
+  const [resourceGroups, setResourceGroups] = useState<Group[]>();
 
   useSortAndFilter(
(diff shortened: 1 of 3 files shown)


static-site/app/groups/[[...groups]]/individual-group-page.tsx
```diff
@@ -211,11 +211,17 @@ export default function IndividualGroupPage({
...
+                <h4 className="centered">
+                  No datasets are available for this group.
+                </h4>
+              ) : (
+                <ListResources
+                  resourceType="dataset"
+                  versioned={false}
+                  fetchResourceGroups={resourcesCallback}
+                />
+              )}
```
Comment (jameshadfield) [inline static-site/app/groups/[[...groups]]/individual-group-page.tsx]: [comment after reading the diff]

This seems like it would be more generally useful and thus better placed within `<ListResources/>`

joverlee521 (approved):

PR #1271, opened 2025-11-25
Use new narratives browser on Groups pages

### LINKED ISSUE #1282: groups: Use ListResources for narratives
## Description

Currently, groups pages ([/groups](https://nextstrain.org/groups) and individual group pages) use the new ListResources component for datasets and the old DatasetSelect component for narratives.

The `/charon/getAvailable` endpoint provides all the necessary data for ListResources to work with narratives. The new component should be used for a more consistent user interface.

## Possible solutions

1. Keep datasets and narratives separate ([PR](https://github.com/nextstrain/nextstrain.org/pull/1271))
2. List them together ([PR](https://github.com/nextstrain/nextstrain.org/pull/1278))

As mentioned in https://github.com/nextstrain/nextstrain.org/pull/1271#issuecomment-3599508976, I'm leaning towards 2 (with some more polish).

## Links

- Initially discussed in https://github.com/nextstrain/nextstrain.org/commit/4f88298c3d65d3a9c32b8635260308e074fc301e#r171173586

victorlin:

narratives

This provides an interface consistent with the dataset browser on the same pages.

Note that this only works for groups pages because `/charon/getAvailable` provides the necessary data. It won't work on core and staging without updating the resource indexer.

## Related issue(s)

Closes #1282

## Checklist

- [x] Checks pass

### DIFF
--- a/static-site/app/groups/[group]/page.tsx
+++ b/static-site/app/groups/[group]/page.tsx
@@ -103,7 +103,13 @@ export default function IndividualGroupPage({
   // resources for listing. Luckily for us the "group name" is the same for both
   async function getResourceGroups(dataResources: DataResource[]): Promise<Group[]> {
     const resources = dataResources.map((dataResource): Resource => {
-      const name = dataResource.request.replace(new RegExp(`^groups/${group}/`), '');
+      const parts = dataResource.request.split('/');
+      let name: string;
+      if (parts[2] === "narratives") {
+        name = parts.slice(3).join('/');
+      } else {
+        name = parts.slice(2).join('/');
+      }
       return {
         name,
         groupName: group,
--- a/static-site/app/groups/available.tsx
+++ b/static-site/app/groups/available.tsx
@@ -60,7 +60,12 @@ export default function Available(): React.ReactElement {
       const parts = dataResource.request.split('/').slice(1);
       const groupName = parts[1]
       if (parts[0] !== "groups" || groupName === undefined) return [];
-      const name = parts.slice(2).join('/');
+      let name: string;
+      if (parts[2] === "narratives") {
+        name = parts.slice(3).join('/');
+      } else {
+        name = parts.slice(2).join('/');
+      }
       return [{
         name,
         groupName,


static-site/components/list-resources/index.tsx
```diff
@@ -315,7 +315,7 @@ function Filter({
   return (
     <div className="filter">
       <Select
-        placeholder={`Filter by keywords in ${resourceType==='dataset'?'dataset':'file'} names`}
+        placeholder={`Filter by keywords in ${displayResourceType(resourceType)} names`}
```
joverlee521: Should make a similar change in a couple places within resource-groups.tsx, e.g.

https://github.com/nextstrain/nextstrain.org/blob/58a37da8401831da63474841d927167e68bc9545/static-site/components/list-resources/resource-group.tsx#L186

Also the resource-groups.tsx just has a couple places that currently hardcodes the word "datasets", e.g. 

https://github.com/nextstrain/nextstrain.org/blob/58a37da8401831da63474841d927167e68bc9545/static-site/components/list-resources/resource-group.tsx#L246

static-site/app/groups/[[...groups]]/group-listing-page.tsx
```diff
@@ -155,13 +132,11 @@ export default function GroupListingPage(): React.ReactElement {
...
       </ScrollableAnchor>
 
-      {dataLoaded && (
-        <DatasetSelect
-          datasets={narratives}
-          columns={datasetColumns}
-          title={"Filter Narratives"}
-        />
-      )}
+      <ListResources
```
Comment (joverlee521) [inline static-site/app/groups/[[...groups]]/group-listing-page.tsx]: _non-blocking_

In brief testing, both instances of `ListResources` share the filter queries set via URL params, so filtering datasets will also filter narratives and vice versa. I could see that as a feature, but it's kind of weird since there are two search inputs controlled by the same URL param.

static-site/app/groups/[[...groups]]/group-listing-page.tsx
```diff
@@ -155,13 +132,11 @@ export default function GroupListingPage(): React.ReactElement {
...
+        resourceType="narrative"
+        versioned={false}
+        fetchResourceGroups={() => getResourceGroups(narratives)}
+      />
```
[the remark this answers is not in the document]
Comment (victorlin) [reply static-site/app/groups/[[...groups]]/group-listing-page.tsx]: Sounds reasonable, done: a8916bfab8fc43444005ac9b15df14b7f7d6938e

static-site/components/list-resources/index.tsx
victorlin: Done as separate commits: eccae57b5cef063feb80188bc20712db08122afa...44f3a9dc0bf568f313f334ae29eb30c8c758e1a0

There's more places where "dataset" and "pathogen" are hardcoded, but they aren't shown for unversioned resource listings. I've left those as-is – it should be obvious what needs fixing if/when version info is added to more pages. I don't think I've missed anything that could appear with this PR's changes, but let me know if you find anything!

victorlin:
#1278 is an alternative to this PR based on @jameshadfield's idea in https://github.com/nextstrain/nextstrain.org/pull/1271#issuecomment-3582815193. I think it's the better approach, but I'll keep this PR open to compare the review apps.

PR #1273, opened 2025-11-26
CI: Small touch ups

victorlin:
Follow-up to #1272

## Checklist

- [x] Checks pass

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -157,7 +157,7 @@ jobs:
 
           # No deploy
           else
-            echo "value=" >> "$GITHUB_OUTPUT"
+            echo "value=" | tee -a "$GITHUB_OUTPUT"
           fi
 
   deploy:


.github/workflows/ci.yml
```diff
@@ -159,6 +159,9 @@ jobs:
           else
             echo "value=" >> "$GITHUB_OUTPUT"
           fi
+      - name: Show outputs
+        run: |
+          echo "heroku-app: ${{ steps.heroku-app.outputs.value }}"
```
joverlee521: _non-blocking_

> It's useful to see output values for debugging purposes. They don't seem
to be accessible in the GitHub workflow logs without the extra step.

You could also use `tee` the output to logs, e.g. 
https://github.com/nextstrain/nextstrain.org/blob/09f7a8865dc603e490840a27744b2177f320e624/.github/workflows/index-resources.yml#L52

PR #1274, opened 2025-11-26
Use dynamic routes for /groups pages

### LINKED ISSUE #1201: Refactor `/groups` routing/file organization

victorlin:

This reorganizes files from a flat structure with everything under a catch-all [[...groups]] to a nested structure with one directory per URL part.

A new catch-all page has been added for /groups/[group]/[...path] to handle nonexistent resources.

The only functional difference with these changes should be the new page-specific titles.

## Related issue(s)

Closes #1201

## Checklist

- [x] Tested all pages locally
- [x] Checks pass

### DIFF
--- a/static-site/app/groups/[group]/page.tsx
+++ b/static-site/app/groups/[group]/page.tsx
@@ -73,6 +73,8 @@ export default function IndividualGroupPage({
   const { group } = params;
 
   useEffect(() => {
+    document.title = `"${group}" Group - Nextstrain`;
+
     async function getGroupInfo(): Promise<void> {
       try {
         const [sourceInfo, availableData] = await Promise.all([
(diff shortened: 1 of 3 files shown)


joverlee521 (approved):
Thanks for following up on this @victorlin!

Issue #1276, opened 2025-12-01
Groups resource listing uses invalid URLs after error


joverlee521:
_Noticed odd behavior when reviewing https://github.com/nextstrain/nextstrain.org/pull/1274, but this behavior is reproducible on prod._

After landing on a non-existent Groups dataset, the Groups resource listing uses invalid URLs.

1. Go to https://nextstrain.org/groups/blab/foo
2. See error message for non-existent dataset
3. Scroll down to click on a dataset, e.g. 229/E, and get directed to bad URL https://nextstrain.org/groups/blab/blab/229e/E

PR #1277, opened 2025-12-01
individual-group-page: use origin-relative URLs for `ListResources`

joverlee521:

Ensures that we don't get an invalid URLs after being redirected to the individual-group-page when user tries to access a non-existent dataset.

Resolves https://github.com/nextstrain/nextstrain.org/issues/1276

## Checklist

- [x] Checks pass


joverlee521:
Tested in review app 
- Go to https://nextstrain-s-group-reso-j5rfbl.herokuapp.com/groups/blab/foo
- Clicking on 229/E dataset successfully goes to https://nextstrain-s-group-reso-j5rfbl.herokuapp.com/groups/blab/229e/E

PR #1281, opened 2025-12-03
Use dynamic routes for pages that use ErrorBanner

victorlin:

This PR has 4 cleanup/prep commits followed by a main commit.

The main commit reorganizes files from a flat structure with everything under a catch-all `[[...<stub>]]` to a nested structure with one directory per URL part.

Note that metadata with page titles are now re-exported from other files. This is to replicate the behavior from custom top-level routers, which did that once for all pages. For each invalid (404) path, it is needed on both (1) the not-found component to set it server-side and (2) the page component to set it client-side.

## Related issue(s)

Follow-up to #1274

## Checklist

- [x] Checks pass


joverlee521 (approved):
Thanks for continuing to push out this pattern, the new directory structure makes sense to me. Only left a couple minor questions.

PR #1285, opened 2025-12-09
Use dynamic routes for /community pages

victorlin:

This PR has (in order) 1 cleanup/prep commit, 1 main commit, and 1 follow-up commit.

The main commit reorganizes files from a flat structure with everything under a catch-all [[...community]] to a nested structure with one directory per URL part.

Note that "narratives" is an optional part in the path. Such logic was easily implemented in the custom router by using Array.shift(), but it can't be encoded into dynamic routes. As a result, there is some boilerplate duplication between community/[user]/ and narratives/[user]/, but the bulk of the content remains in the top-level community/*.tsx files.

## Related issue(s)

Follow-up to https://github.com/nextstrain/nextstrain.org/pull/1274

## Checklist

- [x] [Send proper 404s?](https://github.com/nextstrain/nextstrain.org/pull/1285#discussion_r2608690373) not in this PR
- [x] Checks pass

### DIFF
--- a/static-site/app/community/[user]/page.tsx
+++ b/static-site/app/community/[user]/page.tsx
@@ -10,5 +10,5 @@ export default function Page({
     user: string;
   };
 }): React.ReactElement {
-  return <MissingRepoErrorPage />;
+  return <MissingRepoErrorPage user={params.user} />;
 }
--- a/static-site/app/community/error-page.tsx
+++ b/static-site/app/community/error-page.tsx
@@ -7,14 +7,18 @@ import CommunityPageContent from "./content";
 import { usePathname } from "next/navigation";
 
 /**
- * Renders a generic error page.
+ * Renders an error page for incomplete user-only paths.
  */
-export function MissingRepoErrorPage(): React.ReactElement {
+export function MissingRepoErrorPage({
+  user,
+}: {
+  user: string;
+}): React.ReactElement {
   const errorPath = "nextstrain.org" + usePathname();
   return (
     <>
       <ErrorMessage
-        title={`The community repository, dataset, or narrative "${errorPath}" doesn't exist.`}
+        title={`The path "${errorPath}" is not valid. If you are looking for resources under "${user}", add a repo to the path: "${errorPath}/{repo}"`}
         contents={<p>Here is the community page instead.</p>}
       />
       <CommunityPageContent />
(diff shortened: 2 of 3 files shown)


static-site/app/community/[[...community]]/router.tsx
```diff
@@ -1,135 +0,0 @@
...
- *   over <CommunityRepoPage> with information, status 404
- *
- * Note that the 404 status codes are being set/returned at the
- * Express router code, as a side effect of how that code re-directs
- * requests that don't directly load a dataset or narrative into the
- * Next.js part of the web site. This router code does not make use of
- * the Next.js `notFound()` method because the 404 status has already
- * been set. For the requests that don't display an <ErrorMessage>,
- * the status should _arguably_ be 200, but I'm not going to worry
- * about that at the moment. - @genehack, 1 Aug 2025
```
Comment (victorlin) [inline static-site/app/community/[[...community]]/router.tsx]: Note on previous implementation: this isn't entirely true. 404 is only returned for some routes, and never for `text/html`. Examples:

```sh
curl -I "https://nextstrain.org/community/invalid"
# 200

curl -I "https://nextstrain.org/community/user/invalid"
# 404
curl -I "https://nextstrain.org/community/user/invalid" -H 'accept: text/html'
# 200

curl -I "https://nextstrain.org/community/user/repo/invalid"
# 404
curl -I "https://nextstrain.org/community/user/repo/invalid" -H 'accept: text/html'
# 200
```

This contrasts with `/staging` which returns 404 for `text/html`:

```sh
curl -I "https://nextstrain.org/staging/invalid" -H 'accept: text/html'
# 404
```

Comment (victorlin) [reply static-site/app/community/[[...community]]/router.tsx]: I considered adding another commit to send proper 404 responses, primarily to address the related [testing comment](https://github.com/nextstrain/nextstrain.org/blob/04468c18665f4c5dc13de7cf1bbf1c5c1a7b252d/test/requests.test.js#L338-L347). That was fairly straightforward for https://nextstrain.org/community/user/repo/invalid: 259d795e9f59d0a27c64fd3af9180012c9e38b9f

However, it's complicated for https://nextstrain.org/community/user/invalid – that relies on a client-side fetch to `/charon` to determine whether the repo exists and is public. This could be moved server-side, but I didn't want to deal with that in this PR.

Comment (joverlee521) [reply static-site/app/community/[[...community]]/router.tsx]: Agree it's not worth worrying about the 404s in this PR.

PR #1291, opened 2025-12-31
Restore group name filtering

### LINKED ISSUE #1286: Show group name as a filter option in /groups
Group names should be a filtering option 👇 

Image

Priority: low

victorlin:

The commit "Strip group name in resource list" (c7b44980) had the unintended side effect of removing the group name from filter options.

This is a follow-up that keeps the visual change while allowing group names to be used in filtering. It is accomplished by adding a resource-level property to control which name parts are displayed, using that to hide the first name part on the groups page.

## Related issue(s)

Closes #1286

## Checklist

- [x] Checks pass

### DIFF
--- a/static-site/app/groups/available.tsx
+++ b/static-site/app/groups/available.tsx
@@ -83,11 +83,13 @@ export default function Available(): React.ReactElement {
       const parts = dataset.request.split('/').slice(1);
       const groupName = parts[1]
       if (parts[0] !== "groups" || groupName === undefined) return [];
-      const name = parts.slice(2).join('/');
+      const name = parts.slice(1).join('/');
+      const nameParts = name.split('/');
       return [{
         name,
         groupName,
-        nameParts: name.split('/'),
+        nameParts,
+        displayNameParts: nameParts.slice(1),
         sortingName: name,
         url: dataset.request,
       }];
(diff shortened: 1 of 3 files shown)


static-site/app/groups/available.tsx
```diff
@@ -83,11 +83,12 @@ export default function Available(): React.ReactElement {
       const parts = dataset.request.split('/').slice(1);
       const groupName = parts[1]
       if (parts[0] !== "groups" || groupName === undefined) return [];
-      const name = parts.slice(2).join('/');
+      const name = parts.slice(1).join('/');
       return [{
         name,
         groupName,
         nameParts: name.split('/'),
+        hideFirstNamePart: true,
```
joverlee521: _non-blocking_

Might be more generalizable to provide the `displayNameParts` and let each parent component decide on what should be displayed?

joverlee521 (approved):
The group name filtering doesn't work in the test app, but I _think_ that's due to there only being a single group in the test env. Worked as expected when I tested locally. 

Edit: Yup, the group name filter does not work when there's only a single group since options [filter out search options present in every resource](https://github.com/nextstrain/nextstrain.org/blob/e2ad1aa86948017877abf17f689e96b252dcb104/static-site/components/list-resources/use-filter-options.ts#L56-L57).

static-site/components/list-resources/resource-group.tsx
```diff
@@ -279,9 +279,11 @@ function _addDisplayName(
   const sep = "│"; // ASCII 179
 
   return resources.map((r, i) => {
+    const displayParts = r.displayNameParts ?? r.nameParts;
```
jameshadfield: [aside] have you generally started using `??` over `||` now?

joverlee521 (approved):

PR #1292, opened 2026-01-06
Add tuberculosis to featured analyses

kimandrews:

This PR does the following:
* Updates the tuberculosis tree image (because the previous image was from the [older tuberculosis workflow](https://github.com/nextstrain/tb/tree/dd3ef8a08eca138a0e4e8c26b3b34dc9c4d83139), which is much different from the current workflow)
* Adds tuberculosis to the featured analyses (splash page). The decision about where to place tuberculosis in the order of pathogen cards was [discussed on slack](https://bedfordlab.slack.com/archives/C07U8J5N465/p1767654139894899).

[Splash page preview](https://nextstrain-s-add-tb-to--twz3ue.herokuapp.com/)
[/pathogens preview](https://nextstrain-s-add-tb-to--twz3ue.herokuapp.com/pathogens)

## Related issue(s)
Closes https://github.com/nextstrain/tb/issues/42

## Checklist

- [ ] Checks pass

### DIFF
--- a/static-site/content/featured-analyses.yaml
+++ b/static-site/content/featured-analyses.yaml
@@ -14,6 +14,11 @@
   img: mpox_drc.png
   url: /groups/inrb-mpox/clade-I
 
+- name: Tuberculosis
+  description: Mycobacterium tuberculosis evolution and spread
+  img: tb_tree.png
+  url: /tb/global
+
 - name: All Ebola outbreaks
   description: Evolutionary history of all Zaire ebolavirus outbreaks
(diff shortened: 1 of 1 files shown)


trvrb (approved):
Looks great! I think that we have room for 2 more cards total before the splash page UI becomes too unwieldy. At that point we'll just need to start swapping out existing cards.

Or perhaps we just do a purge at some point soon of a few of the less recent / less important cards to make room for future expansion. 

cards

PR #1299, opened 2026-01-19
[charon/getAvailable] snapshots

jameshadfield:
The new Auspice dataset selector UI <https://github.com/nextstrain/auspice/pull/2030> looks for an optional `snapshots: boolean` in the getAvailable API response to decide whether to expose this in the UI. Over time we may adapt this response (and the UI) to list the known snapshots, but for now a `true` value will result in a freeform text box where you can enter in the desired YYYY-MM-DD snapshot for a given dataset.

Tested locally (Auspice server/client & nextstrain.org server on different localhost ports) -- see https://github.com/nextstrain/auspice/issues/2033.

- [x] Checks pass


jameshadfield:
Merging as there's no changes to behaviour with the current Auspice, and this will allow the forthcoming Auspice release to work as intended

PR #1342, opened 2026-05-21
Fix domain

jsoref:

Replace reference to malicious domain

## Checklist

- [ ] Checks pass


victorlin (approved):
Nice catch, thanks!

PR #1343, opened 2026-05-22
Validate featured analyses

victorlin:
Add runtime validation to the contents of the featured analyses YAML file. Previously, the type assertion ignored issues with missing or incorrectly typed fields. Some issues are still ignored (see comments), but at least this is a more robust TypeScript pattern.

## Checklist

- [x] Checks pass

### DIFF
--- a/static-site/components/splash/index.tsx
+++ b/static-site/components/splash/index.tsx
@@ -5,7 +5,7 @@ import React, { useEffect, useState } from "react";
 import { Tooltip } from "react-tooltip-v5";
 import ScrollableAnchor from "../../vendored/react-scrollable-anchor/ScrollableAnchor";
 
-import featuredAnalyses from "../../content/featured-analyses.yaml";
+import featuredAnalysesYaml from "../../content/featured-analyses.yaml";
 
 import bookOpenIcon from "../../static/logos/fa-book-open-solid.svg";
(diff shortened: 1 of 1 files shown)


static-site/components/splash/index.tsx
```diff
@@ -34,6 +34,9 @@ export interface SplashTile extends GenericTileBase {
   /** url to the dataset on the tile */
   url: string;
 }
+// TypeScript interfaces are not available at runtime, so keep the keys in a
+// runtime value.
+const splashTileKeys = ["name", "description", "img", "url"] satisfies (keyof SplashTile)[];
```
jameshadfield, line 39: [aside]

[This post](https://stackoverflow.com/questions/73909490/new-typescript-satisfies-operator) helped me understand the benefits of `satisfies` rather than just

```ts
const splashTileKeys: (keyof SplashTile)[] = ["name", "description", "img", "url"];
```

Although in this particular case I'm not sure there are any actual benefits except for a more helpful type-hint message from the LSP:

Image

Image

jameshadfield (approved):
Nice!

static-site/components/splash/index.tsx
victorlin: Codex wrote this for me and it seemed more appropriate than `keyof` so I kept it. I think it's slightly more readable too.

Issue #1347, opened 2026-06-08
HTTP 503 for `charon/getDataset?prefix=fetch/` requests to `genome.ucsc.edu` as that domain blocks all AWS IPs


AngieHinrichs:
Hi Nextstrain team!  @ryhisner reported in yatisht/usher#438 that usher.bio (genome.ucsc.edu/cgi-bin/hgPhyloPlace) linkouts to nextstrain.org/fetch/ stopped working.  Links like https://nextstrain.org/fetch/genome.ucsc.edu/trash/hgPhyloPlace/subtreeAuspice1_genome_18fb48_724b90.json give the error "There seems to have been an error accessing that dataset.
Details: Couldn't load JSONs for fetch/genome.ucsc.edu/trash/hgPhyloPlace/subtreeAuspice1_genome_18fb48_724b90.json . Please [contact us](https://nextstrain.org/contact) if you think this is a bug."

In the JS console, there is
<pre>
Failed to load resource: the server responded with a status of 503 ()
</pre>
followed by a longer error including some HTML from /charon/getDataset?prefix=fetch/genome.ucsc.edu/trash/hgPhyloPlace/subtreeAuspice1_genome_18fb48_724b90.json

Recently, genome.ucsc.edu was hit by a DDoS attack from AWS hosts (or possibly other hosts spoofing their IPs) and we made it stop by denying all requests to AWS hosts.  But I think you guys use AWS for a lot -- do you use it for nextstrain.org?  If so, do you get a reasonably stable IP address that we could whitelist, or is it changing all the time underneath DNS?  I will inquire on our end into ways that we might exempt hgPhyloPlace results and/or nextstrain.org from our AWS block.

jameshadfield:
Hey Angie,

>  But I think you guys use AWS for a lot -- do you use it for nextstrain.org?

Yes we do - although transiently in this case, as we use Heroku as our server provider Heroku runs on top of Amazon Web Services (AWS) infrastructure. Those fetch requests get proxied through the server, so I think the USCS server will see the request as coming from an AWS IP.

> If so, do you get a reasonably stable IP address that we could whitelist, or is it changing all the time underneath DNS?

Not sure about this, let us get back to you!

victorlin:
Heroku assigns dynamic and unpredictable IPs to our server, so unfortunately IP-based whitelisting won't work.

[...]

PR #1348, opened 2026-06-08
Add /browser-fetch endpoint

victorlin:
An alternative to /fetch that, instead of proxying the request through the server, sends a client-side request from the user's web browser.

Note that this only works for upstream URLs that support cross-origin requests.

## Checklist
- [ ] Update documentation [Share Analyses through Nextstrain](https://docs.nextstrain.org/en/latest/guides/share/index.html) and [Share via public URLs](https://docs.nextstrain.org/en/latest/guides/share/fetch-via-urls.html)
- [ ] Checks pass

### DIFF
--- a/docs/api-restful.rst
+++ b/docs/api-restful.rst
@@ -354,6 +354,8 @@ The following dataset endpoints exist::
 
     {GET, HEAD, OPTIONS} /fetch/*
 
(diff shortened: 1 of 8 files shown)


victorlin:
Closing per https://github.com/nextstrain/nextstrain.org/issues/1347#issuecomment-4654993482

PR #1363, opened 2026-07-23
Update measles build name format

victorlin:
Follow-up to change in measles repo.¹

¹ https://github.com/nextstrain/measles/commit/474ae9d13c90fc4567cf7bc5e5da3288cb38fcad

Preview: https://nextstrain-s-victorlin--ajeclp.herokuapp.com/measles

## Checklist

- [x] Checks pass

### DIFF
--- a/data/manifest_core.json
+++ b/data/manifest_core.json
@@ -416,9 +416,19 @@
         }
       },
       "measles": {
-        "resolution": {
-          "genome": "",
-          "N450": "",
+        "gene": {
+          "genome": {
+            "region": {
+              "global": "",
+              "default": "global"
+            }
+          },
+          "N450": {
+            "region": {
+              "global": "",
+              "default": "global"
+            }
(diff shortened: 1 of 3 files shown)


data/manifest_core.json
```diff
@@ -416,9 +416,19 @@
         }
       },
       "measles": {
-        "resolution": {
-          "genome": "",
-          "N450": "",
+        "gene": {
+          "genome": {
+            "region": {
+              "global": "",
```
jameshadfield, line 422: No `north-america`? If we push `genome_north-america.json` to nextstrain-data, we should still be able to access it via URL (I think!) so maybe that's ok for a while

victorlin: If we add `north-america` now, it would show in the dataset selector dropdown and 404 on load – not good! It should be added if/when the dataset is pushed.

PR #1365, opened 2026-07-30
Install Auspice with build-time scripts

jameshadfield:
Auspice v3's CLI is now built (transpiled) and thus we need to run npm's scripts during the installation process. See <https://github.com/nextstrain/auspice/pull/2088> for more details

While this PR is targeted at Auspice v3 it should run just fine for v2 as well, just a little slower.

### DIFF
--- a/auspice-client/package-lock.json
+++ b/auspice-client/package-lock.json
@@ -5,7 +5,7 @@
   "packages": {
     "": {
       "dependencies": {
-        "auspice": "^2.73.0",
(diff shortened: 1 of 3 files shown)


jameshadfield:
The single commit here, once merged, will allow our automatic testing PRs for Auspice v3 to work. Auspice v2 installs also work, albeit the build times are slightly slower. 

For posterity, this PR had an extra (now dropped) commit installing Auspice from the v3 branch to confirm that the review app worked.

Will merge shortly.

PR #1377, opened 2026-09-04
Replace old map screenshots

victorlin:

Future map screenshots will likely use the new Mapbox styles that come with Auspice v3. Update all screenshots to reflect the new styles (and latest data) to avoid having a mix of old and new styles on the same page.

## Related issue(s)

Follow-up to #1372

## Checklist

- [x] Checks pass


joverlee521 (approved):
Thanks for thinking of this @victorlin!

